PR #9, opened 2021-01-26
Enroll user in discount group

thekaveman:
This PR implements enrollment of an eligible user into the discount group associated with their eligibility status.

## Overview 

* Discount provider API details stored with [model](https://github.com/cal-itp/benefits/blob/discounts/benefits/core/models.py#L17)
* Discount group is associated with [Eligibility model](https://github.com/cal-itp/benefits/blob/discounts/benefits/core/models.py#L50)
* Simple [forms](https://github.com/cal-itp/benefits/blob/discounts/benefits/discounts/forms.py) bring data and signal events from the client to server
* [Base API client](https://github.com/cal-itp/benefits/blob/discounts/benefits/discounts/api.py#L13) implements client certificate auth used by other clients.

## High-level algorithm

See [server-side components here](https://github.com/cal-itp/benefits/blob/discounts/benefits/discounts/views.py) and [client-side components here](https://github.com/cal-itp/benefits/blob/discounts/benefits/discounts/templates/discounts/index.html)

1. Check for an API access token
   - If exists and valid, continue
   - If exists and expired, refresh
   - If doesn't exist, acquire

2. Present the card tokenization form to user
   - If tokenization succeeds, continue
   - If tokenization fails, present retry screen -> go to 1.
   - If user cancels, go to 1.

3. Using card token, check for customer record
   - If customer exists, update record to allow group enrollment
   - If customer does not exist, create record allowing group enrollment

4. Using customer id, enroll in discount group


thekaveman:
Merging this as-is, will follow up with fixes when integration testing resumes.

PR #10, opened 2021-01-27
Deployment configuration

thekaveman:
## Overview

The new image is based on `python:3.9-slim`, which comes with only a very limited set of tooling on top of the required Python dependencies, reducing the surface area of the image. 

Source code and related assets are copied into the final image with elevated permissions, before changing the user that the image runs under to a non-root user, effectively making source files read-only within the container.

The image has been configured to produce a runnable container that can be started by an ECS Fargate Task. Running the container initializes the Django environment / database, starts `nginx` listening on port 80, and `gunicorn` on a Unix socket. HTTPS is handled by the AWS load balancer.

The Django configuration and settings have been trimmed to only enable those features which are entirely necessary for the application. Django sessions have been configured to use secure cookies on the client instead of storing session data in the Django database, meaning Django stores *absolutely zero* information about the user.

### Still Outstanding

The AWS deployment pipeline is based on our (successful!) tests over in [docker-django-aws-test](https://github.com/cal-itp/docker-django-aws-test).

* [ ] The ECS Task Definition is only a stub that needs to be filled out once we have that environment set up
* [ ] The corresponding GitHub Actions workflow will also need to be updated (including configuring secrets in this repo)
* [x] Update Django settings to configure `ALLOWED_HOSTS`


krainboltgreene (approved):

thekaveman:
Merging this work today ahead of AWS configuration tomorrow.

PR #22, opened 2021-03-13
Allow manual triggers using workflow_dispatch

thekaveman:
Need to merge this into `main` to be able to trigger this workflow manually.


thekaveman:
More info: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #34, opened 2021-04-08
Updated Spanish translations based on feedback

calexity:
Several changes to language including
- using "transporte público"
- making bank card language simpler and clearer
- making age requirements clearer
- updating spelling and accents that were wrong

@thekaveman one question - i feel like some of the language from the "Payment Options" ("Opciones de Pago") page is missing but it's hard for me to tell. Is that in another file?


thekaveman:
@calexity all of the content should be in these PO files. The Payment Options content in on [lines 55-95 in this Spanish PO file](https://github.com/cal-itp/benefits/blob/ae7618123c75ea957cc9389191e392c706bcca75/benefits/locale/es/LC_MESSAGES/django.po#L55..L95). Can you verify it's all there?

thekaveman (approved):

PR #65, opened 2021-05-22
Refactor ECS task and workflow for environments

thekaveman:
Getting ready for a dev, test, and prod deployment. 

Remove environment-specific info from ECS task template, add a step that uses `sed` to fill in with environment secrets.


thekaveman:
Merging now to see if the Actions config will work.

The resulting ECS deploy _should_ fail (or not run at all) since the environment specified has a branch rule for `dev`. If this is the case, create a `dev` branch and see if manually triggering the Action works (it should).

thekaveman:
> The resulting ECS deploy should fail (or not run at all) since the environment specified has a branch rule for `dev`.

Success! https://github.com/cal-itp/benefits/actions/runs/872406063

> Branch "main" is not allowed to deploy to dev due to environment protection rules.

PR #69, opened 2021-06-07
feat: pre-commit improvements, enforce conventional commits

thekaveman:


thekaveman:
Ignoring the `py-up.io/safety-ci` check, as Django has already been updated in `main`.

PR #113, opened 2021-09-08
chore(gitignore): add __pycache__ directory to gitignore

machikoyasuda:
Adds `__pycache__` directory to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 *.mo
 **/data/*.json
 !localhost/data/[client|server].json
-localhost/server/__pycache__/
 static/
 !benefits/static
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,5 +3,6 @@
 *.mo
 **/data/*.json
 !localhost/data/[client|server].json
+localhost/server/__pycache__/
```
machikoyasuda, line 6: @thekaveman Not sure if I should make this directory more flexible with asterisks.

thekaveman (approved):
Nice! I like the broader ignore pattern to catch the directory anywhere it may be!

Issue #109, opened 2021-08-25
Make docs site rebuild only on changes to docs


thekaveman:
Right now the `mkdocs` process and GitHub Pages redeploy happens on all pushes to `dev`.

We should put a filter in the [Action](https://github.com/cal-itp/benefits/blob/dev/.github/workflows/mkdocs.yml) to only trigger for changes in the [docs/](https://github.com/cal-itp/benefits/tree/dev/docs) directory.

PR #114, opened 2021-09-09
Workflow: Only deploy Docs site when docs files change

machikoyasuda:
closes #109 

## How to test this GitHub Action (or, How I tested this GitHub Action)

1. Fork this repository
2. As an extra precaution, delete all unrelated GitHub Action workflows. To make sure unnecessary deploys to Production were absolutely impossible, I deleted the `.github/workflows/ecs-deploy.dev.yml` and `.github/workflows/ecs-deploy-test.yml' files.
3. Merge the PR I am testing into `dev`. I merged `109-docs-action-config`.
4. Click **Actions** tab. Click the button to turn on Actions for this repository. (Sidenote: It's great that GitHub turns Actions off by default on repositories that are forked.)
5. Create new PRs to test the workflow filter: I first created a PR that changes a `/localhost/` file. Then, I created a PR that changes a `/docs/` file. 
6. Merge one PR at a time and look at the **Actions** tab to confirm the desired result.

### When merging a PR to `dev` that changes `/docs/` files

image

After the PR is created, `machikoyasuda-patch-1`, the pre-commit workflow runs. Then, after this PR is merged to `dev`, the `Publish docs` action runs.

### When merging a PR to `dev` that changes `non-/docs/` files

image

After this PR - that only changed a `/localhost/` file - is created and merged to `dev`, the `Publish docs` action is **not** triggered.

PR #115, opened 2021-09-09
Workflow: Only deploy Dev if dev code has changed

### LINKED ISSUE #111: Prevent dev environment deploy for docs changes
Corollary to #109.

We want to prevent the [`dev` deployment](https://github.com/cal-itp/benefits/blob/dev/.github/workflows/ecs-deploy-dev.yml) from occurring for changes that only touch the [docs/](https://github.com/cal-itp/benefits/tree/dev/docs) directory.

machikoyasuda:
close #111 

adds `paths` filter so this dev deploy workflow only runs if at least one file in the explicitly specified list is changed.

## How to test this PR

Follow the instructions from #109 

https://github.com/machikoyasuda/benefits/actions

I made 2 PRs - one PR that changes app code (a i18n copy file) and another that changes a `/docs/` file. 

Only the PR that made changes to app code triggered the Deploy Dev workflow: 
image

image

### DIFF
--- a/.github/workflows/ecs-deploy-dev.yml
+++ b/.github/workflows/ecs-deploy-dev.yml
@@ -5,17 +5,15 @@ on:
   push:
     branches:
       - dev
-    paths-ignore:
-      - 'docs/**'
-      - 'localhost/**'
-      - '.vscode/**'
-      - '.gitignore'
-      - '.github/workflows/mkdocs.yml'
-      - '.github/workflows/pre-commit.yml'
-      - '.pre-commit-config.yaml'
-      - '.pyup.yml'
-      - 'README.md'
-      - 'mkdocs.yml'
+    paths:
+      - '.aws/**'
+      - '.github/workflows/ecs-deploy-*.yml'
+      - 'benefits/**'
+      - 'bin/**'
+      - Dockerfile
+      - gunicorn.conf.py
+      - nginx.conf
+      - requirements.txt
 
 defaults:
   run:
--- a/.github/workflows/ecs-deploy-dev.yml
+++ b/.github/workflows/ecs-deploy-dev.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - dev
+    paths-ignore:
+      - 'docs/**'
 
 defaults:
   run:
--- a/.github/workflows/ecs-deploy-dev.yml
+++ b/.github/workflows/ecs-deploy-dev.yml
@@ -7,6 +7,15 @@ on:
       - dev
     paths-ignore:
       - 'docs/**'
+      - 'localhost/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/ecs-deploy-dev.yml
```diff
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - dev
+    paths-ignore:
+      - 'docs/**'
```
machikoyasuda, line 9: Add `localhost` folder as well

machikoyasuda: Approach: List things we know we don't care about (rather than the opposite - in case we add new files we do care about).

.github/workflows/ecs-deploy-dev.yml
```diff
@@ -5,6 +5,17 @@ on:
...
+      - 'localhost/**'
+      - '.vscode/**'
+      - '.gitignore'
+      - '.github/workflows/mkdocs.yml'
+      - '.github/workflows/pre-commit.yml'
+      - '.pre-commit-config.yaml'
+      - '.pyup.yml'
+      - 'README.md'
+      - 'mkdocs.yml'
```
machikoyasuda, line 18: @thekaveman I wasn't sure if I should add *all* non-apps config files like `.flake8`

thekaveman: Hmmm, good point @machikoyasuda... this kind of blows up my earlier thought about it being easier to list the files we don't care about 🤣 

Alternatively we can list the things we do care about under `paths` like in #114, which may actually be shorter?

```yaml
- '.aws/**'
- '.github/workflows/ecs-deploy-*.yml'
- 'benefits/**'
- 'bin/**'
- Dockerfile
- gunicorn.conf.py
- nginx.conf
- requirements.txt
```

I'm torn. Either way it's a long list... take the approach that makes the most sense to you 👍

thekaveman: The +1 for doing it the same as in #114 (besides being consistent across workflows) is that the explicit `paths` list makes it easier to see what will trigger the workflow, vs. the `paths-ignore` you have to do a bit of mental work to figure out what will trigger.

thekaveman: 🤷

machikoyasuda: @thekaveman Now that I look at the list of files that **do** matter, it makes more sense to me. This way, it also communicates what files actually matter for the app and deployment. Gonna switch to this strategy!

thekaveman (approved):
Yeah this feels cleaner. Nice!

PR #124, opened 2021-09-30
Dynamic localhost ports

### LINKED ISSUE #123: Figure out dynamic Docker ports for localhost
Hardcoding localhost ports to map into Docker containers is unnecessary maintenance overhead, especially when working on multiple projects each with a number of containers/services and trying to choose ports that allow all services to run simultaneously.

Ideally Docker would just pick an available port during container startup and use that. We'd like to get away from managing these values / environment variables:

* `DJANGO_LOCAL_PORT`
* `DOCS_LOCAL_PORT`
* `5678` hardcoded as the debug port for `dev`

The following values will be moving from this project into [`eligibility-server`](/cal-itp/eligibility-server) as part of the work in #121, so we'll want to apply the same approach there as well:

* `5000` hardcoded as the `server` port
* `5789` hardcoded as the debug port for `server`

We'll need to consider:

* Running ad-hoc `docker compose ...` commands
* Running from a devcontainer with associated services

This SO post is a good starting place: https://stackoverflow.com/a/56447023/453168.

thekaveman:
Closes #123 by removing all hardcoded localhost ports and allowing Docker to assign dynamically.

Relevant docs updated to describe how to find the dynamic port for various situations.


docs/getting-started/documentation.md
```diff
@@ -49,7 +49,21 @@ The documentation website can be run locally using Docker Compose:
...
+```
+
+The `PORTS` column should look like:
+
+```console
+0.0.0.0:$PORT->8000/tcp
+```
+
+Where `$PORT` is the dynamically assigned `localhost` port. You can navigate in a browser to <http://localhost:$PORT> to view
+the docs site.
```
machikoyasuda, line 66: I am not able to build or run the docs Docker container locally:

image

docs/getting-started/README.md
```diff
@@ -35,14 +35,30 @@ docker compose build --no-cache client
...
 ```
 
-After initialization, the client is running at `http://localhost:${DJANGO_LOCAL_PORT}`
-(<http://localhost:8000> by default).
+The optional `-d` flag will start in _detatched_ mode and allow you to continue using the terminal session. Otherwise your
+terminal will be attached to the container's terminal, showing the startup and runtime output.
 
-If `DJANGO_ADMIN=true`, the backend administrative interface can be accessed with the superuser you setup at
-<http://localhost:8000/admin>.
+After initialization, the client is running running on `http://localhost` at a port dynamically assigned by Docker. To find the
```
machikoyasuda, line 44: What the Docker Desktop interface looks like: Click **Containers / Apps**, 
image

image

Find $PORT at `http://0.0.0.0:56760/` or `http://localhost:56760/`

docs/getting-started/test-verification-server.md
```diff
@@ -3,10 +3,28 @@
...
+```console
+0.0.0.0:$LOCAL_PORT->5000/tcp, 0.0.0.0:$DEBUG_PORT->5678/tcp
+```
+
+Where `$LOCAL_PORT` and `$DEBUG_PORT` are the dynamically assigned `localhost` ports.
+
+From within another Compose service container, the API is at `http://server:5000` using the service-forwarding features of
+Compose.
+
+In either case, the endpoint `/verify` serves as the Eligibility Verification API endpoint.
```
machikoyasuda, line 30: - I am able to successfully build and run `docker compose build server`, `docker compose up server`

- Note: This log might be confusing to users because the `* Running on http://172.26.0.3:5000/` is now incorrect
image

- When I run the Docker app and the `docker ps` command, I see that the port is assigned to `57463`:
image
image

- When I have both the server and client running via docker-compose, I am able to go through a successful verification flow:
- 
image
image

docs/getting-started/documentation.md
machikoyasuda: I'm actually getting this error on the `dev` branch so maybe it's just me?!?!

thekaveman: Weird, I'm getting the same error... I can't tell if it's a problem with `mkdocs` or a problem in our markdown content somewhere? 

One note: for `docs` there's actually nothing to build, since it's using the same image as `dev`. That one has to be built first, before `docs` can run/up.

machikoyasuda: Works for me now!

machikoyasuda (approved):
💯💯💯

After `Rebuild and Reopen`ing the Dev Container, restarting VSCode (for the Python extension), running the `Django: Benefits Client` debugger and checking the ports with `docker ps`, I was able to open:

- the Docs
- the Client
- the Server

PR #126, opened 2021-10-05
Use launch type debugger

thekaveman:
Maintain the `F5`-to-start sequence, but allow for stop/restart unlike the attach type. We no longer need to expose a debugger port.

The compose service definition was simplified to decouple the app and devcontainer startup, using the `sleep infinity` command ensures the devcontainer stays active if the app/debugger crash.

Since VS Code handles the debugger interaction, we don't need to explicitly install `debugpy`.

## To run this PR

1. Rebuild and Reopen the devcontainer
2. Add a breakpoint to `/healthcheck`, Hit `F5` to launch the debugger, navigate to `/healthcheck`, hit breakpoint
3. Observe you can **stop** the debugger, then **restart** with `F5`

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -15,8 +15,7 @@
             "bash": {
                 "path": "/bin/bash"
             }
-        },
-        "python.pythonPath": "/usr/local/bin/python",
+        }
     },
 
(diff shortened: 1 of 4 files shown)


machikoyasuda (approved):
💯💯💯

PR #148, opened 2021-10-19
Update language on Payment Options page

thekaveman:
Updating "Payment Options" page to include note about supported payment network providers


o-ram:
Spanish translation for the new language below!

Nota: Debe incluir un logo de Visa o Mastercard

PR #170, opened 2021-11-02
Cypress tests for eligibility flow

### LINKED ISSUE #136: Write + document integration Cypress tests
- The test should run in the *newest* code from the pull request, not `dev`.

machikoyasuda:
closes #136 

## What this PR does
- Creates a `fixtures/eligibility.json` file with hard-coded last names and driver's license id from https://github.com/cal-itp/eligibility-server/blob/main/data/server.json#L3 (In the future, this should be further abstracted)
- Add tests that open an Electron browser and goes through the following user interface flows:

1. Clicking button to get to Eligibility page
2. Clicking button to get to Confirmation page
3. ~Confirming an eligible user~ **Note**: This spec is written but it's set to `skip()`. This issue has been created to get this test to pass in the future.
4. Confirming an ineligible user
5. Validating a field with invalid format for license: Tests that the app displays error messages when the server returns one.
6. Validating an empty name field: Tests that the browser validates an empty field and marks a field as `invalid` in the browser natively.
7. Making sure the index page has all the necessary links (but not actually clicking them)

## How to test this PR

- Run all specs
```
cd tests/e2e
npx cypress run
```

- Run one file at a time
```
npx cypress run --spec cypress/integration/eligibility.spec.js
```

- Run one spec only
```
npx cypress run
```

and add `.only` to the spec, like:

`it.only("assertion asjfasdf")`


tests/e2e/cypress/integration/confirm.spec.js
```diff
@@ -0,0 +1,41 @@
...
+    cy.get("#name").type("Garcia")
+    cy.get("input[type='submit']")
+      .click()
+    cy.contains("Check your input. The format looks wrong.")
+    cy.get("#id_sub")
+      .then(($e) => {
+        expect($e).to.have.css("border-color", "rgb(222, 12, 12)")
+      })
+  })
+})
```
thekaveman, line 41: These tests are a great start! Two pieces of feedback:

1. At a higher-level, I think we can combine the `confirm.spec.js` and `eligibility.spec.js` tests into a single `eligibility.spec.js` file, since they are all testing the Eligibility flow of the app.
2. As for the specs themselves, we have to find a way to inject the `A1234567`, `Garcia`, etc. values rather than hard-coding them in the tests, for our know use-cases. This is related to the question of _What server are the tests hitting?_ and needs to be a little more configurable than this.

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,23 @@
...
+
+    cy.contains("Ready to continue")
+      .then(($e) => {
+        expect($e).attr("href").to.have.string("eligibility/confirm")
+      })
+
+    cy.get(".buttons .btn")
+      .click()
+
+    cy.contains("Let’s see if we can confirm your age with the DMV")
```
thekaveman, line 20: Maybe leave `with the DMV` out of the spec?

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    cy.contains("CA driver’s license or ID number *")
```
thekaveman, line 21: Can we test not only that there is a label with this text, but that it is associated with an input?

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  })
```
thekaveman, line 22: Let's also do the same for the Last Name input

tests/e2e/cypress/integration/index.spec.js
```diff
@@ -1,25 +1,44 @@
...
+      })
+  })
+
+  it("Has a payment options page", () => {
+    cy.visit("/")
+
+    cy.contains("Payment Options")
+      .click()
+
+    cy.contains("The contactless symbol is four curved lines")
```
thekaveman, line 32: Let's test the URL, the "Payment Options" heading, and that a button exists that takes user back to index

tests/e2e/cypress/integration/index.spec.js
```diff
@@ -1,25 +1,44 @@
...
+  })
+
+  it("Has a help page", () => {
+    cy.visit("/")
+
+    cy.contains("Help")
+      .click()
+
+    cy.contains("If you need assistance with the site, please reach out to the customer service team for your local transit provider.")
```
thekaveman, line 41: This will be trickier, but on this page we should test that the contact info for the configured agency (or both if coming from the top-level index) is shown.

tests/e2e/cypress/integration/spanish.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+  it("Changes the language to Spanish", () => {
+    cy.visit("/")
+
+    cy.contains("Español")
+      .click()
+
+    cy.contains("La nueva forma de pagar")
+    cy.get("#header").should("not.contain", "Español")
+    cy.get("#header").should("contain", "English")
```
thekaveman, line 17: Let's also verify that the URL stayed the same (i.e. clicking the button didn't take them to another page)

tests/e2e/cypress/integration/spanish.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    cy.contains("English")
+      .click()
+
+    cy.get("#main-content").should("contain", "The new way to pay for transit")
+    cy.get("#main-content").should("not.contain", "La nueva forma de pagar")
```
thekaveman, line 33: Check the button again after changing back

```js
cy.get("#header").should("contain", "Español")
cy.get("#header").should("not.contain", "English")
```

tests/e2e/cypress/integration/spanish.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+    cy.get(".buttons .btn")
+      .first()
+      .click()
+
+    cy.contains("Español")
+      .click()
+
+    cy.get("#main-content").should("contain", "¡Hagámoslo!")
+    cy.get("#main-content").should("not.contain", "Let’s do it!")
```
thekaveman, line 47: I think rather than checking for specific content, we should check that the button exists and when clicking it, it changes to the English button while the page URL stays the same. 

That way if we end up inserting new pages / changing the text/order, this test still passes (which is that the button should be on any page).

thekaveman (changes requested):
These are great! Few notes on organization and some additional items to test.

There's one big thing to figure out and that is, how to inject test values (e.g. `A1234567`) rather than hard-coding them.

tests/e2e/cypress/integration/spanish.spec.js
machikoyasuda: To be resolved in different PR.

tests/e2e/cypress/integration/spanish.spec.js
machikoyasuda: To be resolved in different PR.

tests/e2e/cypress/integration/spanish.spec.js
machikoyasuda: To be resolved in different PR.

tests/e2e/cypress/integration/index.spec.js
machikoyasuda: To be resolved in different PR.

tests/e2e/cypress/integration/index.spec.js
machikoyasuda: To be resolved in different PR.

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,88 @@
...
+  it("Confirmation page has a form with a lastname form label and corresponding form field", () => {
+    cy.contains("Ready to continue")
+      .click()
+    cy.get('input:focus').should('have.length', 0)
+
+    cy.contains("Last name (as it appears on ID) *")
+      .click()
+
+    cy.get('input:focus').should('have.length', 1)
+  })
```
machikoyasuda, line 38: I tried to test that the form fields and labels are semantic (matching `for`/`name` values) in a user-centric way.

A user clicks on the label (`CA driver's license or ID number *`) and the form field should come into focus. For this, I'm testing that 1 form field comes into focus. Is that enough? (I could do an extra check to see that the form field and input are in the same parent div, or that the `name` ('sub') matches the `for` ('sub') -- but those seem like very code-centric things to check -- it's not checking for something a user would notice or be affected by. How else would a user know if a form label and field are correctly paired?)

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,88 @@
...
+    cy.contains("Ready to continue")
+      .click()
+    cy.get("#sub").type("A1234567")
+    cy.get("#name").type("Bob")
+    cy.get("input[type='submit']")
+      .click()
+
+    cy.contains("We can’t confirm your age")
+    cy.contains("You may still be eligible for a discount, but we can’t verify your age")
+  })
```
machikoyasuda, line 61: @thekaveman 

I tried to test that the form fields and labels are semantic (matching `for`/`name` values) in a user-centric way.

A user clicks on the label (`CA driver's license or ID number *`) and the form field should come into focus. For this, I'm testing that 1 form field comes into focus. Is that enough? (I could do an extra check to see that the form field and input are in the same parent div, or that the `name` ('sub') matches the `for` ('sub') -- but those seem like very code-centric things to check -- it's not checking for something a user would notice or be affected by. How else would a user know if a form label and field are correctly paired?)

thekaveman: I like checking that the form field comes into focus. That's the real "goal" of all that other stuff you mentioned - that the label and field are directly related from the User's point of view. Nice!! 👍 👍

thekaveman:
Small clarification, but important from a "understand what we are testing" perspective:

> 5. Validating a field with invalid format for license: Tests that the app checks for valid driver's license/id length and returns an error message and a red bordered field box.

The benefits app actually does not contain this logic. The eligibility API defines a response type that carries some minimal error information: https://docs.calitp.org/eligibility-server/specification/#errors

So our app is just using information coming back from the Eligibility Server.

tests/e2e/cypress/fixtures/eligibility.json
```diff
@@ -0,0 +1,14 @@
+{
+  "eligible": {
```
machikoyasuda, line 2: @thekaveman Any feedback on how I've organized this json file and how I'm using it?

machikoyasuda:
@angela-tran This PR is now ready for review!

I've added instructions on how to run these tests (`eligibility.spec.js` and `index.spec.js`) and what I am trying to test.

tests/e2e/cypress/fixtures/eligibility.json
```diff
@@ -0,0 +1,14 @@
...
+    "name": "Garcia",
+    "sub": "A1234567"
+  },
+  "wrongName": {
```
thekaveman, line 6: Maybe call this one `ineligible` to pair with `eligible`. Since it's not technically "wrong" like the one below is.

tests/e2e/cypress/fixtures/eligibility.json
```diff
@@ -0,0 +1,14 @@
...
+    "name": "Bob",
+    "sub": "A1234567"
+  },
+  "wrongSub": {
```
thekaveman, line 10: maybe `invalidSub` to be clear that it was a user error?

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,90 @@
+const fixtures = require("../fixtures/eligibility.json")
+
+describe("Confirms a user’s eligibility status", () => {
+  beforeEach(() => {
+    cy.visit("/")
+    cy.contains("ABC")
```
thekaveman, line 6: Inject with fixtures

tests/e2e/cypress/integration/eligibility.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      .click()
+
+    cy.get('input:focus').should('have.length', 1)
+  })
+
+  it("From the confirmation page, confirms an eligible user", () => {
+    cy.contains("Ready to continue")
+      .click()
+    cy.get("#sub").type(fixtures.eligible.sub)
+    cy.get("#name").type(fixtures.eligible.name)
```
thekaveman, line 37: this is awesome!! 🎸

tests/e2e/cypress/integration/index.spec.js
```diff
@@ -1,25 +1,34 @@
...
-        cy.contains("Choose your transit provider")
-          .siblings(".btn")
-          .should('not.be.empty')
-          .each(($e) => {
-              expect($e).attr("href").to.match(/\/[a-z]{3,}$/)
-          })
-    })
+  it("Gives user transit provider options", () => {
+    cy.contains("Choose your transit provider")
+      .siblings(".btn")
```
thekaveman, line 8: Can we avoid the CSS selector here? Since we follow with a check of the `href`, we're already assuming `<a>` tags right? Is that assumption better/worse than the CSS class?

tests/e2e/cypress/integration/index.spec.js
```diff
@@ -1,25 +1,34 @@
...
+      .should('not.be.empty')
+      .each(($e) => {
+        expect($e).attr("href").to.match(/\/[a-z]{3,}$/)
+      })
+  })
 
-    it("Takes user to transit provider page", () => {
-        cy.visit("/")
+  it("Takes user to a transit provider page", () => {
+    cy.get(".buttons .btn")
```
thekaveman, line 16: Same as above, is there a better way to get to first button?

thekaveman (changes requested):

thekaveman (changes requested):
Uh oh, I think the force push resolved all the comments I made previously??

I don't think these were addressed:

image

machikoyasuda:
@thekaveman oh weird. here's the new commit i added to address the comments: https://github.com/cal-itp/benefits/pull/170/commits/94caf2c869f49ca27772d81d4a39beefbbcc025d

thekaveman:
Gah! I'm sorry, I think I must not have refreshed my page all the way or something. My fault!

thekaveman (approved):
👍

angela-tran:
I ran them successfully too :+1: 

image

I definitely gleaned more about Cypress from looking over your code and @thekaveman 's comments.

tests/e2e/cypress/integration/index.spec.js
angela-tran: Was resolved in [#176](https://github.com/cal-itp/benefits/pull/176/files?diff=unified&w=0#diff-346f1b6777a6139ef6aed803c8ff4396f25af2c29eb5f64b872de453443e8920)
 - checked URL
 - checked heading
 - checked Back button

tests/e2e/cypress/integration/index.spec.js
angela-tran: Was resolved in #176
- checked that all info is present if coming from top-level index [(line 35 of help.spec.js)](https://github.com/cal-itp/benefits/pull/176/files?diff=unified&w=0#diff-b092ffc585b11e5dcca602107a22ae7f70a62183073a31e62a2fb02e9185bd7dR35)
- checked for info if coming from transit-agency-specific context [(line 45 of help.spec.js)](https://github.com/cal-itp/benefits/pull/176/files?diff=unified&w=0#diff-b092ffc585b11e5dcca602107a22ae7f70a62183073a31e62a2fb02e9185bd7dR45)

PR #172, opened 2021-11-02
Fix Linux devcontainer issue from file permissions

angela-tran:
Closes #171 

- Added lines to `Dockerfile.dev` to use the `calitp` user so that the npm dependencies are installed using that user rather than root
- Added lines in `Dockerfile` to make the Python packages be installed using `calitp` as well

Issue #171, opened 2021-11-02
VS Code Devcontainer fails to set up on Linux


angela-tran:
### Expected Result
The devcontainer is successfully set up.

### Actual Result
An error is thrown when trying to run the `postAttachCommand`, specifically when we're trying to install Cypress.

### Steps to Reproduce
1. Have Linux as your host operating system
2. Make sure you've built the base Docker image `benefits_client:latest`
3. Open the `benefits` folder as a remote container in VS Code

### Output
From console logging:
```
Running the postAttachCommand from devcontainer.json...

[117148 ms] Start: Run in container: /bin/bash localhost/bin/init.sh
[117388 ms] Port forwarding connection from 59748 > 37657 > 37657 in the container.
[117389 ms] Start: Run in container: /root/.vscode-server/bin/6cba118ac49a1b88332f312a8f67186f7f3c1643/node -e 
pre-commit installed at .git/hooks/pre-commit
[WARNING] The 'rev' field of repo 'https://github.com/compilerla/conventional-pre-commit' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.
[INFO] Initializing environment for https://github.com/compilerla/conventional-pre-commit.
...
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.
[119288 ms] [15:54:45] Extracted extension to /root/.vscode-server/extensions/.c2d52185-63c1-4349-ac91-40616d2fdbc8: ms-python.python
[119306 ms] [15:54:45] Renamed to /root/.vscode-server/extensions/ms-python.python-2021.10.1365161279
[119310 ms] [15:54:45] Extracting completed. ms-python.python
[119311 ms] [15:54:45] Extension installed successfully: ms-toolsai.jupyter-keymap
[119311 ms] [15:54:45] Extension installed successfully: ms-toolsai.jupyter-renderers
[119311 ms] [15:54:45] Extension installed successfully: ms-toolsai.jupyter
[15:54:45] Extension installed successfully: ms-python.python
[119316 ms] Extension 'ms-python.python' v2021.10.1365161279 was successfully installed.
[INFO] Initializing environment for https://github.com/psf/black.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
[...]

PR #173, opened 2021-11-04
Make sure locale directory is writable by calitp user

angela-tran:
Related to #171 

The [deploy to dev](https://github.com/cal-itp/benefits/actions/runs/1419184744) failed because the `benefits/locale/**` directories are not writable by the `calitp` user.

[Changes](https://github.com/cal-itp/benefits/pull/172/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L44) from #172 removed the explicit ownership-change on `benefits/locale` and moved the `USER $USER` switch to occur before we `COPY` the source files. We thought this would make `calitp` the owner of the app files/directories, but it turns out [COPY commands are run as root](https://stackoverflow.com/a/44766666/453168).

We actually decided it is better to have `root` be the owner, so in this pull request we just change the permissions for `benefits/locale/**`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,11 +39,18 @@ ENV PATH "$PATH:/home/$USER/.local/bin"
 COPY requirements.txt requirements.txt
 RUN pip install -r requirements.txt
 
-# copy source files
+# copy source files (as root)
 COPY gunicorn.conf.py gunicorn.conf.py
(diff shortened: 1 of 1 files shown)


angela-tran:
Closing this in favor of #174 (more correct branch name)

PR #174, opened 2021-11-04
Make sure locale directory is writable by calitp user

angela-tran:
Related to #171 

The [deploy to dev](https://github.com/cal-itp/benefits/actions/runs/1419184744) failed because the `benefits/locale/**` directories are not writable by the `calitp` user.

[Changes](https://github.com/cal-itp/benefits/pull/172/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L44) from #172 removed the explicit ownership-change on `benefits/locale` and moved the `USER $USER` switch to occur before we `COPY` the source files. We thought this would make `calitp` the owner of the app files/directories, but it turns out [COPY commands are run as root](https://stackoverflow.com/a/44766666/453168).

We actually decided it is better to have `root` be the owner, so in this pull request we just change the permissions for `benefits/locale/**`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,11 +26,25 @@ RUN useradd --create-home --shell /bin/bash $USER && \
     apt-get update && \
     apt-get install -qq --no-install-recommends gettext nginx
 
-# switch to non-root $USER
-USER $USER
-
 # enter app directory
 WORKDIR /home/$USER/app
+
+# copy config files
+COPY gunicorn.conf.py gunicorn.conf.py
+COPY manage.py manage.py
+
+# overwrite default nginx.conf
+COPY nginx.conf /etc/nginx/nginx.conf
+
+# copy source files
+COPY benefits/ benefits/
+COPY bin/ bin/
+
+# ensure $USER can compile messages in the locale directories
+RUN chmod -R 777 benefits/locale
+
+# switch to non-root $USER
+USER $USER
 
 # update PATH for local pip installs
 ENV PATH "$PATH:/home/$USER/.local/bin"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -39,12 +39,19 @@ ENV PATH "$PATH:/home/$USER/.local/bin"
...
-# copy source files
+# copy source files (as root)
 COPY gunicorn.conf.py gunicorn.conf.py
 COPY manage.py manage.py
 COPY benefits/ benefits/
 COPY bin/ bin/
 
+# ensure $USER can compile messages in the locale directories
+USER root
+RUN chmod -R 777 /home/$USER/app/benefits/locale
```
thekaveman, line 50: This change is definitely the one we need to be able to compile the localization messages!!

Confirmed by rebuilding the app container, and entering a `bash` prompt in the container and running the command to compile messages:

image

I wonder if we can find a way to clean this Dockerfile up though? Trying to avoid switching back and forth between `USER` while maintaining a good command ordering in terms of layer cachability.

I think all of these can go _before_ the change to the `calitp` user:

```
# enter app directory
WORKDIR /home/$USER/app

# copy config files
COPY gunicorn.conf.py gunicorn.conf.py
COPY manage.py manage.py

# overwrite default nginx.conf
COPY nginx.conf /etc/nginx/nginx.conf

# copy source files
COPY benefits/ benefits/
COPY bin/ bin/

# ensure $USER can compile messages in the locale directories
RUN chmod -R 777 /home/$USER/app/benefits/locale
```

Then switch, set `PATH`, copy/install requirements, and set entrypoint/command.

My guess is it is likely that the python dependencies change more frequently (e.g. a version update) than the app code at this point, so moving the `pip install` towards the bottom could actually improve our layer caching story?

thekaveman (changes requested):

Dockerfile
angela-tran: Thanks for that feedback! That all sounds correct to me, and I incorporated those changes.

Dockerfile
```diff
@@ -26,27 +26,32 @@ RUN useradd --create-home --shell /bin/bash $USER && \
...
-COPY requirements.txt requirements.txt
-RUN pip install -r requirements.txt
+# overwrite default nginx.conf
+COPY nginx.conf /etc/nginx/nginx.conf
 
 # copy source files
-COPY gunicorn.conf.py gunicorn.conf.py
-COPY manage.py manage.py
 COPY benefits/ benefits/
 COPY bin/ bin/
```
thekaveman, line 41: Since we're editing this file, want to swap the order of these `COPY` commands?

More likely that `benefits/` will change before `bin/`.

thekaveman (changes requested):
This is awesome! 💯 

One more minor change for (maybe?) caching!

thekaveman (approved):
🚀 🎸 🤖

Issue #178, opened 2021-11-10
Eligibility confirmation spec issue


machikoyasuda:
This spec: https://github.com/cal-itp/benefits/pull/170/files#diff-1637a357180023d868d2fca69d3d4e2a68f7b58c61aa71b016977ca95ee672fcR36 is currently set to `skip()` because it does not pass with the way the app is currently set up.

Previously, the Eligibility Server API had fake merchant auth token routes to accept calls from the `abc` transit agency in development mode, but these have since been removed - as of this PR https://github.com/cal-itp/eligibility-server/pull/31/files#diff-e5991fd559847cf92df779bd8b72643e9b64926355be788735b3c1cfcabca211L219

Now, developers cannot get the `eligibility/confirm` page to load with valid test credentials b/c the app looks for valid tokens before it loads the page. There are several ways we can address this:

1. ~Configure some sort of stubbing/mocking in the Cypress test to return fake auth tokens, so that Cypress can open the `eligibility/confirm` page without payment buttons. (This addresses the Cypress issue but not the local dev one).~
2. ~Change the way Django opens this page. https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L26 Have Django instead load the page and then asynchronously load the payment button. (This addresses both Cypress and local dev issue.)~
3. Create a fake Merchant Auth token so the page can load (and where would that code live and how would we do it?) (What we had in the past: https://github.com/cal-itp/benefits/issues/179)

thekaveman:
Closed by #229!

PR #191, opened 2021-11-17
Require confirmed eligibility for enrollment

### LINKED ISSUE #186: Enrollment page is accessible without passing eligibility verification
If a user knows the `/enrollment` endpoint, and has configured their session with a Transit Agency, they can navigate to the credit card enrollment page and completely bypass the EV check. This is a serious bug! 

## Steps to reproduce

1. Visit https://test-benefits.calitp.org
2. Click either of the Agency buttons
3. Manually update the URL to https://test-benefits.calitp.org/enrollment
4. Profit

## Remediation

This page must ensure the user has previously verified eligibility criteria

thekaveman:
## Overview

This fixes #186.

Add a new middleware definition similar to `AgencySessionRequired` that raises an exception when the `session` does not contain a verified eligibility.

Apply this middleware to the enrollment views instead of `AgencySessionRequired` (the new middleware effectively requires an agency in the session as well).

## Notes

We already had the [`session.eligible()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L92) helper for doing this check. This is being used in routing the [`/eligibility/confirm` form response](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L66). This PR builds on that helper to guard the `/enrollment` phase as well.

Additionally, UI test(s) would normally accompany this fix to ensure we don't introduce a regression later. However given the notes in #178 also apply to this situation, we don't currently have a way to load the `/enrollment` page in a fully stubbed/mocked environment.

### DIFF
--- a/benefits/core/middleware.py
+++ b/benefits/core/middleware.py
@@ -26,6 +26,17 @@ class AgencySessionRequired(MiddlewareMixin):
             raise AttributeError("Session not configured with agency")
 
 
+class EligibleSessionRequired(MiddlewareMixin):
+    """Middleware raises an exception for sessions lacking confirmed eligibility."""
+
+    def process_request(self, request):
+        if session.eligible(request):
+            logger.debug("Session has confirmed eligibility")
+            return None
+        else:
+            raise AttributeError("Session has no confirmed eligibility")
+
+
 class DebugSession(MiddlewareMixin):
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
This fix looks good to me.

I locally reproduced the bug and then tested the fix:

image

PR #192, opened 2021-11-18
Add Prettier VSCode extension

machikoyasuda:
## What this PR does
- Add the Prettier VS Code extension
- Set the Prettier extension to `format on save` for CSS, JS (.js, .json), but *not* HTML

## How to test this PR
- Run this container
- Save a JS (.js, .json) file and make sure it gets formatted (Example: Use single quotes instead of double quotes on a file and make sure it does get saved)
- Save a HTML file and make sure it does not get formatted (Example: Save the `alert-bar.html` file and make sure it does not get reformatted)
- Save a CSS file and make sure it does get formatted (Example: Put `h1, h2` on one line, and Prettier should put them on 2 separate lines)
- Test a Python file and make sure it does get properly formatted


.vscode/settings.json
```diff
@@ -1,12 +1,27 @@
...
+  },
   "files.encoding": "utf8",
   "files.eol": "\n",
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
+  "[javascript]": {
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  },
```
thekaveman, line 19: It doesn't work on JSON files for me without this here:

```json
"[json]": {
  "editor.formatOnSave": true,
  "editor.defaultFormatter": "esbenp.prettier-vscode"
},
```

Does it work for you with just the `[javascript]` entry?

Is there a way to specify multiple languages so we don't have to repeat the same settings for `[css]` etc.?

machikoyasuda: On the 2nd note, I tried 
```json
  "[css, javascript, json, html]": {
    "editor.formatOnSave": true,
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
```
 but it doesn't work

thekaveman: 💥 https://github.com/Microsoft/vscode/issues/51935#issuecomment-970436530

thekaveman: _yesterday_

machikoyasuda: Dangggg!

thekaveman (approved):
Good enough for now, we'll adjust as soon as that feature lands in the next VS Code release.

PR #195, opened 2021-11-18
Run UI automation tests for all PRs

### LINKED ISSUE #133: Automate cypress runs for PRs
#122 introduced an initial Integration/End-to-End test setup, but the only way to run tests right now is on localhost by hand.

We'd like to instead run the tests on each PR as a merge check in GitHub Actions. Similar to the work over in cal-itp/eligibility-server#16.

Basic outline is:

* `on` PR to any branch
* Start benefits container, see https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action
* Configure `CYPRESS_baseUrl` and other needed environment
* Run cypress tests, see https://github.com/cypress-io/github-action

thekaveman:
This PR reorganizes the `tests/e2e/cypress` directory in a couple of ways:

* Renames the top-level `cypress/integration` to `cypress/specs`
* Organize tests into two subdirectories based on their type:
  * `integration`: true integration tests that compose multiple systems
  * `ui`: front-end only page/HTML etc. specs

The workflow runs the specs defined in `cypress/specs/ui`, filtered [here in the workflow](https://github.com/cal-itp/benefits/commit/d2a2e1339a6daa863908ae85bd7c5b23e111b7c3#diff-dc16c5eb22367b9954bddcd2ac6aff54fa01f0c2e494749e05b98cd34731fc4dR46-R47):

```yml
spec: |
  cypress/specs/ui/*.spec.js
```

Partially closes #133, but we don't have `integration` tests running yet.


thekaveman:
Seems like this isn't working as expected: https://github.com/addnab/docker-run-action#run-an-image-built-by-a-previous-step

thekaveman:
Following up on my [above comment](https://github.com/cal-itp/benefits/pull/195#issuecomment-972426444): 

The reason we saw "Unable to find the image..." was because Docker was _building_ it, but not _loading_ it in the local images context. The change in a434517 ensures the image is loaded after building.

machikoyasuda (approved):
- Re-ran all tests locally
- Double-checked that docs do not need to be updated

tests/e2e/cypress.json
```diff
@@ -1,4 +1,5 @@
 {
+  "integrationFolder": "cypress/specs",
```
machikoyasuda, line 2: What is this for?

thekaveman: This is the property that Cypress uses as the base for searching for specs to run (relative to this `cypress.json` file).

By default the value is `cypress/integration`, in 57bb78f I changed it to `cypress/specs`. Here's more https://docs.cypress.io/guides/references/configuration#Folders-Files

thekaveman:
I updated our branch protection rules for `dev`, `test`, and `prod` to require these tests pass before merges are allowed:

image

PR #196, opened 2021-11-18
Deploy to test

thekaveman:


thekaveman:
@angela-tran @machikoyasuda can you all please run through https://dev-benefits.calitp.org to make sure that deploy looks good, then we'll push everything to `test`.

angela-tran (approved):
I tested out the following on https://dev-benefits.calitp.org:
 - verifying an eligible user
 - entering ineligible user
 - that JQuery is 3.6.0 and migration plugin is removed
 - `/enrollment` requires eligibility to be in the session

It all worked as expected.

machikoyasuda (approved):
Tested on dev!

PR #202, opened 2021-11-18
Add Lighthouse a11y testing to Test UI workflow

### LINKED ISSUE #167: Add automated Lighthouse accessibility testing for PRs
- Add GitHub action that runs LH on the app built for the PR
- Require Accessibility, Best Practices to pass
- Do the Performance test but don't require it

GitHub Action:
- Run on PRs that touch any `benefits/**` files
- Build Docker image
- Start app w/ Docker at specific port
- Run lighthouse at that specific port

machikoyasuda:
closes #167

## What this PR does 
- Add a step to the existing UI Test GitHub Action that runs `Lighthouse` testing on the homepage and the help page
- Run A11y and Best Practices, on Mobile
- The test will show a "Warn" when the overall Best Practices score is under 90%
- The test will show a blocking "Error" when the overall A11y score is under 100%
- The test results will be uploaded so they can be viewed easily, like this: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1637284608781-6224.report.html and the link can be found in the Action logs https://github.com/cal-itp/benefits/runs/4259113826?check_suite_focus=true#step:7:47 (The report only lasts for 60 days)
- The tests take about 30 seconds
image

## Important notes on Accessibility
- What Lighthouse's a11y test checks:
https://github.com/GoogleChrome/lighthouse/blob/v5.5.0/lighthouse-core/config/default-config.js#L427:L474
image

- What the test doesn't check, and needs to be manually tested:
image
https://developers.google.com/web/fundamentals/accessibility/how-to-review?utm_source=lighthouse&utm_medium=devtools

## Future improvements
- It would be nice to be able to be able to easily test a more complex page, like the form page. In the meantime, developers should be careful to run the Lighthouse a11y audits and do their own manual audits for forms and buttons.

### DIFF
--- a/.github/workflows/tests-ui.yml
+++ b/.github/workflows/tests-ui.yml
@@ -1,4 +1,4 @@
-name: UI automation tests
+name: UI automation and a11y Lighthouse tests
 
 on:
   pull_request:
@@ -46,7 +46,7 @@ jobs:
             -v ${{ github.workspace }}/fixtures:/home/calitp/app/fixtures \
             benefits_client:${{ github.sha }}
 
-      - name: Run tests
+      - name: Run UI automation tests
         uses: cypress-io/github-action@v2
         env:
           CYPRESS_baseUrl: http://localhost:8000
@@ -55,3 +55,9 @@ jobs:
           wait-on: http://localhost:8000/healthcheck
           spec: |
             cypress/specs/ui/*.spec.js
+
+      - name: Run Lighthouse tests for a11y
+        uses: treosh/lighthouse-ci-action@v8
+        with:
+          urls: http://localhost:8000
+          configPath: ".github/lighthouserc.json"
--- a/.github/workflows/tests-ui.yml
+++ b/.github/workflows/tests-ui.yml
@@ -59,5 +59,7 @@ jobs:
       - name: Run Lighthouse tests for a11y
         uses: treosh/lighthouse-ci-action@v8
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-ui.yml
```diff
@@ -1,4 +1,4 @@
-name: UI automation tests
+name: UI automation and a11y Lighthouse tests
```
thekaveman, line 1: This is a little long, can we come up with a simpler name that captures it all?

(See the very bottom): 
image

thekaveman: 👍

machikoyasuda:
Ready for final review @thekaveman

thekaveman (changes requested):
This looks awesome!! So cool you integrated with our existing UI test run!

I do have one piece of feedback on the "trying things out" type of commits. Those are great when you're not sure what is going to happen (esp. with GH Actions!), but can we please rebase to clean up the final PR before merging? I'd rather we don't clutter the history with those types of commits.

I did the exact same thing on #195 - had tons of commits changing one little thing, trying to make the workflow run. Once I got it, I went back and `squashed` those commits down into the "real" commits introducing "real" changes.

machikoyasuda:
@thekaveman Rebased :)

thekaveman (approved):
Amazing!

PR #210, opened 2021-11-19
Secure critical Django cookies

### LINKED ISSUE #201: Secure critical Django cookies
The following can be configured in [`settings.py`](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py).

Setting these ensures the cookies expire and are deleted when the user's browser closes (makes them session cookies)

* [x] [`CSRF_COOKIE_AGE`](https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-cookie-age) = None
* [x] ~[`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/3.2/ref/settings/#session-cookie-age) = None~
* [x] [`SESSION_EXPIRE_AT_BROWSER_CLOSE `](https://docs.djangoproject.com/en/3.2/ref/settings/#session-cookie-age) = True

We also want to prevent the CSRF cookie from being sent in cross-browser requests. E.g. a user clicks on a link to our site, from another site - in this case we don't want any previous CSRF cookie already in their browser sent. It's unlikely to be an issue in our app since there are no user logins etc. but for consistency with our other cookies (session and language).

* [x] [`CSRF_COOKIE_SAMESITE`](https://docs.djangoproject.com/en/3.2/ref/settings/#csrf-cookie-samesite) = "Strict"

angela-tran:
Fixes #201


thekaveman (changes requested):
This part looks awesome!

I just changed the requirements on you 😄 See updated #201, can we make one more setting change for the CSRF token?

thekaveman (approved):
Fantastic! 🍪 🔐

PR #213, opened 2021-11-19
Configure X-XSS-Protection

### LINKED ISSUE #204: Send X-XSS-Protection header
The X-XSS-Protection header can be used to manage certain browser's protection against reflected cross-site scripting (XSS), stopping a page from being loaded if an attack is detected. In modern browsers, the Content-Security-Policy header can provide better protection against XSS and setting X-XSS-Protection might be redundant (#203 tracks CSP implementation).

See more at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

We'll want the following header/value:

```
X-XSS-Protection: 1; mode=block
```

This can be done in a new Middleware and configured in [`settings.py`](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L45) for all requests/responses.

angela-tran:
Fixes #204 

[`SECURE_BROWSER_XSS_FILTER`](https://docs.djangoproject.com/en/3.2/ref/settings/#secure-browser-xss-filter) set to `True` will achieve the desired response header. That setting comes from [`SecurityMiddleware`](https://docs.djangoproject.com/en/3.2/ref/middleware/#x-xss-protection) which we are using in [`settings.py`](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L46), so all we need to do is set it.

## Testing this PR
 - Launch the benefits app
 - Open dev console (F12) and use the Network tab to ensure the response header is as expected (`X-XSS-Protection: 1; mode=block`)
 
image

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -75,6 +75,8 @@ if not DEBUG:
     CSRF_FAILURE_VIEW = "benefits.core.views.csrf_failure"
     SESSION_COOKIE_SECURE = True
 
+SECURE_BROWSER_XSS_FILTER = True
+
 ROOT_URLCONF = "benefits.urls"
 
 template_ctx_processors = [


thekaveman (approved):
Nice!! 💪 I didn't know this was built-in, even better 🚀

PR #216, opened 2021-11-20
Fix browser support in Content Security Policy

thekaveman:
Following up to #211.

Chrome understands the `script-src-elem` directive for JavaScript and `style-src-elem` for CSS. Firefox and Safari don't support either of these.

This PR falls back to the `script-src` and `style-src` directives for wider browser support.

More info: 

* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src-elem
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src-elem
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -209,7 +209,7 @@ CSP_FONT_SRC = ["https://california.azureedge.net/cdt/statetemplate/", "https://
 CSP_FRAME_ANCESTORS = ["'none'"]
 CSP_FRAME_SRC = ["'none'"]
 
-CSP_SCRIPT_SRC_ELEM = [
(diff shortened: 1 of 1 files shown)


angela-tran (approved):
👍 👍 Works for me on FF now

thekaveman:
Thank you for catching this @angela-tran!! 🙏

PR #229, opened 2021-11-24
Delay acquisition of Enrollment access token

thekaveman:
## Background

When a user successfully verifies their eligibility, we redirect the user into the `enrollment` flow. The first step of that flow (on the server) is to acquire an API access token for the Payment Processor's API.

This presents a number of problems, such as that documented in #178 - but also e.g. we lose the ability to give the user a more nuanced error message if their eligibility _was verified_ but the initial connection to the Payment Processor / API token failed.

## What this PR does

This PR delays the request for the API access token until the page has loaded in the user's browser. By introducing a simple `JsonResponse` endpoint and moving the existing code around, it was a minor change to `$.get()` the token async via JavaScript rather than in the Django request.

### DIFF
--- a/benefits/core/viewmodels.py
+++ b/benefits/core/viewmodels.py
@@ -208,16 +208,16 @@ class PaymentProcessor:
     """
     Represents a core.models.PaymentProcessor:
     * model: core.models.PaymentProcessor
-    * access_token: str
+    * access_token_url: str
     * element_id: str
     * color: str
     * [name: str]
     * [loading_text: str]
     """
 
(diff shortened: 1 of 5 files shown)


machikoyasuda (changes requested):
One small change - You can remove the `skip` on `eligibility.spec.js`:

image

And tests will pass!

angela-tran (approved):
I think this approach is much better! Looks good.

machikoyasuda (approved):

Issue #197, opened 2021-11-18
Add MST Courtesy Card Option and Document Info


vykster:
As a rider with a MST Courtesy Card, I want the option to sign up with a MST Courtesy Card so I can enroll in the contactless fare-discount demonstration. 

**Acceptance Criteria**
- [x] Riders can select senior or MST CC discount options. 
- [x] When a rider selects “Senior Discount Program ” and clicks “continue” they will be redirected to the senior discount document page
- [x] When a rider selects “MST Courtesy Cardholder” and clicks “continue” they will be redirected to the courtesy cardholder discount document page. 

**Resources**

- [Figma Large_Eligibility01](https://www.figma.com/file/5IabSdY0xDzwL4gyXhbQ9G/Benefits-(Courtesy-Cards)?node-id=1618%3A4627)
- [Large_Eligibility1 MSTCC](https://www.figma.com/file/5IabSdY0xDzwL4gyXhbQ9G/Benefits-(Courtesy-Cards)?node-id=1618%3A4615) 
- [MST Courtesy Card Holder User Stories Google Doc](https://docs.google.com/document/d/1wxEY4VsxOqqP7FydVXbW9NEzpLznnjpB_IedtzCAHe0/edit#)

vykster:
This story is up to date with current screens linked all changes from MST feedback have been incorporated into screens.

thekaveman:
Closed with the work in #317

PR #230, opened 2021-11-29
[WIP] Refactor data model for more flexibility

thekaveman:
⚠️This is major WIP! ⚠️

## Overview

The main idea here is to refactor what we have defined as [`viewmodels`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/viewmodels.py) into proper Django database models.

The problem we need to solve is that of the hard-coded flows currently found in `views.py` across [`core`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/views.py), [`eligibility`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py) and [`enrollment`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py).

At present these views are responsible for composing what becomes the UI (text content, forms, buttons, etc.) to send to the template for rendering. We'd like that composition to be specified in the configuration data, to allow the views to make decisions and act on the configuration without having to necessarily "know" what any given configuration is/says.

The end result should be a more flexible system that can support e.g. multiple pathways each with its own eligibility form/verification flow.

## End-user text content

A huge implication of this refactor is we need a new way to handle translated text. Currently, python code uses Django's `ugettext` helper to query the `.po` language files based on the `msgid`. [E.g. for the index content title](https://github.com/cal-itp/benefits/blob/dev/benefits/core/views.py#L20):

```python
from django.utils.translation import ugettext as _

# ...

def _index_content_title():
    """Helper returns the content title for the common index page."""
    return _("core.index.content_title")
```

This PR moves most (all?) of that text content into Django's database / model system. The idea is we can leverage [`django-modeltranslation`](https://django-modeltranslation.readthedocs.io/en/latest/) to help us store/query translated text directly to/from the models.

## Open questions

1. How to handle forms in this new paradigm? Right now we only have a few (one for Eligibility, two for Enrollment) and they are "hardcoded" as Django form classes; we anticipate at least one more form to support a different Eligibility flow (see #197)
2. How do the views interact with these new models? E.g. if a user navigates to the homepage, how does the model get pulled out of the database and sent to the template for rendering?
3. Related to the above, can we send these models directly to the template for rendering? Are there any gotchas?
4. What if any text content _doesn't_ move into the Django database / model system? Do we still need the `.po` files?


thekaveman:
On _Outstanding Question 2_ above, here's one (very loose) idea:

* Each `Page` model stores its own permalink, e.g. for the index page we store `/`, for the Help page, we store `/help`.
* A path converter similar to [`TransitAgencyPathConverter`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/urls.py#L14) looks at the incoming request URL, and tries to find a `Page` model in the DB with a corresponding permalink attribute
* A catch-all view method takes the found `Page` model and does whatever needs to happen to render the template

thekaveman:
@angela-tran I think you mentioned class-based views... RE: _Open questions 2 & 3_ above, this seems relevant: https://docs.djangoproject.com/en/3.2/ref/class-based-views/generic-display/#detailview; works similarly to what I described above, just you know, out of the box with Django 🤣

thekaveman:
Rebased on `dev` and added the initial Spanish language translations to new fixture file.

thekaveman:
We decided this approach is more than we need right now.

PR #233, opened 2021-12-01
Introduce explicit order to fixture files

thekaveman:
## What this PR does

Simple rename of the `fixtures/` files - to add a sequence number on the front of each file. This way we can safely use a glob pattern to load all the files at once:

```bash
python manage.py loaddata fixtures/??_*.json
```

while ensuring the load order is always correct to maintain referential integrity (e.g. `EligibilityType` is loaded before `EligibilityVerifier` since the latter has a foreign key reference to the former).


angela-tran (approved):
Changes look good 👍

machikoyasuda (approved):

PR #234, opened 2021-12-02
Refactor CSS: Reduce number of css file calls

machikoyasuda:
closes part of #183

## What this PR does
- Add 2 css classes from [oceanside](https://github.com/Office-of-Digital-Innovation/California-State-Template/blob/master/css/colorscheme-oceanside.css) into benefits/static/css/styles.css
- Remove oceanside.css call from base template

## Why?
- Fewer CSS files to load ==> Lighter page weight and faster web performance

## How I tested this PR
- I manually visually compared all the pages on the site, in both Desktop and Mobile, side by side with this branch and production:
image
image
- Check focus, hover states, check `Skip to Main Content` button visually

## Performance testing

This PR:
image
image

Issue #183, opened 2021-11-16
Performance Lighthouse CI recommendations


machikoyasuda:
This list is up for discussion.

## Performance
- Ensure text remains visible during font load https://web.dev/font-display/?utm_source=lighthouse&utm_medium=devtools (And/or preload Google fonts https://web.dev/preload-optional-fonts/?utm_source=lighthouse&utm_medium=devtools)
- Use WebP instead of JPEGs, PNGs:
![image](https://user-images.githubusercontent.com/3673236/142042190-817fb474-9c71-4acc-896b-8d8d2be811d1.png)
- Eliminate blocking resources
![image](https://user-images.githubusercontent.com/3673236/142042278-51492b1b-e977-4543-a154-f49c610fe5ef.png)

PR #235, opened 2021-12-02
Refactor JS

machikoyasuda:
closes part of #183

## What this PR does
- Remove the California state JS template entirely.
- I _believe_ this was only used to render the up arrow that appears once you scroll. The Benefits app does not need this.

## How to test this PR
- Go through the app in desktop and mobile and look for any regressions

## Performance testing

Shows marked improvement in speed index and first contentful paint from previous PR https://github.com/cal-itp/benefits/pull/234
- Help page
image

- Home page - The image slows it down
image

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -213,7 +213,6 @@ CSP_FRAME_SRC = ["'none'"]
 
 CSP_SCRIPT_SRC = [
     "'unsafe-inline'",
-    "https://california.azureedge.net/cdt/statetemplate/",
     "https://cdn.amplitude.com/libs/",
     "https://code.jquery.com/",
     "*.littlepay.com",
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -206,7 +206,7 @@ CSP_DEFAULT_SRC = ["'self'"]
 
 CSP_CONNECT_SRC = ["'self'", "https://api.amplitude.com/"]
 
-CSP_FONT_SRC = ["https://california.azureedge.net/cdt/statetemplate/", "https://fonts.gstatic.com/"]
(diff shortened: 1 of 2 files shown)


machikoyasuda:
The code in this PR is very simple, but I really want to be sure this JS file wasn't doing anything else for the Benefits app -- aside from adding the scroll-up arrow code. @thekaveman - Do you know what else, if any, it was doing?

thekaveman (changes requested):
This is great - I don't think we need to worry about any implications, we weren't using any of the State Template components - like their navigation menus, accordions, etc. (similar to Bootstrap's components that require their JS). Except for that one "back to top" arrow which is a little silly...

So we can actually tighten up the Content Security Policy a bit, by removing the state template entry from `script-src` directive: https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L216 since we don't need to load scripts from them anymore!

benefits/settings.py
```diff
@@ -213,7 +213,6 @@
 
 CSP_SCRIPT_SRC = [
     "'unsafe-inline'",
-    "https://california.azureedge.net/cdt/statetemplate/",
```
thekaveman, line 216: Perfect! Since this JS file we are removing is responsible for loading State Template fonts, and we don't need those anymore, can we also remove the State Template entry from `CSP_FONT_SRC` please?

thekaveman (changes requested):

thekaveman (commented):
I don't think the `CSP_FONT_SRC` was changed?

thekaveman:
Ah, I was wrong - the fonts are loaded by the State Template CSS file too. So we still need the `CSP_FONT_SRC` settings as it was before. Sorry about that!

thekaveman (approved):

PR #238, opened 2021-12-02
Refactor: Use WebP instead of PNGs

machikoyasuda:
closes part of #183 

This PR is the last of 3 PRs that address performance fixes and refactors for Benefits.

## To Do
- Test in iOS Safari
- Test in Safari
- Test in all other browsers (Android)
- Update image file

### Styles and images
- Replace a PNG in the background with a WebP one set as an ``, while preserving the CSS
- Adds a CSS class with the same CSS, applied to the ``, to replace the in-line styles
- Change the PNG on the Success! page to the same exact WebP image
- Adds corresponding `heights` to this -- having both a width and height lessens the chance of elements shifting around when the page loads
image

## How to test this PR

1. The main image in Desktop and Mobile - Ensure that this image scales up and down correctly
2. Ensure there is an alt tag on the image


benefits/core/templates/core/page.html
```diff
@@ -13,7 +13,8 @@
     <div class="row">
         {% if page.image %}
         {% load static %}
-        <div class="col-lg-6 image" style="background: 48% 75% url('{% static page.image.src %}') no-repeat; background-size: cover;">
+        <div class="col-lg-6 image">
+            <img src="{% static page.image.src %}" class="core-image" alt="Rider tapping a debit card to pay for their trip" >
```
thekaveman, line 17: Maybe this commit didn't make it up yet?

Alt tag needs to be `{% translate "core.index.image" context "image alt text" %}`.

Same as how it's done on the Enrollment success image.

thekaveman (changes requested):
See my one comment on the alt text.

Maybe it's just me, but this image definitely feels grainier / less high-quality 🤷

machikoyasuda:
@thekaveman I can remake the WebP photo file to raise the quality a bit.

machikoyasuda:
Closing for the time being b/c of lack of universal WebP support

PR #244, opened 2021-12-04
Cache headers for static files

thekaveman:
Introduce caching at the `nginx` layer for static file requests (images, CSS, JS, etc.)

Use a static file storage mechanism in Django that silently rewrites file names with a hash of their contents, allowing long cache times and easy cache busting when the contents changes.

See the new headers for `/static`/ files:

image

**NOTE:** when running locally, to get `nginx` to run, you have to have to `docker compose up client` _outside_ the devcontainer. `F5` inside the devcontainer uses the Django debug server.

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -167,6 +167,7 @@ USE_TZ = True
 
 STATIC_URL = "/static/"
 STATICFILES_DIRS = [os.path.join(BASE_DIR, "benefits", "static")]
+STATICFILES_STORAGE = "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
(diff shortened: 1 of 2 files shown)


machikoyasuda:
I am running this locally outside the devcontainer and getting an interesting situation:

image

image

Nginx is returning an old version of the base.html page I was working on in a previous PR. It is trying to fetch a webp image too. I've triple-checked I'm in the correct branch, tried different browsers, and tried turning Docker off and on again. Not sure how I got here!

image

machikoyasuda:
The old CA js file even shows up!

image

machikoyasuda:
Figured it out - had to rebuild the docker container

machikoyasuda (approved):

PR #254, opened 2021-12-07
Fix: more error handling for duplicate enrollment

thekaveman:
The 500 response is used to indicate a duplicate group enrollment with the payment processor. 

In order to determine if a 500 is a a duplicate registration or a "real" server error, we need to ensure the `customer_id` exists in the error details coming back with the 500.


angela-tran (approved):
👍 👍

PR #257, opened 2021-12-08
Fix: enrollment flow for existing customers

thekaveman:
When a customer's card has already been seen by the payment processor (e.g. they have tapped on a device), before registering through `benefits`, the payment processor considers the customer account as _unregistered_ (e.g. unregistered for any discounts) - although the account does exists.

Our current implementation handled this case, by registering any unregistered customers.

However, there is a mismatch in some of the API docs and what happens during real API calls. It turns out that when updating a customer, we need to include the customer's ID in the request payload (in addition to the URL fragment) in order to get it back in the response payload.

Instead of sending `{"is_registered: "true"}` we need to send `{"is_registered: "true", "id": "UUID"}`.

This PR adds the `id` field to customer update calls, and adds some additional checks on the response to raise errors earlier when responses don't look like we expect.

### DIFF
--- a/benefits/enrollment/api.py
+++ b/benefits/enrollment/api.py
@@ -193,7 +193,7 @@ class Client:
             raise ValueError("customer_id")
 
         url = self._make_url(self.payment_processor.customer_endpoint, customer_id)
-        payload = {"is_registered": True}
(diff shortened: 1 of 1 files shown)


angela-tran (approved):
Looks good!

PR #259, opened 2021-12-08
Add workflow to monitor API

### LINKED ISSUE #252: Monitor status of access to payment processor API
We should create something (a script / GitHub action) that will alert us when our access to the payment processor API is broken. This will help us address issues with that integration proactively so that deployments go smoothly.

angela-tran:
Fixes #252 

## How it works
This workflow will alert us if our access to the payment processor is broken. There are two ways it can be triggered:
 - via the configured schedule (set to run at 12:00 UTC)
 - manually through the Actions tab

Secrets need to be configured for this to work. For the cert-file secrets, the expectation is that they are base64-encoded strings.

The general idea is:
 - decode the strings back into files
 - build up the `curl` command to call the API endpoint, with the `-i` flag to include the HTTP header in the output
 - check that the response code is what we expect

Note that we are **not** printing the payload or anything about the certs/endpoint to the console.

## Workflow workarounds
I got very familiar with the GitHub Actions documentation while working on this workflow. 😂 Here are some notes about workarounds I had to do:
 - `jobs.<job_id>.if` does [not support](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability) using the `matrix` context, so I had to unfortunately duplicate the `if` condition on the steps. I used a job-level `env` variable though to reduce the duplication.
 - I looked into using the `jobs.<job_id>.environment` parameter as a way of reusing names for Secrets (i.e. create a GitHub Environment for each API we would want to check, and use generic secret names like `CERT`, `KEY`, `URL`, etc.). There were a few undesirable things about it though, so I went with using [custom variables in the matrix](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#using-environment-variables-in-a-matrix). This does mean though that there will be `(# of APIs) * (# of secrets per API)` secrets names.
    - Undesirable things about using Environments:
       - "Environments" are tied to "Deployment" objects, which will show up in the sidebar of the repo and keep a record. (see [docs](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#how-environments-relate-to-deployments)) I don't think we'd want this API check to be mixed in with our actual deployments: https://github.com/cal-itp/benefits/deployments
       - I also could not find a way to read the job's `environment` value, which I wanted to get for the `if` condition - for example, if the user triggering a manual run selects `qa`, the job/step should check if it is running with `environment: qa` and run or skip accordingly.
 - It seems that `matrix` does not support using expressions, so I used this trick with array notation: https://github.community/t/secrets-in-matrix/17307

### DIFF
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -28,12 +28,14 @@ jobs:
             key: API_CHECK_PROD_KEY
             ca-cert: API_CHECK_PROD_CA_CERT
             url: API_CHECK_PROD_URL
+            data: API_CHECK_PROD_DATA
 
           - name: qa
             cert: API_CHECK_QA_CERT
             key: API_CHECK_QA_KEY
             ca-cert: API_CHECK_QA_CA_CERT
             url: API_CHECK_QA_URL
+            data: API_CHECK_QA_DATA
 
     name: Check API endpoint (${{ matrix.name }})
     steps:
@@ -58,7 +60,7 @@ jobs:
           curl -i --url ${{ secrets[matrix.url] }} \
           --header 'Accept: application/json' \
           --header 'Content-type: application/json' \
-          --data '{"request_access":"CARD_TOKENISATION"}' \
+          --data '${{ matrix.data }}' \
           --cert $temp_dir/cert.pem \
           --key $temp_dir/key.pem \
           --cacert $temp_dir/cacert.ca > $temp_dir/payload.txt
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -1,0 +1,66 @@
+name: Check access to API
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        type: choice
+        description: Select the API environment
+        options: [all, prod, qa]
+  schedule:
+    - cron: "0 12 * * *"
+
+jobs:
+  check-api:
+    runs-on: ubuntu-latest
+    env:
+      SHOULD_RUN: |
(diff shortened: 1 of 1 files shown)


thekaveman:
@angela-tran still digesting all this 😅 

On environments / secrets: you could definitely re-use our existing environments. E.g. we can use our `test` to store secrets for `qa` here, and `prod` for `prod`.

.github/workflows/check-api.yml
```diff
@@ -0,0 +1,66 @@
...
+          echo ${{ secrets[matrix.ca-cert] }} | base64 -d > $temp_dir/cacert.ca
+
+      - name: Call API endpoint
+        if: contains(env.SHOULD_RUN, 'true')
+        run: |
+          temp_dir=$RUNNER_TEMP/${{ matrix.name }}
+          curl -i --url ${{ secrets[matrix.url] }} \
+          --header 'Accept: application/json' \
+          --header 'Content-type: application/json' \
+          --data '{"request_access":"CARD_TOKENISATION"}' \
```
thekaveman, line 61: Let's also pass this payload in as a secret. Not as sensitive, but easier to change that way.

thekaveman (changes requested):
Other than the one small change requested, this is a great approach!

In the deploy workflows, this is how we brought in environment-specific secrets to the jobs: https://github.com/cal-itp/benefits/blob/dev/.github/workflows/deploy-dev.yml#L25. But I think the workarounds here for the environments, given how tied they are to deployments, makes a lot of sense!

thekaveman (approved):

PR #261, opened 2021-12-09
Simplify API check's usage of cert files

angela-tran:
It is not necessary to base64-encode the cert files as a single string.
The workflow will now read the secret, which is expected to be simply the file contents, newlines and all).

### DIFF
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -49,9 +49,18 @@ jobs:
         run: |
           mkdir $RUNNER_TEMP/${{ matrix.name }}
           temp_dir=$RUNNER_TEMP/${{ matrix.name }}
-          echo ${{ secrets[matrix.cert] }} | base64 -d > $temp_dir/cert.pem
-          echo ${{ secrets[matrix.key] }} | base64 -d > $temp_dir/key.pem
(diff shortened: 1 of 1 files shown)


angela-tran:
Oh, good idea with using that branch dropdown. That is nice for testing 👍

angela-tran:
Thanks for catching those errors @thekaveman. I also realized the code to read the event name was slightly wrong (which is why last night's scheduled run just skipped the job steps) and fixed that.

Here's the first successful run: https://github.com/cal-itp/benefits/actions/runs/1561037010 🎉

.github/workflows/check-api.yml
```diff
@@ -41,17 +41,26 @@ jobs:
     steps:
       - name: Echo workflow run information
         run: |
-          echo "Triggering event name: \"${{ github.event.name }}\", \
+          echo "Triggering event name: \"${{ github.event_name }}\", \
           APIs to check: \"${{ github.event.inputs.environment }}\""
```
thekaveman, line 45: Let's get rid of the `\"` from the messages here, it clutters up the log message in the Action run, and we already have the `:` separator to distinguish the message label from the value.

thekaveman (changes requested):
Tiny stylistic change, but otherwise looks good!

thekaveman (approved):

PR #265, opened 2021-12-15
[WIP] Refactor models - incorporate DetailsView

angela-tran:
⚠️This is also a major WIP! ⚠️

This branch is based off the [`refactor/models`](https://github.com/cal-itp/benefits/tree/refactor/models) branch. The goal here was to do some exploratory work on what it would look like to use Django's class-based views (`DetailsView`) with our `Page` model.

### Observations
 - The templates required quite a bit of updating in order to work with the new UI-element Django models.
    - The `ManyToManyFields` will end up as `ManyRelatedManager` objects in the templates, so we have to call `all` in order to get an iterable.
    - The images required a bit more work to build up the right src URL.
 - We have introduced a coupling between `Page`s and agencies via the `slug` property. In other words, if you add a new agency, you must add a page with a matching slug. 
    - Another way to look at it is: it's not as easy now to just use arbitrary test data (e.g. our `abc` and `deftl` transit agency objects). This is why I updated the slugs in our development fixture data to use `mst` and `sacrt`.
 - I needed to add a few fields to the models to get some missing UI pieces - `Page.navigation_title` for the label above buttons, `PageNavigation.classes` to add styling for buttons.

### Missing pieces/open questions
 - Prior to this refactor, each function in `views.py` knew what it should update on the session. In this refactor, each `Page` is analogous to each of those functions, but currently we don't have a way to specify any interaction with the session.
    - We also previously made use of middleware decorators on the view functions. We don't have an equivalent to this currently.
 - How to handle the usages of `django.urls.reverse`?

Also, I limited this exploration to the `core` app, so I didn't work on adding `form` references to `Page`s.

### Trying it out
With these WIP changes, the app is not in a fully working state, but you can start it up and see the `Page` models being rendered. You'll eventually get to the point of seeing `Session not configured with agency` on the `/eligibility` page.

Make sure you start with a clean database
```
rm django.db
./bin/init.sh
```
and clean session (incognito window, for example).

Please excuse the auto-generated migration (I assume it's ok since this is a WIP).


thekaveman:
We decided this approach is more than we need right now.

PR #275, opened 2022-01-04
Chore: reorganize Cypress tests

thekaveman:
This PR makes a minor reorganization of the Cypress test directory to simplify the folder structure and more accurately describe the contents:

* Change the top-level directory from `tests/e2e` to `tests/cypress`; since we don't like using that ambiguous term "e2e"
* Elevate the `fixtures` and `specs` directories to the top-level; there's no reason for an inner `tests/cypress/cypress` directory

The `cypress.json` file and other files that reference these paths (e.g. GitHub Actions) were also updated.


angela-tran (approved):
This reorganization looks good to me

machikoyasuda (approved):
- I rebuilt the app entirely and reinstalled Cypress.
- I ran the app and ran through the main flow
- I ran all the tests
Looks great! I like that we are taking away the name `e2e`.

PR #281, opened 2022-01-06
Tokenize response only for expected statuses

### LINKED ISSUE #279: Better HTTP response code handling in Eligibility API
In the Eligibility API `Client`, we currently attempt to parse the response as if it were a JWT, regardless of response code:

From https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/api.py#L145

```python
try:
    r = requests.get(self.verifier.api_url, headers=self._auth_headers(token))
except requests.ConnectionError:
    raise ApiError("Connection to verification server failed")
except requests.Timeout:
    raise ApiError("Connection to verification server timed out")
except requests.TooManyRedirects:
    raise ApiError("Too many redirects to verification server")
except requests.HTTPError as e:
    raise ApiError(e)

return self._tokenize_response(r)
```

Since input errors on the form are returned as JWTs, the same as success payloads, this code worked fine for 200 and 400 responses. But if the API outright rejects the call with a 403, the above code attempts to parse _that_ response as a JWT, throwing an unhandled exception.

Let's guard the `return self._tokenize_response(r)` to ensure we are only trying to tokenize the expected 200 and 400 responses; other codes should raise an `ApiError`.

angela-tran:
Fixes #279 

### Local testing
I modified my local `eligibility-server` instance to return a 403 for the `/verify` endpoint. Then, I modified the `benefits` compose.yml file so that it would point at `eligibility_server:latest` for the `server` service. I rebuilt all the images and then launched `benefits` with the debugger.

My hard-coded 403 was returned and raised an `ApiError` as expected.

Screenshots:
image

image

### DIFF
--- a/benefits/eligibility/api.py
+++ b/benefits/eligibility/api.py
@@ -154,8 +154,13 @@ class Client:
         except requests.HTTPError as e:
             raise ApiError(e)
 
-        logger.debug("Process eligiblity verification response")
-        return self._tokenize_response(r)
+        expected_status_codes = {200, 400}
+        if r.status_code in expected_status_codes:
+            logger.debug("Process eligiblity verification response")
+            return self._tokenize_response(r)
+        else:
+            logger.info(f"Unexpected eligibility verification response status code: {r.status_code}")
+            raise ApiError("Unexpected eligibility verification response")
 
     def verify(self, sub, name):
         """Check eligibility for the subject and name."""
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
This is great! 🚀 Nice usage of the set literal syntax!

thekaveman:
Sorry, one question:

What do you think about `logger.warning()` for the case when status code is unexpected? That seems important to know about.

angela-tran:
> Sorry, one question:
> 
> What do you think about `logger.warning()` for the case when status code is unexpected? That seems important to know about.

I was definitely going back and forth on that myself. No objections here on using `warning` - I'll go ahead and make that change.

Issue #272, opened 2021-12-21
Text updates for /payment-options (will need Spanish translation)


esquared415:
Please make the following changes on the /payment-options page, as requested by Gillian and Visa (which flagged that Contactless Symbol is a registered trademark and not the symbol we have been showing on this page, which is known as a Contactless Indicator; I have softened this language so we can keep everything lowercase and not use the capitalized marks).

**1. First paragraph**
_Please change **from**:_
Cal-ITP is a new program serving all Californians. In order to activate your discount, you need a bank card that has the contactless symbol on it. The contactless symbol is four curved lines and it looks like this:

_**to:**_
Cal-ITP is a new program serving all Californians. In order to activate your discount, you need a contactless-enabled bank card. Cards that are contactless-enabled have the following symbol (four curved lines) on the front or the back of them:

**2. Second-to-last paragraph** 
_Please change **from**:_
We are working to add additional options for people who do not have access to a bank card. Please check back on this website, or get in touch with your local transit provider.

_**to:**_
Don't have access to a bank card? You can request a contactless card from one of the companies that offer free contactless prepaid debit cards, such as the [Cash App Visa Debit Card](https://cash.app/help/us/en-us/14425-cal-transit) or the [Venmo Mastercard Debit Card](https://venmo.com/about/debitcard/).

_Please note: These changes will need to be translated into Spanish (MST should be able to support; @o-ram can coordinate this ask unless there's another translation solve available)._

o-ram:
Spanish translations @thekaveman 

**Change 1:** Cal-ITP es un nuevo programa que sirve a todos los Californianos.Para activar su descuento, necesita una tarjeta bancaria habilitada sin contacto. Las tarjetas habilitadas sin contacto tienen el siguiente símbolo (cuatro líneas curvas) en la parte delantera o trasera de la tarjeta.

**Change 2**: ¿No tienes acceso a una tarjeta bancaria? Puede solicitar una tarjeta sin contacto a una de las compañías que ofrecen tarjetas de débito prepagas sin contacto gratuitas, como la tarjeta de [Cash App Visa Debit Card](https://cash.app/help/us/en-us/14425-cal-transit) o la [Venmo Mastercard Debit Card](https://venmo.com/about/debitcard/)

thekaveman:
This is not a straightforward text change, as the new copy introduces hyperlinks where there were none previously. This will take some time to figure out how to implement cleanly.

PR #290, opened 2022-01-18
Release 2022.01.1

thekaveman:


machikoyasuda (approved):
- Ran through successful and unsuccessful flows to completion
- Checked copy changes from https://github.com/cal-itp/benefits/issues/272

angela-tran (approved):
Ran through the entire flow here too 👍

PR #292, opened 2022-01-24
Refactor: read custom Content Security Policy config from env

thekaveman:
Makes updates to deployed apps easier, since we can just update the environment and restart - vs. having to change the code and redeploy. 

⚠️ Update your local `.env` file with the new settings. Add the values that were previously hard-coded in `settings.py` using the new environment variables introduced in this PR.


machikoyasuda:
[@thekaveman Lighthouse Best Practice test score is coming up short ](https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1643057744369-18541.report.html#best-practices)

machikoyasuda (changes requested):
Got this working locally and ran thru the app successfully 👍 

`.env.test` will have to be updated so that Lighthouse/Cypress tests can be run in the correct environment.

angela-tran (approved):
These changes look good to me 👍 Definitely makes it easier to update those settings.

Props to @machikoyasuda for mentioning the Lighthouse errors!

machikoyasuda (approved):

PR #293, opened 2022-01-25
Feat: Add rate limiting

thekaveman:
Simple rate limiting feature can be applied on per-view basis with a function decorator:

## What this PR does
```python
# in views.py

from django.utils.decorators import decorator_from_middleware
from benefits.core.middleware import RateLimit

@decorator_from_middleware(RateLimit)
def view_func(request):
    # etc.
```

Requires configuring new `.env` settings:

```env
# how many requests to allow per time period
DJANGO_RATE_LIMIT=5

# the request methods to rate limit
DJANGO_RATE_LIMIT_METHODS=GET,POST,PUT,DELETE

# the amount of time until the rate limit resets, in seconds
DJANGO_RATE_LIMIT_PERIOD=30
```

## How to run this PR locally
- Remember to get the latest .env file

## How to test this PR
- Make sure a _normal, real, human_ user can go through all the regular flows (eligible user, ineligible user, format error validation, Spanish language)
- Testing the rate limit: Simulating a rate limit exceeds scenario on the eligibility confirmation form page should take the user to the rate limit exceeds error page


thekaveman:
Rebased on `dev` after merging CSP changes in #292 to resolve conflicts.

angela-tran:
### Notes on the CI tests
In 63a1f6d, I set DEBUG to true so that we would avoid a 500 error.
After pushing 8a3e937, the CI tests passed, but Lighthouse tests [failed on accessibility](https://github.com/cal-itp/benefits/actions/runs/1753086410) because of the big red debug banner.

I investigated what was causing the original 500 error when DEBUG is false (which is our preferred setting anyways) and found it was from the changes made in ed9d97ec and 4f52866 where the `beforeEach` was changed to use `visit` instead of `click`-ing through each page.

[`cypress.visit`](https://docs.cypress.io/api/commands/visit#Requirements) will resolve when the [`load` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) fires.  Our [.env.tests](https://github.com/cal-itp/benefits/blob/76d2fe8a70cc023bfd16793810c1b54ba901b104/tests/cypress/.env.tests) leaves out `'https://cdn.amplitude.com/libs/`so we get a 500 for that resource.

The options are to switch back to using `click` or to add amplitude to `.env.tests`. I went with switching back to `click`, and the tests pass now: https://github.com/cal-itp/benefits/runs/4959215945?check_suite_focus=true

thekaveman:
@angela-tran thanks for that explanation and good find!

Avoiding the `.click()` is good because we want (need?) to be able to test parts of the app without having to go through the entire user flow as a real user would. Init the session by `cy.visit("/")` and then being able to `cy.visit()` the page we care about directly is useful.

I think it's OK to have the amplitude URLs configured in `.env.tests`. I purposefully left it out since I didn't think we need it, but it seems like we do to get the site to load correctly, so that's fine. The good news is, we don't configure an amplitude API key in CI, so we don't actually send bogus events up.

angela-tran:
The attempt to use `cy.visit` didn't work, so we'll just go with `cy.click` for now. 😣

angela-tran (approved):

machikoyasuda (approved):

PR #297, opened 2022-01-26
Feat: Show error page when rate limit is reached

machikoyasuda:
Note: This PR merges into the main `feat/rate-limit` PR.

## What this PR does
- Adds code to the `RateLimit` middleware mixin method that uses `viewmodels.ErrorPage` to return a 400 bad request error and this page

## Feature
Adds a small feature where: When a user reaches the rate limit on the eligibility confirmation page (by exceeding the number of accepted requests per minute), the user will be taken to this error page with a `Return home` button. The `Return home` button will take them back to the `/eligibility/` page

image

### DIFF
--- a/benefits/core/middleware.py
+++ b/benefits/core/middleware.py
@@ -5,12 +5,13 @@ import logging
 import time
 
 from django.http import HttpResponse, HttpResponseBadRequest
+from django.template import loader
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
👍 👍

PR #326, opened 2022-03-16
TransitAgency model allows for multiple verifiers

### LINKED ISSUE #320: Update TransitAgency model to ref many EligibilityVerifier
## Background

This is the central change outlined in #317, here scoped to allow the model refactor to happen in a non-breaking way that can be built upon in subsequent tasks.

## Tasks

- [x] Convert the [`eligibility_verifier` field](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L120) from the single  `ForeignKey` to a `ManyToManyField` and rename it to the plural `eligibility_verifiers`
- [x] Update the [initial migration](https://github.com/cal-itp/benefits/blob/dev/benefits/core/migrations/0001_initial.py#L84) since we rebuild the database from scratch each deploy
- [x] Update the [fixtures](https://github.com/cal-itp/benefits/blob/dev/fixtures/04_transitagency.json) to rename the field and wrap the existing value in a list `[]`
- [x] Update usages of the `TransitAgency.eligibility_verifier` attribute in python code to `TransitAgency.eligibility_verifiers[0]` (for now)
- [x] Update the [Eligibility API Client](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/api.py#L117) to accept a `verifier` parameter in addition to `agency`, instead of assuming the `verifier` from the `agency`
- [x] Update [initialization of the Client](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L88) with `agency=agency` and  `verifier=agency.eligibility_verifiers[0]` (for now)

angela-tran:
Closes #320 

Made all the changes mentioned in #320. Also changed `TransitAgency.types_to_verify` to take in a verifier.

### Testing
Manually tested to make sure that the eligibility flow still works. ✔️

### DIFF
--- a/benefits/core/migrations/0001_initial.py
+++ b/benefits/core/migrations/0001_initial.py
@@ -81,7 +81,7 @@ class Migration(migrations.Migration):
                 ("jws_signing_alg", models.TextField(help_text="The JWS-compatible signing algorithm.")),
                 ("payment_processor", models.ForeignKey(on_delete=models.deletion.PROTECT, to="core.paymentprocessor")),
                 ("eligibility_types", models.ManyToManyField(to="core.EligibilityType")),
-                ("eligibility_verifier", models.ForeignKey(on_delete=models.deletion.PROTECT, to="core.eligibilityverifier")),
+                ("eligibility_verifiers", models.ManyToManyField(to="core.eligibilityverifier")),
                 # fmt: on
             ],
         ),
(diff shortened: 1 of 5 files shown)


thekaveman (approved):
This looks good! I think I would have gone for maybe a few separate commits to introduce the changes to different areas of the app, but this is a small enough scope that it's fine.

thekaveman (approved):
Confirmed I could rebuild the database using `bin/init.sh` and run through eligibility verification locally with these changes.

Issue #323, opened 2022-03-16
Create VerifierSessionRequired middleware that expects a verifier to exist in session


thekaveman:
## Background

Once the session tracks the selected verifier in #321, we can make use of that infrastructure to put guards on certain view functions that require a verifier to be selected. The first step is to create a new middleware class that enforces the requirement.

This is similar to how the [`AgencySessionRequired`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/middleware.py#L20) and [`EligibileSessionRequired`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/middleware.py#L68) middleware are used.

## Tasks

- [x] Create a new middleware class like `VerifierSessionRequired` inheriting from `MiddlewareMixin`, see the other `*SessionRequired` as examples
- [x] In `process_request()`, check `session.verifier()` for the request.
  - If `None`, raise an error to stop the request
  - Otherwise return `None` to allow the request to continue
- [x] Apply this middleware to the following views to enforce that a verifier is needed:
  - [x]  [`eligibility:index`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L16)
  - [x]  [`eligibility:confirm`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L46)

PR #330, opened 2022-03-17
Feat: track verifier in session

### LINKED ISSUE #321: Track EligibilityVerifier in user session
## Background

As part of #317 the application will need maintain the user's selected `EligibilityVerifier`.

We can do this using the existing [`session` framework](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py) which keeps everything local to the user in secure cookies.

## Tasks

- [x] Add a [new key `_VERIFIER = "verifier"`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L28) for use elsewhere
- [x] Create a new function `verifier()` that uses the key to get the `EligibilityVerifier` instance corresponding to the saved value from the session; see [`agency()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L35) and [`eligibility()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L87) as examples
- [x] Update the [`update()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L192) function to accept a new kwarg `verifier=None`; if not None and it is a valid `EligibilityVerifier` instance, store the `id` in the request's session using the key
- [x] Update the [`reset()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L134) function to also clear the verifier
- [x] Update the [`context_dict()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L52) function to add the new key/value - this shows up in the debug bar
- [ ] Store the agency's first verifier in session (for now) when an agency is configured
- [ ] Retrieve the verifier from session to give to the API client

angela-tran:
Closes #321

### DIFF
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -77,6 +77,12 @@ class EligibilityVerifier(models.Model):
         """jwcrypto.jwk.JWK instance of this Verifier's public key"""
         return self.public_key.jwk
 
+    @staticmethod
+    def by_id(id):
+        """Get an EligibilityVerifier instance by its ID."""
+        logger.debug(f"Get {EligibilityVerifier.__name__} by id: {id}")
+        return EligibilityVerifier.objects.get(id=id)
+
 
 class PaymentProcessor(models.Model):
     """An entity that processes payments for transit agencies."""
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -66,6 +66,7 @@ def context_dict(request):
         _TOKEN: token(request),
         _TOKEN_EXP: token_expiry(request),
         _UID: uid(request),
+        _VERIFIER: verifier(request),
     }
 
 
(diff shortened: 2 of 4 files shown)


thekaveman (approved):
Looks great!! 👍

thekaveman (changes requested):
Actually, sorry - I'm going to go a bit off scope from #321.

Can you also make the change to store/retrieve the verifier from session in the views? For now you can store in the same place that the agency is stored (use the first verifier as usual). And then retrieve before the API Client is initialized.

Then we can test/see that the verifier is really being tracked in the session.

angela-tran:
> Actually, sorry - I'm going to go a bit off scope from #321.
> 
> Can you also make the change to store/retrieve the verifier from session in the views? For now you can store in the same place that the agency is stored (use the first verifier as usual). And then retrieve before the API Client is initialized.
> 
> Then we can test/see that the verifier is really being tracked in the session.

Sure, not a problem

thekaveman:
Added those items to #321. Thanks! This will get us in better position for #323 too.

thekaveman (approved):
This is great. Now running locally, I can see that the session starts without an `agency` or a `verifier`:

image

And then both are set once the `agency` is selected:

image

And going back to the homepage correctly resets the `verifier`!

PR #336, opened 2022-03-22
Feat: dynamic verification form

### LINKED ISSUE #324: Update eligibility verification form to use dynamic data
## Background

This builds on #208 and #319 to incorporate specific `EligibilityVerifier` properties into the [eligibility verification form](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/forms.py#L15)

## Tasks

* [x] Create an `__init__()` method on the form class accepting a new kwarg `verifier` of type `EligibilityVerifier`
* [x] If the `verifier.form_sub_pattern` field has a value, add a `pattern` attribute with the value to the `sub` input via [Widget.attrs](https://docs.djangoproject.com/en/3.2/ref/forms/widgets/#django.forms.Widget.attrs)
* [x] If the `verifier.form_name_max_length` field has a value, add a `maxlength` attribute to the `name` input via Widgets.attrs
* [x] Use attributes from the `verifier` as `msgid` keys for labels etc.
* [x] Rework the [`eligibility:confirm` view](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L46) to initialize the `Page` viewmodel and eligibility verification form from the selected verifier

angela-tran:
Closes #324 

## Testing
After reinitializing your database and recompiling language files, the form at `eligibility/confirm` should look exactly the same as before.

image


angela-tran:
I made sure to consider the behavior of `gettext_lazy` vs `gettext` as mentioned in https://github.com/cal-itp/benefits/issues/324#issuecomment-1071880332, and I believe no change is needed to the way we're using those in `eligibility/forms.py` and `eligibility/views.py`

thekaveman:
👀

thekaveman (approved):
👍 Looks good. I was able to run through verification locally, form worked as expected and showed the expected content.

machikoyasuda (approved):

PR #335, opened 2022-03-22
Feat: create a separate eligibility start page

thekaveman:
This closes #332.

## Testing this PR

1. Delete your local database and compiled language files:

  ```bash
  rm django.db
  
  rm benefits/locale/**/*.mo
  ```

2. Reinitialize

  ```bash
  bin/init.sh
  ```

3. Launch the app with `F5`, navigate to the homepage, and select an agency.
4. Manually enter the url `/eligibility/start` and you should see the same page/content as `/eligibility`:

`/eligibility`:

image

`/eligibility/start`:

image

PR #337, opened 2022-03-22
Feat: use EligibilityVerifierSelectionForm

### LINKED ISSUE #322: Django form to capture user's EligibilityVerifier selection
## Background

Building on #320 and #332, once a `TransitAgency` has multiple `EligibilityVerifier` references, the user needs a way to select which one they will use for verification.

Django has the [`ChoiceField`](https://docs.djangoproject.com/en/3.2/ref/forms/fields/#choicefield) which can use the `widget=RadioSelect` to render as a list of radio buttons. The field takes a list of `choices`, each a tuple where the first item is the value and the second is the label:

```
from django import forms

choices = [("value1", "Label One"), ("value2", "Label Two")]
field = forms.ChoiceField(choices=choices, widget=forms.RadioSelect)
```

See e.g. https://stackoverflow.com/a/5925240/453168

## Tasks

* [x] Create a new form class like `EligibilityVerifierSelectionForm` in [`benefits/eligibility/forms.py`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/forms.py); form should `POST` to `eligibility:index`
* [x] Add a `ChoiceField` using `widget=RadioSelect`
* [x] Create an `__init__()` method accepting an `agency` parameter
* [x] Seed the `ChoiceField.choices` from the `agency.eligibility_verifiers` property, using the above `choices` structure as an example
* [ ] Rework the [`eligibility:index` view](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L16) to handle both `GET` and `POST`
  * for `GET`: If agency has multiple verifiers, initialize form with the selected agency's verifiers and send to the template; otherwise store the one verifier in session and immediately redirect to `eligibility:start`
  * for `POST`: store the selected verifier in session and redirect to `eligibility:start`

angela-tran:
Closes #322

This is dependent on changes from #335 

### Changes in this PR:
 - items listed in #322
 - removed hard-coded verifier selection from [`core/views.py`](https://github.com/cal-itp/benefits/pull/337/files#diff-2e5e461b0d86b207a87071f948630540090d2d020e2b0e053b5486e6ac5c82fdL64) and removed `VerifierSessionRequired` from [`index`](https://github.com/cal-itp/benefits/pull/337/files#diff-bd503c63c9a7b07070e81c9df074adafc509da84db6e270d3c91c54e282218b2L16) function in `eligibility/views.py` (since that is where the verifier will now be saved to the session)
 - added a PO file entry for the `EligibilityVerifierSelectionForm`'s submit button and for the selection page content title


thekaveman:
@angela-tran I know this is still WIP, but I'm wondering if it would be possible to avoid the bulk of the changes that happened in the `*.po` files here? It's going to make for a more difficult merge of the eventual #338 PR.

Maybe just have this PR focus on the specific adds/changes needed for the selection page, and not all the line-number, reorganization, and timestamp changes?

thekaveman:
@angela-tran I know this is still WIP, but I'm wondering if it would be possible to avoid the bulk of the changes that happened in the `*.po` files here? It's going to make for a more difficult merge of the eventual #338 PR.

Maybe just have this PR focus on the specific adds/changes needed for the selection page, and not all the line-number, reorganization, and timestamp changes?

angela-tran:
@thekaveman I think that's a really good idea. I will do some rebasing and editing to make as minimal of changes to the `*.po` files. I think what that'll look like is just the addition of an entry for the submit button (which the Figma mockup has as saying "Continue"/"Continuar").

angela-tran:
@thekaveman As I've been looking into the remaining work to finish out the selection form, I'm starting to feel like we should split it off into a separate issue. The remaining work is all frontend and might require bigger changes or decisions. Here are the details:

In order to get _from_ the current state in this PR _to_ what's in the Figma mockup, the questions to answer are:
image


- How should we get that paragraph under the form?
   - Currently, `paragraphs` from `viewmodels.Page` is rendered above the `form`. See [page.html](https://github.com/cal-itp/benefits/blob/e872bd1244931f563a18adea000192f4a23ef03d/benefits/core/templates/core/page.html#L45)
- `RadioSelect` in Django 3.2 renders as `<li>` elements. We probably don't want that? Should we add something to `page.html` so that we have more granular control over how radio buttons are shown? (Django docs have an [example](https://docs.djangoproject.com/en/3.2/ref/forms/widgets/#radioselect) of looping over the widgets.)
- Rendering the `selection_label_description` may involve adding some HTML to the display value in our `choices` tuple. My first step would be to try out the approach in https://stackoverflow.com/questions/2204696/setting-help-text-for-each-choice-in-a-radioselect.

thekaveman:
@angela-tran I think you're right, a separate issue to split out the frontend work makes a lot of sense here. I have some ideas so I'll write up a starter issue and you can edit with what I miss.

For this PR, rendering the unstyled radio button list with the intended functionality is good 👍

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -227,6 +227,14 @@ msgctxt "image alt text"
...
 msgstr "Bank card icon with contactless symbol"
 
+#: benefits/eligibility/forms.py:23
+msgid "eligibility.form.continue"
+msgstr "Continue"
+
+#: benefits/eligibility/forms.py:24
+msgid "eligibility.form.continuing"
+msgstr "Continuing"
+
```
thekaveman, line 237: Referring to the [Figma file for this page](https://www.figma.com/file/5IabSdY0xDzwL4gyXhbQ9G/Benefits-(Courtesy-Cards)?node-id=1618%3A4627), there is no second-state for this button. (It's a simple form anyway)

benefits/eligibility/forms.py
```diff
@@ -6,12 +6,29 @@
...
+class EligibilityVerifierSelectionForm(forms.Form):
+    """Form to capture eligibility verifier selection."""
+
+    action_url = "eligibility:index"
+    method = "POST"
+
+    verifier = forms.ChoiceField(label="", widget=forms.RadioSelect)
+
+    submit_value = _("eligibility.form.continue")
+    submitting_value = _("eligibility.form.continuing")
```
thekaveman, line 24: See my other comment - there aren't two states for this form. Can just use the same `msgid` here.

benefits/eligibility/views.py
```diff
@@ -8,38 +8,44 @@
...
-        button=viewmodels.Button.primary(text=_("eligibility.pages.index.button"), url=reverse("eligibility:confirm")),
-    )
+    eligibility_start = reverse("eligibility:start")
 
-    return PageTemplateResponse(request, page)
+    if request.method == "POST":
+        form = forms.EligibilityVerifierSelectionForm(data=request.POST, agency=agency)
+
+        if not form.is_valid():
+            return None
```
thekaveman, line 29: If the form is not valid (unlikely, but let's pretend...) we don't want to return `None` here - that doesn't give the user a chance to correct the form. We'd just give them an empty response.

We want to return the same as if they `GET` the page the first time, only with the same form they submitted instead of a new one. This is how [`confirm()` works](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L65) - dealing with the potential `None` that `_verify()` may return.

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -227,6 +227,14 @@ msgctxt "image alt text"
 msgid "core.icons.bankcard"
 msgstr "Bank card icon with contactless symbol"
 
+#: benefits/eligibility/forms.py:23
+msgid "eligibility.form.continue"
+msgstr "Continue"
```
thekaveman, line 232: We used to use `msgid "eligibility.pages.index.button"` for this, can we just update that text instead of creating a new entry?

benefits/locale/es/LC_MESSAGES/django.po
```diff
@@ -229,6 +229,14 @@ msgctxt "image alt text"
...
 msgstr "Icono de tarjeta bancaria"
 
+#: benefits/eligibility/forms.py:23
+msgid "eligibility.form.continue"
+msgstr "Continuar"
+
+#: benefits/eligibility/forms.py:24
+msgid "eligibility.form.continuing"
+msgstr "Continuando"
+
```
thekaveman, line 239: Same comments as above in English file

thekaveman (changes requested):

angela-tran:
Rebased my commits onto `dev`, and changed the target branch to `dev`

angela-tran:
The `eligibility` UI test won't pass until #335 is merged in. (`reverse("eligibility:start")` throws an exception, so the tests can't find the "Continue" button.)

angela-tran:
Closing and re-opening to rerun the checks

thekaveman:
You can also re-run from the failed action page:

image

thekaveman:
@angela-tran I think we can just skip the failing tests for now - we know we'll have to update them in #339 anyway.

angela-tran:
Thank you @thekaveman! I added `.skip()` to those tests.

thekaveman (changes requested):
I found out what the issue is, and if you rebase on `dev` again you can make the tests pass:

Since there is only 1 verifier, the `eligibility:index` view is forwarding directly to the `eligibility:start` view. This view is using the `eligibility.pages.start.button` msgid for the button, which still has the old "Ready to continue" value.

thekaveman:
I'll clean up these msgids a bit more in #344 once this is merged 🙏

thekaveman (approved):

PR #345, opened 2022-03-25
Feat: custom RadioSelect widget

### LINKED ISSUE #347: Enhance form rendering: Radio button list
This is a follow-on to #322, to enhance how forms are rendered in the [`page` template](https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/page.html)

@angela-tran [notes](https://github.com/cal-itp/benefits/pull/337#issuecomment-1076882758):

> `RadioSelect` in Django 3.2 renders as `<li>` elements. We probably don't want that? Should we add something to `page.html` so that we have more granular control over how radio buttons are shown? (Django docs have an [example](https://docs.djangoproject.com/en/3.2/ref/forms/widgets/#radioselect) of looping over the widgets.)

image

The [State Template form components](https://template.webstandards.ca.gov/sample/form.html) has some example HTML for what might work here?

```html
<!--Bootstrap4 Custom Radios-->
<div class="form-check">
    <input class="form-check-input" type="radio" name="flexRadioDefault" id="customRadio1">
    <label class="form-check-label" for="customRadio1">
        Toggle this custom radio
    </label>
</div>
<div class="form-check">
    <input class="form-check-input" type="radio" name="flexRadioDefault" id="customRadio2" checked="">
    <label class="form-check-label" for="customRadio2">
        Or toggle this other custom radio
    </label>
</div>
```

The comment indicates this is just [Bootstrap 4 radios](https://getbootstrap.com/docs/4.0/components/forms/#radios)?

@angela-tran [suggests](https://github.com/cal-itp/benefits/pull/337#issuecomment-1076882758):

> Rendering the `selection_label_description` may involve adding some HTML to the display value in our `choices` tuple. My first step would be to try out the approach in https://stackoverflow.com/questions/2204696/setting-help-text-for-each-choice-in-a-radioselect.

image

angela-tran:
Closes #347

## Goals of this PR
 - [x] Override Django 3.2's rendering of radio buttons as `<li>` elements
 - [x] Style the radio buttons similarly to the sizes, colors, and spacing in the Figma mockup
 - [x] Support displaying a description under the label for the radio button
 - [x] Make sure validation occurs if a submission is made without a selection

## Changes
To accomplish the goals listed above, this PR adds our own custom widget derived from Django's `RadioSelect`. `EligibilityVerifierSelectionForm` is then configured to use that instead.

This custom `RadioSelect` overrides the `template_name` and `option_template_name` fields so that it uses our own templates, also introduced in this PR. 

The commit history here shows the various implementations that were explored. First we tried using styles from the [CA Web Template](https://template.webstandards.ca.gov/sample/form.html) and then one from the [U.S. Web Design System](https://designsystem.digital.gov/components/radio-buttons/). The one that gave us the most flexibility, simplicity, and compatibility with browser-validation was based off an example in [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#styling_radio_inputs).

### ~Preview / screenshots~ Outdated screenshots 
**(see [comments ](https://github.com/cal-itp/benefits/pull/345#issuecomment-1083954378) for newer screenshots)**

You should be able to checkout this branch and see the styled radio buttons. 

**Here is the blank form:**
https://user-images.githubusercontent.com/25497886/160190844-0dbcb87b-fd26-4298-96db-aa50cfc1aa5b.png

**If you click "Continue" without choosing an option, here is the default validation rendering from Django:**
https://user-images.githubusercontent.com/25497886/160191002-9121ba60-a77c-4ca5-a084-b4c0d0977d31.png

**Here is the form with a selected option:**
https://user-images.githubusercontent.com/25497886/160191056-f5dd8841-310a-4ea3-945e-af0f7f498d63.png

Figma file for reference: https://www.figma.com/file/5IabSdY0xDzwL4gyXhbQ9G/Benefits-(Courtesy-Cards)?node-id=1618%3A4627


angela-tran:
@thekaveman I have this as a draft PR because I anticipate that there will probably be tweaks to this. Could you take a look at these changes so far though and let me know what you think?

angela-tran:
I'm going to rebase this onto the latest `dev` commit.

thekaveman (changes requested):
Good overall approach on the widget implementation.

Couple of changes:

* Feels strange to have the template files in `core` and the field class in `eligibility`; seems better if everything was in `core` (it is a general implementation). Then just import the class in `eligibility.forms`
* Can we rename to hide the implementation detail? We build on top of CA State Template today, but maybe something else tomorrow. `benefits_radio_select` or even just `radio_select` (same for the class)

benefits/core/templates/core/widgets/radio_select.html
```diff
@@ -0,0 +1,19 @@
+{% with id=widget.attrs.id %}
+<div{% if id %} id="{{ id }}"{% endif %} class="radio-container">
+  {% for group, options, index in widget.optgroups %}
```
thekaveman, line 3: I see a comment in the `RadioSelect` widget class: 

> `# this implementation does not support groups from ChoiceWidget.optgroups`

So do we need this loop here?

benefits/eligibility/forms.py
```diff
@@ -27,6 +27,13 @@ def __init__(self, agency: models.TransitAgency, *args, **kwargs):
         choices = [(v.id, _(v.selection_label)) for v in agency.eligibility_verifiers.all()]
         self.fields["verifier"].choices = choices
 
+        choice_descriptions = []
+        for verifier in agency.eligibility_verifiers.all():
+            if verifier.selection_label_description:
+                choice_descriptions.append((verifier.id, _(verifier.selection_label_description)))
+
```
thekaveman, line 34: This could also be done in a list comprehension, like above for `choices`

thekaveman: And looking at how this is used in the `RadioSelect` widget, it seems like a `dict` would be the more appropriate data structure? `{verifier.id: _(verifier.selection_label_description)}` and then you can lookup by `id` later, vs. having to loop over the whole thing.

benefits/core/widgets.py
```diff
@@ -15,3 +16,26 @@ def __init__(self, pattern=None, placeholder=None, **kwargs):
...
+
+    def __deepcopy__(self, memo):
+        obj = super().__deepcopy__(memo)
+        obj.choice_descriptions = copy.copy(self.choice_descriptions)
+        return obj
+
+    def create_option(self, name, value, label, selected, index, subindex, attrs):
+        option = super().create_option(name, value, label, selected, index, subindex, attrs)
+        # this implementation does not support groups from ChoiceWidget.optgroups
+        for choice_value, choice_description in self.choice_descriptions:
```
thekaveman, line 37: See comments below about using a `dict` for `choice_descriptions` instead of list of tuples - will make this code simpler

thekaveman (changes requested):
A few questions and thoughts on the `choice_descriptions` implementation.

The styles are 💣 🔥 nice job!!!

benefits/static/css/styles.css
```diff
@@ -144,6 +144,45 @@ footer {
   height: 75px;
 }
 
+.radio-container {
+  display: block;
+  margin-left: 15%;
+}
+
+.radio-label {
```
machikoyasuda, line 152: Both the radio label and the radio input itself should have `cursor: pointer`, like https://github.com/uswds/uswds/commit/73ec4d8059af7b8ee67130e07a6bf21cace1c77a, so that when a Desktop user hovers their mouse over the 🔘 radio input or the radio label (the text), the mouse cursor turns into a 👆

machikoyasuda (dismissed):
Great work! Just one small CSS change, based on https://designsystem.digital.gov/components/radio-buttons/

benefits/core/widgets.py
```diff
@@ -15,3 +16,26 @@ def __init__(self, pattern=None, placeholder=None, **kwargs):
...
+    option_template_name = "core/widgets/radio_select_option.html"
+
+    def __init__(self, choice_descriptions=(), *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.choice_descriptions = list(choice_descriptions)
+
+    def __deepcopy__(self, memo):
+        obj = super().__deepcopy__(memo)
+        obj.choice_descriptions = copy.copy(self.choice_descriptions)
+        return obj
```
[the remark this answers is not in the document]
angela-tran: Since our `RadioSelect` subclass introduces `choice_descriptions` as an instance variable, we want to make sure that when anyone makes a deep copy of a `RadioSelect` object, that copy will have its own instance of `choice_descriptions` rather than sharing the instance from the original `RadioSelect`. (The call to `super().__deepcopy__(memo)` on line 30 is handling the other instance variables, of course.)

Notably, when Django instantiates [forms](https://github.com/django/django/blob/main/django/forms/forms.py#L107) and their [fields](https://github.com/django/django/blob/main/django/forms/fields.py#L142), it uses deep-copying, so there would presumably be unexpected behavior / bugs if we didn't have this.

These Django tickets have examples of the kind of unexpected behavior that would occur:
- https://code.djangoproject.com/ticket/25085
- https://code.djangoproject.com/ticket/12048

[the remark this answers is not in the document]
angela-tran: (To be fair, our current usage of `RadioSelect` wouldn't expose the bug.)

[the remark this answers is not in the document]
thekaveman: @angela-tran Thanks for this background info! 🙏

benefits/core/templates/core/widgets/radio_select.html
thekaveman: Thanks for the explanation and deep dive into Django form widgets @angela-tran! We aren't supporting groups but this is how to loop over all options for display.

benefits/core/widgets.py
angela-tran: Thanks for pairing with me on all this through screenshare earlier! Commenting here with some links here that explain more on what these "groups" are and to document that we decided we don't need to support these for now.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#field-choices-named-groups
https://code.djangoproject.com/ticket/20931
https://github.com/django/django/blob/3.2/django/forms/widgets.py#L587
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

thekaveman (approved):
It looks amazing!

image

image

PR #352, opened 2022-03-28
Fix: devcontainer command syntax

### LINKED ISSUE #351: Fix devcontainer command syntax
The current syntax for the [devcontainer `command`](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/compose.yml#L21) is broken on newer versions of Docker/Compose (at least Docker `4.6.1 (76265)`), making the devcontainer unable to start. This does not affect the startup of the application, nor does it affect the build of either the application or devcontainer (all of which succeed).

We need to change to the list-based `command` style and pass the `"-c"` option so [`/bin/bash`](https://github.com/cal-itp/benefits/blob/dev/Dockerfile#L57) knows the rest of the arguments are the commands to run.

See this SO post for more: https://stackoverflow.com/a/62313159

thekaveman:
Fixes #351


machikoyasuda (approved):

angela-tran:
I'm currently having issues with opening the devcontainer when I'm on branch `fix/devcontainer-cmd`. That was on my older version of Docker Desktop (4.5.1), and I've now tried with the latest version as well. Still seeing errors.

image
image


The devcontainer does open for me when I'm on the `dev` branch.
image

thekaveman:
@angela-tran that error looks similar from yesterday.

In Docker Desktop, **Delete** the `benefits_devcontainer` group (or one by one if that doesn't work)

image

Then locally, 

```bash
docker compose build --no-cache dev
```

And finally `Reopen in Container`... that should work?

angela-tran:
I see the same behavior as you now @thekaveman - it was right after I went into Docker Desktop > Settings > General, and checked "Use Docker Compose V2".

image

Now the devcontainer opens for me on `fix/devcontainer-cmd` but not on `dev`.

angela-tran (approved):

PR #364, opened 2022-03-29
Feat: show layout image by default

### LINKED ISSUE #362: Update main layout to show large image by default
The application layout received an update with the Courtesy Cards feature set; the default is to show the large "rider tapping" image on the left and the content on the right, for *most* pages (including the forms).

However the Help and Payment Options pages are still full screen / no image, so we need the option to "turn it off".

thekaveman:
Closes #362

### DIFF
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -20,11 +20,6 @@ def _index_content_title():
     return _("core.pages.index.content_title")
 
 
-def _index_image():
-    """Helper returns a viewmodels.Image for the common index page."""
-    return viewmodels.Image("ridertappingbankcard.png", pgettext("image alt text", "core.pages.index.image"))
-
-
 def _index_paragraphs():
     """Helper returns the content paragraphs for the common index page."""
     return [_("core.pages.index.p[0]"), _("core.pages.index.p[1]"), _("core.pages.index.p[2]")]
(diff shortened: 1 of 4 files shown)


machikoyasuda (approved):
image
image
image

machikoyasuda (approved):

benefits/core/views.py
```diff
@@ -20,11 +20,6 @@ def _index_content_title():
     return _("core.pages.index.content_title")
 
 
-def _index_image():
-    """Helper returns a viewmodels.Image for the common index page."""
-    return viewmodels.Image("ridertappingbankcard.png", pgettext("image alt text", "core.pages.index.image"))
```
angela-tran, line 25: This made me realize that we weren't and aren't using the `Image.alt` value in `page.html`. 🤔 Unrelated to this ticket though.

angela-tran (approved):
These changes look good 👍

benefits/core/views.py
thekaveman: It's true - this is [implemented as a CSS `background` property](https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/page.html#L16) rather than an `` tag, which is why there is no alt attribute (and how the Accessibility checker passes).

If you can find anything related to using images in CSS background and alt attributes, or maybe there's a special `aria-` attribute we can use instead? Go ahead and file an issue since yeah, we weren't using it before either, and shouldn't prevent this PR from being merged.

PR #378, opened 2022-03-30
Tests: Update UI tests for multiple verifier flow

machikoyasuda:
closes #339 

### Tests
- Modifies existing UI specs for Eligibility Confirmation page, by having the user select `DEFtl` Transit (which is set to be the single-verifier path)
- Adds UI specs for the Verifier selection page, testing for:

After a user selects `ABC` Transit, a multiple verifier path:

1. Whether a user can click `Continue` to get to the Verifier selection page
2. Whether a user sees 2 radio buttons, with `MST Courtesy Cardholder` and `Senior Discount Program`*
3. Whether a user gets a validation message if they click `Continue` without choosing a radio button**
4. Whether a user can click `Continue` to get to the Eligibility start page, after selecting a radio button

- Modify the existing Eligibilty confirmation page spec to:

1. Add the actions of selecting a radio button (`Senior Discount Program`) to the `beforeEach`, so that the specs can run on the proper page
2. Add the next step of clicking `Continue` on the other specs so the specs can run on the proper page

*Note: The text of the radio buttons are hard-coded.
**Note: See extended note below

### Documentation
- update Cypress documentation to include instructions on how to run Cypress _outside_ of the devcontainer - which allows developers to use the Cypress Test Runner user interface.

### Chores
- update `.gitignore`
- check in latest `package-lock.json` for Cypress

## Development notes

1. In order to get the `eligibility-server` service running locally, I had to revert the image back to a version from several weeks ago: https://github.com/cal-itp/eligibility-server/pkgs/container/eligibility-server/13177499, before https://github.com/cal-itp/eligibility-server/pull/58 was merged

```diff
--- a/.devcontainer/compose.yml
+++ b/.devcontainer/compose.yml
@@ -36,7 +36,7 @@ services:
       - ../:/home/calitp/app:cached
 
   server:
-    image: ghcr.io/cal-itp/eligibility-server:main
+    image: ghcr.io/cal-itp/eligibility-server@sha256:337d5b2beb1e458980be49a778efd4a47f8daa8decc5e8329c0d528596e2f196
     ports:
       - "5000"
```

2. A browser bug was revealed during the development of this ticket.

On my Apple M1 Pro chip Macbook Pro from 2021, on all three browsers (Safari, Chrome, Firefox) and on Cypress Test Runner UI (Electron, Chrome, Firefox), at the Verifier Selection page, I am able to click `Continue` without having selected a radio button. The page then refreshes and shows me this screen: 
image
 
[...]

PR #333, opened 2022-03-18
Feat: middleware to require verifier in session

thekaveman:
Closes #323

PR #385, opened 2022-03-31
Fix: use verifier's translation strings on eligibility unverified page

### LINKED ISSUE #379: "Unverified" page should use translation fields on EligibilityVerifier
### To reproduce
Go through the eligibility verification flow such that you will see the "unverified" page that an ineligible user would see.

### Expected result
The page shows with translated content.

### Actual result
The page shows with translation strings.

image

### Thoughts
I think we forgot to update the `unverified` function in the eligibility app's views when we did #338.

thekaveman:
Fixes #379.

## What this PR does

A couple of small changes to the `eligibility/views.py`, `unverified()` view function:

1. Require a verifier in the session - using the new `VerifierSessionRequired` middleware from #333 
2. Get the verifier from session before creating the `Page` viewmodel
3. Translate `msgid` from the verifier for the `Page` viewmodel

## Results

Here's the result of failing the eligibility check for "ABC" with "Senior Discount":

image

And for "ABC" with "Courtesy Cards":

image

## Testing note

In order to get a server to run that you can test (and fail) an eligibility verification request against, modify `.devcontainer/compose.yml` entry for `server` to use a different image:

```diff
   server:
-   image: ghcr.io/cal-itp/eligibility-server:main
+   image: ghcr.io/cal-itp/eligibility-server@sha256:337d5b2beb1e458980be49a778efd4a47f8daa8decc5e8329c0d528596e2f196
```

See the note in https://github.com/cal-itp/benefits/pull/378#issue-1186966293

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -156,6 +156,7 @@ def verified(request, verified_types):
 
 
 @decorator_from_middleware(middleware.AgencySessionRequired)
+@decorator_from_middleware(middleware.VerifierSessionRequired)
 def unverified(request):
     """View handler for the unverified eligibility page."""
 
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -166,11 +166,13 @@ def unverified(request):
     agency = session.agency(request)
     buttons = viewmodels.Button.agency_contact_links(agency)
 
+    verifier = session.verifier(request)
+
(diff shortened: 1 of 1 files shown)


machikoyasuda (approved):
Confirmed locally --


ABC with Senior and invalid:

image

ABC with Courtesy Card and invalid:

image

PR #389, opened 2022-04-01
Fix: add realistic CSP settings to sample env

thekaveman:
The CSP entries in the `.env.sample` weren't realistic, and meant that when running the app locally, it doesn't load all the styles.

This PR adds better defaults to the `.env.sample` for a couple of key CSP settings.


angela-tran (approved):
🚀

PR #390, opened 2022-04-01
Feat: re-introduce static Docker port config

thekaveman:
For the upcoming SSO integration, we need to be able to pin a static localhost port for testing with the dev environment.

But only for the app container and devcontainer, the others can still be dynamic.

This only affects the _host machine_ port and only at development time; internally in the containers, the ports are unaffected.


angela-tran (approved):
I read a bit from https://docs.docker.com/compose/networking/, and this looks good!

Issue #377, opened 2022-03-30
Update analytics events for Login.gov flow


thekaveman:
## Background

We'll be introducing some new steps in the flow to support the Login.gov integration. Also the login requirement is being moved further into the overall application flow, so begin in our app flow before hitting the Login.gov flow.

Crucially, we need to understand any "drop off" points, where the user may either get distracted before starting the Login.gov flow, or abandon the application altogether during the Login.gov sign on.

## Tasks

* [x] "started sign in" event for clicking the "Continue to Login.gov" button
  * [x] `event_property` for the `auth_provider=OAUTH_CLIENT_NAME`
* [x] "finished sign in" after successful authorization
  * [x] `event_property` for the `auth_provider=OAUTH_CLIENT_NAME`
* [x] Update the "clicked link" event to also cover URLs containing a hash (e.g. `#payment-options` if the user clicks `/help#payment-options`)
* [x] "started sign out" for when the "Sign out of Login.gov" button is clicked (#368)
    * [x] `event_property` for `auth_provider=OAUTH_CLIENT_NAME`
* [x] "finished sign out" for after the user is redirected back into `post_logout` (see #436 implementation)
    * [x] `event_property` for `origin`
    * [x] `event_property` for `auth_provider=OAUTH_CLIENT_NAME`

thekaveman:
I started some of the basics for this in #452 - feel free to build off that PR if you get to this before I can come back to it.

thekaveman:
@angela-tran thanks for picking this back up!

Based on the implementation we're going towards in #436 - can you mirror the `started/finished sign in` events with `started/finished sign out` with the same attributes? Instead of just the single `signed out` that I listed above (and started with in #452).

angela-tran:
Re: this bullet point,
> * [ ] Update the "viewed page" event to include a custom `event_property` for the URL fragment / hash (e.g. `#payment-options` if the user clicks `/help#payment-options`)

I'm seeing from various sources that the anchor in the request URL is not sent to the server.
 - [Get anchor tag from Django request URL](https://stackoverflow.com/questions/65249426/get-anchor-tag-from-django-request-url)
 - [Is the anchor part of a URL being sent to a web server?](https://stackoverflow.com/questions/3067491/is-the-anchor-part-of-a-url-being-sent-to-a-web-server)

Confirmed in the debugger that we don't have that URL fragment available on the server side. Here I went to the `help#about` URL:
[...]

Issue #375, opened 2022-03-30
Implement OIDC sign in flow


thekaveman:
## Background

Before the user can verify their information with the DMV API, they must authenticate with the Login.gov SSO system. This task is to implement the [OpenID Connect](https://openid.net/connect/) (OIDC) flow -- built on OAuth 2.0 -- that Login.gov supports.

There are two steps:

1. The user is sent from the Benefits application to the OIDC auth server, where they enter credentials. Upon successful login, the user is redirected back to the Benefits application with an *access token*
2. The access token from the prior step is sent back to the OIDC auth server for validation

Since the Benefits application (server code) must track that the user is signed in (see #370), we need both of these steps to occur on the server in order to trust the results of the access token validation.

Once the access token has been validated, the user is effectively "signed in" from the point of view of the Benefits app.

Build on #380 which places the button on the page.

## Tasks

* [x] When the user clicks the "Continue with Login.gov" button, OIDC redirect to the auth server
* [x] Implement a callback view to receive the OIDC callback redirect (after user logs in), which is a `GET` with querystrings containing the access token
* [x] If the access token is validated, update the user's session to set `auth = True`

PR #392, opened 2022-04-04
Feat: AuthProvider model

### LINKED ISSUE #372: Model settings for SSO
## Background

We'll need to use a configuration object for a JavaScript library that helps implement the Login.gov flow. 

### Configuration data

Here's what the configuration object eventually needs to look like in JavaScript:

```js
// base address of Benefits application
const baseAddress = "https://localhost:8000";

// configuration object
var config = {
  // authorization server address
  authority: "https://login.gov",
  
  // this is registered application id with auth server
  client_id: "UUID",
  
  // post-login, user is redirected here
  redirect_uri: baseAddress + "/callback.html",
  
  // post-logout, user is redirected here
  post_logout_redirect_uri: baseAddress + "/index.html",
  
  // must be "code" to use the authorization-code flow
  response_type: "code",
  
  // space-delimited list
  // since you do not need any information (claims) about the user, "openid" is the only needed scope.
  // If you need refresh tokens, you must add the "offline_access" scope
  scope: "openid",
};
```

## Tasks

* [x] Build out the list of necessary config items (keys, URLs, endpoints, etc.) - *see above ⬆️*
* [x] Create a model like `AuthProvider` with text fields for the above list, all required
  * [ ] Consider storing URL parameters as the name of the app route, e.g. `eligibility:confirm` so we can use the Django helpers like `reverse()` and `{% url %}` and not duplicate raw paths in data 
* [x] Add a `ForeignKey` for `AuthProvider` to the `EligibilityVerifier` model, make it optional
* [x] Create a wrapper bool attribute on the `EligibilityVerifier` that denotes whether Auth is configured (the **"feature flag"**) by checking if there is a linked provider or not
* [x] Create attributes to track the Sign in + Sign out buttons  `msgid`s

angela-tran:
Closes #372


thekaveman:
@angela-tran regarding the `silent*` fields - I think I mistakenly left these in from the sample. We probably won't need this feature (to silently renew in the background) so let's disregard these fields for now.

thekaveman:
To your question about fixtures: is that maybe the reason the tests are failing? We should be able to start up the app with these changes, so if you need to add fixture(s) to make that work, go for it!

thekaveman:
@angela-tran looking at this...

benefits/core/models.py
```diff
@@ -95,6 +110,10 @@ def public_jwk(self):
         """jwcrypto.jwk.JWK instance of this Verifier's public key"""
         return self.public_key.jwk
 
+    @property
+    def requires_authentication(self):
+        return self.auth_provider.all().exists()
```
thekaveman, line 115: `self.auth_provider` is defined as a `ForeignKey`, not a `ManyToMany`

It should either have a value that is an instance of `AuthProvider`, or be `None`.

thekaveman (changes requested):
Minor fix on the feature flag

thekaveman (changes requested):
I added one more item to the checklist on #372: occurred to me that we'll need `msgid` for the Sign in and Sign out buttons.

Can you add attributes to `AuthProvider` for those please?

benefits/core/models.py
angela-tran: Thanks for catching this. I was going off docs about resolving Django `QuerySet`s and didn't catch that `all()` doesn't make sense on `ForeignKey` fields 🤦 

Is the check for "is an instance of `AuthProvider`" necessary? I'm thinking we can just check that it is not `None`.

angela-tran:
@thekaveman It is much easier to test code changes when we have an `AuthProvider` fixture so I'm thinking of creating a `0002_authproviders.json` file and bumping up the other fixture file numbers. I'm not sure what to use as values for the fixture though. Do you have any thoughts or advice?

benefits/core/models.py
thekaveman: Yeah I think you're right, the `None` check is sufficient!

thekaveman:
Yeah let's add the sample auth provider as a fixture here too 👍 

Sample values don't matter as much, you can take some of what was in the issue description, or make something up (as long as it is representative).

And since you're doing that, let's add a `name` attribute as well. We can use it in e.g. #377

thekaveman:
@angela-tran if adding the sample fixture is a blocker here, we can do that in a follow-up. This PR is needed by a lot of other work so want to get it merged ASAP.

thekaveman (approved):

angela-tran:
I added a bullet point to #375 to create the `AuthProvider` fixture.

PR #396, opened 2022-04-04
Track sign in status in user session

### LINKED ISSUE #370: Track sign in status in user session
## Background

Once the user successfully signs in to their Login.gov account, we will receive a response from the authentication server indicating their status. We can store the fact the the user is then "signed in" in their session using the existing [`session` framework](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py) which keeps everything local to the user in secure cookies.

This task is just about setting up the session code to be able to track this, but not actually getting the status in there (see #373 for that).

Check out #321 / #330 as an example of a similar addition last Sprint.

## Tasks

- [x] Add a [new key `_AUTH = "auth"`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L26) for use elsewhere
- [x] Create a new function `auth()` that uses the key to get the stored `bool` corresponding to whether the user is signed in via or not
- [x] Update the [`update()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L192) function to accept a new kwarg `auth=False`; if it's a bool, store in the request's session using the key
- [x] Update the [`reset()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L134) function to set the key to `False` by default
- [x] Update the [`context_dict()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L52) function to add the new key/value - this shows up in the debug bar

machikoyasuda:
closes #370 

## What this PR does

- [x] Add a [new key `_AUTH = "auth"`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L26) for use elsewhere
- [x] Create a new function `auth()` that uses the key to get the stored `bool` corresponding to whether the user is signed in via or not
- [x] Update the [`update()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L192) function to accept a new kwarg `auth=False`; if it's a bool, store in the request's session using the key
- [x] Update the [`reset()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L134) function to set the key to `False` by default
- [x] Update the [`context_dict()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/session.py#L52) function to add the new key/value - this shows up in the debug bar
- Adds new logs messages: `Get session auth`, `Update session auth`

## Screenshots

- Auth key/value pair now in DEBUG_MODE: `"auth": "None"`

image

### DIFF
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -53,7 +53,7 @@ def active_agency(request):
 
 def auth(request):
     """Get the auth from the request's session, or None"""
-    logger.debug("Get auth token")
+    logger.debug("Get session auth")
     return request.get(_AUTH)
 
 
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -51,6 +51,12 @@ def active_agency(request):
     return a and a.active
 
 
+def auth(request):
+    """Get the auth from the request's session, or None"""
+    logger.debug("Get auth token")
+    return request.get(_AUTH)
+
+
 def context_dict(request):
     """The request's session context as a dict."""
     logger.debug("Get session context dict")
@@ -138,6 +144,7 @@ def reset(request):
     """Reset the session for the request."""
     logger.debug("Reset session")
     request.session[_AGENCY] = None
+    request.session[_AUTH] = None
     request.session[_ELIGIBILITY] = None
     request.session[_ORIGIN] = reverse("core:index")
     request.session[_TOKEN] = None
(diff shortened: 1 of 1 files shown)


machikoyasuda:
Ready for review @thekaveman @angela-tran

thekaveman:
@machikoyasuda looking at this now

benefits/core/session.py
```diff
@@ -192,11 +201,24 @@ def uid(request):
     return u
 
 
-def update(request, agency=None, debug=None, eligibility_types=None, origin=None, token=None, token_exp=None, verifier=None):
+def update(
+    request,
+    agency=None,
+    auth=False,
```
thekaveman, line 207: The default here should be `None`.

We may call `update()` after the user has successfully signed in (to update something else); if a value for `auth` wasn't given, the default of `False` would effectively sign the user out.

The logic correctly checks for a `bool` parameter, so `None` is the safer default here.

benefits/core/session.py
```diff
@@ -137,6 +145,7 @@ def reset(request):
     """Reset the session for the request."""
     logger.debug("Reset session")
     request.session[_AGENCY] = None
+    request.session[_AUTH] = None
```
thekaveman, line 148: Here we may want to default to `False`, to be explicit that on session reset the user is not signed in.

thekaveman (changes requested):
Looks good! One small swap of the defaults

machikoyasuda:
@thekaveman Thank you! Updated PR to resolve your comments.

thekaveman (approved):
👍

PR #397, opened 2022-04-04
feat(styles): styles to match mockup

### LINKED ISSUE #369: Minor style enhancements
A number of light enhancements to the overall site look and feel

## Tasks

* [x] Update the main banner background color to `#2b6597` - this comes from the [`.bg-primary` override](https://github.com/cal-itp/benefits/blob/dev/benefits/static/css/styles.css#L72) in our stylesheet
* [x] Update the logo `.svg` files to match the above change, swapping `#046b99` for `#2b6597`
* [x] Primary action buttons (e.g. "Get Started" on the `core:agency_index` page or "Check Status" on the `eligibility:confirm` page should remain `#046b99`
* [x] Primary action buttons (see above) should be displayed in ALL CAPS
* [x] Outline action buttons (e.g. the agency selection buttons on the `core:index` page) should remain  `#046b99`
* [x] Link buttons (e.g. the agency selection buttons on the `core:index` page) should remain  `#046b99`
* [x] Update the footer background color (`.global-footer`) to `#212121`
* [x] Verify font sizes, weights, etc. match the Mocks

Jorden1420:
Resolves #369

updated font size, colors


benefits/static/css/styles.css
```diff
@@ -178,6 +192,11 @@ footer {
   margin-bottom: 1rem;
   padding: 1rem;
   width: 100%;
+  text-transform: uppercase;
```
thekaveman, line 195: This feels like the wrong place for the `text-transform` rule.

We want to transform `.btn-primary`, not necessarily every `.btn-lg` -- evidenced by the "undo" rule below for the `.btn-outline-primary`.

benefits/static/css/styles.css
```diff
@@ -178,6 +192,11 @@ footer {
...
+}
+
+.buttons a.btn.btn-lg.btn-outline-primary {
+  text-transform: inherit;
```
thekaveman, line 199: If the `.btn-primary` is targeted for uppercase, this rule is unneeded.

benefits/static/css/styles.css
```diff
@@ -163,6 +167,16 @@ footer {
...
 }
 
+.container.content h1 {
+  font-size: 24px;
+  line-height: 30px;
+}
+
+.container.content p {
+  font-size: 18px;
+  line-height: 27px;
```
thekaveman, line 177: I think these rules belong in the [`p` block](https://github.com/cal-itp/benefits/blob/dev/benefits/static/css/styles.css#L41) up at the top of the document

benefits/static/css/styles.css
```diff
@@ -163,6 +167,16 @@ footer {
   margin-bottom: 2rem;
 }
 
+.container.content h1 {
+  font-size: 24px;
+  line-height: 30px;
```
thekaveman, line 172: I think these rules belong in the [`h1` block](https://github.com/cal-itp/benefits/blob/dev/benefits/static/css/styles.css#L37) up at the top of the document

thekaveman (changes requested):
Few minor changes to where rules show up; I think the text transform still needs some work - should be simpler.

benefits/static/css/styles.css
```diff
@@ -65,11 +65,15 @@ footer {
   padding-bottom: 1rem;
 }
 
+footer a {
```
machikoyasuda, line 68: For the sake of specificity, could you target this style for `footer .footer-links a`? That way, in case we ever add different links to the footer, outside of this `<ul>`, we won't have to override the `a` element entirely.

thekaveman:
@Jorden1420 please remember to link the Issue # (in this case, #369) to the PR so that it closes automatically when the PR is merged. I have done that in this case but also see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

We would also appreciate some examples as to what this PR changes and what we should expect when running the code locally - a comment with a couple of screenshots is helpful.

machikoyasuda:
Great first PR @Jorden1420!

I am comparing the app with the Figma mocks and noticed a few things that weren't mentioned in the Issue that I wanted to point out here:

- [ ] The header is now 80px high on desktop.
image

- [ ] The footer is now 40px high on desktop. To achieve this, the `.with-image footer` should have `padding-top: 0.138rem; padding-bottom: 0.138rem;`
image
image

- [ ] The Español button dimensions have also changed on Figma, to 126x38
image

- [ ] The main button is now 60px high.

benefits/static/css/styles.css
machikoyasuda: Also,  letter spacing is now 0.05em: `letter-spacing: 0.05em;`

benefits/static/css/styles.css
machikoyasuda: Add `letter-spacing: 0.05em;`

benefits/static/css/styles.css
Jorden1420: The reason outline primary stays inherit is because in the mockup the outline buttons do not have an uppercase so here I told the text transform to remain inherit. Do you want all btn outlines to also have text transform of uppercase?

Jorden1420:
Pushed new commit to this pull request based on recommendations.

Jorden1420 (commented):
- header has been moved to global styles with requested changes
- paragraph has been moved to global styles with requested changess
- footer has been changed for more specificity for font size
- action buttons have been given width and height based on recommendations
- header was given height 80 on desktop sizes
- changed footer padding

thekaveman (approved):
LGTM 👍 Thanks @Jorden1420!

@machikoyasuda can you please take one more look?

machikoyasuda (approved):
Looks good!

PR #400, opened 2022-04-05
fix: update package-lock.json

afeld:
I didn't run an update or anything; these showed up after starting the devcontainer 🤷

### DIFF
--- a/tests/cypress/package-lock.json
+++ b/tests/cypress/package-lock.json
@@ -408,6 +408,7 @@
       "integrity": "sha512-w0q/enDHhPLq44ovMGdQeeDLvwxwavsJX7oQGYt/LrBlYsyaxyDnp6z3QzFut/6kLLKnlcUVJLrpB7KBfgG/RA==",
       "dev": true,
       "dependencies": {
+        "colors": "1.4.0",
         "string-width": "^4.2.0"
       },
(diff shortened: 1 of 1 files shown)


thekaveman:
This shows up almost every time. If you can figure out why that would be great! We tend to update it when updating the tests.

thekaveman:
You can just `git restore` on it to get rid of any local changes.

afeld:
If it shows up every time, why wouldn't we commit it?

afeld:
Again, this doesn't update the package, just reflects dependencies that should have been in the lock file in the first place (but weren't for some reason).

thekaveman:
Here's a recent commit where these exact lines where added: https://github.com/cal-itp/benefits/pull/378/commits/a59837ce2ed71bf38c7b092f79479dfc36306dda

And in the same PR, a later commit where they were removed: https://github.com/cal-itp/benefits/pull/378/commits/f05551087ecc3e5c5859881c8c0fe04bcf0b9709

I don't know why these lines in particular continue to show up and disappear (no matter what we commit). If you can find the root cause, I would definitely like to know what it is!

afeld:
Gotcha. My guess is a NPM CLI version difference. Won't worry about it for now. Thanks!

PR #401, opened 2022-04-05
Fix: add missing third-party domains to CSP settings

afeld:
Fixes browser warnings when loading the page locally.


thekaveman (approved):
👍

PR #406, opened 2022-04-05
style: Simplify application index page

### LINKED ISSUE #366: Simplify application index page
## Background

We expect most users to come in to the app through a direct link to their agency's entry page, In other words, via [`core:agency_index`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/views.py#L61) rather than [`core:index`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/views.py#L39). Since every user has to eventually go through the agency entry page and will read the content, we'd like to simplify the view on the application index to help avoid confusion and direct the user into the main flow. 

### New look

image

## Tasks

* [x] #433
* [ ] #432
* [ ] #431

MattRenda:
Resolves #366 

Updates the core.pages.index.content_title text in English and Spanish

Removes all paragraph content from the viewmodel in core:index

Updates the two buttons side-by-side

## Screenshots

image
image

### DIFF
--- a/benefits/core/templates/core/page.html
+++ b/benefits/core/templates/core/page.html
@@ -54,7 +54,8 @@
 
             {% block buttons %}
             {% if page.buttons|length_is:"0" %}{% else %}
-                <div class="buttons">
+                <h2 style="margin-top:20vh; font-weight:bold">{{page.choose_provider}}</h2>
+                <div style="display:flex; justify-content: space-between" class="buttons">
                     {% for b in page.buttons %}
                         {% include "core/includes/button.html" with button=b %}
                     {% endfor %}
--- a/benefits/core/templates/core/page.html
+++ b/benefits/core/templates/core/page.html
@@ -54,8 +54,7 @@
 
             {% block buttons %}
             {% if page.buttons|length_is:"0" %}{% else %}
-                <h2 style="margin-top:20vh; font-weight:bold">{{page.choose_provider}}</h2>
-                <div style="display:flex; justify-content: space-between" class="buttons">
+                <div class="buttons">
                     {% for b in page.buttons %}
                         {% include "core/includes/button.html" with button=b %}
                     {% endfor %}
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -30,6 +30,11 @@ def _index_url():
     return reverse("core:index")
 
 
+def _index_choose_provider():
+    """Helper returns the content title for the common index page."""
+    return _("core.pages.index.chooseprovider")
+
+
 @middleware.pageview_decorator
 def index(request):
     """View handler for the main entry page."""
@@ -39,11 +44,11 @@ def index(request):
     agencies = models.TransitAgency.all_active()
     buttons = [viewmodels.Button.outline_primary(text=a.short_name, url=a.index_url) for a in agencies]
     buttons[0].classes.append("mt-3")
-    buttons[0].label = _("core.pages.index.chooseprovider")
+    buttons[1].classes.append("mt-3")
 
     page = viewmodels.Page(
         content_title=_index_content_title(),
-        paragraphs=_index_paragraphs(),
+        choose_provider=_index_choose_provider(),
         buttons=buttons,
         classes="home",
     )
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -30,11 +30,6 @@ def _index_url():
     return reverse("core:index")
 
 
-def _index_choose_provider():
-    """Helper returns the content title for the common index page."""
-    return _("core.pages.index.chooseprovider")
-
-
 @middleware.pageview_decorator
 def index(request):
     """View handler for the main entry page."""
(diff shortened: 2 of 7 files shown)


benefits/core/templates/core/page.html
```diff
@@ -54,7 +54,8 @@ <h1>{{ page.content_title }}</h1>
 
             {% block buttons %}
             {% if page.buttons|length_is:"0" %}{% else %}
-                <div class="buttons">
+                <h2 style="margin-top:20vh; font-weight:bold">{{page.choose_provider}}</h2>
+                <div style="display:flex; justify-content: space-between" class="buttons">
```
machikoyasuda, line 58: Feel free to create your own new classes and styles in `styles.css` as necessary. We prefer styles in `styles.css` over inline styles like this.

machikoyasuda:
Please create and use CSS classes/styles in `styles.css`, rather than using inline styles. Aside from that, everything looks good to me.

benefits/core/templates/core/page.html
thekaveman: I'll also add, doing this here in the `page.html` template re-styles every button in the app, and not just the homepage buttons.

I think the right approach is to create a new `index.html` template that `{% extends "core/page.html" %}` and overrides the rendering of buttons for this page specifically.

thekaveman (changes requested):
Left a [similar comment](https://github.com/cal-itp/benefits/pull/406/files#r844336368) on @machikoyasuda's review, but restating again:

I think the better approach here is to create a new template (`.html` file) specifically for the homepage, so you can override the rendering of buttons there - rather than adding a new attribute to the `viewmodels` for only this one page. It will also let you get the classes in (see Machiko's comments) specifically for these side-by-side homepage buttons.

See e.g. how the [`help.html` template](https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/help.html) `{% extends "core/page.html" %}`, then uses the `{% block %}` definitions from `page.html` to place content or override content in certain areas.

Then in the view you can [specify which template file to use](https://github.com/cal-itp/benefits/blob/dev/benefits/core/views.py#L90).

angela-tran (changes requested):
I had a chance to review this today 😄 and left some comments about `viewmodel.Page`.

benefits/static/css/styles.css
```diff
@@ -200,7 +200,10 @@ footer {
   width: 100%;
   text-align: center !important;
 }
-
+.provider {
```
thekaveman, line 203: This class doesn't say enough about what this is or why we have this style.

I'd like it to be a little more descriptive like `.index-agencies-label` or similar.

thekaveman (changes requested):
One minor change to the class for the label, to be more descriptive.

In general, please also be mindful of what files/changes are included in your commits. The latest commit included a change to the file `tests/cypress/package-lock.json` ... we have a long standing issue with this file constantly getting changed/updated and you didn't do anything wrong to cause the change in that file.

But it also wasn't part of what you were working, so the file shouldn't have been included in the commit. Not a big deal this time but please keep it in mind.

benefits/static/css/styles.css
```diff
@@ -177,7 +177,7 @@ footer {
   font-weight: 500;
   margin-bottom: 1rem;
   padding: 1rem;
-  width: 100%;
+  width: 49%;
```
angela-tran, line 180: This is affecting other pages. I _think_ putting `width: 100%` back and moving `width: 49%` into your style class should solve it, but please check it.

Comparison of the agency-index page on your branch vs. the `dev` branch:
image

image

benefits/static/css/styles.css
```diff
@@ -263,13 +266,6 @@ footer {
     margin-top: 8rem;
     margin-bottom: 7rem;
   }
-
-  .container.content input[type="submit"],
-  .container.content .btn {
-    width: 70%;
-    margin-left: 15%;
-    margin-right: 15%;
-  }
```
angela-tran, line 272: Are you sure these are ok to remove? i.e. is it possible there were other pages that should have these styles?

benefits/core/views.py
```diff
@@ -12,7 +12,7 @@
 
 def PageTemplateResponse(request, page_vm):
     """Helper returns a TemplateResponse using the common page template."""
-    return TemplateResponse(request, "core/page.html", page_vm.context_dict())
+    return TemplateResponse(request, "core/index.html", page_vm.context_dict())
```
angela-tran, line 15: `PageTemplateResponse` is used in many places, so you are potentially changing a lot more than you should for the effect you are trying to get (to make the `index` function use the `core/index.html` template).

The return statement of [`index`](https://github.com/cal-itp/benefits/pull/406/files#diff-2e5e461b0d86b207a87071f948630540090d2d020e2b0e053b5486e6ac5c82fdR51) is what you should be changing. Feel free to ask me any questions if you feel unsure on this!

angela-tran (changes requested):
Thanks for addressing the `viewmodels.Page` field @MattRenda.

While re-reading your changes, I noticed a few other issues and left comments below. I think it would be super helpful to us reviewers if you could break up your changes into more atomic commits. With smaller commits that each have their own specific message, it is easier to understand the intent behind your changes.

angela-tran:
Also, I don't see this change: `Updates the core.pages.index.content_title text in English and Spanish`

The screenshot on https://github.com/cal-itp/benefits/issues/366 shows what the title text should be.

benefits/core/views.py
thekaveman: I didn't notice this, thanks for catching!

angela-tran:
@machikoyasuda offered the perspective that the scope of #366 might be too large for one PR. She worked on breaking it up into smaller tickets.
 - #431 
 - #432 
 - #433

PR #402, opened 2022-04-05
Feat: add auth button to Eligibility start page

thekaveman:
Closes #380.

This PR just adds the button with a minimal amount of styling for the button itself; the placement on the page still needs to coordinate with #395 

To get this to work, we also needed sample AuthProvider data, which is included as a fixture here. Rebuild your database to get these changes:

```bash
rm django.db

bin/init.sh
```

The existing `eligibility.spec.js` UI tests were re-written for the non-`AuthProvider` path, to keep the tests close to their original state as possible. A new UI test block was introduced for the Eligibility start page, with a test that verifies the Login.gov button appears.

## What it looks like

**Normal**

image

**Hover**

image

Issue #373, opened 2022-03-30
Consume authentication server response data


thekaveman:
We need a way to take the results of the Login.gov flow (whatever indicates that the user was successful in that flow) and store it via #370.

The scope of this ticket is handling the submission back to our app on the Django side. The sign on button itself will be implemented in #375.

In any case, this PR needs to build upon #375's solution to get the OIDC client library.

## More information from sample

In the sample Login.gov integration app, after the user successfully authenticates with Login.gov, the auth flow sends the user to a `redirect_uri`. This page must call a library function `signinRedirectCallback()` to complete the login.

In the sample, this is implemented with a simple `callback.html` file like:

```html
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title></title>
</head>
<body>
    <script src="libs/oidc-client.js"></script>
    <script>
        new Oidc.UserManager({ response_mode: "query" }).signinRedirectCallback().then(function () {
            window.location = "index.html";
        }).catch(function (e) {
            console.error(e);
        });
    </script>
</body>
</html>
```

## Tasks

* [ ] Create a [new urlpath](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/urls.py) `auth_redirect`
* [ ] Create a [new form](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/forms.py) `AuthRedirect`
* [ ] Create a [new template](https://github.com/cal-itp/benefits/tree/dev/benefits/eligibility/templates/eligibility) `auth-redirect.html`
  * [ ] Extend the `core/page.html` template
  * [ ] Following the above example, submit the form on the page (instead of setting the `window.location`). See [`enrollment:index`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L40) for a similar example
* [ ] Create a new view function `auth_redirect(request)`:
  * [ ] `GET` should send the form to the template, see e.g. [`eligibility:start`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L83)
  * [ ] `POST` should update the session to `auth = True` (#370) and redirect to `eligibility:confirm`

thekaveman:
Closing this in favor of a server-only, see revised #375.

Issue #371, opened 2022-03-30
Create middleware that expects user to be authenticated


thekaveman:
## Background

Once the session tracks the sign in status in #370, we can make use of that infrastructure to put guards on certain view functions that require a user to be authenticated with Login.gov. The first step is to create a new middleware class that enforces the requirement.

This is similar to how the [`AgencySessionRequired`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/middleware.py#L20) and [`EligibileSessionRequired`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/middleware.py#L68) middleware are used.

See also #323 / #333 for a very similar task from last Sprint.

We also need the feature flag from #372.

## Tasks

- [x] Create a new middleware class like `AuthSessionRequired` inheriting from `MiddlewareMixin`, see the other `*SessionRequired` as examples
- [ ] In `process_request()`:
  - If the session's verifier is empty, or does not have auth enabled, return `None` to allow the request to continue
  - If the session's verifier has auth enabled, and the session's auth status is `False`, raise an error to stop the request
  - Otherwise return `None` to allow the request to continue
- [ ] Apply this middleware to the following views to enforce that the user must be signed-in:
  - [ ]  [`eligibility:confirm`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L90)
  - [ ]  [`enrollment:token`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py#L96)
  - [ ]  [`enrollment:index`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py#L109)

thekaveman:
#414 was merged, and since it was built on #409, it closed this issue.

Reopening since we still need this.

PR #407, opened 2022-04-06
Feat: implement auth button with OIDC

### LINKED ISSUE #375: Implement OIDC sign in flow

thekaveman:
Closes #375.

## Choosing a client library

We need to reference an OIDC Client library, which does the heavy lifting for authentication with Login.gov. There were at least three options:

* The original, but deprecated since June 2021 [`oidc-client-js`](https://github.com/IdentityModel/oidc-client-js)
* A TypeScript fork of the original, still actively maintained [`oidc-client-ts`](https://github.com/authts/oidc-client-ts)
* Perhaps more overall users, but not as recently updated [`AppAuth-JS`](https://github.com/openid/AppAuth-JS)

Since the original was initially recommended but no longer maintained, we decided on the TypeScript fork `oidc-client-ts`

## Referencing the client library

`oidc-client-ts` is an NPM package and expects to be consumed through that ecosystem; `node` and `npm` had to be installed in the _application_ container in order to bring this dependency in. The installation was moved from the _devcontainer_, where it was previously needed only for the Cypress tests.

This is tricky though: Django needs static files (like JavaScript, CSS, images) to live inside the `benefits/static/` directory. This is because During [application startup, Django compiles static assets](https://github.com/cal-itp/benefits/blob/dev/bin/init.sh#L27) for serving, the behavior differing between app container and devcontainer runs. Read more about [Deploying static files in Django](https://docs.djangoproject.com/en/3.2/howto/static-files/deployment/). 

Since `npm` installs its packages in `node_modules/` at the root of the project directory (repo), we have to do some creative copying of files during the right time (app container build, devcontainer postAttach) to get the library in the right folder, making it referenceable from a template. Entries in `.gitignore` ensure we don't commit the library as part of our source code.

## Using the client library

The configuration data needed to make use of the library is modeled by the `AuthProvider` model, which was defined in #392 and a sample added in #402. The `EligibilityVerifier` model was extended with a "feature flag" that indicates if auth is required: we make use of that here to opt-in to requiring auth and referencing the client library.

When auth is required, the `AuthProvider` model instance is sent to the Start page template, where it can be used to generate the JavaScript block that will execute when the user eventually clicks the button.

## Result of this PR

Assuming the correct `AuthProvider` settings (e.g. edit the sample added in #402), clicking the "Continue with Login.gov" button takes the user through the login flow with Login.gov. After successfully signing in, the user is redirected to the (currently pre-configured) `redirect_uri`, which does not yet exist in this app. See #373.

## Reviewing this PR / after merging

You'll want to rebuild both the application and devcontainers on your host machine after pulling these changes in. If you're updating your `fixtures/` to configure an `AuthProvider`, delete the Django database as well:

```
rm django.db

cd .devcontainer/

docker compose build --no-cache dev
```

Then *Reopen in Devcontainer* with VS Code.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,12 +2,8 @@ FROM benefits_client:latest
 
 USER root
 
-# install node.js
-# see https://github.com/nodesource/distributions#installation-instructions
-
-RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash
-
-RUN apt-get install -qq nodejs npm libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev \
+# install Cypress prereqs and other dev tooling
+RUN apt-get install -qq libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev \
     libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb curl git jq ssh
 
 USER $USER
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -1,14 +1,18 @@
 #!/usr/bin/env bash
 set -eu
 
-# initialize hook environments
+echo [initialize hook environments]
 pre-commit install --install-hooks --overwrite
 
-# manage commit-msg hooks
+echo [manage commit-msg hooks]
 pre-commit install --hook-type commit-msg
 
-# copy Login.gov client library
+echo [install npm packages for app]
+npm ci
+mkdir -p benefits/static/js
 cp node_modules/oidc-client-ts/dist/browser/oidc-client-ts* benefits/static/js
 
-# install cypress
-cd tests/cypress && npm install && npx cypress install
+echo [install npm packages for tests]
+cd tests/cypress
+npm ci
+npx cypress install
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -7,5 +7,8 @@ pre-commit install --install-hooks --overwrite
 # manage commit-msg hooks
 pre-commit install --hook-type commit-msg
 
+# copy Login.gov client library
+cp node_modules/oidc-client-ts/dist/browser/oidc-client-ts* benefits/static/js
+
 # install cypress
 cd tests/cypress && npm install && npx cypress install
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
 set -eu
 
+echo [install npm packages for app]
+npm ci
+mkdir -p benefits/static/js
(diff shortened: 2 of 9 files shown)


afeld:
Curious: did you look into doing the implementation server-side? We'll need to have the rest of the OIDC logic there anyway, and seems like it would be more consistent with the app having minimal JavaScript.

Happy to pair on this — especially since https://github.com/cal-itp/benefits/issues/371 is so closely related — if helpful!

thekaveman:
Closing this PR, we're going to do this differently using `Authlib` from the server.

afeld:
Cross-posting [an earlier message from Slack](https://cal-itp.slack.com/archives/C037Y3UE71P/p1649343121439999) to be able to reference it later:

> we've been moving towards integrating the [login.gov](http://login.gov/) OpenID Connect flow in the Django app directly, but we could alternatively go the proxy direction … where the redirects and all that happen before a request makes its way to Django. some options:
>
> - [Amazon Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-oidc-flow.html) / [ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/listener-authenticate-users.html)
> - [Cloudflare Zero Trust](https://developers.cloudflare.com/cloudflare-one/identity/idp-integration/generic-oidc/)
> - [Google Cloud Identity Platform](https://cloud.google.com/identity-platform)
> - [OpenResty](https://github.com/zmartzone/lua-resty-openidc) (NGINX)
> - [NGINX Plus](https://github.com/nginxinc/nginx-openid-connect)
> - [OAuth2 Proxy](https://oauth2-proxy.github.io/oauth2-proxy/)
>
> NGINX has [a diagram that shows how that flow would work](https://github.com/nginxinc/nginx-openid-connect#description), which would basically apply to all of them.

PR #414, opened 2022-04-10
Feat: implement OAuth/OIDC sign on with Authlib

### LINKED ISSUE #375: Implement OIDC sign in flow

thekaveman:
Closes #375.

Replaces #407 (we're doing this all on the server now).
Replaces #409 (we don't need Django's auth system/user models).

## Reviewing this PR / after merge

1. Delete `django.db`
2. Update `.env` with new variables / values from CDT sample
3. Rebuild app container `docker compose build --no-cache client`
4. Rebuild devcontainer `docker compose build --no-cache dev`

See more at https://docs.calitp.org/benefits/configuration/oauth


benefits/oauth/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+_oauth = OAuth()
+
+
+@decorator_from_middleware(middleware.AgencySessionRequired)
+@decorator_from_middleware(middleware.VerifierSessionRequired)
+def login(request):
+    verifier = session.verifier(request)
+    if not verifier.requires_authentication:
+        return redirect(session.origin(request))
```
afeld, line 18: While this is presumably working now, I worry a bit about the complexity of our session states, and having bugs that put users in incompatible states. Will noodle on that a bit as we go.

As an example of how to simplify, do we actually need to check `AgencySessionRequired`, `VerifierSessionRequired`, and `verifier.requires_authentication` here? Ditto for `authorize`.

benefits/oauth/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+    oauth_client = _oauth.create_client(verifier.auth_provider.name)
```
afeld, line 20: I'm assuming this [verifies the token](https://developers.login.gov/oidc/#token-response), though [the Authlib docs](https://docs.authlib.org/en/latest/client/frameworks.html?highlight=authorize_access_token#parsing-id-token) don't that say that explicitly.

benefits/oauth/urls.py
```diff
@@ -0,0 +1,11 @@
+from django.urls import path
+
+from . import views
+
+
+app_name = "oauth"
+urlpatterns = [
+    # /oauth
+    path("login", views.login, name="login"),
+    path("authorize", views.verify, name="authorize"),
```
angela-tran, line 10: I'm getting an error when the devcontainer tries to run `bin/init.sh`. It's saying there is no `verify` function in `oauth/views.py`. I'm guessing that should be `views.authorize` instead?

```
  File "/home/calitp/app/benefits/oauth/urls.py", line 10, in <module>
    path("authorize", views.verify, name="authorize"),
AttributeError: module 'benefits.oauth.views' has no attribute 'verify'
```

benefits/oauth/views.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if not oauth_client:
+        return redirect(session.origin(request))
```
angela-tran, line 23: I don't think there's anything wrong with doing this redirect, but would it be more appropriate to raise an exception instead? It seems like we'd get into this branch if we forgot to add the auth provider's configuration to `.env` - i.e. this is a state we don't expect to be in.

angela-tran (changes requested):
This approach looks good to me! I noticed a few things causing errors, so I left suggestions as well as a question.

benefits/eligibility/views.py
```diff
@@ -58,12 +58,11 @@ def index(request):
...
     """View handler for the eligibility verification getting started screen."""
 
-    session.update(request, eligibility_types=[])
+    session.update(request, eligibility_types=[], origin=reverse("eligibility:start"))
     verifier = session.verifier(request)
 
     if verifier.requires_authentication:
-        auth_provider = verifier.auth_provider
-        button = viewmodels.Button.external(text=_(auth_provider.sign_in_button_label), url="#", id="login")
+        button = viewmodels.Button.external(text=_(verifier.sign_in_button_label), url=reverse("oauth:login"), id="login")
```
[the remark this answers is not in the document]
thekaveman: Yeah this was a mistake, shouldn't have been committed. I'm pushing a fix for this in a sec!

benefits/oauth/views.py
thekaveman: Hmm... yeah this definitely indicates a bad configuration, mismatch between the fixtures and settings.py or something. Exception is probably a better idea! 👍

afeld: Should we have a fallback one? Seems a bit strange that someone getting to the `login` URL (one way or another) might not actually be logged in.

benefits/settings.py
```diff
@@ -70,9 +71,10 @@ def _filter_empty(ls):
 CSRF_COOKIE_HTTPONLY = True
 CSRF_TRUSTED_ORIGINS = _filter_empty(os.environ["DJANGO_TRUSTED_ORIGINS"].split(","))
 
-SESSION_COOKIE_SAMESITE = "Strict"
+SESSION_COOKIE_SAMESITE = "Lax"
```
afeld, line 74: Can you say more about this change?

afeld (commented):
We are definitely going to want some tests for this functionality; want to do them in this pull request, or create a separate issue?

benefits/core/migrations/0001_initial.py
```diff
@@ -29,17 +29,7 @@ class Migration(migrations.Migration):
...
-                ("client_id", models.TextField(help_text="The application ID registered with auth server")),
-                ("redirect_uri", models.TextField(help_text="Post-login, the user is redirected here")),
-                ("post_logout_redirect_uri", models.TextField(help_text="Post-logout, the user is redirected here")),
-                ("response_type", models.TextField(help_text='Must be "code" to use the authorization-code flow')),
-                (
-                    "scope",
-                    models.TextField(
-                        help_text='Space-delimited list. If you need refresh tokens, you must add the "offline access" scope'
-                    ),
-                ),
```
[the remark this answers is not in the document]
thekaveman: Yep the bulk of these moved to env/settings

benefits/oauth/views.py
thekaveman: We definitely need to check `VerifierSessionRequired` since we need the `verifier` later to get the `oauth_client`.

`AgencySessionRequired` isn't technically needed / since `verifier` comes from the agency originally. But it's a valid sanity check - the user shouldn't be here if they don't have an agency in their session.

On `requires_authentication`... I could go either way. Theoretically the user doesn't get here unless their verifier requires auth, so we could assume that.

benefits/oauth/views.py
thekaveman: Not sure if you meant this comment for another line? This line just creates the client based on the settings.

benefits/oauth/views.py
```diff
@@ -0,0 +1,44 @@
...
+@decorator_from_middleware(middleware.AgencySessionRequired)
+@decorator_from_middleware(middleware.VerifierSessionRequired)
+def authorize(request):
+    verifier = session.verifier(request)
+    if not verifier.requires_authentication:
+        return redirect(session.origin(request))
+
+    oauth_client = _oauth.create_client(verifier.auth_provider.name)
+
+    token = oauth_client.authorize_access_token(request)
```
afeld, line 38: I'm assuming this [verifies the token](https://developers.login.gov/oidc/#token-response), though [the Authlib docs](https://docs.authlib.org/en/latest/client/frameworks.html?highlight=authorize_access_token#parsing-id-token) don't that say that explicitly.

benefits/oauth/views.py
afeld: Oops! Indeed.

benefits/settings.py
thekaveman: Yeah there was a big Slack thread about this, will tag you in again.

TLDR; with `Strict` the user loses their Django session between leaving our app to sign in with Login.gov, and coming back into our app from the Login.gov redirect. This is because `Strict` disallows cookies from external domains and so the session cookie is lost.

`Lax` allows the cookie to travel with the user and be sent back to us by Login.gov, as long as the request is "safe" i.e. GET, which it is.

benefits/eligibility/views.py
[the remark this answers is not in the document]
thekaveman: In my testing, clicking the button when already logged in just took me straight through the flow (didn't make me login again).

benefits/oauth/views.py
angela-tran: @afeld Do you mean a fallback that results in the user effectively being seen as logged in? I'm not sure I see the need for that since, other than the case that's handled by `if not verifier.requires_authentication:`, we would always want the user to go through a real authentication process. (Maybe I misunderstood your suggestion though?)

thekaveman: I was a bit confused by @afeld wording too - I think the question is about having a fallback OAuth client to use, in the case that we can't find the one named by the `auth_provider`.

I'm inclined to throw the Exception rather than try to figure out a fallback client. It really shouldn't happen / be needed, and we should like to know if the configuration is bad.

Agreed that the redirect back out (possibly indicating a successful login) is probably not the best approach.

benefits/eligibility/views.py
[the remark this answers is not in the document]
thekaveman: Fixed in [264b65f](https://github.com/cal-itp/benefits/pull/414/commits/264b65f84e1c62faea655c78ac308e2c2ef03ebe)

benefits/oauth/urls.py
thekaveman: Fixed in [5ff71c9](https://github.com/cal-itp/benefits/pull/414/commits/5ff71c9a6a2c927f3e5dbf14fb66a942ef2b4f4f)

benefits/oauth/views.py
thekaveman: So addressing both of your comments in this thread and @afeld [other related comments](https://github.com/cal-itp/benefits/pull/414#discussion_r847693486):

I moved the `client_name` parameter in `_oauth.create_client()` here to be a setting rather than a property of the verifier. This simplifies the views, to where they don't really have to care about the `session`. We basically have a "on" / "off" for OAuth for the app, rather than per-`verifier`.

This also means means an Exception is thrown if these view functions are called and an OAuth client wasn't configured.

benefits/oauth/views.py
thekaveman: I posted [a follow-up](https://github.com/cal-itp/benefits/pull/414#discussion_r847845386) to this and other comments in another thread - refactored the code to address simplification suggestion

benefits/core/migrations/0001_initial.py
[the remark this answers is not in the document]
thekaveman: Even more moved in [fc413a6](https://github.com/cal-itp/benefits/pull/414/commits/fc413a6265426704982e59f742130ba614103eab)

benefits/settings.py
afeld: Moving the comment into the code for longevity:

```suggestion
# With `Strict`, the user loses their Django session between leaving our app to sign in with Login.gov, and coming back into our app from the Login.gov redirect. This is because `Strict` disallows cookies from external domains and so the session cookie is lost.
#
# `Lax` allows the cookie to travel with the user and be sent back to us by Login.gov, as long as the request is "safe" i.e. GET, which it is.
SESSION_COOKIE_SAMESITE = "Lax"
```

benefits/oauth/views.py
```diff
@@ -0,0 +1,37 @@
...
+def authorize(request):
+    if not oauth_client:
+        raise Exception("No OAuth client")
+
+    token = oauth_client.authorize_access_token(request)
+
+    if token is None:
+        return redirect(session.origin(request))
+    else:
+        session.update(request, auth=True)
```
afeld, line 36:
```suggestion
        # we are intentionally not storing anything about the user, including their token
        session.update(request, auth=True)
```

thekaveman: This comment may change if we don't do the logout immediately after - we'll need to store the token. Thoughts?

benefits/settings.py
thekaveman: See [842540b](https://github.com/cal-itp/benefits/pull/414/commits/842540b57e93f3c12c68723328afd478f4de5b39)

benefits/oauth/views.py
afeld: Agreed; just describing the current state.

benefits/oauth/views.py
angela-tran: Agreed that it would be nice for the docs to say it more clearly. Looking through the source code, I think it does verify the token -- feel to check my reading of it:
 - `django_client`'s `OAuth` class inherits from `BaseOAuth` and is set up to use certain [classes](https://github.com/lepture/authlib/blob/master/authlib/integrations/django_client/__init__.py#L8)
 - [`_oauth.create_client`](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/registry.py#L68) will use `DjangoOAuth2App` due to our [configuration](https://github.com/cal-itp/benefits/pull/414/files#diff-d71405aa44f1e899a9cd920e5187aa1d02affdbd0da472eca95f814b0a8c1210R166-R170)
 - [`oauth_client.authorize_access_token`](https://github.com/lepture/authlib/blob/master/authlib/integrations/django_client/apps.py#L57) will internally call the [`parse_id_token` function](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/sync_openid.py#L32) that it gets from inheriting `OpenIDMixin`
 - `parse_id_token` [decodes](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/sync_openid.py#L66) the `id_token` [into claims](https://github.com/lepture/authlib/blob/098bb16d7068953280bbef456bee3ff5954f19e7/authlib/jose/rfc7519/jwt.py#L86), and [calls `validate`](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/sync_openid.py#L76) on those decoded claims
    - The implementation we'd run through is [`JWTClaims.validate`](https://github.com/lepture/authlib/blob/098bb16d7068953280bbef456bee3ff5954f19e7/authlib/jose/rfc7519/claims.py#L91)

angela-tran: Hmm, as I read through `JWTClaims,validate`, it looks like it's just checking if fields are present? Not sure where to find the verification we're hoping to see.

afeld:
> Update `.env` with new variables / values from CDT sample

Can we put those instructions in the repository somewhere?

benefits/eligibility/views.py
```diff
@@ -58,12 +58,12 @@ def index(request):
...
     """View handler for the eligibility verification getting started screen."""
 
-    session.update(request, eligibility_types=[])
+    session.update(request, eligibility_types=[], origin=reverse("eligibility:start"))
     verifier = session.verifier(request)
 
     if verifier.requires_authentication:
         auth_provider = verifier.auth_provider
-        button = viewmodels.Button.external(text=_(auth_provider.sign_in_button_label), url="#", id="login")
+        button = viewmodels.Button.external(text=_(auth_provider.sign_in_button_label), url=reverse("oauth:login"), id="login")
```
angela-tran, line 66: 😕 I am getting a `NoReverseMatch` error from this `reverse("oauth:login")`. Do you have any ideas on what might be going on?

image

thekaveman:
> > Update `.env` with new variables / values from CDT sample
> 
> Can we put those instructions in the repository somewhere?

Most appropriate would probably be the [`Getting Started` docs](https://github.com/cal-itp/benefits/tree/dev/docs/getting-started)? A new page/section talking about this OAuth configuration.

benefits/eligibility/views.py
thekaveman: So this is probably related to [@afeld comment above](https://github.com/cal-itp/benefits/pull/414#issuecomment-1097083063), in that the [`.env` needs to be updated with the new keys/values](https://github.com/cal-itp/benefits/pull/414/files#diff-e43af9f217e778e159273c8b2c475150122ad0ccf0525cb2bce1edd237ca5f03) and reread.

We could put a check here to ensure that the `OAUTH_CLIENT_NAME` setting has a value or raise an Exception, which would give a clearer error message than this one.

benefits/oauth/views.py
thekaveman: Here's what `token` looks like after the call to `authorize_access_token()`:

image

So that contains some parsed `userinfo`.

benefits/eligibility/views.py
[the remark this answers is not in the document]
thekaveman: See [0da50ab](https://github.com/cal-itp/benefits/pull/414/commits/0da50ab24804edbf512e527e5340480532c85996) for omitting this button if the user is logged in from the app's perspective.

benefits/eligibility/views.py
angela-tran: > We could put a check here to ensure that the `OAUTH_CLIENT_NAME` setting has a value or raise an Exception, which would give a clearer error message than this one.

I think this is a good idea. I vote we do this ☑️

angela-tran (commented):
@thekaveman I like your idea to add an explicit check for the presence of OAuth configuration in the `eligibility/start` view. I think once that is added and you are able to run through the authentication flow with this current iteration of the code, I'd say this is ready to merge.

I am currently not able to test out the authentication flow locally, but as @afeld mentioned in Slack, this probably has more to do with the fluctuating state of the CDT authorization server.

angela-tran (dismissed):

afeld (approved):
Will be nice to have this in since so many things hang off of it. My only hesitation with this being merged is if we are unable to log in due to configuration issues with the authorization server. If we're willing to accept that, let's do it!

thekaveman:
Holding on the merge until everyone can get their local environment configured

angela-tran (dismissed):

thekaveman:
Ah, looks like I need to fix the merge conflicts. Will do so.

angela-tran (dismissed):
Oh lol, I thought it would let you resolve the merge conflict without needing a re-review, but makes sense! Looks good

afeld (approved):

angela-tran (approved):

PR #425, opened 2022-04-12
refactor: speed up Docker image rebuild times

afeld:
Switching branches was getting time-consuming; this moves the expensive operation of installing the Python packages earlier in the `Dockerfile` so that changes to application/configuration files don't require all packages to be reinstalled.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,12 @@ RUN useradd --create-home --shell /bin/bash $USER && \
 
 # enter app directory
 WORKDIR /home/$USER/app
+# switch to non-root $USER
+USER $USER
+
+# install python dependencies
+COPY requirements.txt requirements.txt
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -28,6 +28,12 @@ RUN useradd --create-home --shell /bin/bash $USER && \
 
 # enter app directory
 WORKDIR /home/$USER/app
+# switch to non-root $USER
+USER $USER
```
thekaveman, line 32: Part of the idea of delaying the switch in `USER` was to copy the app source files using root; so they are readable but not writeable by the container process.

Should we be concerned about that?

thekaveman (commented):
Generally in favor of this! One comment on how we copy the source files in to the image.

thekaveman (approved):
👍

Issue #428, opened 2022-04-13
Place Login.gov sign out button


thekaveman:
## Background

Once the user goes through the Login.gov flow and has successfully signed in to their Login.gov account, they are redirected back to the Benefits application.

We want to give the user a visual indication that "something has changed" with their status; we also want to provide a safe off-ramp for users that may not be able to complete the flow before stepping away from their browser session.

This task is just about placing the button with the right visual look/feel on the correct pages.

## Button placement, look, and feel

As a guide to where to place a sign out button, here are the latest mockups where this appears:

![image](https://user-images.githubusercontent.com/25497886/165826758-b1a10169-2740-4a40-9e24-78b2d7a354f7.png)

![image](https://user-images.githubusercontent.com/25497886/165826768-7b1dc959-72d3-4944-95c8-37884ba877fb.png)

## Tasks

* [x] Display on `eligibility:confirm`
* [x] Display on `enrollment:index`
*  ~Display on `enrollment:success`~ `enrollment:success` uses a different visual that was implemented by #466
* [x] Only display when auth is enabled, and the user is logged in

afeld:
@srhhnry working on updated mockups.

machikoyasuda:
These pages will also need the Log out button:

- [ ] Enrollment unverified
- [ ] Eligibility unverified

PR #436, opened 2022-04-14
Feat: Implement OAuth/OIDC and app sign out

### LINKED ISSUE #368: Implement OIDC sign out
## Background

Once the user goes through the Login.gov flow and has successfully signed in to their Login.gov account, they are redirected back to the Benefits application.

We want to give the user a visual indication that "something has changed" with their status; we also want to provide a safe off-ramp for users that may not be able to complete the flow before stepping away from their browser session. The button will be styled and placed in #428

With #370, the user session will track whether they have authenticated or not. The implementation for #375 is server-side, and "sign out" has to happen for both the OAuth server and the benefits application.

## Sign in, sign out

@afeld writes:

> Since we don’t have a need to leverage their token past the initial authentication, we _could_ log them out from Login.gov immediately after the user reaches the callback on the Django app.

### Button

* [ ] Point to endpoint / view in `oauth` app to handle updating session to `auth = False` (build on #414)
* [ ] Redirect back to `eligibility:start`

### Sign in flow

* [ ] Update `authorize()` view to instead [redirect user through `end_session_endpoint`](https://github.com/lepture/authlib/issues/331#issue-838728145)

thekaveman:
Closes #368.

Consolidated #464 into this PR.

## Summary

This PR adds a new route `/oauth/logout` that signs the user out of all 3 environments: Login.gov, the auth server, and the Django app.

We have to store the `id_token` that we get back as part of the `/oauth/authorize` callback, in order to later be able to sign the user out. Because of this, session management was updated to differentiate between the `oauth_token` and the `enrollment_token`.

The sign out flow is:

1. User clicks the sign out button (#428)
2. `/oauth/logout` retrieves the (now stored) OAuth token from user's Django session
3. User's Django session OAuth token is cleared 
4. A redirect URI is generated for `/oauth/post_logout`
5. User is redirected through the auth server's `end_session_endpoint` with their token and the redirect URI
6. Once back in the app, `/oauth/post_logout` redirects the user to `session.origin()`

We have a number of places the user can sign out from:

* The `/eligibility/confirm` page - redirects back to `/eligibility/start`
* The `/enrollment` page - redirects back to `/enrollment`
* The `/enrollment/success` page - redirects back to `/enrollment/success`

* ~Each of these post-logout redirect URIs has to be registered with the auth server~

🔎 **Important to call out:** once the user passes eligibility verification (they are verified as eligible for the discount) we don't _need_ to make them sign in with OAuth again (since the "OAuth gate" has already been crossed and the protected page already completed).

When the user makes it to the `/enrollment` phase, their session will indicate they have been verified as eligible. So on sign out, we can redirect them back into the `/enrollment` phase vs. making them restart with `/eligibility`.

## Tests

`logout` test added in [`5272f5b`](https://github.com/cal-itp/benefits/pull/436/commits/5272f5b82156560aeedd5670a7a5bdeb2effd918):

* Creates a mock of `_deauthorize_redirect` to avoid going to the real auth server
* Verifies `oauth_token` is cleared from session
* Verifies the token and correct `post_logout` redirect URI are sent to `_deauthorize_redirect` 

`post_logout` test added in [`d360f48`](https://github.com/cal-itp/benefits/pull/436/commits/d360f48379514d24ff7a7e4f746dfa18eff07ae5):

* Updates session `origin`
* Checks that `post_logout` redirects to `origin`

`_deauthorize_redirect` tests added in [`3f8db89`](https://github.com/cal-itp/benefits/pull/436/commits/3f8db89cfd997dae4c897e5a25b53f21e63bdef1):

* Load mock server metadata
* Checks redirect to `end_session_endpoint` with correct `id_token_hint` and `post_logout_redirect_uri`

`_generate_redirect_uri` tests added in [`2e291f2`](https://github.com/cal-itp/benefits/pull/436/commits/2e291f2c50007e743fb19aba389bdee41149d4a8):

* HTTP redirect is generated for `localhost`
* HTTPS redirect is generated otherwise

Also added a general helper fixture [`session_request`](https://github.com/cal-itp/benefits/pull/436/files#diff-b4308b616ed5ec54421fc66ee8b3a76c4d2f4272ea20a48808c7f6a9d3f00a32) based on @afeld work in #445 to initialize a mock request with a session.

Note I added the [`pytest-mock`](https://github.com/pytest-dev/pytest-mock/) package to `tests/pytest/requirements.txt` so you will want to `pip install` before running this test.

## Reviewing

1. Run the app locally with the configuration for the dev CDT auth server
2. Sign in through the Login.gov sandbox with your test account
3. Once back in the app, there are 3 different logout scenarios (see list above) - manually update the URL to `/oauth/logout` for each:
    * from the Eligibility Confirm page
    * from the Enrollment index (update `compose.yml` for `server`, `image: ghcr.io/cal-itp/eligibility-server@sha256:337d5b2beb1e458980be49a778efd4a47f8daa8decc5e8329c0d528596e2f196` and restart devcontainer to get a server running. `A1234567` / `Garcia` for test data)
    * from the Enrollment success (after passing Eligibility, manually update URL to `/enrollment/success`)
4. You should briefly see a "You are now logged out" message from the auth server, before being redirected back to the corresponding page (see above list)

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -110,7 +110,7 @@ def start(request):
             ),
         )
 
-        if not session.auth(request):
+        if not session.oauth_token(request):
             button = viewmodels.Button.external(
                 text=_(verifier.auth_provider.sign_in_button_label),
                 url=reverse("oauth:login"),
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -198,10 +198,9 @@ def verified(request, verified_types):
 
     analytics.returned_success(request)
 
-    enrollment_index = reverse("enrollment:index")
-    session.update(request, eligibility_types=verified_types, origin=enrollment_index)
+    session.update(request, eligibility_types=verified_types)
 
-    return redirect(enrollment_index)
+    return redirect("enrollment:index")
 
 
 @decorator_from_middleware(middleware.AgencySessionRequired)
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -48,18 +48,31 @@ def authorize(request):
     if token is None:
         logger.warning("Could not authorize OAuth access token")
         return redirect(ROUTE_START)
-    else:
-        # we are intentionally not storing anything about the user, including their token
-        logger.debug("OAuth access token authorized")
-        session.update(request, auth=True)
-        return redirect(ROUTE_CONFIRM)
+
+    logger.debug("OAuth access token authorized")
+
+    # we store the id_token in the user's session
+    # this is the minimal amount of information needed later to log the user out
+    session.update(request, oauth_token=token["id_token"])
+
+    return redirect(ROUTE_CONFIRM)
 
 
 def logout(request):
-    """View implementing application signout, e.g. updating the Django session."""
+    """View implementing OIDC and application sign out."""
 
-    session.update(request, auth=False)
-    return redirect(ROUTE_START)
+    # overwrite the oauth session token, the user is signed out of the app
+    token = session.oauth_token(request)
+    session.update(request, oauth_token=False)
+
+    origin = session.origin(request)
+    redirect_uri = request.build_absolute_uri(origin)
+
+    logger.debug(f"OAuth end_session_endpoint with redirect_uri: {redirect_uri}")
+
+    # send the user through the end_session_endpoint, redirecting back to
+    # their origin within the app (where they clicked the sign out button)
+    return _deauthorize_redirect(token, redirect_uri)
 
 
 def _deauthorize_redirect(token, redirect_url):
@@ -76,7 +89,7 @@ def _deauthorize_redirect(token, redirect_url):
     metadata = oauth_client.load_server_metadata()
     end_session_endpoint = metadata.get("end_session_endpoint")
 
-    params = dict(id_token_hint=token["id_token"], post_logout_redirect_uri=redirect_url)
+    params = dict(id_token_hint=token, post_logout_redirect_uri=redirect_url)
     encoded_params = urlencode(params)
     end_session_url = f"{end_session_endpoint}?{encoded_params}"
 
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -2,6 +2,7 @@ import logging
 
 from django.shortcuts import redirect
 from django.urls import reverse
+from django.utils.http import urlencode
 
 from authlib.integrations.django_client import OAuth
 
@@ -59,3 +60,24 @@ def logout(request):
 
     session.update(request, auth=False)
     return redirect(ROUTE_START)
+
+
+def _deauthorize_redirect(token, redirect_url):
+    """Helper implements OIDC signout via the `end_session_endpoint`."""
+
+    # Authlib has not yet implemented `end_session_endpoint` as the OIDC Session Management 1.0 spec is still in draft
+    # See https://github.com/lepture/authlib/issues/331#issuecomment-827295954 for more
+    #
+    # The implementation here was adapted from the same ticket: https://github.com/lepture/authlib/issues/331#issue-838728145
+
+    if not oauth_client:
+        raise Exception("No OAuth client")
+
+    metadata = oauth_client.load_server_metadata()
+    end_session_endpoint = metadata.get("end_session_endpoint")
+
+    params = dict(id_token_hint=token["id_token"], post_logout_redirect_uri=redirect_url)
+    encoded_params = urlencode(params)
+    end_session_url = f"{end_session_endpoint}?{encoded_params}"
+
+    return redirect(end_session_url)
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -52,3 +52,10 @@ def authorize(request):
         logger.debug("OAuth access token authorized")
         session.update(request, auth=True)
         return redirect(ROUTE_CONFIRM)
+
+
+def logout(request):
+    """View implementing application signout, e.g. updating the Django session."""
+
+    session.update(request, auth=False)
+    return redirect(ROUTE_START)
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -32,7 +32,7 @@ def login(request):
         raise Exception("No OAuth client")
 
     route = reverse(ROUTE_AUTH)
-    redirect_uri = request.build_absolute_uri(route)
+    redirect_uri = _generate_redirect_uri(request, route)
 
     logger.debug(f"OAuth authorize_redirect with redirect_uri: {redirect_uri}")
 
@@ -67,7 +67,7 @@ def logout(request):
     session.update(request, oauth_token=False)
 
     route = reverse(ROUTE_POST_LOGOUT)
-    redirect_uri = request.build_absolute_uri(route)
+    redirect_uri = _generate_redirect_uri(request, route)
 
     logger.debug(f"OAuth end_session_endpoint with redirect_uri: {redirect_uri}")
 
@@ -102,3 +102,15 @@ def _deauthorize_redirect(token, redirect_url):
     end_session_url = f"{end_session_endpoint}?{encoded_params}"
 
     return redirect(end_session_url)
+
+
+def _generate_redirect_uri(request, redirect_path):
+    redirect_uri = str(request.build_absolute_uri(redirect_path)).lower()
+
+    # this is a temporary hack to ensure redirect URIs are HTTPS when the app is deployed
+    # see https://github.com/cal-itp/benefits/issues/442 for more context
+    # this follow-up is needed while we address the hosting architecture
+    if not redirect_uri.startswith("http://localhost"):
+        redirect_uri = redirect_uri.replace("http://", "https://")
+
+    return redirect_uri
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -83,7 +83,7 @@ def post_logout(request):
     return redirect(origin)
 
 
-def _deauthorize_redirect(token, redirect_url):
+def _deauthorize_redirect(token, redirect_uri):
     """Helper implements OIDC signout via the `end_session_endpoint`."""
 
     # Authlib has not yet implemented `end_session_endpoint` as the OIDC Session Management 1.0 spec is still in draft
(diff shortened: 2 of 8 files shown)


afeld:
> Currently built on top of https://github.com/cal-itp/benefits/pull/414

[Changed the base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) to reflect that.

benefits/oauth/views.py
```diff
@@ -47,8 +48,49 @@ def authorize(request):
...
+
+def logout(request):
+    """View implementing OIDC and application sign out."""
+
+    # overwrite the oauth session token, the user is signed out of the app
+    token = session.oauth_token(request)
+    session.update(request, oauth_token=False)
+
+    origin = session.origin(request)
+    redirect_uri = request.build_absolute_uri(origin)
```
afeld, line 69: Rather than passing different endpoints to the auth server and having to ensure each is registered, what do you think about making a single logout callback endpoint, then redirect to the `origin` from there? This would mirror what we do on the login side.

afeld (changes requested):
Nice work! Tried out that first test locally, which worked great. Some thoughts:

- Did we decide not to go with logging them out from Login.gov right away, thus avoiding the need to store the token? This all looks good, but I wish we didn't need all the extra complexity.
- I would feel much more better having all this auth code in there with corresponding automated tests, and those are going to get harder to write the longer we wait. How would you feel about pairing on tests for this pull request tomorrow?

benefits/oauth/views.py
thekaveman: I really like this idea 👍

thekaveman: Done in [`512c2f4`](https://github.com/cal-itp/benefits/pull/436/commits/512c2f47a14bb3ca26f083afad19afaa48f80f2e)

angela-tran (commented):
These code changes look good to me, and I also like @afeld's idea to have one redirect URI.

I _think_ since that was a very recent change, the CDT auth server hasn't been updated with `/oauth/post_logout`, so I have yet to test a successful logout-redirect. I'll be on the lookout in Slack for that update 👀

thekaveman:
`logout` test added in [`eb0e040`](https://github.com/cal-itp/benefits/pull/436/commits/eb0e04002bcc3b71a91a5013e99a7b4d773db18c):

* Creates a mock of `_deauthorize_redirect` to avoid going to the real auth server
* Verifies `oauth_token` is cleared from session
* Verifies the token and correct `post_logout` redirect URI are sent to `_deauthorize_redirect`

benefits/oauth/views.py
```diff
@@ -47,8 +49,68 @@ def authorize(request):
...
+    #
+    # The implementation here was adapted from the same ticket: https://github.com/lepture/authlib/issues/331#issue-838728145
+
+    if not oauth_client:
+        raise Exception("No OAuth client")
+
+    metadata = oauth_client.load_server_metadata()
+    end_session_endpoint = metadata.get("end_session_endpoint")
+
+    params = dict(id_token_hint=token, post_logout_redirect_uri=redirect_uri)
```
afeld, line 100: So strange that it's called `id_token_hint` when it's in fact the `id_token`.

tests/pytest/conftest.py
```diff
@@ -1,7 +1,34 @@
...
+@pytest.fixture
+def session_request(request, rf):
+    """
+    Fixture creates a new Django request object and initializes its session.
+
+    Optionally use @pytest.mark.request_path(path: str) to customize the request's path.
+    """
+
+    # see if a request_path was provided via marker
+    marker = request.node.get_closest_marker("request_path")
```
afeld, line 21: Hmm, having to specify the path outside of the test is strange to me. I could also imagine test cases where we'd need to simulate multiple requests. I personally prefer the helper function approach 🤷

tests/pytest/oauth/test_views.py
```diff
@@ -0,0 +1,56 @@
...
+import pytest
+
+
+@pytest.mark.request_path("/oauth/logout")
+def test_logout(mocker, session_request):
+    # logout internally calls _deauthorize_redirect
+    # this mocks that function and a success response
+    # and returns a spy object we can use to validate calls
+    message = "logout successful"
+    spy = mocker.patch("benefits.oauth.views._deauthorize_redirect", return_value=HttpResponse(message))
```
afeld, line 16: This works, but could we mock `load_server_metadata()` instead and then check the `Location` header on the `response` for a more comprehensive test?

tests/pytest/oauth/test_views.py
```diff
@@ -0,0 +1,56 @@
...
+
+@pytest.mark.request_path("/oauth/post_logout")
+def test_post_logout(session_request):
+    origin = reverse("core:index")
+    session.update(session_request, origin=origin)
+
+    result = post_logout(session_request)
+
+    assert result.status_code == 302
+    assert result.url == origin
```
afeld, line 38: If the intention is to keep the user logged into Benefits, we should confirm that in this test. Also, let's add a test for no `origin`.

tests/pytest/requirements.txt
```diff
@@ -1,2 +1,3 @@
 pytest
 pytest-django
+pytest-mock
```
afeld, line 3: Nice. I had only only quickly glanced at it and missed the "undoing the mocking automatically after the end of the test" part, which is useful.

benefits/oauth/views.py
```diff
@@ -47,8 +49,68 @@ def authorize(request):
...
+    # send the user through the end_session_endpoint, redirecting back to
+    # the post_logout route
+    return _deauthorize_redirect(token, redirect_uri)
+
+
+def post_logout(request):
+    """View routes the user to their origin after sign out."""
+
+    origin = session.origin(request)
+    return redirect(origin)
```
afeld, line 83: > When the user makes it to the `/enrollment` phase, their session will indicate they have been verified as eligible. So on sign out, we can redirect them back into the `/enrollment` phase vs. making them restart with `/eligibility`.

Hmmmm. I know we had a lot of back and forth about what "logging out" means, but I would expect a "sign out" link to clear my logged-in state (session) on the current site (Benefits).

Also, the stored `oauth_token` from the auth server will no longer be valid, but that's still how we're tracking the auth state in Benefits. Therefore, once we get the logout link placed, that link will:

- Appear to users even though they logged out, and
- Not work if the token is rejected by the auth server
    - I haven't tested this, and even if it doesn't, I could see that changing in an upgrade.

afeld (changes requested):

benefits/eligibility/views.py
```diff
@@ -198,10 +198,9 @@ def verified(request, verified_types):
 
     analytics.returned_success(request)
 
-    enrollment_index = reverse("enrollment:index")
-    session.update(request, eligibility_types=verified_types, origin=enrollment_index)
+    session.update(request, eligibility_types=verified_types)
```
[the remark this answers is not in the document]
thekaveman: This was the only place where `origin` was updated in a view different from what the new `origin` being set was, which didn't make a lot of sense.

This update was moved to the `enrollment:index` view: https://github.com/cal-itp/benefits/pull/436/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R112

tests/pytest/conftest.py
thekaveman: Can you say more about your hesitation here?

The path is specified next to / on top of the test definition...

```python
@pytest.mark.request_path("/some/path")
def test_something(session_request):
    # now session_request is initialized for /some/path and with a session

```

In many cases (like, all of these tests plus everything in #474 so far) the path actually doesn't matter? We just need a request with a session initialized.

Doing this as a fixture means it can easily be injected into any test and is already ready to go, no extra set up needed in the test itself.

tests/pytest/oauth/test_views.py
thekaveman: But... that is not the intention? This is the logout code, the user is signing out of everything. 

> This PR adds a new route /oauth/logout that signs the user out of all 3 environments: Login.gov, the auth server, and the Django app.

benefits/oauth/views.py
thekaveman: > I would expect a "sign out" link to clear my logged-in state (session) on the current site (Benefits).

Yes, the Benefits logged-in state is cleared: https://github.com/cal-itp/benefits/pull/436/files/2e291f2c50007e743fb19aba389bdee41149d4a8#diff-046f7cf43c214ae0ab405c72f331b032dd2c52f707e95f6fb9e33f35f07f7170R67

> Appear to users even though they logged out

I'm not sure where you're getting this from? #428 is pretty clear: *"Only display when auth is enabled, and the user is logged in"*

tests/pytest/oauth/test_views.py
thekaveman: That feels like a different test, of the `_deauthorize_redirect` function. This test is ensuring our logout function is handling app sign out correctly.

I will write another test for that function 👍

thekaveman: Added in [`3f8db89`](https://github.com/cal-itp/benefits/pull/436/commits/3f8db89cfd997dae4c897e5a25b53f21e63bdef1)

tests/pytest/conftest.py
thekaveman: How about both!? 

See [`384b332`](https://github.com/cal-itp/benefits/pull/436/commits/384b332f0feda3d066e5f508dbe02e3f34eda370)

benefits/eligibility/views.py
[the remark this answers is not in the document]
afeld: Rewording to confirm I understand: The `origin` should have been set at the previous view they visited, so we don't need to set it again here. Do I have that right?

benefits/oauth/views.py
afeld: > the Benefits logged-in state is cleared

D'OH. My bad, totally missed that. Sorry!

benefits/eligibility/views.py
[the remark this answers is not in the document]
thekaveman: Correct - when the user gets to this view `eligibility:verified`, they have already gone through `eligibility:start` which is the last "restore point" until the next one at `enrollment:index`.

tests/pytest/oauth/test_views.py
afeld: Sorry, this was tied to my misunderstand of where/whether the `oauth_token` was cleared.

thekaveman: Just thought, something to add to this test - the user should already be logged out of their Django session. That would have made this more clear from the beginning, apologies! Adding...

afeld (approved):
Nice work!

angela-tran (approved):
I was able to see the sign-out redirect locally 👍

thekaveman:
Thank you @afeld for all of the thoughtful review and insisting on comprehensive tests 🙏 (will write some more for login in a follow-up #476)

Thank you @angela-tran for helping us work through all the redirect issues 💪

Issue #435, opened 2022-04-14
Finalize CSS styles


angela-tran:
A number of pull requests on the Login.gov milestone have introduced new CSS styles and tweaks:
 - #397 
 - #402 
 - #403

Ones that have yet to be merged and will add more CSS styles/changes are:
 - #395
 - #429

As a part of reviewing the design with @srhhnry and @Indiajar and of maintaining our code's integrity, we should go through our styles and pages to make sure we all agree with the implementation.

---
### Issues we know need to be addressed
@machikoyasuda and I noticed some unexpected behavior after #397 was merged into `dev`. This is not an exhaustive list, but here are some things I remember us seeing:
 - Changes to the header style cause our view in DEBUG mode to not render well
    - Prior to any Login.gov changes, the DEBUG bar did not impact the layout computation (see screens below)

**On commit 4a7dcff0d2ad9fcf58634c40e76fddfdd33efa40**, the layout behaves well even when the DEBUG bar is present.
| With debug bar             |  Without debug bar |
| -------------------------  | ----------------------|
| ![image](https://user-images.githubusercontent.com/25497886/163443618-aa88af73-391e-4b6d-8fbe-e9d1cf6f2733.png)  | ![image](https://user-images.githubusercontent.com/25497886/163449754-776d33c4-6c17-40c5-929d-6e646ea4d29d.png) |
| ![image](https://user-images.githubusercontent.com/25497886/163443751-e54ce404-00fc-4ef5-bc6a-8216012e158a.png) | ![image](https://user-images.githubusercontent.com/25497886/163449786-4d8a16c4-fd69-46d1-86d8-551d014937a1.png)  |

**The current HEAD of `dev` (2788a8d8ef0cfeb9429653c12a4dcb74d1f94523)**, which contains the new styles
| With debug bar             |  Without debug bar |
| -------------------------  | ----------------------|
| ![image](https://user-images.githubusercontent.com/25497886/163447945-0eebd9d3-2e64-4d15-9891-dee6a6e6d7d7.png) |  ![image](https://user-images.githubusercontent.com/25497886/163448938-c0170e55-8f6b-457c-a9cf-cbea57c05865.png) |
| ![image](https://user-images.githubusercontent.com/25497886/163448730-4fbc0ac5-d433-4a18-95f0-a477757afbfb.png) | ![image](https://user-images.githubusercontent.com/25497886/163448851-b640b9dc-509b-4cf6-b491-70ea3257135b.png) |

[...]

PR #444, opened 2022-04-16
Style: agency index

angela-tran:
Related to #435 , but does not close it entirely.

This PR makes the `agency_index` page better match the Figma mockup. 

I noticed that #397 did not actually change the font-size, letter-spacing, and line-height for this page (at least for the resolution/screen size that I normally use). This was because the `@media (min-width: 992px)` styles were overriding the changes made in #397. I fixed that (which probably fixes it for other pages too?) and then added styles for better spacing on the title, paragraphs, and button label.

| Before | After |
| ------- | ------ |
| image | image |

That's on Firefox with Windows set to 150% scaling and 1920 x 1200 resolution:
image

At 125% scaling, the **After** screen looks like this:
image

Compare the 125% scaling screenshot with the Figma mockup:
| How I see it at 125% scaling on Windows | Figma mockup |
| --------------------------------------------- |----------------- |
| image | image |

### DIFF
--- a/benefits/core/templates/core/agency_index.html
+++ b/benefits/core/templates/core/agency_index.html
@@ -1,5 +1,6 @@
 {% extends 'core/page.html' %}
 {% load i18n %}
+{% block classes %}{{ block.super |add:" agency-index"}}{% endblock %}
 
 {% block container_content %}
 <h1>{{ page.content_title }}</h1>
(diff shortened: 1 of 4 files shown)


machikoyasuda (approved):

thekaveman (approved):
Nice pattern of adding the `agency-index` class with the template override, then adding styles for that specific class!

PR #447, opened 2022-04-18
Fix: add SECURE_PROXY_SSL_HEADER for nginx

### LINKED ISSUE #442: Ensure OAuth redirect URLs are secure
We construct the OAUth redirect URLs using the Django [`reverse()`](https://docs.djangoproject.com/en/3.2/ref/urlresolvers/#reverse) helper and [`request.build_absolute_uri()`](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.build_absolute_uri).

This is done in e.g. the [`login()` view](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/views.py):

```python
ROUTE_AUTH = "oauth:authorize"

route = reverse(ROUTE_AUTH)

redirect_uri = request.build_absolute_uri(route)
```

The docs for `build_absolute_uri` say:

> Mixing HTTP and HTTPS on the same site is discouraged, therefore [build_absolute_uri()](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.build_absolute_uri) will always generate an absolute URI with the same scheme the current request has. If you need to redirect users to HTTPS, it’s best to let your Web server redirect all HTTP traffic to HTTPS.

When running in AWS, we have the Load Balancer sitting in front of all requests; this is the layer where HTTP --> HTTPS redirection happens and where the cert that the browser sees is installed.

Then the request gets to the application container, where it first hits `nginx`. This is where requests for static files end; application requests are forwarded through to `gunicorn` over a Unix socket. Strictly speaking, Django does not receive an HTTPS request.

## Potential solution

The [`SECURE_PROXY_SSL_HEADER`](https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header) Django setting seems to be aimed at helping this situation, where `nginx` sets a header with an indicated value on the request forwarded into the application via gunicorn, to tell the app if the request was originally secure or not.

In the app's [`nginx.conf`](https://github.com/cal-itp/benefits/blob/dev/nginx.conf#L55) we do set that header on forwarded requests.

## Todo

* [ ] Ensure the current `nginx.conf` setting overwrites any header value coming in with the request and/or clear that header first
* [ ] Add the `SECURE_PROXY_SSL_HEADER` to `settings.py` when `DEBUG = False`

thekaveman:
Closes #442 

The nginx reverse proxy sits in front of the application in deployed environments, and SSL terminates before getting to Django. 

We already add this header in the [nginx configuration](https://github.com/cal-itp/benefits/blob/dev/nginx.conf#L55) using the value of the request `$scheme`.

Adding the setting allows Django to determine if the original request was secure or not, if the header value (set by nginx on the request) matches the setting value (`https`).

See https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -90,6 +90,14 @@ if not DEBUG:
 
 SECURE_BROWSER_XSS_FILTER = True
 
+# the NGINX reverse proxy sits in front of the application in deployed environments
+# SSL terminates before getting to Django, and NGINX adds this header to indicate
+# if the original request was secure or not
+#
+# See https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header
+if not DEBUG:
+    SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
 ROOT_URLCONF = "benefits.urls"
 
 template_ctx_processors = [


angela-tran (approved):
I read through the Django documentation you linked as well as the NGINX docs and our nginx.conf file. This looks good to me

afeld (approved):

PR #450, opened 2022-04-18
fix(header): move debug bar out of <header>, remove header height css

machikoyasuda:
#435 

Fixes most of 435.

Moves the `DEBUG BAR` _out_ of the `<header>` HTML element, which has a lot of CSS added on it. 

## Test this PR

image

Open any page _with_ this PR, and then open another tab of the same page -- with the Debug bar entirely removed. (You can remove it by commenting out the Debug bar HTML code, or also removing it from the DOM within DevTools). The spacing of everything under the header _should remain the same_.

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -37,10 +37,6 @@
             {% include "core/includes/noscript.html" %}
             </noscript>
 
-            {% if debug %}
-            {% include "core/includes/debug.html" %}
-            {% endif %}
-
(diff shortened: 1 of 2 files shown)


thekaveman:
I'll just add @machikoyasuda we can put this e.g. at the bottom of the page, under the footer, if that would be more helpful?

machikoyasuda:
@thekaveman I considered that too, but this placement is fine - HTML/CSS-wise too.

angela-tran:
Reviewing this / launching locally 👀

angela-tran (approved):
Works very well now 👍

PR #451, opened 2022-04-18
Refactor: remove paragraphs from index

### LINKED ISSUE #432: Remove all paragraph content from the viewmodel in `core:index`
From https://github.com/cal-itp/benefits/issues/366
image

angela-tran:
Closes #432

| Before (`dev` branch)  | After |
| ------- | ------ |
| image | image |

**UI tests still pass**
image

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -174,6 +174,10 @@ footer .footer-links a {
   background: #212121;
 }
 
+.home label {
+  margin-top: 20vh;
+}
+
 .agency-index h1 {
   margin-bottom: 7vh;
 }
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -174,7 +174,7 @@ footer .footer-links a {
   background: #212121;
 }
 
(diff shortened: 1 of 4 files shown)


benefits/static/css/styles.css
```diff
@@ -174,6 +174,10 @@ footer .footer-links a {
   background: #212121;
 }
 
+.home label {
```
machikoyasuda, line 177: I would change the `margin-top: 20vh` to be applied to the outer parent `.buttons` class instead of adding it to `label`. Also, it's a good practice generally to prefer adding spacing styles to classes (not to all HTML elements). 
image

Otherwise, this pr looks great!

machikoyasuda (changes requested):
Requesting 1 small change

benefits/static/css/styles.css
angela-tran: Thank you for this feedback and the tip on good practices! 🎨 

I agree this makes more sense:
image

machikoyasuda (approved):
🎉

PR #452, opened 2022-04-18
Feat: add analytics for auth flows

### LINKED ISSUE #377: Update analytics events for Login.gov flow

thekaveman:
Closes part of #377.

## Done

* [x] Define basic event class `OAuthEvent`
* [x] Implement a few specific `OAuthEvent` subclasses
* [x] Define helpers for sending these events
* [x] Some minor clean up of analytics event classes project-wide 
* [x] Send the events at the correct points in the flow
* [x] Add tests for sending the events at correct points in the flow

### DIFF
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -8,6 +8,7 @@ from django.utils.http import urlencode
 from authlib.integrations.django_client import OAuth
 
 from benefits.core import session
+from . import analytics
 
 
 logger = logging.getLogger(__name__)
@@ -27,6 +28,8 @@ ROUTE_POST_LOGOUT = "oauth:post_logout"
 
 
 def login(request):
+    analytics.started_sign_in(request)
+
     if not oauth_client:
         raise Exception("No OAuth client")
 
@@ -54,6 +57,8 @@ def authorize(request):
     # we store the id_token in the user's session
     # this is the minimal amount of information needed later to log the user out
     session.update(request, oauth_token=token["id_token"])
+
+    analytics.finished_sign_in(request)
 
     return redirect(ROUTE_CONFIRM)
 
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -66,6 +66,8 @@ def authorize(request):
 def logout(request):
     """View implementing OIDC and application sign out."""
 
+    analytics.started_sign_out(request)
+
     # overwrite the oauth session token, the user is signed out of the app
     token = session.oauth_token(request)
     session.logout(request)
@@ -82,6 +84,8 @@ def logout(request):
 
 def post_logout(request):
     """View routes the user to their origin after sign out."""
+
+    analytics.finished_sign_out(request)
 
     origin = session.origin(request)
     return redirect(origin)
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -28,10 +28,10 @@ ROUTE_POST_LOGOUT = "oauth:post_logout"
 
 
 def login(request):
-    analytics.started_sign_in(request)
-
     if not oauth_client:
         raise Exception("No OAuth client")
+
+    analytics.started_sign_in(request)
 
     route = reverse(ROUTE_AUTH)
     redirect_uri = _generate_redirect_uri(request, route)
--- a/tests/pytest/oauth/test_views.py
+++ b/tests/pytest/oauth/test_views.py
@@ -22,6 +22,8 @@ def test_login(mocker, app_request):
     mock_client = mocker.patch.object(benefits.oauth.views, "oauth_client")
     mock_client.authorize_redirect.return_value = HttpResponse("authorize redirect")
 
+    mocker.patch.object(benefits.oauth.views, "analytics")
+
     assert not session.logged_in(app_request)
 
     login(app_request)
@@ -50,6 +52,8 @@ def test_authorize_success(mocker, app_request):
     mock_client = mocker.patch.object(benefits.oauth.views, "oauth_client")
     mock_client.authorize_access_token.return_value = {"id_token": "token"}
 
+    mocker.patch.object(benefits.oauth.views, "analytics")
+
     result = authorize(app_request)
 
     mock_client.authorize_access_token.assert_called_with(app_request)
@@ -67,6 +71,8 @@ def test_logout(mocker, app_request):
     message = "logout successful"
     spy = mocker.patch("benefits.oauth.views._deauthorize_redirect", return_value=HttpResponse(message))
 
+    mocker.patch.object(benefits.oauth.views, "analytics")
+
     token = "token"
     session.update(app_request, oauth_token=token)
     assert session.oauth_token(app_request) == token
@@ -81,9 +87,11 @@ def test_logout(mocker, app_request):
 
 
 @pytest.mark.request_path("/oauth/post_logout")
-def test_post_logout(app_request):
+def test_post_logout(mocker, app_request):
     origin = reverse("core:index")
     session.update(app_request, origin=origin)
+
+    mocker.patch.object(benefits.oauth.views, "analytics")
 
     result = post_logout(app_request)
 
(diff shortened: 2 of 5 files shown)


angela-tran:
❌ Looking into the failing tests...
image

angela-tran:
📝 Reminder that we'll need to update our schema definition in `cal-itp/data-infra` to pick up these new event-properties

angela-tran:
I moved some of the Todo items that were in this PR's description out into a separate PR: #488 

Those items were more related to tracking pageviews for the Login.gov anchor link and new pages.

tests/pytest/oauth/test_views.py
```diff
@@ -30,6 +32,17 @@ def test_login(mocker, app_request):
     assert not session.logged_in(app_request)
 
 
+def test_login_analytics(mocker, app_request):
```
angela-tran, line 35: Should I be using `@pytest.mark.request_path(...` here? I'm confused on what that is actually meant to do or express.

tests/pytest/oauth/test_views.py
```diff
@@ -30,6 +32,17 @@ def test_login(mocker, app_request):
...
+
+    mock_client = mocker.patch.object(benefits.oauth.views, "oauth_client")
+    mock_client.authorize_redirect.return_value = HttpResponse("authorize redirect")
+
+    login(app_request)
+
+    spy.assert_called_once()
+
+
 @pytest.mark.request_path("/oauth/login")
```
angela-tran, line 46: This confused me more: should this be `"oauth/authorize"` instead of `"oauth/login"`?

tests/pytest/oauth/test_views.py
afeld: Nah, shouldn't matter unless the request path is used somewhere.

tests/pytest/oauth/test_views.py
afeld: Yep, good catch.

```suggestion
@pytest.mark.request_path("/oauth/authorize")
```

tests/pytest/oauth/test_views.py
thekaveman: If you want (need) the `app_request` to point to a specific URL, it's a [mark supported by that fixture](https://github.com/cal-itp/benefits/blob/dev/tests/pytest/conftest.py#L20)

Not needed if the test doesn't "care" about the request path.

tests/pytest/oauth/test_views.py
```diff
@@ -80,17 +108,44 @@ def test_logout(mocker, app_request):
...
+    spy.assert_called_once()
+
+
 @pytest.mark.request_path("/oauth/post_logout")
-def test_post_logout(app_request):
+def test_post_logout(mocker, app_request):
     origin = reverse("core:index")
     session.update(app_request, origin=origin)
 
+    mocker.patch.object(benefits.oauth.views, "analytics")
```
afeld, line 133: This doesn't seem to be necessary if the test function is decorated with `@pytest.mark.django_db`. I'd go with that to be more consistent with other tests, test more of the code, and avoid crowding the test body.

afeld (approved):
Left some comments, but none are blockers.

tests/pytest/oauth/test_views.py
thekaveman: Yeah I think the better answer here is to probably just delete these, since it is more confusing than necessary

thekaveman: (you and @afeld are correct, but the value isn't actually used...)

benefits/oauth/views.py
```diff
@@ -27,6 +28,8 @@
 
 
 def login(request):
+    analytics.started_sign_in(request)
```
thekaveman, line 31: Can we move this to after the (possible) `raise` - I don't think we should send an event if the app is somehow misconfigured.

thekaveman (commented):
FYI I can't approve this since I originally opened it.

benefits/oauth/views.py
angela-tran: Sure, I'm ok with this. 

Would we want the same thing for `logout`? I would need to pull
```python
if not oauth_client:
   raise Exception("No OAuth client"
```
out of `_deauthorize_redirect` and to the top of `logout`. Then move the analytics line after the possible `raise`.

thekaveman: Hmm pulling that out doesn't feel right? And `logout` still logs the user out of the the app, even if this part fails, right? (If the user `started` but didn't `finish` that might give us an indication that something was wrong in between)

I think no, but you go ahead and make the call whatever makes the most sense.

angela-tran: Oh, I see the difference now. I agree with your reasoning, so I'll move it for login only.

thekaveman: Nice, thank you!

tests/pytest/oauth/test_views.py
angela-tran: Thanks for the confirmation @afeld !

If the value isn't actually used, then I would also vote to remove them.

thekaveman:
> 📝 Reminder that we'll need to update our schema definition in `cal-itp/data-infra` to pick up these new event-properties

Created cal-itp/data-infra#1432

tests/pytest/oauth/test_views.py
angela-tran: I agree that that's an alternative. I would prefer though to use `@pytest.mark.django_db` only when the test is for functionality that actually requires the database. 

These tests, e.g. `test_logout`, all of a sudden "want" to use the database only because sending an analytic event requires reading from the session which requires trying to query for models like `TransitAgency` and `EligibilityVerifier`. (The test actually doesn't even set up an agency or verifier to be read from the session, so we'd be querying the database just to get back nothing.) Since the intention of the test is to cover login/logout rather than analytics, I think mocking analytics makes that intention clearer.

All that said, I do wish I could use a decorator or something to mock it instead of crowding the test body.

thekaveman: > All that said, I do wish I could use a decorator or something to mock it instead of crowding the test body.

I think we should be able to do this with a fixture/mark combo?

Similar to whatever the `@pytest.mark.django_db` is doing to signal that we "need the db" - can we have something like `@pytest.mark.mock_analytics`? Not necessary for this PR I think.

tests/pytest/oauth/test_views.py
angela-tran: Removed them in [fd3ddc5](https://github.com/cal-itp/benefits/pull/452/commits/fd3ddc529ffe067b4c4ca7ddbb2261b09b4bf239)

thekaveman (commented):
General point of feedback: if possible, the `test_*_analytics` tests shouldn't repeat all of the body of the other tests - just enough to test that the analytics function was called.

afeld (approved):
Not _super_ worried about DRY-ing up the tests, personally — we can always refactor later, or merge the analytics checks into the other tests.

tests/pytest/oauth/test_views.py
angela-tran: Created #503

PR #454, opened 2022-04-18
Chore: run pytest in CI

afeld:
Also add configuration to run Django tests in GitHub Actions. Want a baseline test that works as a starting place for adding tests in https://github.com/cal-itp/benefits/pull/445. Merged in https://github.com/cal-itp/benefits/pull/455 to build off of those.

### DIFF
--- a/.github/workflows/tests-pytest.yml
+++ b/.github/workflows/tests-pytest.yml
@@ -1,0 +1,34 @@
+name: Pytest
+
+on: push
+
+jobs:
+  pytest:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Install system packages
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y gettext
+
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.9'
+          cache: pip
+          cache-dependency-path: '**/requirements.txt'
+      - name: Install Python dependencies
+        run: pip install -r requirements.txt -r tests/pytest/requirements.txt
+
+      - name: Load environment variables
+        uses: cardinalby/export-env-action@v1
+        with:
+          envFile: .devcontainer/.env.sample
+
+      - name: Run setup
+        run: ./bin/init.sh
+
+      - name: Run tests
+        run: pytest
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -1,0 +1,7 @@
+import pytest
+
+
+@pytest.fixture(scope="session")
+def django_db_setup():
+    # use existing database since it's read-only
+    pass
--- a/tests/pytest/test_views.py
+++ b/tests/pytest/test_views.py
@@ -1,0 +1,10 @@
+import pytest
+from django.urls import reverse
+
+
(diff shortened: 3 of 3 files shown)


.github/workflows/tests-django.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Create environment file
+        run: cp .env.sample .env
+        working-directory: .devcontainer
+
+      - name: Build base image
+        run: docker compose build client
+        working-directory: .devcontainer
```
afeld, line 18: I wanted to keep this as simple as possible (to start) so didn't bother with the fancier Buildx-based Docker config that other workflows have; we can switch to that now or later.

afeld:
Getting the same test failure here as locally:

```
  File "/home/calitp/app/benefits/core/views.py", line 36, in index
    buttons[0].classes.append("mt-3")
IndexError: list index out of range
```

This seems to mean that no agencies are active…?

https://github.com/cal-itp/benefits/blob/1f8b89f933d75c6d4b4e8a240a61fcd4b3ec17bc/benefits/core/views.py/#L34-L37

Assuming this is because the database isn't set up or something?

thekaveman:
We started a branch to add these types of tests [`django-tests-spike`](https://github.com/cal-itp/benefits/tree/django-test-spike) - can you take a look and build on that set up?

afeld:
Current error:

```
$ python manage.py migrate
…
sqlite3.OperationalError: unable to open database file
```

My theory is this is a volume issue, where the volume isn't writable from the Docker engine that the Compose container is running in.

thekaveman:
Putting this in draft while it is still in progress - so we don't get notified for review every change 😁

thekaveman:
My strong advice would be - get the tests to pass locally first, we can worry about CI after. As you're running up against, it can be more of a pain to get the Docker stuff all working together in CI.

But if they are passing locally at least we know we have them and they "should" work.

afeld:
> so we don't get notified for review every change

Sorry about the noise! Have been unable to reproduce the error locally, hence needing to do so many pushes.

thekaveman:
I'm thinking more... does the container/app actually need to be running to run these tests?

Would it work to have job steps that install requirements, run `bin/init.sh`, and run pytest -- directly in the GH Actions runner?

afeld:
> run pytest -- directly in the GH Actions runner

Good call. I was keep consistent by sticking with the container-based setup, but this was much easier.

.github/workflows/tests-pytest.yml
```diff
@@ -0,0 +1,34 @@
+name: Pytest
+
+on: push
+
+jobs:
+  run:
```
thekaveman, line 6: Sorry, one change - can we make this `pytest` or something - easier to reference in required checks and other places.

thekaveman (changes requested):

.github/workflows/tests-pytest.yml
afeld: Oh, strange it doesn't include the workflow name. Done!

thekaveman (approved):
Thank you, this is super helpful!

After merge I'll update the branch protection to require this check as well.

PR #455, opened 2022-04-19
Chore: set up app-level tests

afeld:
This was started a little while ago by @angela-tran; creating a pull request to increase visibility.


afeld:
I created the pull request so it won't let me approve, but I think we should merge this as a good starting place for our tests in Python. Nice work @angela-tran!

thekaveman (approved):

PR #457, opened 2022-04-20
Fix/css - Header, Footer, Container spacing

machikoyasuda:
## What this PR does
- In desktop mode, the footer is now always 40px
- In all modes, the header is now always 80px
- On all pages (except for the Help page), the content container is now always the full height of the screen, minus 40px+80px.
- On all pages (except for the Help page), the content container's inner padding will now be set on the container itself, rather than set by the contents of the container.
- On all pages, the content container's main Button will have margins.

## Developer experience
- New class in `page.html`, `.main-row`, a child of `.main_content .container-fluid`.
- CSS: The styles.css page now has comments for various sections: Header, Footer, Main Primary Container (Applies to all pages, except Help page), Help Page
- `.container.content` padding: XL size screens (padding: 4rem 5rem), L size screens (padding: 4rem 3rem), S/M screens (padding: 3rem 1rem 2rem 1rem)

### DIFF
--- a/benefits/core/templates/core/page.html
+++ b/benefits/core/templates/core/page.html
@@ -10,7 +10,7 @@
 
 {% block main_content %}
 <div class="container-fluid">
-    <div class="row">
+    <div class="row main-row">
         {% if page.noimage %}
         <div class="container">
         {% else %}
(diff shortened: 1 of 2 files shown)


thekaveman:
@machikoyasuda if you rebase on `dev` the `pytest` check will be able to run

PR #458, opened 2022-04-20
Refactor: help page styles

### LINKED ISSUE #456: Fix styles on Help page: button width
We got feedback from @srhhnry and @Indiajar in our meeting on 2022-04-15. This ticket is for:
 - Go back button should match Figma mockup (needs to be not so wide)

angela-tran:
Closes #456


thekaveman (changes requested):
Can we find a way to refactor the `button.html` include, to cover the `link.html` use-case? Maybe I am missing something? The only difference I spot is the `role="button"` attribute on the former and not the latter.

angela-tran:
> Can we find a way to refactor the `button.html` include, to cover the `link.html` use-case? Maybe I am missing something? The only difference I spot is the `role="button"` attribute on the former and not the latter.

In `page.html`, the `includes "core/links.html"` is wrapped by a `<div class="links">`. This is so the styles for the links vs. buttons are separated out. @machikoyasuda and I were finding it difficult to accomplish what we need with one class.

benefits/static/css/styles.css
```diff
@@ -174,6 +173,12 @@ footer .footer-links a {
   background: #212121;
 }
 
+.links div {
```
machikoyasuda, line 176: Try to avoid ever setting a css style on an entire html element (like `div`). You can add a class here like `links-div` (or something descriptive to describe what div is).

benefits/static/css/styles.css
```diff
@@ -115,6 +113,7 @@ footer .footer-links a {
 
 .btn-link {
   color: #046b99;
+  font-weight: bold;
```
machikoyasuda, line 116: This declaration changes the Eligibility Start page

image

So maybe the `font-weight: bold` should style should be declared somewhere where it only touches the Help page links

benefits/static/css/styles.css
```diff
@@ -2,7 +2,7 @@
 
 body {
   font-weight: 400;
-  font-size: 16px;
+  font-size: 18px;
```
machikoyasuda, line 5: This also affects the Eligibility Start page. I would declare this font-size 18 somewhere where it _only_ touches the Help page
image

angela-tran: Ahh! Thanks for catching these. I forgot to go back through the other pages to check on this type of thing. This should help me remember. 🙏

thekaveman:
> > Can we find a way to refactor the `button.html` include, to cover the `link.html` use-case? Maybe I am missing something? The only difference I spot is the `role="button"` attribute on the former and not the latter.
> 
> In `page.html`, the `include "core/includes/links.html"` is wrapped by a `<div class="links">`. This is so the styles for the links vs. buttons are separated out. @machikoyasuda and I were finding it difficult to accomplish what we need with one class.

Oh yeah, I'm not talking about how/where the include is used - I'm talking about the definition of the include. Can't we make a single include work for both use-cases?

Looking closer, we have:

```diff
{% if link.url %}
+ <div>
{% if link.label %}
<label>{{ link.label }}</label>
{% endif %}
<a {% if link.id %}id="{{ link.id }}"{% endif %}
    {% if link.target %}target="{{ link.target }}"{% endif %}
    {% if link.rel %}rel="{{ link.rel }}"{% endif %}
    class="{{ link.classes | join:" " }}"
    href="{{ link.url }}"
-  role="button"
    >
    {{ link.text }}
</a>
+ </div>
{% endif %}
```
As the only difference between these two definitions - I'd like to find a way to consolidate the definitions.

machikoyasuda:
Hey Angela - I have another suggestion:

To split up this PR into 2 parts:

PR 1 - _Just_ change the Back button - the changes for this part in this PR is ready to go 

image


PR 2 - Change the way the Agency Name: Agency phone number part is displayed

angela-tran:
Good idea @machikoyasuda . I opened #463 with the commit to make the button smaller. In this PR, I went ahead and dropped that commit and will work on the agency links / other tweaks to the page.

angela-tran:
I created a new issue #473  to track this work on agency links and font sizes. I'm going to switch to working on #377, so feel free to take #473 if I haven't come back to it.

PR #461, opened 2022-04-20
Button: Large Button centering/height fixes

machikoyasuda:
This CSS change affects:

- _Every_ button/button-like link (**Get started**, **Continue**, **Check status,** **Back**), in the Main Container (`.container.content .btn-lg`) now has a height of 60px. All of these buttons have a height of 60px with text vertically and horizontally centered.
- This means it does not affect the Español/English button in the Header. (The Español/English button _also_ uses `.btn-lg` class, but this button is 38px high.

| Figma | After | 
| ------ | ------|
| image | image |

To test this PR:
- Check _all_ buttons on all pages.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -373,7 +373,8 @@ footer .footer-links a {
 .container.content .btn-lg {
   font-weight: 500;
   margin-bottom: 1rem;
-  padding: 1rem;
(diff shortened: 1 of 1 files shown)


angela-tran (approved):
I was able to confirm this fixed the centering/height for all buttons

PR #463, opened 2022-04-20
Style: Help page button width

angela-tran:
Part of #456, only addresses the button

Note that all buttons in the app are using the same style, so we have to be careful to not introduce regressions/undesired results.

Some details from my commit message:
> The main goal here was to make the "Go back" button on the Help page
> smaller. Most of the buttons in the Figma mockup are around 250px, so
> set that as min-width. Use margin:auto for better centering.


machikoyasuda (approved):
I confirmed that there are no other button size regressions and I checked for this case too:

image

PR #464, opened 2022-04-21
Fix: Ensure HTTPS redirect_uri when app is deployed

thekaveman:
This is a follow-up to #442 (PR #447).

The current AWS architecture has SSL terminating _before_ the request reaches nginx. So the fix in #447 doesn't have the desired effect.

This is a temporary fix that just looks at the redirect URI, and when it _does not_ contain `localhost`, enforce HTTPS.

Note: targeting the `feat/oauth-signout` branch for now until #436 is merged.

### DIFF
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -32,7 +32,7 @@ def login(request):
         raise Exception("No OAuth client")
 
     route = reverse(ROUTE_AUTH)
-    redirect_uri = request.build_absolute_uri(route)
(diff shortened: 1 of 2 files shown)


afeld (changes requested):
Great! Mind adding a couple tests?

thekaveman:
I'm going to put these commits in with #436 since they are working on similar enough areas, and now both have extensive tests in place that build upon one another.

Issue #465, opened 2022-04-21
Update "Sign in with Login.gov" button


thekaveman:
We're awaiting a new mock from @srhhnry, which will seek to incorporate some of the Login.gov branding.

PR #470, opened 2022-04-21
Redesign: Sign in with Login.gov button + Eligibility Start page alignment

machikoyasuda:
#465 

## What this PR does
- Adds `Login.gov` SVG file, from [18F's Identity Style Guide docs](https://github.com/18F/identity-style-guide): https://design.login.gov/brand/#logo
- Updates both English and Spanish label copy (final translations needed) for the button to `Create an account or sign in with`
- Adds a 60px high button that reads has a `Login.gov` SVG, also in `#ffffff`. Button background is `#046b99` and changes to `#2f4c65` on hover. Within the `<a>` link in the DOM, there is text that can be read by screenreaders, but is not visible in the browser:
image

## Code changes
- Removes the previous `Button#external` view model class/method and replaces it with a a Login.gov-specific view model method. This new method, `Button#login` should only be used to create a Login.gov button.
- By default, the `Button#login` method adds the following:

1. `#login` ID, which adds the Login.gov SVG
2. `Login.gov` fallback text (This fallback text is the same for both English and Spanish languages.)
- The `fallback_text` attribute on Button can be used separately from this `login` button method as well, by others in the future.

## How to test this PR
- Developers: Make sure to run `bin/init.sh` before you test this locally (or else the button will read `Continue with login.gov login.gov`
- Test that this button renders as it should on Desktop, Mobile
- Test that you can hover over this button with your mouse to get the hover effect
- Test that you can focus onto this button with keyboard only
- Ensure no regressions on DEFTL flow without Login.gov

## Button design & composition

Button
image

Hover
image

On the page, in various mobile widths

| iPhone 5         | iPhone 6     | iPhone SE | iPhone XS, 12 Mini|
|--------------|-----------|------------|------------|
|image|image|image|image        |

image
image
image


machikoyasuda:
Setting this PR as Ready for Review, so the code changes can be reviewed. The final design decisions are still TBD from Design/Product.

thekaveman (approved):
LGTM!

Nice job with the fallback text approach to ensure screen reader support.

machikoyasuda:
@thekaveman @srhhnry @Indiajar Ready for re-review

benefits/eligibility/views.py
```diff
@@ -112,10 +112,10 @@ def start(request):
         )
 
         if not session.logged_in(request):
-            button = viewmodels.Button.external(
-                text=_(verifier.auth_provider.sign_in_button_label),
+            button = viewmodels.Button.login(
+                label=_(verifier.auth_provider.sign_in_button_label),
+                text="",
                 url=reverse("oauth:login"),
```
thekaveman, line 118: Since this is specifically the `login` button, maybe move this `url=` to the `Button.login()` helper? Same as how the `id` was moved in there, it's part of how the button is defined right?

benefits/eligibility/templates/eligibility/start.html
```diff
@@ -1,5 +1,9 @@
 {% extends "core/page.html" %}
 
+{% block classes %}
+    {{ block.super |add:" eligibility-start"}}
```
thekaveman, line 4: This is cool 🕶️

benefits/eligibility/views.py
machikoyasuda: I thought about that, but for https://github.com/cal-itp/benefits/pull/493, I want to be able to make a Log _out_ button with the same method too, so I opted to keep `url` defined out of the method.

thekaveman: That makes way more sense, thank you! 🙏

thekaveman (approved):

srhhnry:
Hey @machikoyasuda the mobile sizing and placement mostly looks good to me. Any reason why the justification is to the left for some and center for others? 

For the desktop view, the login button should be flush on the right with the container edge (Español button) and the body text/icons arrangement should be spaced out more horizontally. I'm attaching a screenshot from figma
Screen Shot 2022-04-27 at 12 22 11 PM

machikoyasuda:
Updated with button aligned with the Spanish button:

image
image

thekaveman (approved):
I'm not complaining and it certainly doesn't block - but this PR now includes some additional work beyond the button. Maybe update the title/description?

PR #471, opened 2022-04-21
Fix: use recommended settings import approach

thekaveman:
Before this PR, we were importing settings variables directly from our own `benefits.settings` module.

According to Django best practice on [Using settings in Python code](https://docs.djangoproject.com/en/3.2/topics/settings/#using-settings-in-python-code), instead we should import the `django.conf.settings` object and use attributes off of that.

This PR updates everywhere we had imported `benefits.settings` or variables to use the recommended approach.

Also updated our docs guidance around using settings in application code.

### DIFF
--- a/benefits/core/admin.py
+++ b/benefits/core/admin.py
@@ -1,10 +1,10 @@
 """
 The core application: Admin interface configuration.
 """
-from benefits.settings import ADMIN
+from django.conf import settings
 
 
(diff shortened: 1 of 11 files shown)


angela-tran (approved):
These changes look good to me. I also launched the app on this branch, and everything seems to be working correctly.

PR #472, opened 2022-04-21
Fix: Spanish copy on agency index

machikoyasuda:
This PR fixes a current bug where `core.pages.agency_index.p[0]/help#about` is being displayed instead of the actual copy on the Agency Index page.

|Current | Fixed   |
|--------|--------|
|image|image|


machikoyasuda:
@angela-tran Ready for review

angela-tran (approved):
🙏 Thanks @machikoyasuda !

PR #480, opened 2022-04-25
Refactor: Agency link style/template for Help page, Enrollment Unverified page

### LINKED ISSUE #473: Fix styles on Help page: agency links and font sizes
We got feedback from @srhhnry and @Indiajar in our meeting on 2022-04-15. This ticket is for:
 - Transit agency links should match Figma mockup

Also, the Help page was implemented in #403 assuming the styles from #397 took care of the font sizing, letter spacing, etc. As shown on #444 for the agency index page, this was not the case, so we should go back through and compare the current implementation against the Figma mockup and make appropriate fixes.

machikoyasuda:
closes #473 

## What this PR does
- Creates a new includes template that is specific to adding Transit Agency phone number links. This new template creates the HTML necessary to display the **`Transit Agency Name:`** as plain HTML and phone number as a link. The template uses the regular button includes template to generate the Back button.
- Use this new template in the Eligibility#Unverified page.

## Screenshots

- One agency

image

- Two agencies

image

- Unverified agency page

image


benefits/core/templates/core/includes/help-links.html
```diff
@@ -0,0 +1,26 @@
+<div class="buttons-agencies">
+    {% for agency_button in buttons %}
+        {% if "agency" in agency_button.classes %}
+            {% if agency_button.label != "phone" %}
+                <strong>{{agency_button.label}}:</strong>
+                {% comment %} <a target="{{ agency_button.target }}"
+                    rel="{{ agency_button.rel }}"
+                    href="{{ agency_button.url }}">
+                    {{ agency_button.label }}:
+                </a> {% endcomment %}
```
machikoyasuda, line 10: This code creates a _link_ to the agency, but it's commented out because a link is not part of the Figma designs.

thekaveman: Let's try and resolve this question before committing commented code to `dev`... I'd rather not if we can get a solid answer.

machikoyasuda:
@srhhnry @Indiajar Here are the designs implemented for the Help page:

https://user-images.githubusercontent.com/3673236/165137845-cbcc59d1-d67d-42b5-a290-fcad85394663.png

Should the link to the transit agency be removed entirely?

benefits/core/templates/core/help.html
```diff
@@ -3,48 +3,47 @@
...
+                <h2 id="littlepay">{% translate "core.pages.help.littlepay" %}</h2>
+                <p>{% translate "core.pages.help.littlepay.p[0]" %}</p>
+                <p>{% blocktranslate with website="https://littlepay.com" %}core.pages.help.littlepay.p[1]{{ website }}{% endblocktranslate %}</p>
+
+                <h2 id="questions">{% translate "core.pages.help.questions" %}</h2>
+                <p>{% translate "core.pages.help.questions.p[0]" %}</p>
+
+                {% block buttons %}
+                    {% include "core/includes/help-links.html" with buttons=page.buttons %}
+                {% endblock buttons %}
```
angela-tran, line 40: I realize that I wrote the ticket geared towards the Help page, so let me know if you think this is in or out of scope for this PR...

The agency links are also shown on some error pages, and the Figma mockup shows them as one line with just a phone number as well: https://www.figma.com/file/5IabSdY0xDzwL4gyXhbQ9G/Benefits-(Login.gov)?node-id=2083%3A8058

image

Would we just need to move this change to `page.html` instead? And rename the includes to `links.html`?

machikoyasuda: Thank you for this reminder @angela-tran!

I will test this page too and make any changes necessary.

machikoyasuda: Upcoming --> image

machikoyasuda:
Ready for re-review, with new designs from Figma @thekaveman @angela-tran

thekaveman (dismissed):

thekaveman (changes requested):
Couple of points of feedback:

## `agency-links.html` include

1. It feels weird that the `agency-links.html` include is responsible for rendering the `Go Back` button. I know they are all technically links, but the rendering is so different and `Go Back` has nothing to do with the agency / agencies.

2. Maybe this is what you were alluding to with the `Page` viewmodel getting in the way, but I would almost rather see the help page template doing exactly what we need vs. trying to do more in the `page.html`, using only simple includes where their reuse value is high across the project.

    I know this this is the opposite of what you and @angela-tran [talked about in this thread](https://github.com/cal-itp/benefits/pull/480/files#r858066458), seeing it in practice makes me want to simplify and avoid doing too much in `page.html`.

## Link style


*(I realize you are following the mock here)*

Are these... links? Like clickable links? Because they do not look like it at all, and it seems kind of counter-productive UX to _only_ style these links differently from all others?

If the hypothesis is that we want to make the usability... not as obvious... out of a concern about users getting distracted/offboarded - this is just not a concern we need to deal with right now. There have been a total of 5 clicks on an agency phone number since launch on December 8, and none since December 29:

Agency phone number clicks

machikoyasuda:
@thekaveman 

1. Yeah - it's definitely convoluted. The reason why the code is not directly in the `help` template (which would be much simpler), is because the same content and design (agency links with a Back button) are also on the Eligibility Unverified page. And I didn't want to also re-write a brand new template just for the Eligibility Unverified page, at least right now. Because of having to balance both the needs of the Help page _and_ the Eligibility Unverified page, the `Back` button is now also in the template. 

2. Design-wise, yeah, I'm not sure why they look different/have no link indicators. That's up to Design/UX.

benefits/eligibility/views.py
```diff
@@ -217,16 +217,17 @@ def unverified(request):
     # tel: link to agency phone number
     agency = session.agency(request)
     buttons = viewmodels.Button.agency_contact_links(agency)
+    buttons.append(viewmodels.Button.home(request, _("enrollment.buttons.retry")))
```
thekaveman, line 220: Can we please rename this `enrollment.buttons.retry` text to `core.buttons.retry` since it is now used in multiple places?

thekaveman, line 220: This wasn't the focus of this ticket, but it revealed the issue nonetheless...


Now that the Unverified page has a Back button, there is a possible flow where the user is already signed in with Login.gov and they land on the Start page. We don't make them login again, but show a "Continue" button instead, which needs to be styled to right-align like the Login.gov button does:

image

thekaveman (changes requested):

thekaveman:
> The reason why the code is not directly in the help template (which would be much simpler), is because the same content and design (agency links with a Back button) are also on the Eligibility Unverified page. And I didn't want to also re-write a brand new template just for the Eligibility Unverified page.

I hear you, there are definitely tradeoffs. Would a Back button include help here? 

Imagine a brand new dev is coming into this codebase, they aren't looking at the `agency-links` include for how the Back button is rendered, right? I want that to be pulled out / more explicit.

And I'll say this again: in general we _should_ be moving towards individual template files, even for basic pages like Unverified. We've all learned the hard way that trying to do much in `page.html` causes problems later. Better to have purpose-specific templates.

benefits/eligibility/views.py
machikoyasuda: Made new issue: https://github.com/cal-itp/benefits/issues/511

machikoyasuda: Done

machikoyasuda:
> > The reason why the code is not directly in the help template (which would be much simpler), is because the same content and design (agency links with a Back button) are also on the Eligibility Unverified page. And I didn't want to also re-write a brand new template just for the Eligibility Unverified page.
> 
> I hear you, there are definitely tradeoffs. Would a Back button include help here?
> 
> Imagine a brand new dev is coming into this codebase, they aren't looking at the `agency-links` include for how the Back button is rendered, right? I want that to be pulled out / more explicit.
> 
> And I'll say this again: in general we _should_ be moving towards individual template files, even for basic pages like Unverified. We've all learned the hard way that trying to do much in `page.html` causes problems later. Better to have purpose-specific templates.

Agreed! Definitely for any new UI changes moving forward in the upcoming sprints, we should inline improvements and refactors for how the layouts are done. Seeing that we're at the end of the sprint though, I am leaning towards leaving this PR as it is.

thekaveman (approved):
Opened #512 so I don't block this anymore. I really hate adding new tech debt / messiness but I get it, we're nearing the end.

PR #474, opened 2022-04-22
Tests: session management

thekaveman:
Closes #226 

*(Decided to talk about #334 more / leave it out of this PR)*

## Test coverage

```bash
$ coverage report --include benefits/core/session.py
Name                       Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------
benefits/core/session.py     165      0     30      0   100%
```

## Coverage HTML report

This PR adds a helper script `tests/pytest/coverage.sh` that:

1. Runs the tests with `coverage`
2. Generates an HTML report in the app's `static/` directory

The report can be viewed by launching the app and navigating  to `http://localhost:$DJANGO_LOCAL_PORT/static/htmlcov/index.html`

PR #486, opened 2022-04-26
Test: default user session values

thekaveman:
Follow-on to #474.

Fixes "default" tests for the 3 user-info session functions:

* `did` (device ID)
* `start` (session start time)
* `uid` (user ID)

Checks that in each case, `session.reset()` is called when no value is initially present, so that a value is always returned.

### DIFF
--- a/benefits/core/analytics.py
+++ b/benefits/core/analytics.py
@@ -25,13 +25,17 @@ class Event:
     def __init__(self, request, event_type, **kwargs):
         """Analytics event of the given type, including attributes from request's session."""
         self.app_version = VERSION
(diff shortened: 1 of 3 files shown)


afeld (approved):

angela-tran:
Looking at this 👀

angela-tran (commented):
👍 These are good tests to have

Non-blocking suggestion: I think we talked about maybe adding comments/docstrings to give more context about why we need `start`, `uid`, `did` and why they are treated differently. This PR seems like a good opportunity to do that.

angela-tran (approved):

PR #487, opened 2022-04-26
Test: add tests for middleware

afeld:
This pull request builds out tests for various views that hit the various middleware, bringing the overall test coverage from [57%](https://github.com/cal-itp/benefits/runs/6179499569?check_suite_focus=true#step:8:90) to [65%](https://github.com/cal-itp/benefits/runs/6184144403?check_suite_focus=true#step:8:102). I didn't touch:

- The analytics-related middleware, since it seems like things around that are changing
- `RateLimit`, since I'm not sure we want to keep that as is: https://github.com/cal-itp/benefits/issues/502

I went with view-level tests rather than ones testing the middleware directly, since that "fed two birds with one scone," exercising the views _and_ the middleware that restrict them. If folks feel strongly about testing narrower slices, I can change them.

### DIFF
--- a/tests/pytest/core/test_middleware_login_required.py
+++ b/tests/pytest/core/test_middleware_login_required.py
@@ -13,6 +13,19 @@ from tests.pytest.conftest import initialize_request
 login_path = reverse("oauth:login")
 
 
+@pytest.fixture
+def mocked_view():
+    def test_view(request):
+        pass
+
+    return create_autospec(test_view)
+
+
+@pytest.fixture
+def decorated_view(mocked_view):
+    return decorator_from_middleware(LoginRequired)(mocked_view)
+
+
 def require_login(request):
     verifier = EligibilityVerifier.objects.filter(auth_provider__isnull=False).first()
     assert verifier
@@ -21,34 +34,22 @@ def require_login(request):
 
 
 @pytest.mark.django_db
-def test_login_auth_required(rf):
-    @decorator_from_middleware(LoginRequired)
-    def test_view(request):
-        pass
-
-    mock = create_autospec(test_view)
-
+def test_login_auth_required(rf, mocked_view, decorated_view):
     blocked_path = "/some/blocked/path"
     request = rf.get(blocked_path)
     initialize_request(request)
     require_login(request)
 
-    response = test_view(request)
+    response = decorated_view(request)
 
-    mock.assert_not_called()
+    mocked_view.assert_not_called()
 
     assert response.status_code == 302
     assert response.headers["Location"] == login_path
 
 
 @pytest.mark.django_db
-def test_login_auth_not_required(rf):
-    @decorator_from_middleware(LoginRequired)
-    def test_view(request):
-        pass
-
-    mocked_view = create_autospec(test_view)
-
+def test_login_auth_not_required(rf, mocked_view, decorated_view):
     blocked_path = "/some/blocked/path"
     request = rf.get(blocked_path)
     initialize_request(request)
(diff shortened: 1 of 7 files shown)


tests/pytest/core/test_middleware_session_req.py
```diff
@@ -0,0 +1,43 @@
...
+    decorated_view = decorator_from_middleware(AgencySessionRequired)(mocked_view)
+
+    request = rf.get("/something")
+    initialize_request(request)
+
+    agency = TransitAgency.objects.first()
+    session.update(request, agency=agency)
+
+    decorated_view(request)
+    mocked_view.assert_called_once()
```
afeld, line 23: Testing this middleware directly is a bit redundant with the client-level tests below… Thoughts on whether it's useful to keep?

tests/pytest/core/test_middleware_login_required.py
```diff
@@ -21,34 +34,22 @@ def require_login(request):
...
-    mock = create_autospec(test_view)
-
+def test_login_auth_required(rf, mocked_view, decorated_view):
     blocked_path = "/some/blocked/path"
     request = rf.get(blocked_path)
     initialize_request(request)
     require_login(request)
 
-    response = test_view(request)
+    response = decorated_view(request)
```
afeld: This was a bug; it wasn't actually calling the decorated view, hence the mock not getting called.

tests/pytest/eligibility/test_views.py
```diff
@@ -0,0 +1,57 @@
...
+    verifier = agency.eligibility_verifiers.first()
+    assert verifier
+    mock.return_value = verifier
+
+
+@pytest.mark.django_db
+def test_index_with_agency(mocker, client):
+    agency = TransitAgency.objects.first()
+    assert agency
+    with_agency(mocker, agency)
```
thekaveman, line 28: I think you could say `set_agency(mocker)` here to shortcut these 3 lines

tests/pytest/core/test_middleware_session_req.py
thekaveman: Yeah agreed the direct client tests below seem better.

tests/pytest/core/test_middleware_login_required.py
```diff
@@ -0,0 +1,56 @@
...
+    assert verifier.requires_authentication
+    session.update(request, verifier=verifier)
+
+
+@pytest.mark.django_db
+def test_login_auth_required(app_request, mocked_view, decorated_view):
+    require_login(app_request)
+    response = decorated_view(app_request)
+
+    mocked_view.assert_not_called()
```
thekaveman, line 30: Is this the same `mocked_view` that was `decorated` ... because of the order of the fixture args here?

tests/pytest/core/test_middleware_session_req.py
thekaveman: So are you planning to remove these `pass` placeholder tests?

thekaveman (changes requested):
Minor opportunity to use one of the helper functions you created, otherwise just a few general comments/questions. 

Looking great! I like the "two birds one stone" of getting the middleware and view tested.

tests/pytest/eligibility/test_views.py
afeld: Good catch, thanks!

tests/pytest/core/test_middleware_login_required.py
afeld: [pytest fixtures are cached on a test-by-test basis](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixtures-can-be-requested-more-than-once-per-test-return-values-are-cached); order doesn't matter.

thekaveman (approved):
Nice!

PR #491, opened 2022-04-26
fix: clean out old test coverage results

afeld:
Otherwise, it's easy to refresh the page and see old results. Broken out of https://github.com/cal-itp/benefits/pull/487.


thekaveman (approved):
Good idea 👍

PR #492, opened 2022-04-26
fix: replace deprecated logger warning calls

afeld:
Was getting the warning

```
  /home/calitp/app/benefits/enrollment/views.py:62: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
    logger.warn(f"card_tokenize_url: {context['payment_processor'].card_tokenize_url}")
```


thekaveman (approved):
Funny that these two crept in at different times and places. Thanks!

PR #398, opened 2022-04-05
Fix: Docker tweaks

afeld:
A handful of small changes/fixes — see commits.

https://github.com/cal-itp/benefits/pull/390 changed the Django app to no longer be served via a dynamic port; the documentation changes are a follow-up to that pull request.

PR #496, opened 2022-04-27
fix: remove deprecated Docker Compose volume option

afeld:
https://stackoverflow.com/a/70957456/358804

Extracted from https://github.com/cal-itp/benefits/pull/398.


thekaveman (dismissed):
I honestly don't remember why those were there in the first place.

What about the `:ro` option on the `.aws` volume (line 11) - does this matter? The linked SO post/GH issue mentions it as one of the deprecated, but... technically we don't want to write

afeld:
Fixed!

thekaveman (approved):

PR #499, opened 2022-04-27
Feat: upload the coverage report as an artifact in CI

thekaveman:
Initial step in publishing the coverage report as part of our docs site.


afeld (approved):
Was able to download and view, thanks!

PR #514, opened 2022-04-28
fix(css): right-align continue button on eligibility start

### LINKED ISSUE #511: Continue button style
This wasn't the focus of this ticket, but it revealed the issue nonetheless...

Now that the Unverified page has a Back button, there is a possible flow where the user is already signed in with Login.gov and they land on the Start page. We don't make them login again, but show a "Continue" button instead, which needs to be styled to right-align like the Login.gov button does:

image

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/480#discussion_r861119593_

machikoyasuda:
image

image

closes #511 

## How to test this PR
- Check Eligibility Start page in MST Courtesy Cardholder flow
- Check Eligibility Start page in Senior Discount flow by going through Login.gov, entering an invalid/unverified credential, clicking Try Again and going _back_ to Eligibility Start _after_ already being logged in to Login.gov

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -302,6 +302,11 @@ footer .footer-links a {
   justify-content: right;
 }
 
+.eligibility-start .container.content .buttons .btn {
+  margin-left: initial;
+  margin-right: initial;
+}
+
 .eligibility-start .container.content .buttons label {
   margin: 0 0.8rem 0 0;
   line-height: 1;


thekaveman (approved):
I know it looks a little funky on the 2 item Start page, but we didn't really get a design for that one, and it won't be visible this go around anyway. So all good 👍

PR #500, opened 2022-04-27
Docs: publish the coverage report with docs site

thekaveman:
Download the coverage report artifact (#499) and send it with the `mkdocs` publish.

The latest coverage report from dev will be available at https://docs.calitp.org/benefits/tests/coverage

PR #515, opened 2022-04-28
Fix: mkdocs runs after Pytest

thekaveman:
Following on to #500 and #509, trying to publish the latest `coverage` report to our docs site.

#509 introduced what amounted to a race condition between the `mkdocs` and `pytest` workflows - if `pytest` took too long, `mkdocs` would run and deploy before the `coverage` report was updated.

This PR makes the relationship explicit: `mkdocs` now runs after `pytest` has run and is successful, in addition to running in response to real docs updates.

Read more about triggering with `workflow_run`: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -1,13 +1,18 @@
 name: Publish docs
 on:
   workflow_dispatch:
+  workflow_run:
+    workflows:
+      - Pytest
+    types:
+      - completed
+    branches:
+      - dev
   push:
     branches:
       - dev
     paths:
-      - 'benefits/**'
       - 'docs/**'
-      - 'tests/pytest/**'
       - 'mkdocs.yml'
       - '.github/workflows/mkdocs.yml'
 
@@ -15,6 +20,7 @@ jobs:
   docs:
     name: Publish docs
     runs-on: ubuntu-latest
+    if: github.event.workflow_run == null || github.event.workflow_run.conclusion == 'success'
     steps:
       - name: Checkout
         uses: actions/checkout@v3


.github/workflows/mkdocs.yml
```diff
@@ -1,20 +1,26 @@
...
       - 'docs/**'
-      - 'tests/pytest/**'
       - 'mkdocs.yml'
       - '.github/workflows/mkdocs.yml'
 
 jobs:
   docs:
     name: Publish docs
     runs-on: ubuntu-latest
+    if: github.event.workflow_run == null || github.event.workflow_run.conclusion == 'success'
```
thekaveman, line 23: @afeld does this syntax look correct for the case where it is not `workflow_run`? 

Another possibility is `github.event_name != 'workflow_run'`, not sure if that's better / more correct?

afeld: Hmm. I haven't set up dependencies between GitHub Actions workflows before, so would need to do some research myself. For starters, seems like the `push` trigger above is redundant with the `workflow_run`, since the latter would be coming from the mkdocs one that already happens on `push`, right? If it works, it works 🤷 

I suggest we don't spend _too_ much time finagling the deployment; worst case, we can either:

- Smush the mkdocs stuff into the same workflow/job as pytest
- Direct folks to download the artifact from the [GitHub Actions runs](https://github.com/cal-itp/benefits/actions/workflows/tests-pytest.yml?query=branch%3Adev)

and call it a day.

thekaveman: Yeah agreed. I was more talking about the syntax of the `if` specifically, but I think this one now works as intended.

The `push` is still necessary to re-publish the docs site when only docs change (like fixing a typo or adding a new page/section).

`workflow_run` makes the docs site re-publish when python code changes and `coverage` is updated in the `dev` branch.

PR #516, opened 2022-04-28
Feat: sign out link

### LINKED ISSUE #428: Place Login.gov sign out button

angela-tran:
Closes #428 

### Summary
 - Adds new context processors
    - `authentication`: Adds information about authentication requirement and status to context for use in templates
    - `auth_provider`: Adds information about auth_provider to context
 - Adds a new includes `sign-out-link.html` that shows only if authentication is required and user is signed in
 - Adds usage to the pages that need it
    - `eligibility:confirm`
    - `eligibility` unverified
    - `enrollment:index`
    - `enrollment:retry`

### DIFF
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -2,6 +2,7 @@
 The core application: context processors for enriching request context data.
 """
 from django.conf import settings
+from django.urls import reverse
 
 from . import session
 
@@ -9,6 +10,38 @@ from . import session
 def analytics(request):
     """Context processor adds some analytics information to request context."""
     return {"analytics": {"api_key": settings.ANALYTICS_KEY, "uid": session.uid(request), "did": session.did(request)}}
+
+
+def authentication(request):
+    """Context processor adds authentication information to request context."""
+    verifier = session.verifier(request)
+
+    if verifier:
+        return {
+            "authentication": {
+                "required": verifier.requires_authentication,
+                "logged_in": session.logged_in(request),
+                "sign_out_route": reverse("oauth:logout"),
+            }
+        }
+    else:
+        return {}
+
+
+def auth_provider(request):
+    """Context processor adds auth_provider information to request context."""
+    verifier = session.verifier(request)
+
+    if verifier and verifier.requires_authentication:
+        auth_provider = verifier.auth_provider
+        return {
+            "auth_provider": {
+                "sign_in_button_label": auth_provider.sign_in_button_label,
+                "sign_out_button_label": auth_provider.sign_out_button_label,
+            }
+        }
+    else:
+        return {}
 
 
 def debug(request):
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -17,29 +17,18 @@ def authentication(request):
     verifier = session.verifier(request)
 
     if verifier:
-        return {
-            "authentication": {
-                "required": verifier.requires_authentication,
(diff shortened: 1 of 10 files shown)


machikoyasuda:
## Desktop link alignment

The key is that the `Sign out of Login.gov` button needs to be: 1) Aligned with the **`Español`** button, on the right-most side of the `container` and also 2) Not affect the `<h1>` of the rest of the app page's containers. This button's container is absolutely positioned to always be 22px below the Header, and uses the same `container` class as the Header to justify the `Sign out of Login.gov` link the same way the  **`Español`** button is. I added a z-index to ensure the link is always clickable on top of any other divs that may pop up.

Note: We don't have mobile designs yet, so mobile may need to have different CSS styles.

Wider screen

image

Responsive narrower screen

image

benefits/core/context_processors.py
```diff
@@ -11,6 +12,38 @@ def analytics(request):
...
+def auth_provider(request):
+    """Context processor adds auth_provider information to request context."""
+    verifier = session.verifier(request)
+
+    if verifier and verifier.requires_authentication:
+        auth_provider = verifier.auth_provider
+        return {
+            "auth_provider": {
+                "sign_in_button_label": auth_provider.sign_in_button_label,
+                "sign_out_button_label": auth_provider.sign_out_button_label,
```
thekaveman, line 40: I feel like these two properties can be combined with the `authentication` context no?

Instead of immediately returning the dict, build it first so you can optionally add the extra keys:

```python
if verifier:
    data = {
        "required": verifier.requires_authentication,
        "logged_in": session.logged_in(request),
        "sign_out_route": reverse("oauth:logout"),
    }
    
    if verifier.requires_authentication:
        # add the other fields
    
    return { "authentication": data }
```

thekaveman (changes requested):
This is looking great so far! Nice use of context processors, that seems like a clean and easy way to do it 👍 

One minor refactor I think could be made there.

benefits/core/context_processors.py
angela-tran: Done in 1a12137

angela-tran:
> Note: We don't have mobile designs yet, so mobile may need to have different CSS styles.

👍 I agree -- no need to make changes for mobile in this PR. I did notice some odd behavior though and want to just note here so we remember to look into it later.

On this branch, I see a lot of whitespace / vertical scroll for some reason:
image
 
I don't see it on our dev instance:
image

angela-tran:
I checked all four pages in Desktop, and the styling looks great @machikoyasuda. Thank you!

thekaveman (approved):
This is great.

Definitely non-blocking, but it also looks a little funky on certain pages on super wide desktop:

(Maybe we need to expand the container in the header as well?) 

image

PR #520, opened 2022-04-29
Fix: enrollment retry button msgid

angela-tran:
msgid was changed in #480 to `core.buttons.retry`, but this usage was leftover


thekaveman (approved):
Great catch 🙏

thekaveman (approved):

PR #532, opened 2022-04-29
Fix: signout link on Eligibility Confirm resubmission page

angela-tran:
### Steps to reproduce
From @machikoyasuda in Slack:

> ***Sign out button issue*** When I go through Login.gov, I get to the Eligibility Confirm page and see the Sign out of Login.gov button. Then, I make a mistake on my form (either put in a bad ID number or a bad last name), the page refreshes and shows me my errors. At that point, the Sign out of Login.gov link disappears

image

### Expected result / screenshot of fixed behavior
image


thekaveman (approved):
Nice fix.

Ideally we get rid of all usages of `PageTemplateResponse` and always return a specific template file.

Issue #405, opened 2022-04-05
Eligibility confirm form button does not show the "loading" state


thekaveman:
This is what the button currently looks like after clicking "Check Status":

![Screenshot 2022-05-01 195521](https://user-images.githubusercontent.com/1783439/166295035-898ec9a7-823a-4e75-9922-916a54dbf189.png)

This is what it should look like (the payment connection button - disabled state):

![image](https://user-images.githubusercontent.com/1783439/166295135-03d1c5fc-da31-41a0-b845-cabf7a16ab92.png)

PR #534, opened 2022-05-03
fix(button): 

machikoyasuda:
fix https://github.com/cal-itp/benefits/issues/405

Google Recaptcha seems to be removing the `disabled` state from the button HTML (related issue: https://stackoverflow.com/questions/59546714/google-recaptcha-enables-disabled-submit-button#59567003). Instead of disabling the button with HTML, I'm using a Bootstrap CSS class to do it. The CSS class uses `pointer-events: none` (which has high cross-broswer compatibility: https://caniuse.com/pointer-events)

- add `.disabled` class to override recaptcha removing disabled state

### DIFF
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -68,7 +68,8 @@
                     $("button[data-action='submit']", this)
                         .attr("role", "status")
                         .attr("disabled", "true")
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
Very nice!

angela-tran (approved):

PR #544, opened 2022-05-04
Mobile: Image and Sign Out Link

### LINKED ISSUE #535: Mobile: New button design for Sign Out
image

machikoyasuda:
closes #535 

### Code changes
- Move image logic from the HTML template and the Page object into CSS only
- Add new updated image display styles for screens under the 992px width breakpoint.

### Design changes
- Image on mobile screens is now set to 240px in height, from the bottom. This allows the hand, the card with the VISA logo, the contactless pay logo on the card reader all to be shown: 
image
- The button now has 2px solid #2b6597 borders, with a 3px border radius and 10px top and bottom/30px left and right padding, resulting in a button that's 231px by 46px. The button is 20px below the bottom of the image. image

## Testing this PR, for Engineers:

- You must turn off Debug mode or hide the Debug bar, or else the button gets placed like this.
image

Add this code or delete the element from Developer Tools.
```css
.bg-danger {
  display: none !important;
}
```

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -551,7 +551,7 @@ footer .footer-links a {
     background-position: center bottom;
     height: 240px;
     background-size: cover;
-    margin-bottom: 40px;
+    margin-bottom: 60px;
   }
 
   .signout-row .container .signout-link {
@@ -569,7 +569,7 @@ footer .footer-links a {
   }
 
   .signout-row {
-    top: 368px;
+    top: 308px;
   }
 
   /* Eligibility Start */
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -75,6 +75,11 @@ main .main-row {
   min-height: calc(100vh - 120px);
 }
 
+main .main-row .col-lg-6.image {
+  background: 48% 75% url("/static/img/ridertappingbankcard.png") no-repeat;
+  background-size: cover;
+}
+
 footer {
   margin-top: auto;
   padding-top: 1rem;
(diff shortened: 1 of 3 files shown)


benefits/static/css/styles.css
```diff
@@ -75,6 +75,11 @@ main .main-row {
   min-height: calc(100vh - 120px);
 }
 
+main .main-row .col-lg-6.image {
+  background: 48% 75% url("/static/img/ridertappingbankcard.png") no-repeat;
```
thekaveman, line 79: Since this path moved to the stylesheet, can we please clean up the [`core.viewmodels.Page`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/viewmodels.py#L144), which also sets a path to this image.

benefits/static/css/styles.css
```diff
@@ -540,6 +545,34 @@ footer .footer-links a {
 }
 
 @media (max-width: 992px) {
+  /* Mobile With Image */
+
+  .with-image main .main-row .col-lg-6.image {
+    background: center bottom url("/static/img/ridertappingbankcard.png")
```
thekaveman, line 551: Why is this image path repeated? Can't we use `background-image` once and change whatever else is needed in the media queries?

thekaveman (changes requested):

benefits/static/css/styles.css
machikoyasuda: Removed

thekaveman (dismissed):
Thank you!

machikoyasuda:
Tomorrow: Will do an in-person design review with Design Team, and then merge after that.

thekaveman (approved):

machikoyasuda:
Now on Staging:

image
image
image

Issue #448, opened 2022-04-18
Set up error monitoring


afeld:
[Discussion](https://cal-itp.slack.com/archives/C013W8RUAMU/p1649779283917369)

_As a developer on the Benefits application, I want to know when users encounter errors so that I can ensure the service is reliable._

- [ ] Check with DMV about what they have set up
- [x] Write up what we're looking to do

Issue #141, opened 2021-10-15
Refactor eligibility verification into reusable package


thekaveman:
## Overview

We have `benefits` and the [`eligibility-server`](/cal-itp/eligibility-server) each with Python code that implements one half of the [Eligibility Verification API](https://docs.calitp.org/eligibility-server/specification/).

We could consolidate these two disparate halves into a single Python package that would:

* Allow for easier testing, outside the context of a web application
* Keep the constructing and parsing of messages consistent
* Consolidate dependency management into a single package

Note that last point is particularly relevant as we are now in the process of [upgrading dependencies in the server](https://github.com/cal-itp/eligibility-server/issues/26) to match recently posted updates to the _same dependencies_ here (#131, #132).

The corollary tracking issue in server is cal-itp/eligibility-server#38

## Benefits specifics

[`benefits/eligibility/api.py`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/api.py) is where the client-side implementation lives. 

[`RequestToken`](/cal-itp/benefits/blob/dev/benefits/eligibility/api.py#L30) and [`ResponseToken`](/cal-itp/benefits/blob/dev/benefits/eligibility/api.py#L71) are almost entirely independent from any Django-based context or references. Where they do make use of Django objects (e.g. a [`TransitAgency`](/cal-itp/benefits/blob/dev/benefits/core/models.py#L106) model is passed in) we could instead pass the needed data (e.g. the agency's ID), directly.

[`Client`](/cal-itp/benefits/blob/dev/benefits/eligibility/api.py#L114) is only slightly more tied into the Django-specifics and could be easily refactored as well.

thekaveman:
Closing this now that we have fully refactored the Eligibility API code + dependencies out of Benefits. Nice work @angela-tran!!

PR #575, opened 2022-05-11
Eligibility API tests

### LINKED ISSUE #550: Add tests for eligibility API
Broken out from https://github.com/cal-itp/benefits/issues/413.

angela-tran:
Closes #550


angela-tran:
New test is failing in CI because it fails to connect to an eligibility server.

I think we'll either need to 
 - change the `test-pytest` workflow to start up and access an `eligibility-server` container, retracting from the approach taken in https://github.com/cal-itp/benefits/pull/454, OR
 - mock the `eligibility-server` response, which may require pulling in some code for us to then just rip back out when we do #141

thekaveman:
We definitely don't want `pytest` tests hitting anything real - e.g. #553 just disabled `pytest` ability to make external network requests.

Don't worry about refactoring things later for #141 - the task here is to get tests running for the _current_ Eligibility API code.

afeld:
I think mocking is the way to go, even if we end up moving/copying it to the other repo soon.

angela-tran:
@thekaveman @afeld How much coverage do you think we should strive for here? There are several error paths with the Eligibility API code - for example, when [making the request](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L147-L154), when [getting back the response](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L161-L162), and various points when parsing the response ([reading response text ](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L83),  [decrypting the signed token](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L90-L93), [verifying the signature](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L101-L104)). Just want to make sure we actually want all that coverage before I start writing tests for all those different errors.

Also, I was able to mock the eligibility-server response. Does the approach of using [HTTPretty](https://httpretty.readthedocs.io/en/latest/) look ok to you?

angela-tran:
> @thekaveman @afeld How much coverage do you think we should strive for here? There are several error paths with the Eligibility API code - for example, when [making the request](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L147-L154), when [getting back the response](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L161-L162), and various points when parsing the response ([reading response text ](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L83), [decrypting the signed token](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L90-L93), [verifying the signature](https://github.com/cal-itp/benefits/blob/84ddadf/benefits/eligibility/api.py#L101-L104)). Just want to make sure we actually want all that coverage before I start writing tests for all those different errors.
> 
> Also, I was able to [mock the eligibility-server response](https://github.com/cal-itp/benefits/pull/575/files#diff-2795879aa11b38997f60af6530d235b16a86c46202b936f926abe6f5595698b2R77). Does the approach of using [HTTPretty](https://httpretty.readthedocs.io/en/latest/) look ok to you?

Re: coverage, I decided that the cases related to decrypting/verifying signatures don't need to be done here and would be better/easier to write tests for in the [eligibility-api](https://github.com/cal-itp/eligibility-api) repo.

afeld:
> How much coverage do you think we should strive for here?

Just to throw a number out there, I'm happy around 90% per file and overall. Arguably should be higher given while we are without error monitoring (#448). That said, some tests are better than no tests (for this part of the codebase), so I would merge this sooner than later.

> want to make sure we actually want all that coverage before I start writing tests for all those different errors

I say "write tests until you're sick of it" 😉 You may find the tests easier to write if you refactor smaller functions out of the Eligibility views and write unit tests against those… that refactored code would then more easily get turned into a standalone package. In other words, it may be easiest to build out the eligibility-api package in this repository as just standalone Python files, then move it to the other repository once it's stabilized a bit.

> I was able to [mock the eligibility-server response](https://github.com/cal-itp/benefits/pull/575/files#diff-2795879aa11b38997f60af6530d235b16a86c46202b936f926abe6f5595698b2R77).

👏 

> Does the approach of using [HTTPretty](https://httpretty.readthedocs.io/en/latest/) look ok to you?

For sure!

angela-tran:
Ok, thanks for the feedback @afeld! I'll put this PR back into draft and see what additional coverage I can get in.

angela-tran:
We went from 27% coverage to 95% coverage for `eligibility/api.py`!



| Before | After |
| ------- | ------ |
| image | image |

tests/pytest/eligibility/test_views.py
```diff
@@ -52,3 +67,201 @@ def test_start_without_verifier(mocker, client):
...
+        #         "iss": "test-server",
+        #         "iat": int(datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc).timestamp()),
+        #         "eligibility": ["type1"],
+        #     },
+        #     verifier.jws_signing_alg,
+        #     _get_jwk("server.key"),
+        #     verifier.jwe_encryption_alg,
+        #     verifier.jwe_cek_enc,
+        #     _get_jwk("client.pub"),
+        # ), id='TokenError("Invalid JWS token")'),
```
afeld, line 226: Can we put this in an issue instead? Avoids clutter in the codebase and makes it more clearly a TODO.

afeld (changes requested):
Just the one comment. Otherwise, good to go!

tests/pytest/eligibility/test_views.py
angela-tran: Created #585

afeld (approved):

thekaveman (commented):
I don't want to block since this is definitely better than the low coverage in this area that we had before!

This feels like combining a few different concerns into the same tests though, namely how the form/view work, and how the API client works. I am also not a fan of what is essentially a reimplementation of the token encryption and signing logic as a helper for test setup. That feels... bad.

I would think the view tests should use a fully mocked `client`, not even getting to the point of having to mock HTTP requests, encryption, etc. These tests would ensure the view logic is correct when various inputs and outputs of the API are processed.

Separate tests for the Eligibility API / client / etc. could get in to the mechanics of working with the request and response tokens, encryption keys, servers, and the like.

angela-tran:
> This feels like combining a few different concerns into the same tests though, namely how the form/view work, and how the API client works.

This is a really good point, and I'm glad you pointed this out. I've created #595 to cover this, which should be done as a part of the Eligibility API refactor (should not be left behind as technical debt).

Thank you @afeld and @thekaveman for your thorough reviews and feedback! 🙏

PR #579, opened 2022-05-12
Feat: redirect to agency's index if only one agency

### LINKED ISSUE #566: Auto select Transit Agency when only 1 Active
Simplify the flow for a user, and avoid having to deal with the design of a single button on the `core:index` page.

When there is only 1 active transit agency, and user lands on `core:index`, forward to the active agency's `core:agency_index` page.

angela-tran:
Closes #566

### Summary
 - Updates the core app's `index` function to redirect to the agency's index if there is only one active agency.
 - Also added a unit test for behavior of homepage with single agency.
 - Also launched the app and tested manually.
image

### DIFF
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -2,6 +2,7 @@
 The core application: view definition for the root of the webapp.
 """
 from django.http import HttpResponseBadRequest, HttpResponseNotFound, HttpResponseServerError
+from django.shortcuts import redirect
 from django.template import loader
 from django.template.response import TemplateResponse
 from django.urls import reverse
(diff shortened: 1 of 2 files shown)


benefits/core/views.py
```diff
@@ -31,20 +32,25 @@ def index(request):
...
-        content_title=_index_content_title(),
-        buttons=buttons,
-        classes="home",
-    )
-
-    return PageTemplateResponse(request, page)
+    if len(agencies) == 1:
+        agency = agencies[0]
+        return redirect(agency.index_url)
+    else:
```
thekaveman, line 40: I almost think we could live without this `else`. 

The previous `return` will end the flow, so don't really need to worry about branching the `if`.

Not a blocker.

tests/pytest/core/test_views.py
```diff
@@ -9,8 +9,14 @@
...
-    assert response.status_code == 200
-    assert "transit" in str(response.content)
+
+    agencies = TransitAgency.all_active()
+    if agencies.count() == 1:
+        assert response.status_code == 302
+        assert response.url == agencies[0].index_url
+    else:
+        assert response.status_code == 200
+        assert "transit" in str(response.content)
```
thekaveman, line 20: I think we do want a new test here, since this is introducing new behavior.

`TransitAgency.all_active()` can be mocked in one case to return just a single entry, to test this change.

thekaveman (changes requested):

tests/pytest/core/test_views.py
angela-tran: I hadn't thought about mocking `TransitAgency.all_active()`. Nice! I'll work on making that change.

thekaveman (approved):
This looks great! I tested it out locally by only having a single active agency `ABC` and was immediately redirected to their landing page 👍

PR #598, opened 2022-05-18
Refactor: simplify env file, better default settings

### LINKED ISSUE #498: Better defaults for environment variables
New environment variables being added to the application cause an issue where people's `.env` files get out of sync, and hit confusing errors. It seems like most of us can use the [`.env.sample`](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/.env.sample) as-is, so what if instead of having to make a copy of that file regularly, we instead had defaults that made the application work (at least in development) without modification? That way, the `.env` just gets used for overrides, making it very clear what has been changed from the defaults.

thekaveman:
Closes #498.

## Reviewing this PR

1. Reopen the repository in your local machine, and checkout this branch
2. Save a copy of your current `.env` for posterity, then create a new one `cp .env.sample .env`
3. `docker compose build client`
4. Rebuild and reopen the devcontainer

## Acceptance criteria

* [x] `F5` launches the app in debug mode, allows hitting breakpoints
* [x] You can verify eligibility for the non-auth flow
* [x] The auth flow gives the expected error of "Cannot connect to `http://example.com`" when clicking the "Continue with Login.gov" button
* [x] All the Pytests pass: `tests/pytest/run.sh`
* [x] All the Cypress UI tests pass: `npm run cypress:ui`
    * easier to run these outside the devcontainer, against `docker compose up -d client`

## Update your `.env` after this is merged

After this is merged, for day-to-day development and testing, you should only need a few values in your `.env`:

```env
# OAuth config
DJANGO_LOCAL_PORT=**** # localhost port registered with CDT auth server
DJANGO_OAUTH_AUTHORITY=**** # CDT auth server root domain
DJANGO_OAUTH_CLIENT_ID=**** # client id for CDT auth server
```


machikoyasuda:
After both this and #603 tickets are merged in, we should update the benefits-secrets repo with the sample `.env` file.

machikoyasuda (approved):

.devcontainer/.env.sample
```diff
@@ -12,32 +9,9 @@ AWS_BUCKET=bucket-name
...
-DJANGO_RECAPTCHA_VERIFY_URL=https://www.google.com/recaptcha/api/siteverify
 DJANGO_SECRET_KEY=secret
-DJANGO_RATE_LIMIT=0
-DJANGO_RATE_LIMIT_METHODS=GET,POST,PUT,DELETE
-DJANGO_RATE_LIMIT_PERIOD=0
-
-DJANGO_OAUTH_AUTHORITY=https://example.com
-DJANGO_OAUTH_CLIENT_ID=client-id-123
-DJANGO_OAUTH_CLIENT_NAME=client_name
-DJANGO_OAUTH_SCOPE=openid
```
machikoyasuda, line 15: Should we keep these two in here?

```
# Django config
DJANGO_OAUTH_AUTHORITY=**** # CDT auth server root domain
DJANGO_OAUTH_CLIENT_ID=**** # client id for CDT auth server
```

.devcontainer/compose.yml
```diff
@@ -31,7 +31,7 @@ services:
     entrypoint: mkdocs
     command: serve --dev-addr "0.0.0.0:8000"
```
afeld, line 32: Better to change this to 8001 for consistency?

benefits/settings.py
```diff
@@ -168,14 +168,16 @@ def _filter_empty(ls):
 
 # OAuth configuration
 
-OAUTH_CLIENT_NAME = os.environ.get("DJANGO_OAUTH_CLIENT_NAME")
+OAUTH_AUTHORITY = os.environ.get("DJANGO_OAUTH_AUTHORITY", "http://example.com")
+OAUTH_CLIENT_NAME = os.environ.get("DJANGO_OAUTH_CLIENT_NAME", "benefits-oauth-client-name")
+OAUTH_CLIENT_ID = os.environ.get("DJANGO_OAUTH_CLIENT_ID", "benefits-oauth-client-id")
```
afeld, line 173: At risk of over-complicating things: should it to throw an error if settings like these aren't configured in production? Not a blocker; happy to create a follow-up issue.

.devcontainer/.env.sample
afeld: In my humble opinion, the ideal is for the app to Just Work™️ in development without any special configuration, so we shouldn't leave anything in the `.env.sample` file.

benefits/settings.py
angela-tran: We are currently doing that check, but it is deep within the flow (it's when we [show the button to the user](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L95-L96)). I think it makes sense to check it more upfront, though we'd need to figure out the appropriate place to do it. A follow-up issue sounds good to me.

.devcontainer/compose.yml
thekaveman: This is the internal port... but sure? I don't know if `mkdocs` expects to be running on a certain port or not.

benefits/settings.py
angela-tran: > A follow-up issue sounds good to me.

Er, that is - if the consensus is that we want to do this 😄

docs/configuration/environment-variables.md
```diff
@@ -127,6 +161,10 @@ Django's primary secret, keep this safe!
 
 ### `DJANGO_TRUSTED_ORIGINS`
 
+!!! warning "Deployment configuration"
+
+    You must change this setting when deploying the app to a non-localhost domain
+
```
angela-tran, line 167: These call-outs for deployment configuration settings are so helpful! 👏

.devcontainer/compose.yml
thekaveman: Changed in [ff47855](https://github.com/cal-itp/benefits/pull/598/commits/ff47855768355d154513d267f01bbe34116c253e), went back to a dynamic local port but `8001` for the internal port to be clear we're different from the app.

.devcontainer/.env.sample
angela-tran: @afeld Are you thinking that we shouldn't have `.env.sample` at all?

afeld: That's the (eventual?) goal I'm suggesting, yep.

thekaveman: I think we could actually get there if it weren't for the `DJANGO_LOCAL_PORT`. Thoughts on hardcoding to `11369` for the CDT auth server?

The others can be set in the `compose.yml` file directly.

Then maybe `.env.sample` goes back to being purely an example of all the possible environment variables, but we ditch the docs that instruct developers to first create an `.env` from it?

angela-tran: > Thoughts on hardcoding to 11369 for the CDT auth server?

I'd be ok with that as the default.

> Then maybe .env.sample goes back to being purely an example of all the possible environment variables, but we ditch the docs that instruct developers to first create an .env from it?

Sounds good. I'd also be ok with `.env.sample` not having _all_ the possible environment variables, but rather ones that could be helpful to override maybe? I don't feel strongly either way, just an idea.

angela-tran (dismissed):
I think these changes are good to merge in. Nice work on this simplification @thekaveman

machikoyasuda (dismissed):
👏

thekaveman:
Fine, I lied - the changes actually were not so bad. Got rid of the `.env.sample` entirely!

thekaveman:
OK done, rebased on latest `dev`

tests/pytest/core/test_session.py
```diff
@@ -200,7 +200,8 @@ def test_rate_limit_time_default(app_request):
     t0 = session.rate_limit_time(app_request)
     t1 = int(time.time())
 
-    assert t1 >= t0
+    # rate limit expiry time should be in the future
+    assert t0 >= t1
```
angela-tran, line 204: Nice catch with fixing this 💥

angela-tran (approved):
🚀

tests/pytest/core/test_session.py
thekaveman: It was only through adding a more sane default in the `settings.py`, and having the test run (and fail!) that I caught it 🙄

afeld (approved):
Nice work!

Issue #501, opened 2022-04-27
Refactor view helpers out of view files


afeld:
Our `views.py` files combine both functions that are used as views, plus functions that they use as helpers. When scanning those files, I've found it a bit confusing for both to be present… requires _just a little extra brainpower_ to distinguish which is which. The proposal here is to move any functions that _aren't_ views to a different file, e.g. an adjacent `view_helpers.py`. Thoughts?

PR #606, opened 2022-05-20
Refactor: replace internal EV API code with eligibility-api package

### LINKED ISSUE #596: Consume `eligibility-api` package instead of internal code
The scope of this issue is to replace usage of the internal `eligibility/api.py` code with code from the extracted [eligibility-api](https://github.com/cal-itp/eligibility-api) package. Since that package has not been published to PyPI yet (see https://github.com/cal-itp/eligibility-api/issues/22), it is ok to point at the GitHub repo for now.

angela-tran:
Closes #596


Dockerfile
```diff
@@ -24,7 +24,9 @@ RUN useradd --create-home --shell /bin/bash $USER && \
     chown -R $USER /home/$USER && \
     # install server components
     apt-get update && \
-    apt-get install -qq --no-install-recommends gettext nginx
+    apt-get install -qq --no-install-recommends gettext nginx && \
+    # install git for use when installing dependencies from VCS (https://pip.pypa.io/en/stable/topics/vcs-support/)
+    apt-get install -qq --no-install-recommends git
```
angela-tran, line 29: Quick notes: Installing git here is necessary to use pip's [VCS support](https://pip.pypa.io/en/stable/topics/vcs-support/) in `requirements.txt` (see [line 5](https://github.com/cal-itp/benefits/pull/606/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R5) in `requirements.txt`).

I don't expect this line to be around long-term, so I wanted to separate it out and make it stick out so that it doesn't look like a long-term change. 

If there are any strong feelings on this though, let me know! (For example, maybe you'd rather we just put `git` at the end of line 27. Or maybe the comment should be more explicit about this being temporary.)

.devcontainer/Dockerfile
```diff
@@ -8,7 +8,7 @@ USER root
 RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash
 
 RUN apt-get install -qq nodejs npm libgtk2.0-0 libgtk-3-0 libgbm-dev libnotify-dev \
-    libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb curl git jq ssh
+    libgconf-2-4 libnss3 libxss1 libasound2 libxtst6 xauth xvfb curl jq ssh
```
angela-tran, line 11: Similar to my note on Dockerfile about the short-term need to install git in the app container, I'd love to know if others are ok with change here. We would just need to remember to put this back later.

angela-tran:
This is a WIP because I'm waiting on cal-itp/eligibility-api#25 to be merged in. Once that's done, I can remove the hash from `requirements.txt` here.

Dockerfile
thekaveman: Thanks for the reasoning around keeping it on a separate line, I was going to comment and ask about that.

Maybe just create an issue to follow-up and remove this when you consume the PyPI package.

requirements.txt
```diff
@@ -2,6 +2,7 @@ Authlib==1.0.1
 cryptography==37.0.2
 Django==3.2.13
 django-csp==3.7
+git+https://github.com/cal-itp/eligibility-api@54f93e20ed2adb797b7bcd94a14b16fb3f8c2342#egg=eligibility_api # WIP: pointing at specific commit with types_to_verify fix
```
thekaveman, line 5: I recommend we point at `main` so we don't have to constantly update this reference here.

Each time the container is rebuilt it will grab the latest `main` from API. So we just have to make sure that it always stays stable.

benefits/eligibility/views.py
```diff
@@ -9,11 +9,13 @@
 from django.utils.decorators import decorator_from_middleware
 from django.utils.translation import pgettext, gettext as _
 
+from eligibility_api import api
```
thekaveman, line 12: Related to cal-itp/eligibility-api#26, I feel like we should be importing `client` (or even `Client` directly) here, not `api`, since the whole thing is `eligibility_api`.

Since the API spec is split into a client/server model, it makes sense to organize the code around that too.

tests/pytest/eligibility/test_views.py
```diff
@@ -13,7 +13,7 @@
 
 from benefits.core import session
 from benefits.core.models import TransitAgency, EligibilityVerifier
-from benefits.eligibility.api import ApiError, TokenError
+from eligibility_api.api import ApiError, TokenError
```
thekaveman, line 16: See my other comment related to cal-itp/eligibility-api#26 and organization within the package.

thekaveman (changes requested):

Dockerfile
angela-tran: Created #612 for this and put it in the "Blocked" column since it requires cal-itp/eligibility-api#22 to be done.

.devcontainer/Dockerfile
angela-tran: Noted in #612 that we need to put this back later.

thekaveman (approved):
This is a good start.

I think this could be a little cleaner in the view, e.g. keeping the `api.py` module to wrap all the `Client` creation and types expansion etc.

But we can revisit later or in #501.

PR #629, opened 2022-05-27
Mobile: New footer design

### LINKED ISSUE #539: Mobile: Footer - Redesign Footer for Mobile
There are new designs for this now:

image
image

- Add stacked footer https://designsystem.digital.gov/components/footer/ - 40 px height links with 2px solid white border bottom, on all but the last link
- Mobile header should be 80px high
- Mobile button changes: Width and height

machikoyasuda:
closes #539 

## What this PR does
- [x] Add stacked footer, similar to https://designsystem.digital.gov/components/footer/ with 50 px height links with 2px solid white border bottom, on all but the last link
- [x] Mobile header should be 80px high
- [x] Need to confirm with @srhhnry - Are there any changes necessary for the link hover style?* See below

## How to test this PR
- Test out links on mobile: That you can click the _whole bar_ to get to the link
- Test out the footer style on Desktop, iPad and any other size: Should be no regressions to the non-mobile footer design

## Screenshots

- Mobile: Smallest Phone width
image

- Tablet Wide - Tablet will not have mobile-style footer.
image

- Tablet Height - Tablet will not have mobile-style footer.
image

- Desktop
image


srhhnry:
Hey, so I did some research and referred to the USWDS protocol. I think we should treat these as links on a dark background, which means they get an action color of the Cal-ITP blue (Primary #046b99), which changes to the Hovers and Active State color (Hover #2f4c65). Additionally, I set the text as underlined to indicate links. Also, I bumped up the footer size from 40 to 50 height given our older demographic. This is all now reflected in the Figma file, shown below. 

I think eventually we'll use a more standard set of colors for text linking (blue and purple) but this is on brand and consistent for now. 


XS_Introduction

machikoyasuda:
> One (very minor) note - hover state isn't so much of a thing in mobile, right? There's no way to hover.
> 
> Active state, yes!

Yeah I mean the active state! It still is pretty prominent cos when you click on the link, the color/style changes.

machikoyasuda:
@srhhnry Got it!

I will increase the height to 50px.

The link colors, both the link (`#046B99`) and the hover (`#2f4c65`) do not have enough color contrast with the background (`#212121`). Since the font size is below 18, we need to be WCAG AAA color contrast compliant:

https://contrastchecker.com/?c=046B99&b=212121 Ratio 2.74

https://contrastchecker.com/?c=046b99&b=212121 Ratio 1.8

This tool is useful in finding the closest accessible color: https://accessible-colors.com/

srhhnry:
Thanks for looking into the color contrast--that's super helpful and I hadn't incorporated into my process yet. 

Okay, so for footer links, I've updated the text colors. The new default footer link color (#0590CD) is a compliant variation of the Cal-ITP blue. The selected/active color (#9B74D7) is a compliant variation on the USDWS recommended purple (their thinking is that this color is pretty standard and recognizable around the internet). 

image

benefits/static/css/styles.css
```diff
@@ -86,8 +86,13 @@ footer {
   padding-bottom: 1rem;
 }
 
-footer .footer-links a {
+footer.global-footer .footer-links a {
   font-size: 16px;
+  color: #0590cd;
+}
+
+footer.global-footer .footer-links a:hover {
```
angela-tran, line 94: This worked when I tested on Chrome in [device mode](https://developer.chrome.com/docs/devtools/device-mode/). 👍 

When I tested in Firefox in [Responsive Design Mode](https://firefox-source-docs.mozilla.org/devtools-user/responsive_design_mode/index.html) though, I didn't ever see the purple color. When I "tap" the link, it seems like the `:focus` state is taking effect, and the link turns an orange color which is coming from the CA Web Template styles. 

Here was the best screenshot I could take of it:
image

And here is simulating that state using devtools:
image

I think to be safe, we could add a selector for the `:focus` state here?

angela-tran (commented):
These changes look good to me. I left a note about something I noticed in Firefox only. Not sure if you'll see it in FF too and if it's a priority.

benefits/static/css/styles.css
machikoyasuda: Great catch, thank you!

Added the `focus` and `active` selectors.

angela-tran (approved):

angela-tran:
It worked in Firefox!

Issue #595, opened 2022-05-18
Separate out testing of `eligibility/confirm` view from testing of Eligibility API


angela-tran:
Tests were written in #575 to increase test coverage of `eligibility/api.py`. They were enough to test the functionality; however, the tests are covering too many units within one test.

For this issue, let's take those tests and separate those concerns out into tests that are focused on one unit at a time.

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/575#pullrequestreview-974647176_

> I don't want to block since this is definitely better than the low coverage in this area that we had before!
>
> This feels like combining a few different concerns into the same tests though, namely how the form/view work, and how the API client works. I am also not a fan of what is essentially a reimplementation of the token encryption and signing logic as a helper for test setup. That feels... bad.
> 
> I would think the view tests should use a fully mocked `client`, not even getting to the point of having to mock HTTP requests, encryption, etc. These tests would ensure the view logic is correct when various inputs and outputs of the API are processed.
> 
> Separate tests for the Eligibility API / client / etc. could get in to the mechanics of working with the request and response tokens, encryption keys, servers, and the like.

PR #639, opened 2022-06-04
Refactor: bring back eligibility API helper module

### LINKED ISSUE #622: Move fine-grained API-specific logic out of eligibility view
> This is a good start.
>
> I think this could be a little cleaner in the view, e.g. keeping the `api.py` module to wrap all the `Client` creation and types expansion etc.
> 
> But we can revisit later or in #501.

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/606#pullrequestreview-983660165_

thekaveman:
Simplifies logic in the Eligibility views by wrapping the API Client and request/response parsing into a helper function.

Closes #622

### DIFF
--- a/benefits/eligibility/api.py
+++ b/benefits/eligibility/api.py
@@ -1,0 +1,47 @@
+"""
+The eligibility application: Eligibility Verification API helpers.
+"""
+
+from django.conf import settings
+
+from eligibility_api.client import Client
+
+from benefits.core import session
+
+
+def get_verified_types(request, form):
+    """
+    Helper calls the eligibility verification API with user input.
+    Returns None and updates form with user input error(s).
+    Returns a list of verified eligibility types, or an empty list when no types were verified.
+    """
+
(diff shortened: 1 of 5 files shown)


afeld (approved):
Looking at the code coverage, there are a couple of untested paths in `benefits/eligibility/views.py#confirm`. How would you feel about filling those in as part of the refactor, to ensure nothing broke?

Otherwise, looks good!

thekaveman:
Totally agree that we want to fill in those gaps. Looks like coverage dropped from 65% to 62% with this PR (wonder if we can automate this "change in coverage" check/info?)

#595 captures what you're talking about I think? (and would be where I would want to do it). Feel free to update there with more info if needed.

angela-tran:
👀

angela-tran (approved):
These changes look great -- the `confirm` view code is more readable, and these new tests are very clear. 👍

PR #661, opened 2022-06-09
IAL2 Login.gov Flow: Redirect users to Eligibility Start

### LINKED ISSUE #656: Engineering: Back-end - Combine logic for Agency Index and Getting Started page
_As a MST Senior Rider, when I go to benefits.calitp.org, I should see the new Introduction page design and content._

- Change the app logic to the following -
1. When there is only 1 transit agency, go directly to the agency index page (/mst)
2. From the agency index page, go directly to the verifier selection page (/eligibility/)
3. When there is only 1 eligibility verifier, go directly to the only start page (Login.gov verifier)

- Write tests for this new logic
- Update sample data for this new logic (??) so we can adequately test this in Dev, Staging

## How it is now

```mermaid
graph LR
    A[core:index] -->|redirect| B(core:agency_index)
    B -->|user clicks button| C(eligiblity:index)
    C -->|redirect| D(eligibility:start)
```

## After this is implemented

```mermaid
graph LR
    A[core:index] -->|redirect| B(core:agency_index)
    B -->|*redirect*| C(eligiblity:index)
    C -->|redirect| D(eligibility:start)
```

machikoyasuda:
closes #656 

As a senior with a Social Security Number, when I go to benefits.calitp.org, I should get redirected to the Eligibility Start page.

## What this PR does
- Applies this new view logic: When there is only 1 agency and 1 verifier, redirect the user to the eligibility start page
- [x] Adds Pytest for this view logic

## To test this feature locally

You will have to change your database data to reflect a scenario in which there is only 1 agency and 1 verifier. There are 2 ways to do this -

### Code way

1. Remove `DefTL` agency from the `05_transitagency.json` file
2. For the `ABC` agency, Change `"eligibility_verifiers": [1,2]` to `"eligibility_verifiers": [1]`
3. Run `bin/init.sh` - This should remove your current database, create a new database and run the necessary migrations
4. Go to `localhost:8000`
5. The site should redirect you to `localhost:8000/eligibility/start`

### Admin UI way

1. Turn on Admin by adding `DJANGO_ADMIN=true`
2. Run `bin/init.sh` and set your username/password
3. Go to http://localhost:9000/admin/ 
4. Go to http://localhost:9000/admin/core/transitagency/1/change/- Uncheck **Active**, to mark ABC Transit as inactive.
5. Click **Save**
image

### DIFF
--- a/tests/pytest/core/test_views.py
+++ b/tests/pytest/core/test_views.py
@@ -25,6 +25,23 @@ def test_homepage_single_agency(mocker, client):
 
 
 @pytest.mark.django_db
+def test_homepage_single_agency_single_verifier(mocker, client):
+    # Agency set to DEFTl, which only has 1 verifier
+    agencies = TransitAgency.all_active()[1:]
+    mocker.patch("benefits.core.models.TransitAgency.all_active", return_value=agencies)
+
+    assert len(agencies) == 1
+    assert len(agencies[0].eligibility_verifiers.all()) == 1
+
+    path = reverse("core:index")
+    response = client.get(path)
+
+    assert response.status_code == 302
+    assert response.url == reverse("eligibility:start")
+    # assert response.url == agencies[0].index_url
+
+
+@pytest.mark.django_db
 def test_help(client):
     path = reverse("core:help")
     response = client.get(path)
--- a/tests/pytest/core/test_views.py
+++ b/tests/pytest/core/test_views.py
@@ -36,7 +36,7 @@ def test_homepage_single_agency_single_verifier(mocker, client):
     # https://docs.djangoproject.com/en/3.2/topics/testing/tools/#making-requests
     assert response.redirect_chain[0] == ("/deftl", 302)
     assert response.redirect_chain[1] == ("/eligibility/", 302)
-    assert response.redirect_chain[2] == ("/eligibility/start", 302)
+    assert response.redirect_chain[-1] == ("/eligibility/start", 302)
 
 
 @pytest.mark.django_db
--- a/tests/pytest/core/test_views.py
+++ b/tests/pytest/core/test_views.py
@@ -30,15 +30,13 @@ def test_homepage_single_agency_single_verifier(mocker, client):
     agencies = TransitAgency.all_active()[1:]
     mocker.patch("benefits.core.models.TransitAgency.all_active", return_value=agencies)
 
-    assert len(agencies) == 1
-    assert len(agencies[0].eligibility_verifiers.all()) == 1
+    response = client.get(reverse("core:index"), follow=True)
 
-    path = reverse("core:index")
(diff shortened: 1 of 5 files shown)


thekaveman:
I'd say if you want to update the sample fixtures to be more inline with the use-case we're working on, that's fine.

machikoyasuda:
@thekaveman I _do_ want to do that, but I also _don't_ want to break existing Cypress / Pytest tests. So I was thinking, we'd all just adjust our local databases?

tests/pytest/core/test_views.py
```diff
@@ -24,6 +24,23 @@ def test_homepage_single_agency(mocker, client):
...
+
+    assert len(agencies) == 1
+    assert len(agencies[0].eligibility_verifiers.all()) == 1
+
+    path = reverse("core:index")
+    response = client.get(path)
+
+    assert response.status_code == 302
+    assert response.url == reverse("eligibility:start")
+    # assert response.url == agencies[0].index_url
```
machikoyasuda, line 41: @angela-tran Any ideas what is wrong with this spec? On lines 33/34, I'm checking that `agencies` is set to have 1 agency with 1 verifier -- so the data seems correct to me. The `response.url` is coming back as `/mst`, though, though I expect it to be going to `eligibility/start`.

angela-tran: I believe what's happening is that when you call this code on lines 36 and 37,
```python
    path = reverse("core:index")
    response = client.get(path)
```
you're only going to get the first redirect. So it's only going through the first two pages:
```mermaid
graph LR
    A[core:index] -->|redirect| B(core:agency_index)
```

To make the pytest-django `client` fixture follow redirects, we can pass `follow=True` when calling `get`.

_From https://docs.djangoproject.com/en/3.2/topics/testing/tools/#making-requests:_
> If you set `follow` to `True` the client will follow any redirects and a `redirect_chain` attribute will be set in the response object containing tuples of the intermediate urls and status codes.

I tried it out locally, and here's what `redirect_chain` looks like:
image

thekaveman (approved):
This looks good, and sends the user on the correct flow!

One thought: having only 1 verifier is not _really_ the reason we are forwarding to the `eligibility:index` page (the verifier selection page).

We forward to `eligibility:index` because we're trying to get to the `eligibility:start` where the new combined instructions are. Having only 1 verifier is what makes that _second redirect_ happen.

So the redirect could be even simpler. Not a blocker, but something to consider - as this builds in a certain logic that doesn't necessarily map to how we're thinking about the flow. This will probably have to change once we have more than one verifier anyway, so again not a big deal.

```diff
-if len(agency.eligibility_verifiers.all()) == 1:
-  return redirect(reverse("eligibility:index"))
+return redirect(reverse("eligibility:index"))
```

PR #662, opened 2022-06-10
Have all resources be managed by Terraform

### LINKED ISSUE #618: Configure infrastructure as code
There are a lot of resources out there explaining the benefits of managing infrastructure as code (IaC) — [here is one](https://techcommunity.microsoft.com/t5/fasttrack-for-azure/the-benefits-of-infrastructure-as-code/ba-p/2069350). [Azure has strong support for Terraform](https://docs.microsoft.com/en-us/azure/developer/terraform/overview), and given that we can see everything that would need to be managed as part of the[resource group](https://portal.azure.com/#@digitalca.onmicrosoft.com/resource/subscriptions/9bdb8e29-156f-4fc9-a1fe-1bb6a915a4f0/resourceGroups/RG-CDT-PUB-VIP-CALITP-D-001/overview), I think the lift would be relatively low.

afeld:
…in the production resource group. Closes #618.


terraform/app_service.tf
```diff
@@ -0,0 +1,80 @@
...
+  resource_group_name = data.azurerm_resource_group.prod.name
+  os_type             = "Linux"
+  sku_name            = "P2v2"
+
+  lifecycle {
+    ignore_changes = [tags]
+  }
+}
+
+# app_settings, sticky_settings, and storage_account are managed manually through the portal since they contain secrets
```
afeld, line 13: Possible we could (at least partially) move these into Terraform, but I didn't want to spend a ton of time on it.

terraform/app_service.tf
```diff
@@ -0,0 +1,80 @@
...
+    ignore_changes = [app_settings, sticky_settings, tags]
+  }
+}
+
+resource "azurerm_linux_web_app_slot" "dev" {
+  name       = "dev"
+  https_only = true
+
+  # hack to match the casing of what's returned by Azure
+  app_service_id = replace(replace(azurerm_linux_web_app.main.id, azurerm_linux_web_app.main.name, lower(azurerm_linux_web_app.main.name)), data.azurerm_resource_group.prod.name, lower(data.azurerm_resource_group.prod.name))
```
afeld, line 53: Since the string cases don't match, it things the Resource ID has changed, even though Azure treats them case-insensitively. It then fails when trying to change, since you (in this case) can't change what App Service instance a slot is associated with.

terraform/main.tf
```diff
@@ -15,12 +15,11 @@ terraform {
 }
 
 provider "azurerm" {
-  # temporary workaround for permissions issue
-  skip_provider_registration = true
```
afeld, line 19: Fixed by @cal-itp/cdt-devsecops.

terraform/uptime/main.tf
```diff
@@ -1,13 +1,8 @@
-data "azurerm_application_insights" "benefits" {
-  name                = "AI-CDT-PUB-VIP-CALITP-P-001"
-  resource_group_name = var.resource_group_name
-}
```
afeld, line 4: Move this out of the module, since the resource should only be created once.

terraform/app_service.tf
```diff
@@ -0,0 +1,80 @@
...
+
+resource "azurerm_linux_web_app" "main" {
+  name                = "AS-CDT-PUB-VIP-CALITP-P-001"
+  location            = data.azurerm_resource_group.prod.location
+  resource_group_name = data.azurerm_resource_group.prod.name
+  service_plan_id     = azurerm_service_plan.main.id
+  https_only          = true
+
+  site_config {
+    ftps_state = "AllAllowed"
```
afeld, line 23: Not sure that we want to keep this enabled; will address in a follow-up.

terraform/app_service.tf
```diff
@@ -0,0 +1,80 @@
...
+      }
+    }
+  }
+
+  lifecycle {
+    ignore_changes = [app_settings, sticky_settings, tags]
+  }
+}
+
+resource "azurerm_linux_web_app_slot" "dev" {
```
thekaveman, line 48: So to configure our `test` slot, we'll just add another entry like this one and then run the terraform commands?

terraform/app_service.tf
```diff
@@ -0,0 +1,80 @@
...
+    vnet_route_all_enabled = true
+  }
+
+  identity {
+    identity_ids = []
+    type         = "SystemAssigned"
+  }
+
+  logs {
+    detailed_error_messages = false
```
thekaveman, line 66: Do you know what this does? Why `false`?

thekaveman (commented):
Should I expect to see anything about the custom domain and SSL configuration in here? That is a per-slot setting.

terraform/app_service.tf
afeld: Mind if we find out as a follow-up?

thekaveman (approved):

afeld:
Added custom domain config (on the App Service side) in https://github.com/cal-itp/benefits/pull/690. We don't have the ability to modify Front Door, so still need to go through DevSecOps for that. Created an issue around certificates: https://github.com/cal-itp/benefits/issues/704

PR #649, opened 2022-06-06
Fix(ci): use importlib as import mode

angela-tran:
Fixes #648 

See notes on the issue itself for more context.

Issue #648, opened 2022-06-06
Fix pytest runs in GH Actions


thekaveman:
`pytest` is currently failing in GH Actions on `dev` (and elsewhere):

* Last know good run on `dev`: https://github.com/cal-itp/benefits/actions/runs/2418089818
* Most recent run (failure) on `dev`: https://github.com/cal-itp/benefits/actions/runs/2449795228

## To Reproduce

Push new commits to a PR/branch to trigger `pytest`, see:

* #647 and the check run: https://github.com/cal-itp/benefits/runs/6763507049?check_suite_focus=true
* #638 and a recent check from from a rebase/push: https://github.com/cal-itp/benefits/runs/6762605601?check_suite_focus=true

## Expected behavior

These tests should pass in CI, especially as nothing is changing with Python code or the Python testing environment/dependencies.

## Additional context

Tests are passing locally in the devcontainer for @thekaveman and @angela-tran

thekaveman:
On the last run that worked on `dev`:

* [Resolved to use Commit `01b35d9`](https://github.com/cal-itp/benefits/runs/6679612653?check_suite_focus=true#step:5:11) from `eligibility-api`
* [Commit `01b35d9`](https://github.com/cal-itp/eligibility-api/commit/e27a2160)

On the first run that failed on `dev`:

* [Resolved to use Commit `404697e`](https://github.com/cal-itp/benefits/runs/6762724922?check_suite_focus=true#step:5:11) from `eligibility-api`
* [Commit `404697e`](https://github.com/cal-itp/eligibility-api/commit/404697e)

The [diff between them](https://github.com/cal-itp/eligibility-api/compare/01b35d9...404697e) gives no indication that this would have anything to do with the failure here.

angela-tran:
### Confusion
[...]

Issue #635, opened 2022-06-03
Login.gov IAL2 for age verification (seniors)


vykster:
**Date**: June 2022
**Writers**: @thekaveman @vykster @machikoyasuda @angela-tran @srhhnry @Indiajar @o-ram @jarrettkrumrei 
**Status**: In-Progress
**Priority**: [No. 1 per the Roadmap](https://docs.google.com/document/d/1IFoa8Ye0IXwGXXwxFjMrm1s3617Dbv6l-E-aCB0kgnA/edit#)

## Background
Login.gov is a digital identity management tool developed by the federal government and in use for a growing number of applications, recently this tool has been made available to state governments. Login.gov offers two account types, with the second an IAL2-like account users can "proof" their identity. Many older adults already use login.gov to access their SSI benefits. Integrating with a login.gov account allows us a reliable online source of someone's status as qualified for a senior program (65 years or older). Federal funding for transit includes requirements of operators that they offer senior discounts to riders. 

## Decision drivers
Age verification needs to be behind an identity proofed flow, age needs to be verified by a third-party. 

## Decision outcome

Given we are already integrating with the login.gov to create accounts, migrating to the the login.gov IAL2* identity proofed account type and using [user attributes](https://developers.login.gov/attributes/) to verify age allows us a rapid and controlled timeline without waiting for other the DMV to implement the id-linking tools and provides user experience than is possibly less confusing. Current limitations of this approach include the complex account registration process and strict requirement that users have a social security number, as from our research we know not all riders eligible for the senior discount programs have social security numbers. However given the use of [Login.gov for Supplemental Security Income (SSI)](https://secure.ssa.gov/RIL/SiView.action) we expect out of all groups of transit riders, older adults to have the most familiarity with login.gov.

We will need to accommodate flows for three classes of errors: 

* Eligibility not verified
* Login.gov failure
* Payment card not verified

## Acceptance criteria

- [x] Acceptance criteria

## Definition of Done

Benefits is live in production, older adults (age 65+) can login with an IAL2* proofed Login.gov account, and continue on to enroll their contactless payment card with Littlepay.

## User Stories

- [x] As a new Login.gov user, I want to understand what is required for an IAL2* proofed account, so that I know what to expect in the sign-up process
- [x] As a new Login.gov user, I want to proof my identity with Login.gov using IAL2*, so that I can verify my age in Benefits
- [x] As an existing Login.gov user, I want to login to my IAL2* proofed account, so that I can verify my age in Benefits
[...]

PR #668, opened 2022-06-13
Views: cleanup and testing

thekaveman:
## Background

With #635 we'll make some changes to the way the views work and the user flows through the app. A number of outstanding issues highlight some cleanup that can be done to the views, and the need to refactor existing tests while increasing coverage.

The following sections describe the major changes in this PR.

## Refactor view helpers

* _Closes #501_

Every function in each app's `views.py` file is now _actually a Django view_. Some helpers were refactored into the view functions as they weren't as widely used; in other cases helper module(s) were introduced.

## Increase view test coverage

* _Closes #504_
* _Closes #595_
* _Address [#639 comment (review)](https://github.com/cal-itp/benefits/pull/639#pullrequestreview-997184030)_

Prior to this PR, the [test coverage on `dev`](https://docs.calitp.org/benefits/tests/coverage/) was at `72%` overall, with a range of coverage for the different app views:

| Module | statements | missing | excluded | branches | partial | coverage |
| -- | -- | -- | -- | -- | -- | -- |
| benefits/core/views.py | 81 | 23 | 0 | 16 | 1 | 71% |
| benefits/eligibility/views.py | 98 | 31 | 0 | 24 | 9 | 62% |
| benefits/enrollment/views.py | 89 | 49 | 0 | 20 | 1 | 38% |
| benefits/oauth/views.py | 64 | 3 | 0 | 12 | 4 | 91% |
| **Total** | 1438 | 356 | 0 | 396 | 54 | **72%** |

This PR brings the overall test coverage up to `80%`, with each app's views fully covered:

| Module | statements | missing | excluded | branches | partial | coverage |
| -- | -- | -- | -- | -- | -- | -- |
| benefits/core/views.py | 81 | 0 | 0 | 16 | 0 | 100% |
| benefits/eligibility/views.py | 105 | 0 | 0 | 24 | 0 | 100% |
| benefits/enrollment/views.py | 89 | 0 | 0 | 18 | 0 | 100% |
| benefits/oauth/views.py | 40 | 0 | 0 | 2 | 0 | 100% |
| **Total** | 1452 | 245 | 0 | 390 | 37 | **80%** |

This PR also refactored _most_ view tests to favor using the [Django test `client`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#the-test-client) over importing and calling the view function directly. This allows for middleware processing and generally keeps the tests more focused (and works better with the cleaned up fixtures described in the next section).

Some view functions are still called directly where it was necessary (e.g. to test the `bad_request()` view in `benefits.core.views`).

## Normalize test helpers/fixtures

* _Closes #503_

In #648 we saw pytests fail in GH Actions because of how we were importing helpers from the root `tests/pytest/conftest.py` in lower-level test files. The fix in #649 got the tests to pass in CI, but the issue recently came back in local pytest runs.

This PR amends all helper functions across all tests and `conftest.py` to use the `@pytest.fixture` decorator, and amends all test functions to ["request" fixtures following the usual pytest semantics](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#requesting-fixtures).

Further, sub-directory tests (e.g. `tests/pytest/oauth/`) can define their own `conftest.py` with reusable fixtures scoped to that directory.


angela-tran:
Reviewing this 👀

thekaveman:
Let me push the rebase on `dev` before you submit a review!

thekaveman:
@angela-tran done. (Maybe I already had? added one more tiny refactor of the tests based on #656)

tests/pytest/eligibility/test_views.py
```diff
@@ -1,204 +1,272 @@
...
-    agency, verifier = set_verifier(mocker)
+@pytest.mark.usefixtures("mocked_eligibility_auth_request", "mocked_session_eligibility")
+def test_confirm_get_verified(client, mocked_session_update):
+    path = reverse(ROUTE_CONFIRM)
+    response = client.get(path)
 
-    httpretty.register_uri(httpretty.GET, "http://localhost/verify", status=404)
+    assert response.status_code == 302
+    assert response.url == reverse(ROUTE_ENROLLMENT)
+    mocked_session_update.assert_called_once()
```
angela-tran, line 202: Some tests in `eligibility/test_views.py` are failing for me because they assert that `session.update` is called once, but if you have DEBUG mode set to True, the `DebugSession` middleware will call `session.update`, causing the test to fail.

image

image

image

Should we disable the `DebugSession` middleware when running these tests?

tests/pytest/enrollment/test_views.py
```diff
@@ -1,30 +1,222 @@
...
+    return {"card_token": "tokenized_card"}
+
+
+@pytest.fixture
+def invalid_form_data():
+    return {"invalid": "data"}
+
+
+@pytest.mark.django_db
+def test_token_ineligibile(client):
```
angela-tran, line 29: Small typo here (`ineligibile`)

angela-tran (commented):
I'm still reviewing (on commit 16 out of 37). I can go ahead and add some small comments I have so far.

tests/pytest/eligibility/test_views.py
thekaveman: Interesting! I had not run the tests with `DJANGO_DEBUG=true` sorry about that! Will look into this.

thekaveman: @angela-tran do you have `DJANGO_DEBUG=True` in your `.env` file?

angela-tran: @thekaveman Yes - it's `DJANGO_DEBUG=true`. When I change it to `DJANGO_DEBUG=false` and rebuild the devcontainer, re-run the tests, all the tests pass.

thekaveman: OK, can we take this as a follow-up then?

With #598 we added that env override to `launch.json`, so the app always launches in debug mode from the devcontainer `F5`, regardless of the `.env` file variable. You don't _actually_ need this variable in your local `.env` anymore.

Here it looks like the `.env` file variable is impacting the test runs, and I think we could go either way in supporting the tests under debug mode and not? Since we shouldn't be running the app in debug mode anywhere out there.

angela-tran: Thanks for reminding me that `launch.json` can help us easily develop in with DEBUG mode on AND run the tests locally with DEBUG mode off. I'd say we don't even need a followup if we want to go ahead and say the tests don't support DEBUG mode.

angela-tran (approved):
I made it through. I didn't see anything incorrect in these refactors, and it seems like a lot was simplified here.

afeld (approved):
This is all good stuff, though it was a lot to reason about in a single pull request. Would be great to have smaller chunks for future refactors.

Nice work!

angela-tran:
If you'd like to remove `httpretty` from the test `requirements.txt`, I think it is safe to do so 🙏

benefits/enrollment/api.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, response, requested_id, payload=None):
         self.customer_ids = list(payload)
         self.updated_customer_id = self.customer_ids[0] if len(self.customer_ids) == 1 else None
         self.success = requested_id == self.updated_customer_id
+        self.message = "Updated customer_id does not match enrolled customer_id" if not self.success else ""
```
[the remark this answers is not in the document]
thekaveman: Yes, it used to be in the [`_enroll()` view helper](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py#L90).

I moved it to the `api` module since the error really is coming from the API call (and the [very specific logic of parsing it](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/api.py#L88)), and it felt a little weird to have the _error message_ defined in another place.

thekaveman:
This is the only commit that had a real meaningful change with the rebase: [351164c](https://github.com/cal-itp/benefits/pull/668/commits/351164cdb13656d02c14b91e0f568659bdef9564)

angela-tran (approved):

PR #669, opened 2022-06-13
format Python files using black extension

afeld:
~~This does a few things:~~

- ~~Sets the default formatter to be Prettier, which will also apply it to things like Markdown.~~
- Installs the Black extension to enable formatting Python files on save.
- ~~Formats on save by default, which is fine because formatting HTML is still disabled, protecting the Jinja templates.~~
- ~~Formats the settings files (with Prettier).~~

The rest has been moved to https://github.com/cal-itp/benefits/pull/746.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,6 +22,7 @@
     "eamodio.gitlens",
     "mhutchie.git-graph",
     "ms-python.python",
+    "ms-python.black-formatter",
     "ms-python.vscode-pylance",
     "esbenp.prettier-vscode"
   ]
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,10 +8,11 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
+  // https://github.com/microsoft/vscode-black-formatter/issues/82#issuecomment-1155492426
   "[python]": {
-    "editor.defaultFormatter": "ms-python.python"
+    "editor.defaultFormatter": "ms-python.black-formatter"
   },
-  "python.formatting.provider": "black",
+  "python.formatting.provider": "none",
   "python.languageServer": "Pylance",
   "python.linting.enabled": true,
   "python.linting.flake8Enabled": true,
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get install -qq curl jq ssh
 USER $USER
 
 RUN python -m pip install --upgrade pip && \
-    pip install black flake8 pre-commit
+    pip install flake8 pre-commit
 
 COPY docs/requirements.txt docs/requirements.txt
 RUN pip install -r docs/requirements.txt


.devcontainer/devcontainer.json
```diff
@@ -34,6 +22,7 @@
     "eamodio.gitlens",
     "mhutchie.git-graph",
     "ms-python.python",
+    "ms-python.black-formatter",
```
thekaveman, line 25: What is the difference between installing the extension and [installing `black` directly](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/Dockerfile#L16) as part of the devcontainer build? My understanding is the `black` formatting has already been working for a while now...

.vscode/settings.json
```diff
@@ -13,24 +8,11 @@
...
   "files.trimTrailingWhitespace": true,
-  "[javascript]": {
-    "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "[json]": {
-    "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
   "python.formatting.provider": "black",
```
thekaveman, line 11: I'm pretty sure this is what tells VS Code to use `black` - see comment above.

.devcontainer/devcontainer.json
afeld: `black` is installed as a package, but wasn't actually being run from the editor. The extension does the latter. Steps to reproduce, without this pull request:

1. Modify a Python file in a way that should be undone with formatting (adding whitespace or whatever).
1. Save — the file doesn't change.
1. From the command palette, run `Format Document`.

   Screen Shot 2022-06-13 at 1 18 18 PM

1. This dialog appears:

   Screen Shot 2022-06-13 at 1 18 26 PM

.vscode/settings.json
afeld: Correct, but it appears the extension is needed as well.

.devcontainer/devcontainer.json
thekaveman: I don't see this behavior on the existing `dev`

1. Modify a Python file
2. Save - the file is formatted with `black`

save-black

afeld: Weird. Can you confirm you don't have the `black` extension installed? @machikoyasuda @angela-tran Do you get formatting on save, or no?

thekaveman: Confirmed

image

angela-tran: @afeld Formatting on save has been working for me as well. I wonder if there might be some difference with Macs?

afeld: Mind trying rebuilding on this branch? If that works for you all, mind if we add the extension, even if it only makes a difference for me (for some reason)?

angela-tran: There are some troubleshooting tips here: https://code.visualstudio.com/docs/python/editing#_troubleshooting-formatting

angela-tran: I can also rebuild on your branch and make sure that having the extension would be ok

thekaveman:
Let's remove `black` from the devcontainer build process and just use the extension since it seems to work for more cases!

.devcontainer/devcontainer.json
angela-tran: I followed these steps:


1. Modify a Python file in a way that should be undone with formatting (adding whitespace or whatever).
1. Save — the file doesn't change.
1. From the command palette, run `Format Document`.
1. I get a different dialog
image

When I click "Configure" and choose "Black Formatter", this is added to `.vscode/settings.json`, and code is formatted on save:
```
  "[python]": {
    "editor.defaultFormatter": "ms-python.black-formatter"
  }
```

afeld: Asked for clarification: https://github.com/microsoft/vscode-black-formatter/issues/82

thekaveman (commented):
I'm now getting this error message when I _Rebuild and Reopen_ the devcontainer on this branch:

image

I'm also _not_ getting Format on Save anymore:

no-format-on-save

afeld:
Good gracious. Asked in https://github.com/microsoft/vscode-black-formatter/issues/82#issuecomment-1155478052. Does format on save work after you click `Install and Reload`?

thekaveman:
Going back to draft while this is sorted out to remove from the review queue.

afeld:
Split this pull request into two to hopefully allow the other changes to get in.

How are we feeling about this pull request? [The `Install and Reload` problem is a broader bug](https://github.com/microsoft/vscode/issues/149309); dunno if that's a deal-breaker for folks or not. Again, I'm not writing enough Python on this project that having to install the extension manually is a huge deal for me, so if everyone else is happy with the current setup, fine to just close this.

.vscode/settings.json
```diff
@@ -8,10 +8,11 @@
   "files.insertFinalNewline": true,
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
+  // https://github.com/microsoft/vscode-black-formatter/issues/82#issuecomment-1155492426
```
afeld, line 11: [FYI that this file supports comments.](https://code.visualstudio.com/Docs/languages/json#_json-with-comments)

afeld:
On second thought, since I got formatting Python on save to work without the extra extension, this doesn't add value, and having to `Install and Reload` is annoying.

Issue #495, opened 2022-04-27
Speed up time to development environment


afeld:
Coming into this project, I experienced some frustration at how long it was taking to rebuild the devcontainer when I would switch branches. I learned some shortcuts:

- Don't worry about rebuilding the container for most branch changes
- If the dependencies change, just run:

    ```sh
    pip install -r requirements.txt -r tests/pytest/requirements.txt
    ```

but it still takes a while to get the devcontainer ready when first starting it. Ideas for how to speed this up:

- [ ] Move [`postAttach.sh`](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/postAttach.sh) commands into `Dockerfile`, where possible
- [ ] Run the SQLite database in-memory, avoiding the need to delete the database in some cases

thekaveman:
> Move [postAttach.sh](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/postAttach.sh) commands into Dockerfile, where possible

I (think?) this is mostly not possible because of Docker volumes. Anything that we install during Docker build in the source directories, like `node_modules/` or `.git/hooks`, will be overwritten when the devcontainer volume mounts to the host machine. Maybe there is another way around this?

> Run the SQLite database in-memory, avoiding the need to delete the database in some cases

Cool idea / approach! Definitely interested in seeing more about this.

> Move Cypress (and thus Node.js?) stuff to its own container

This is probably the biggest part of the time it takes when recycling the devcontainer.

Part of the issue I ran into before when looking at this is with container interaction between the devcontainer and other containers that aren't started as part of the devcontainer start up process - it seems harder to get this to work.

@angela-tran @machikoyasuda and I have talked about it a number of times and usually come back to it being worth it to have Cypress installed in the devcontainer. See also the notes above about `postAttach` which apply to the Cypress stuff too - we can't pre-initialize too much in the Docker build step or it is overwritten by the volume mount.

> Move nginx to its own container

I'm curious about this one - nginx is installed in the app container and used in the deployed environments, it doesn't come in to play for development time in the devcontainer. But definitely impacts overall Docker image build time.

> Don't start docs service by default

Sure. Probably the least time consuming out of the items on the this list, but also easy enough to start on its own.

afeld:
> > Move [postAttach.sh](https://github.com/cal-itp/benefits/blob/dev/.devcontainer/postAttach.sh) commands into Dockerfile, where possible
>
> not possible because of Docker volumes … Maybe there is another way around this?

[...]

PR #671, opened 2022-06-13
Chore: Faster devcontainer start

afeld:
Punchline up front: This **speeds up devcontainer start time by around five minutes** on my machine.

The start time of a devcontainer was becoming a regular annoyance for me, enough so that I was finding myself working outside of it, which kinda defeats the purpose. I spent a bit of time to improve a couple of things that would make a big difference:

- Install pre-commit dependencies at `dev` image build time, which is infrequent/cached
- Don't bother installing Cypress in the devcontainer, since it wasn't being used (and maybe wasn't even working?) in there anyway. @machikoyasuda Can you confirm?

While it doesn't complete _all_ the tasks listed, I'm going to say this closes https://github.com/cal-itp/benefits/issues/495 as being "good enough."

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,3 +20,10 @@ RUN pip install -r docs/requirements.txt
 
 COPY tests/pytest/requirements.txt tests/pytest/requirements.txt
 RUN pip install -r tests/pytest/requirements.txt
+
+# install pre-commit environments in throwaway Git repository
+# https://stackoverflow.com/a/68758943
+COPY .pre-commit-config.yaml .
+RUN git init . && \
+    pre-commit install-hooks && \
+    rm -rf .git
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -3,6 +3,3 @@ set -eu
 
 # initialize pre-commit
 pre-commit install --overwrite
-
-# install cypress
-cd tests/cypress && npm install && npx cypress install
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -1,11 +1,8 @@
 #!/usr/bin/env bash
 set -eu
 
-# initialize hook environments
-pre-commit install --install-hooks --overwrite
-
-# manage commit-msg hooks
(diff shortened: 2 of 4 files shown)


thekaveman (approved):
Looks good to me. I do think we should have an alternative to running Cypress locally - maybe another container (though I think there are some challenges there).

I would like @machikoyasuda to weigh in especially from the Cypress POV.

machikoyasuda (approved):
Good for me 👍 

I confirmed that Cypress runs correctly with the existing Documentation instructions: 
image

The one thing is that everyone now must be able to run Cypress locally - @angela-tran @thekaveman @afeld -- can y'all run the Cypress specs locally?

afeld:
> I do think we should have an alternative to running Cypress locally - maybe another container

[Cypress does support Docker Compose](https://www.cypress.io/blog/2019/05/02/run-cypress-with-a-single-docker-command/#Docker-compose), though any time X11 is involved, I think it may be more trouble than it's worth 🙃

thekaveman:
> any time X11 is involved, I think it may be more trouble than it's worth 🙃

Oh goodness yes! I meant a container to run the CLI-based tests, not the UI!

afeld:
> can y'all run the Cypress specs locally?

I can run it, though I get some failures. Might be worth a screen share to troubleshoot at some point, since I feel like I've heard others say the same.

> I meant a container to run the CLI-based tests, not the UI

Since we're already going to recommend people install Cypress locally and GitHub Actions runs it directly in the runner, not sure there's a ton of value. We can re-add later if needed.

angela-tran:
I realize I'm a bit late to this convo and that the branch has already been merged, which is totally fine. 

I just wanted to echo @thekaveman 's preference for having an alternative to running Cypress locally. I would use Cypress in the devcontainer to run tests with the CLI. The thing that didn't work from the devcontainer was launching the test runner UI. Hopefully we have some time in the future to look into defining a separate container.

thekaveman:
Also a late realization: this PR did not remove any of the node/npm installation steps that were only required because Cypress was installed in there too! Building the devcontainer from scratch still takes a while with all these steps... will post a quick follow-up.

PR #679, opened 2022-06-13
Feat: verifier oauth fields

### LINKED ISSUE #678: Add text fields to the `EligibilityVerifier` model to hold scope/claims info
We need two additional text fields on the `EligibilityVerifier` to hold configuration data for that verifier when they require authentication.

* [ ] `auth_scope` - Optional text field containing the additional scope(s) (space separated) to send (we always send `openid`)
* [ ] `auth_claim` - Optional text field containing the claim to validate

thekaveman:
Closes #678 

We needed two fields here after learning that the `scope` we send and the `claim` that is returned are slightly different. 

A follow-up could be to refactor the `AuthProvider` model fields into the `EligibilityVerifier` (since there are only 2 extra fields), and change the way `requires_authentication()` works to simplify the models a bit. Saving that for another time 😉


benefits/core/migrations/0001_initial.py
```diff
@@ -103,13 +118,62 @@ class Migration(migrations.Migration):
...
+        migrations.AddField(
+            model_name="eligibilityverifier",
+            name="public_key",
+            field=models.ForeignKey(
+                help_text="The Verifier's public key, used to encrypt requests targeted at this Verifier and to verify signed responses from this verifier.",  # noqa
+                on_delete=django.db.models.deletion.PROTECT,
+                related_name="+",
+                to="core.pemdata",
+            ),
+        ),
```
angela-tran, line 178: The changes to this file look fine to me. It does make me curious though: what steps did you follow to get this result, and should that be our norm be for updating this file? 

(Based on the comment at the top that this was generated, I would guess the steps were to delete the file and then run the `manage.py` module with something like `makemigrations`?)

angela-tran: Ah thanks for including that -- I should've checked the commit messages. Re: writing docs on this, I'd be ok with it being part of this PR or as follow up. Either way works for me!

thekaveman: Took a follow-up in #683

PR #682, opened 2022-06-14
chore: add DevSecOps as CODEOWNERS for Terraform stuff

afeld:
@cal-itp/cdt-devsecops This means you will be notified on relevant pull requests. Mind commenting here to confirm that works for you?


thekaveman:
I don't see a value in adding this at this time.

PR #684, opened 2022-06-14
Feat: use oauth claims

### LINKED ISSUE #641: Confirm eligibility in the `oauth.authorize` callback

angela-tran:
Closes #641 

 * Follows **Option 1** from #641 
    * on successful OAuth login, add returned claims to the session
    * then in the eligibility confirm view, see if the session's verifier can verify eligibility using that claim
 * Refactors `eligibility/api` to `eligibility/verify` which contains api-based verification logic and oauth-based verification logic
    * @thekaveman and I found that this made more sense than having one method on the `EligibilityVerifier` model
 * Refactor `EligibilityVerifier` to have one `EligibilityType` so that putting `auth_scope` and `auth_claim` on the verifier directly makes more sense
 * Adds unit tests to preserve coverage of existing code and add coverage of new code


benefits/oauth/views.py
```diff
@@ -54,9 +54,16 @@ def authorize(request):
...
-    # this is the minimal amount of information needed later to log the user out
-    session.update(request, oauth_token=token["id_token"])
+    # We store the id_token in the user's session. This is the minimal amount of information needed later to log the user out.
+    id_token = token["id_token"]
+    # We store the returned claim in case it can be used later in eligibility verification.
+    verifier = session.verifier(request)
+    if verifier.auth_claim:
+        userinfo = token.get("userinfo")
+        claim_value = token.get("userinfo").get(verifier.auth_claim) if userinfo else None
+        claim = {verifier.auth_claim: claim_value} if claim_value else None
```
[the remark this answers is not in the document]
angela-tran: The `claim_value` is just `True`. It seems more meaningful to store the full key-value pair (`{"is_senior": True}`).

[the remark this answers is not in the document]
thekaveman: Ahh interesting OK I didn't realize it came back that way. It still might be easier to store the `auth_claim`? Then it is just a simple string compare later with `verifier.auth_claim == session.oauth_claim()`.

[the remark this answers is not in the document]
angela-tran: Sure, that sounds like a fine way to do it. I'll make the changes for that soon. (I have some rebasing to do with all the recent merges 😁)

benefits/core/models.py
```diff
@@ -108,6 +110,43 @@ def by_id(id):
         logger.debug(f"Get {EligibilityVerifier.__name__} by id: {id}")
         return EligibilityVerifier.objects.get(id=id)
 
+    def get_verified_types(self, form=None, agency=None, oauth_claim=None):
```
thekaveman, line 113: Since it can be tricky to know which combination of kwargs to use here, I propose we split this function into two:

* One for API verification w/`form` and `agency`
* Another for oauth verification w/`oauth_claim`

angela-tran:
Had a very productive pair-programming session with @thekaveman to add tests for the new claims-based verification logic.

The only thing left to do here is refactor `EligibilityVerifier` to hold one eligibility type instead of many.

benefits/oauth/views.py
```diff
@@ -43,9 +43,16 @@ def authorize(request):
...
-    # this is the minimal amount of information needed later to log the user out
-    session.update(request, oauth_token=token["id_token"])
+    # We store the id_token in the user's session. This is the minimal amount of information needed later to log the user out.
+    id_token = token["id_token"]
+    # We store the returned claim in case it can be used later in eligibility verification.
+    verifier = session.verifier(request)
+    if verifier.auth_claim:
+        userinfo = token.get("userinfo")
+        claim_value = token.get("userinfo").get(verifier.auth_claim) if userinfo else None
+        claim = verifier.auth_claim if claim_value else None
```
thekaveman, line 53: @angela-tran I think this branch is something we still need to cover in the tests, yes?

angela-tran:
With my latest commit d8e8dc7, the branch is now back at 100% coverage for all views.

image

angela-tran:
Depending on which one gets merged in first, either #698 or this PR will need to update some tests to use the `mock_oauth_client_instance` fixture that is introduced in #698.

benefits/oauth/views.py
angela-tran: Added coverage here: https://github.com/cal-itp/benefits/pull/684/commits/40bc88a4eb4e0404c00141cd5d2d7f8590236890

angela-tran:
@thekaveman In https://github.com/cal-itp/benefits/pull/684/commits/bb04d2155dbd6c3c66a2345020bc689d5b3eed79 I made some changes that we didn't discuss, but I noticed the need for them when looking at the fixture data. It seemed incorrect to me that the `abc` TransitAgency fixture had only `"eligibility_types" : [1]` and yet had `eligibility_verifiers` where one has `"eligibility_type": 1` and another has `"eligibility_type": 2`. The motivating question behind my changes was "Why does an oauth-based verifier get to ignore the agency's `eligibility_types`?", with the answer being it shouldn't.

thekaveman:
@angela-tran LMK when you're ready for review here

angela-tran:
Rebased and updated tests that needed to use the new `mocked_oauth_client_instance` fixture. This is ready for review

tests/pytest/oauth/test_views.py
```diff
@@ -71,6 +73,40 @@ def test_authorize_success(mocked_oauth_client_instance, mocked_analytics_module
     assert result.url == reverse(ROUTE_CONFIRM)
 
 
+@pytest.mark.django_db
+def test_authorize_success_with_claim(mocked_session_verifier_auth_required, mocked_oauth_client_instance, app_request):
+    # mocked_session_verifier_auth_required is a fixture that mocks benefits.core.session.verifier(request)
+    # call it here, passing a None request, to get the return value from the mock
+    verifier = mocked_session_verifier_auth_required(None)
```
thekaveman, line 80: Not a blocker, but the other syntax I used on that new fixture could be used here too (I might go back and clean all these up, this seems more clear):

```python
verifier = mocked_session_verifier_auth_required.return_value
```

thekaveman (approved):
This looks great! Thanks for all your work with the back and forth on implementation!

Issue #655, opened 2022-06-08
Engineering: Front-end / Mobile - Redesign the Enrollment, Enrollment Success, Eligibility pages to a no-image / single-column design.


machikoyasuda:
- Redesign the Enrollment, Enrollment Success, Eligibility pages to a no-image / single-column design
- Make sure to check the design with and without the Sign out button. 
- For Desktop, no changes are necessary.

<img width="1073" alt="image" src="https://user-images.githubusercontent.com/3673236/172735631-3f158904-212b-4c95-8143-408cea7ef832.png">

thekaveman:
Closed by #685

PR #685, opened 2022-06-14
Mobile: One column layout for certain mobile pages

machikoyasuda:
#655 

## What this PR does
- Creates a new class `no-image-mobile` that shows the side bar image on Desktop, but hides it on mobile
- Apply this class to the following pages: Enrollment, Enrollment Success, Enrollment Retry

## How to test
- http://localhost:9000/enrollment/ with logout button
image
image

image
image

- http://localhost:9000/enrollment/success
image
image

- Enrollment Success with Log out button
image
image


thekaveman:
Regarding the conflicts: this is the only change that is really in conflict (based on the refactors in #668): https://github.com/cal-itp/benefits/pull/685/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R34

The `_index()` helper function where you made that change was mostly copy/pasted into the `index()` view function.

machikoyasuda:
@thekaveman ready for re-review

benefits/static/css/styles.css
```diff
@@ -383,13 +382,19 @@ footer.global-footer .footer-links a:active {
 /* Enrollment Success */
 
 .enrollment-success #login {
-  padding: 0.3rem 0.6rem 0.1rem 0.6rem;
-  vertical-align: middle;
+  padding: 0.1rem 0.5rem 0.3rem 0.6rem;
+  display: inline-block;
+  padding-top: 0;
```
thekaveman, line 387: does this `padding-top: 0` undo the `padding: 0.1rem ...` from above?

thekaveman (approved):
Looks good! Button is in the correct place and 1 column on mobile is great.

Not blocking but I do wonder about all the extra space between the bottom of the content and the footer on mobile:

image

No matter what screensize, I can't see the footer without scrolling - maybe we could address this some time?

machikoyasuda:
> Looks good! Button is in the correct place and 1 column on mobile is great.
> 
> Not blocking but I do wonder about all the extra space between the bottom of the content and the footer on mobile:
> 
> image
> 
> No matter what screensize, I can't see the footer without scrolling - maybe we could address this some time?

The reason why this screen looks like it has a lot of margin is because there is the invisible Connect Card button there, taking up space, but not visible:

image

benefits/static/css/styles.css
machikoyasuda: will fix this in the next pr

Issue #621, opened 2022-05-24
Use production Login.gov


afeld:
## Acceptance Criteria

- [x] On the production site, when a user clicks the sign in button, they get directed to https://secure.login.gov.

## Additional context

There is currently only a development authorization server, which uses the Login.gov sandbox for authentication.

## What is the definition of done?

- [x] CDT gets contract signed
- [x] Production configuration for authorization server is complete with Login.gov
- [x] Production authorization server available
  - [x] Production Benefits app configured to point to it

thekaveman:
This is complete on https://benefits.calitp.org.

Leaving open until the rest of cal-itp/benefits-secrets#13 is complete, as we may need to swap back and forth for other testing.

thekaveman:
We did it!

PR #693, opened 2022-06-15
chore: reflect the sticky settings in Terraform

afeld:
No change to configuration, just moving more things into Terraform.


terraform/app_service.tf
```diff
@@ -40,8 +40,32 @@ resource "azurerm_linux_web_app" "main" {
...
+      "APPINSIGHTS_PROFILERFEATURE_VERSION",
+      "APPINSIGHTS_SNAPSHOTFEATURE_VERSION",
+      "APPLICATIONINSIGHTS_CONFIGURATION_CONTENT",
+      "APPLICATIONINSIGHTS_CONNECTION_STRING",
+      "ApplicationInsightsAgent_EXTENSION_VERSION",
+      "DJANGO_ALLOWED_HOSTS",
+      "DJANGO_OAUTH_AUTHORITY",
+      "DJANGO_OAUTH_CLIENT_ID",
+      "DJANGO_SECRET_KEY",
+      "DJANGO_TRUSTED_ORIGINS",
```
thekaveman, line 55: From [`linux_web_app#sticky_settings`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#sticky_settings):

> [app_setting_names](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#app_setting_names) - (Optional) A list of app_setting names that the Linux Web App will not swap between Slots when a swap operation is triggered.

I think most of these `DJANGO_*` settings should _not_ be sticky, they correspond to the slots.

On this list, the only one that may be the same across slots is `DJANGO_OAUTH_CLIENT_ID` (but even that could change once we finish with #621).

Not on this list, but could be added (they are not currently slot settings):

* `DJANGO_INIT_PATH` - now that we have the mounts normalized for each env, this is the same)
* `DJANGO_OAUTH_CLIENT_NAME` - just the internal app identifier for the OAuth client

thekaveman (changes requested):

afeld:
The `plan`:

```
      ~ sticky_settings {
          ~ app_setting_names       = [
                # (3 unchanged elements hidden)
                "APPLICATIONINSIGHTS_CONFIGURATION_CONTENT",
              - "APPLICATIONINSIGHTS_CONNECTION_STRING ",
              + "APPLICATIONINSIGHTS_CONNECTION_STRING",
                "ApplicationInsightsAgent_EXTENSION_VERSION",
              - "DJANGO_ALLOWED_HOSTS",
              - "DJANGO_LOG_LEVEL",
              - "DJANGO_OAUTH_AUTHORITY",
                "DJANGO_OAUTH_CLIENT_ID",
              - "DJANGO_SECRET_KEY",
              - "DJANGO_TRUSTED_ORIGINS",
              + "DJANGO_OAUTH_CLIENT_NAME",
              + "DJANGO_INIT_PATH",
                "DiagnosticServices_EXTENSION_VERSION",
                # (7 unchanged elements hidden)
            ]
            # (1 unchanged attribute hidden)
        }
```

Not sure why there was whitespace after `APPLICATIONINSIGHTS_CONNECTION_STRING` (in the state), but it was giving an error when I left it in 🤷

thekaveman (approved):
👍

thekaveman:
@afeld Maybe I'm confused... this seems to have done exactly the opposite of what I thought we wanted? Here's a screenshot from the `dev` slot - the ✔️ indicates it is a Slot setting (so varies by slot), and these are all the ones that should actually be _sticky_ (not vary) -- right??

image

afeld:
Attempting to clarify in https://github.com/cal-itp/benefits/pull/736.

Issue #689, opened 2022-06-15
Machiko's list of fixes to make


machikoyasuda:
- Double check all English copy
- On mobile, the icons are on a separate row in the updated comp

thekaveman:
Adding one more from https://dev-benefits.calitp.org

![image](https://user-images.githubusercontent.com/1783439/173742664-f04fc30c-0613-4736-b105-8a23a59c0103.png)

srhhnry:
In no particular order

- Touch targets on desktop Help URLs span the page, which can create a misleading or unexpected hover effect
- On mobile, we took out the text link "Learn more about Cal-ITP benefits" entirely, as it links to the same location as the Help button at the bottom
- Just checking that the link color was updated for footers (#73B3E7)
- On mobile, the icons are on a separate row in the updated comp

thekaveman:
@machikoyasuda here's another weird case: the user is already logged in to Login.gov, but goes back to the Start page somehow. We show a "Continue" button instead of the Login.gov button again, and it looks like we need a little spacing on that one.

## Desktop

![image](https://user-images.githubusercontent.com/1783439/173957835-06635d48-fa2b-47ce-92bb-a5999fe2aa0e.png)

## Mobile

![image](https://user-images.githubusercontent.com/1783439/173957874-2daa49e2-9bf5-452a-9c20-751a0b5b5e04.png)

machikoyasuda:
- Update new photos https://drive.google.com/drive/folders/1a94hVd_MzFVZgruK2jG0_6eQ_vcIxT_n

PR #697, opened 2022-06-15
Fix: Eligibility Start - Links

machikoyasuda:
closes other part of #689 

```
- Touch targets on desktop Help URLs span the page, which can create a misleading or unexpected hover effect
- On mobile, we took out the text link "Learn more about Cal-ITP benefits" entirely, as it links to the same location as the Help button at the bottom
```


thekaveman (approved):
Tested locally, looks great

PR #698, opened 2022-06-16
Feat: dynamic OAuth scope from verifier

### LINKED ISSUE #640: Re-configure OAuth `scope` setting for age verification
## Background

In #598 we simplified the developer setup experience and runtime configuration by creating better defaults in `settings.py` and only using the environment / `.env` file for specific overrides.

Part of that effort saw the OAuth `scope` parameter move from an environment setting to be hard-coded in `settings.py`, which was fine for the basic IAL1 Single Sign-On use-case of #363.

We need to be able to send additional scopes for IAL2 / #635. So we need this setting to be configurable and not explicitly hard-coded to only the single `openid` scope.

## Additional context

The scope we need to send looks like `age_verification` or `age_verification:is_senior` for the specific Senior use-case. So in total we'll send two scopes (they are space separated):

```
openid age_verification:is_senior
```

## Thoughts

The additional scope is basically a stand-in for our `EligibilityType` in the [Eligibility Verification API Request](https://docs.calitp.org/eligibility-api/specification/#request). One approach could be to model needed scope(s) on the `EligibilityType` and inject into the OAuth request from the configured agency/verification type(s).

## Decision

* Use verifier's scope field when configuring the OAuth client
* See #678 for the definition of the scope field
* See #641 for how we'll verify eligibility

thekaveman:
Closes #640.

The refactor in #668 make this a lot easier - it created a helper function `benefits.oauth.client.instance()` to get the client instance. This PR makes two changes to that helper:

1. Each returned instance is now distinct; before it was a singleton
2. The helper accepts a `scope` param and adds it to the list of scopes configured in the client's `client_kwargs` attribute

The hardcoded scope (`openid`) in `settings.py` was removed in favor of always adding it in this helper.

### DIFF
--- a/benefits/oauth/client.py
+++ b/benefits/oauth/client.py
@@ -8,9 +8,14 @@ from authlib.integrations.django_client import OAuth
 logger = logging.getLogger(__name__)
 
 
-def instance():
+_OPENID_SCOPE = "openid"
+
+
+def instance(scope=None):
     """
     Get an OAuth client instance using the OAUTH_CLIENT_NAME setting.
+
+    Optionally configure with a (space-separated) scope.
     """
     if settings.OAUTH_CLIENT_NAME:
         logger.debug(f"Using OAuth client configuration: {settings.OAUTH_CLIENT_NAME}")
@@ -20,5 +25,12 @@ def instance():
         client = oauth.create_client(settings.OAUTH_CLIENT_NAME)
     else:
         raise Exception("OAUTH_CLIENT_NAME is not configured")
+
+    scopes = [_OPENID_SCOPE]
+
+    if scope:
+        scopes.append(scope)
+
+    client.client_kwargs["scope"] = " ".join(scopes)
 
     return client
--- a/benefits/oauth/client.py
+++ b/benefits/oauth/client.py
@@ -5,25 +5,20 @@ from django.conf import settings
 from authlib.integrations.django_client import OAuth
 
 
-_OAUTH_CLIENT = None
-
-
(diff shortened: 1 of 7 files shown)


benefits/oauth/client.py
```diff
@@ -5,25 +5,32 @@
...
+        client = oauth.create_client(settings.OAUTH_CLIENT_NAME)
+    else:
+        raise Exception("OAUTH_CLIENT_NAME is not configured")
+
+    scopes = [_OPENID_SCOPE]
+
+    if scope:
+        scopes.append(scope)
+
+    client.client_kwargs["scope"] = " ".join(scopes)
```
afeld, line 34: Is there any risk to doing this modification after the client is created? Worried it could lead to a subtle bug.

tests/pytest/oauth/test_client.py
```diff
@@ -16,14 +16,37 @@ def test_instance_no_oauth_client_name():
...
+    assert scope1 in client_scope
+    assert client._OPENID_SCOPE in client_scope
+
+    scope2 = "scope2"
+    oauth_client2 = client.instance(scope2)
+    client_scope = oauth_client2.client_kwargs["scope"]
+
+    assert scope1 not in client_scope
+    assert scope2 in client_scope
+    assert client._OPENID_SCOPE in client_scope
```
afeld, line 43: Hmm, not sure I would check internal properties like this. Could we generate the redirect URL and check that?

afeld (approved):
A couple questions/suggestions, not blockers.

Issue #683, opened 2022-06-14
Docs to describe model/migrations update process


thekaveman:
The changes to this file look fine to me. It does make me curious though: what steps did you follow to get this result, and should that be our norm be for updating this file? 

(Based on the comment at the top that this was generated, I would guess the steps were to delete the file and then run the `manage.py` module with something like `makemigrations`?)

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/pull/679#discussion_r896968973_

PR #710, opened 2022-06-17
Refactor: models and migrations

thekaveman:
* Remove the `help_text` from model definitions, retain as comments
* Helper script to regenerate the migrations file
* Docs for how we handle this


angela-tran:
I think this is for #683, so I'll go ahead and link that issue to this PR

angela-tran (approved):
The docs and helper script look great!

PR #712, opened 2022-06-17
Refactor: oauth client config

### LINKED ISSUE #707: Switch (back) to static OAuth client config
In https://github.com/cal-itp/benefits/pull/698, we switched from having the `scope` specified under `AUTHLIB_OAUTH_CLIENTS` in settings to specifying it in the client after creation. I worry this could lead to some subtle bug, where Authlib is doing something special with the scope on client initialization.

An approach that would be more in line with Authlib's recommended setup would be to specify multiple clients in settings: one for each (verifier) use case, I guess? That would be something like `ial1` and `ial2_age`.

## Acceptance Criteria

- [ ] `scope` is specified in settings, rather than at runtime

thekaveman:
Closes #707 

I followed an approach adapted from this SO answer: https://stackoverflow.com/a/64174413

It made sense to consolidate all configuration values into the same place, and I went with the `AuthProvider` models / fixtures. There are no more env vars or settings in `settings.py`, but OAuth is still configured statically:

1. Fixtures specify `client_name`, `client_id`, `authority`, `scope`
2. On container start up, database built once, creating models from fixtures
3. On app start up, register an OAuth client for each `AuthProvider`
4. In views and elsewhere, create OAuth client instances from the registry with the user's `EligibilityVerifier.auth_provider`

Note I'm currently basing this on #710, so it would be good to get that one merged in first.


benefits/oauth/client.py
```diff
@@ -1,36 +1,52 @@
...
+from benefits.core.models import AuthProvider
+
 
 logger = logging.getLogger(__name__)
 
+oauth = OAuth()
 
-_OPENID_SCOPE = "openid"
 
+def client_kwargs(scope=None):
```
angela-tran, line 17: Would this function ever be used outside of this module? If not, I suggest prefixing with a `_`

benefits/oauth/client.py
```diff
@@ -1,36 +1,52 @@
...
-def instance(scope=None):
+    `scope` should be a space-separated list of scopes to add.
     """
-    Get an OAuth client instance using the OAUTH_CLIENT_NAME setting.
+    scopes = ["openid", scope] if scope else ["openid"]
+    return {"code_challenge_method": "S256", "scope": " ".join(scopes)}
+
 
-    Optionally configure with a (space-separated) scope.
+def server_metadata_url(authority):
```
angela-tran, line 27: Same comment as for the function above this one: Would this function ever be used outside of this module? If not, I suggest prefixing with a `_`

benefits/oauth/apps.py
```diff
@@ -0,0 +1,18 @@
...
+from django.apps import AppConfig
+
+
+class OAuthAppConfig(AppConfig):
+    name = "benefits.oauth"
+    label = "oauth"
+    verbose_name = "Benefits OAuth"
+
+    def ready(self):
+        from .client import oauth, register_providers
```
afeld, line 13: You meant to do this import in a function? If so, mind adding a comment explaining why?

angela-tran (commented):
I really like that this refactor puts the client configuration on `AuthProvider` - it's much more aligned with our original reasoning for creating that model. This approach seems more robust.

benefits/oauth/client.py
```diff
@@ -1,36 +1,52 @@
...
+    Register OAuth clients into the given registry, using configuration from AuthProvider models.
+    """
+    logger.info("Registering OAuth clients")
 
-    if scope:
-        scopes.append(scope)
+    providers = AuthProvider.objects.all()
 
-    client.client_kwargs["scope"] = " ".join(scopes)
+    for provider in providers:
```
afeld, line 44:
```suggestion
    # https://stackoverflow.com/a/64174413
    for provider in providers:
```

benefits/oauth/client.py
thekaveman: You're right, it probably would not! 👍

benefits/oauth/client.py
thekaveman: 👍

afeld (changes requested):
A couple little things. Nice work!

benefits/oauth/client.py
thekaveman: Rebasing, will add this.

angela-tran (dismissed):
💯

thekaveman:
@angela-tran rebased on `dev` 

Also raising `Exception` in `oauth` views if the client is not configured (related to my earlier trouble with registration, helps 
know about a missing client earlier and be more explicit about it)

angela-tran:
> Also raising Exception in oauth views if the client is not configured (related to my earlier trouble with registration, helps
know about a missing client earlier and be more explicit about it)

👍 This will definitely be more helpful for troubleshooting

angela-tran (approved):

afeld (approved):

PR #718, opened 2022-06-21
Feat: oauth unverified

### LINKED ISSUE #713: Redirect IAL2 users to unverified paths
Just like when API eligibility verification fails, and we send users to the `eligibility:unverified` page, we need to do the same for when OAuth claims verification fails.

## Acceptance Criteria

- [ ] As a user, when I cancel the Login.gov sign-up / IAL2 proofing process and am redirected back to Benefits, I see the unverified page
- [ ] As a user under 65 years of age, when I log in to my proofed Login.gov account and am redirected back to Benefits, I see the unverified page
- [ ] As a user 65 years of age or older, when I log in to my proofed Login.gov account and am redirected back to Benefits, I see the enrollment page

## Additional context

Currently unverified / canceled users see the confirm form. The configuration for the DMV confirm form will not be present at initial launch of this feature.

angela-tran:
Closes #713 

### Main changes here were:
 - add a branch in `eligibility/confirm` so that if the request is a `GET` _and_ the verifier uses auth verification _and_ the types come back as an empty list, we redirect to `eligibility/unverified` **instead of** a template with a form.
   - A `uses_auth_verification` method was added to the `EligibilityVerifier` model to more explicitly check that it should try to verify eligibility with claims returned from the auth provider.
 - add a `oauth/cancel` endpoint which redirects to `eligibility/unverified`
   - this is the endpoint that the auth server should redirect to if the user cancels out of identity proofing

### Other changes made:
 - allow null for fields meant for API-based verification
 - fix migration script to allow `auth_provider` field to be null
 - tweaks to msgids, fixtures, and tests


angela-tran:
Because we renamed some of the `msgid`s with `dmv` --> `oauth`, there are some missing `msgid`s when I run through the flow with our sample fixtures. We could either add those `msgid`s or change the fixture to use the renamed id.

Screenshot 2022-06-22 at 14-05-29 Transit Benefits Select a discount option Cal-ITP

Screenshot 2022-06-22 at 14-05-19 Transit Benefits eligibility pages unverified dmv title Cal-ITP

thekaveman:
@angela-tran let's no worry about making Cypress tests pass here please. We don't want to add back things that just confuse more later, so no need to add DMV msgids that aren't being used. We will delete/skip the tests instead.

angela-tran:
> @angela-tran let's no worry about making Cypress tests pass here please. We don't want to add back things that just confuse more later, so no need to add DMV msgids that aren't being used. We will delete/skip the tests instead.

Ok, sounds good. I'll drop that commit and force push

angela-tran:
Pushed a fix for this: 

One of the Cypress test failures caught a bug -- `verifier.uses_auth_verification` throws an error if the verifier's `auth_provider` is `None`. 
The test is `"Multiple verifier, no AuthProvider: Eligibility confirmation form spec"` -> `"Has a Courtesy Card number form label and corresponding field"`
image


Reproduced this locally by choosing the MST Courtesy Card verifier.
image

thekaveman (approved):
Thank you for finishing this off!

thekaveman:
I am going to update the fixture files in Azure `dev` / `test` now that this is merged.

Issue #631, opened 2022-06-01
Get traffic logs to appear for Azure apps


afeld:
User story: _As a developer on the Benefits app, I need to see logs appear for the Benefits app to help with troubleshooting._

## To Reproduce

There are various places to see (and various ways to configure?) logs in Azure. Click into each of the following:

- [Log Analytics workspaces](https://portal.azure.com/#view/HubsExtension/BrowseResource/resourceType/Microsoft.OperationalInsights%2Fworkspaces)
- [Log stream](https://portal.azure.com/#@digitalca.onmicrosoft.com/resource/subscriptions/9bdb8e29-156f-4fc9-a1fe-1bb6a915a4f0/resourceGroups/RG-CDT-PUB-VIP-CALITP-D-001/providers/Microsoft.Web/sites/AS-CDT-CALITP-D-001/logStream)
- [(Diagnostic?) logs](https://portal.azure.com/#@digitalca.onmicrosoft.com/resource/subscriptions/9bdb8e29-156f-4fc9-a1fe-1bb6a915a4f0/resourceGroups/RG-CDT-PUB-VIP-CALITP-D-001/providers/Microsoft.Web/sites/AS-CDT-CALITP-D-001/analytics)

All seem to be empty, or at least not showing as much as I would expect.

## Expected behavior

I expect to see logs similar to what appears locally when running the server, e.g.

```
[01/Jun/2022 00:07:43] INFO django.server:161 "GET /abc HTTP/1.1" 200 9414
```

## Additional context

- Possible there are Azure permissions issues at play here.
- One solution may be to send application logs the same way we send errors for https://github.com/cal-itp/benefits/issues/448: [through OpenCensus to Azure Monitor](https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python)

thekaveman:
I don't have permissions to run this command. Sent to @cal-itp/cdt-devsecops team to see if they can run it.

PR #719, opened 2022-06-22
Send logs to Azure Monitor

afeld:
Closes https://github.com/cal-itp/benefits/issues/631. Keeps the logging otherwise the same.

Used the following instructions, to start:

- https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python#configure-logging-for-django-applications
- https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python-request#tracking-django-applications

## Before merge

- [x] Set `APPLICATIONINSIGHTS_CONNECTION_STRING` in `dev`

Added tasks for each environment in https://github.com/cal-itp/benefits/pull/734.

### DIFF
--- a/benefits/logging.py
+++ b/benefits/logging.py
@@ -1,0 +1,28 @@
+def get_config(level):
+    return {
+        "version": 1,
+        "disable_existing_loggers": False,
+        "formatters": {
+            "default": {
+                "format": "[{asctime}] {levelname} {name}:{lineno} {message}",
+                "datefmt": "%d/%b/%Y %H:%M:%S",
+                "style": "{",
+            },
+        },
+        "handlers": {
+            "default": {
+                "class": "logging.StreamHandler",
+                "formatter": "default",
+            },
+        },
+        "root": {
+            "handlers": ["default"],
+            "level": level,
+        },
+        "loggers": {
+            "django": {
+                "handlers": ["default"],
+                "propagate": False,
+            },
+        },
+    }
--- a/benefits/logging.py
+++ b/benefits/logging.py
@@ -1,5 +1,8 @@
+import os
+
+
 def get_config(level):
-    return {
+    config = {
         "version": 1,
         "disable_existing_loggers": False,
         "formatters": {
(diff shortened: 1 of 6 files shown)


docs/deployment/infrastructure.md
```diff
@@ -63,6 +63,18 @@ flowchart LR
...
 We have [ping tests](https://docs.microsoft.com/en-us/azure/azure-monitor/app/monitor-web-app-availability) set up to notify about availability of the dev, test, and prod deployments. Alerts go to [#benefits-notify](https://cal-itp.slack.com/archives/C022HHSEE3F).
 
+## Logs
+
+We send application logs to [Azure Monitor Logs](https://docs.microsoft.com/en-us/azure/azure-monitor/logs/data-platform-logs). To find them:
+
+1. [Open Application Insights.](https://portal.azure.com/#view/HubsExtension/BrowseResource/resourceType/microsoft.insights%2Fcomponents)
+1. Click the resource corresponding to the environment.
+1. In the navigation, under `Monitoring`, click `Logs`.
+1. In the Query Editor, type `requests` or `traces`, and click `Run`.
```
afeld, line 73: Seems that traffic logs are sent to the former, stdout/stderr is sent to the latter.

thekaveman (approved):
A couple tests for the new configuration helper could be nice, but this looks good. Confirmed I still see the expected console output locally.

benefits/logging.py
```diff
@@ -0,0 +1,47 @@
...
+            "class": "opencensus.ext.azure.log_exporter.AzureLogHandler",
+            # send all logs
+            "logging_sampling_rate": 1.0,
+        }
+
+        # create custom logger
+        # https://github.com/census-instrumentation/opencensus-python/issues/1130#issuecomment-1161898856
+        config["loggers"]["azure"] = {
+            "handlers": ["azure"],
+            "level": "INFO",
```
thekaveman, line 44: Sorry I just noticed this - do we want to hardcode `INFO` here?

thekaveman: I'm not thinking about the local use-case, but the Azure use-case -- in `dev` (at least, and probably `test` and `prod` while we get things set up), we'll want `DEBUG` level output.

thekaveman (approved):

PR #738, opened 2022-06-24
Manage app settings through Terraform

afeld:
This is a proof of concept showing how we could manage app settings through Terraform. I'm still a little confused how best to do this with [swapping](https://docs.microsoft.com/en-us/azure/app-service/deploy-staging-slots), since it seems like the app settings would get changed outside of Terraform. Might take a bit of experimentation.


afeld:
Closing for now; will likely re-open/recreate as part of https://github.com/cal-itp/benefits/issues/830.

PR #744, opened 2022-06-24
allow the SHA of the deployed image to be viewed

afeld:
Also refactored the deploy pipelines to all use a single file. Can't confirm the pipeline works until we merge, unfortunately.


thekaveman (approved):
Looks great.

Just want to confirm we are sure `${{ github.ref_name }}` contains only the branch name and nothing else? I remember looking through this a while ago and thought I read something about there being more like `refs/HEAD/{branch}`?? I could be thinking of something else.

.github/workflows/deploy.yml
```diff
@@ -13,13 +15,16 @@ defaults:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    environment: dev
-    concurrency: dev
+    environment: ${{ github.ref_name }}
```
afeld, line 18: 90% sure `ref_name` is what we want.

afeld: https://github.com/github/docs/issues/15319 suggests it's correct. Let's find out.

afeld: Appears [it worked](https://github.com/cal-itp/benefits/runs/7047472242?check_suite_focus=true#step:6:9)!

thekaveman: Nice 👍 https://dev-benefits.calitp.org/static/sha.txt

PR #746, opened 2022-06-27
chore: improve formatting behavior

afeld:
Split out from https://github.com/cal-itp/benefits/pull/669, excluding the use of the new extension.

This does a few things:

- Sets the default formatter to be Prettier, which will also apply it to things like Markdown.
- Formats on save by default, which is fine because formatting HTML is still disabled, protecting the Jinja templates.
- Formats the settings files (with Prettier).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,10 +1,5 @@
 {
-  "[css]": {
-    "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "editor.defaultFormatter": null,
-  "editor.formatOnSave": false,
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[html]": {
     "editor.formatOnSave": false
   },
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -13,24 +8,14 @@
...
   "files.trimTrailingWhitespace": true,
-  "[javascript]": {
-    "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
-  },
-  "[json]": {
-    "editor.formatOnSave": true,
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.python"
```
afeld, line 12: This setting fixes format on save for Python files for me, since it overrides my global setting of `"editor.defaultFormatter": "ms-python.black-formatter"`.

angela-tran (approved):
Format-on-save still works for me with these changes. I also saw the addition of format-on-save for other files like Markdown and JSON files. ✅ 

I think @machikoyasuda will test this out as well

machikoyasuda (approved):

Issue #188, opened 2021-11-17
Use DJHTML to lint HTML, Django, Jinja tempaltes


machikoyasuda:
Created this ticket after #187 added Prettier linting for CSS and JS/JSON.

(Or find another tool for linting Django-flavored HTML)

machikoyasuda:
Another option: https://github.com/Riverside-Healthcare/djlint

thekaveman:
Yet another option: https://github.com/rtts/djhtml

PR #761, opened 2022-06-30
chore(pre-commit): add DJHTML pre-commit for indenting html template files

machikoyasuda:
#188 

Adds https://github.com/rtts/djhtml to pre-commit

This extension will indent HTML files.

To test this locally:

```
$ pip install djhtml
$ djhtml -i template.html
```

To see changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,3 +45,9 @@ repos:
     hooks:
       - id: prettier
         types_or: [javascript, css]
+
+  - repo: https://github.com/rtts/djhtml
+    rev: 'v1.5.1'  # replace with the latest tag on GitHub
+    hooks:
+      - id: djhtml
+      - id: djcss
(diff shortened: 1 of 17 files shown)


angela-tran:
Tried it out locally, and it works well 👍

thekaveman:
@machikoyasuda so this added a pre-commit check, what about a "on save" linter/formatter? Does it do that too? Kind of like how Prettier and Black run on save (but also in pre-commit).

machikoyasuda:
@thekaveman DJHTML does not have an `on save` command by default, but we can add it to VS Code like this: https://stackoverflow.com/questions/65320298/vscode-custom-code-formatting-and-auto-indention-for-languagedjango-template-in/68667543#68667543 

```
I use the extension [run on save](https://marketplace.visualstudio.com/items?itemName=emeraldwalk.RunOnSave) and add similar configuration in settings.json
```

```json
"emeraldwalk.runonsave": {
    "commands": [
        {
            "match": "\\.html$",
            "isAsync": true,
            "cmd": "./path/to/venv/bin/djlint ${file} --reformat --quiet" 
        }
    ]    
}
```

Issue #672, opened 2022-06-13
Complete penetration tests


afeld:
Follow-up to https://github.com/cal-itp/benefits/issues/521.

## Acceptance Criteria

- [ ] We are made aware of open security vulnerabilities

## Tasks

- [x] Update [testing plan](https://docs.google.com/document/d/1MDtGR01OkdEi9kOQVNM0dsHfYgvOoEiDqVZ2JHYbNng/edit#)
- [x] Reach back out to Bugcrowd about updated schedule: would like to start testing Wed 6/22
- [x] BugCrowd
  - [x] Test started
- [x] Jemurai
  - [x] Test started
  - [x] Findings received
  - [x] Create issues for any findings

afeld:
Testing by Jemurai has started; waiting to hear about timeline from BugCrowd.

afeld:
Jemurai scans completed; [artifacts](https://drive.google.com/drive/folders/1Zb2risjywexN71l4D2jqe_cTiUUuHKVc). Still waiting on BugCrowd to start.

afeld:
BugCrowd bounty has begun.

afeld:
New requirement from the CDT ISO that we have a penetration test performed by "A California Military Department". @jarrettkrumrei is getting more details.

thekaveman:
CMD completed; [artifacts](https://drive.google.com/drive/folders/1S7kRq-F7WBw9e2Fcbf-Z7SlXSZtR4S3r?usp=sharing).

Presented all findings to CDT ISO, collating remediations/notes for final approval next week Thursday August 11.

afeld:
Will track that approval as part of https://github.com/cal-itp/benefits/issues/621.

PR #770, opened 2022-07-08
deploy latest changes to test

afeld:
Might as well get them in before BugCrowd kicks off (#672) so they have the latest and greatest.


thekaveman (approved):
👍

Issue #776, opened 2022-07-11
Clean up Django .po files


machikoyasuda:
The purpose of this ticket is to clean up our naming conventions and names on the msgIDs for the Django .po files. 

- [x] Remove `dmv` `mst`
- [x] Perhaps rename `oauth` to something more descriptive? Or document what path `oauth` refers to
- [x] Come up with a better naming system for the current `mst` (courtesy card) flow pages

angela-tran:
I have some additional ideas on cleaning up our process around .po files:

1. I'd like for us to have a workflow that makes use of Django's `makemessages` utility. It seems like we lost that when we had to move some `msgid`s to our fixtures since running `makemessages` would comment out the ones that it can't see in our fixtures. One solution could be to have separate PO files for the fixture `msgid`s and have a script that combines those with the result of `makemessages`.
2. Avoid reusing `msgid`s. Even if the content is the same, we should define a separate `msgid`/`msgstr` entry so that reading through template files and .PO files is easier. One example of a confusing re-usage is [`agency_index.p[3]` on the `eligibility/start.html` template](https://github.com/cal-itp/benefits/pull/664#discussion_r897226484).

machikoyasuda:
Use `mst_cc` for new CC copy

machikoyasuda:
Closing - Addressed in https://github.com/cal-itp/benefits/pull/958 https://github.com/cal-itp/benefits/pull/966 https://github.com/cal-itp/benefits/pull/969

Issue #338, opened 2022-03-22
Refactor language files for multiple verifiers


thekaveman:
## Background

We need message strings for multiple verifiers

## Tasks

* [x] Create a new [verifier fixture](https://github.com/cal-itp/benefits/blob/dev/fixtures/02_eligibilityverifier.json) from the existing one
  * [x] With a new `pk` and `fields.name`
  * [x] Adjust both to have different `msgid`, e.g. `eligibility.pages.confirm.dmv.*` and `eligibility.pages.confirm.mst.*`
* [x] Adjust the `django.po` files with new `msgid` and new sample content for the Courtesy Cards verifier for each language
* [x] Wire the above fixture to one of the [sample agencies](https://github.com/cal-itp/benefits/blob/dev/fixtures/04_transitagency.json)

## Testing

Once #322 is merged

* [x] Remove your local db and local compiled language files, then run `bin/init.sh`
  * [x] Observe that the migrations run correctly, the fixture data is loaded, and the language files are compiled
* [x] Open the app locally and select the agency configured above with multiple verifiers
  * [x] Observe the `eligibility:index` view with different verifier selections
* [x] Select one of the verifiers
  * [x] Observe the verifier-specific `eligibility:start` view
  * [x] Confirm verification of a test user
* [x] Go back to `eligibility:index` and select the other verifier
  * [x] Observe the verifier-specific `eligibility:start` view
  * [x] Confirm verification of a test user

PR #785, opened 2022-07-13
Chore: improve PO file workflow

angela-tran:
One of multiple PRs for #776 

This PR attempts to improve our workflow for adding and removing `msgid`s/`msgstr`s to our `.po` files. 

The main problem with our current process is that we manually add entries to our `.po` files because not all our `msgid`s are in our code (see [#338](https://github.com/cal-itp/benefits/issues/338) for the reasoning for that) so we can't leverage Django's `makemessages`. As a result of this:
  - entries that become unused are not automatically removed
  - entries are not automatically organized in any sort of logical ordering or grouping
  - the generated comments containing line-number references from previous runs of `makemessages` are out-of-date and not useful

Another problem currently is that all the possible fixture `msgid`s have to be added to `django.po` even if your configuration of fixtures only needs a subset of them.

### Syntax-highlighting
There's a VS Code extension that does PO file syntax-highlighting and also allows you to Ctrl+Click on comments that contain line-number references to jump over to that code.

### Helper script
The bash script `bin/makemessages.sh` helps keep `msgid`s in our .po files up-to-date and organized.

It first runs `makemessages` which goes through our Django templates and code to find `msgid`s and also removes unused entries. Then, the script will look under `benefits/locale/<locale code>/fixture_messages/` for additional .po files to append to `django.po` for each locale.

This will make it easier for us to maintain and work with our .po files because:
 - unused entries are removed
 - entries are in a logical order (`makemessages` seems to go through each app's templates and then views)
 - the comments above each entry give context on where the entry is used
 - fixture `msgid`s are maintained in separate files (they still need to ultimately end up in `django.po`, but all editing should be done in their individual files)

#### Adding new entries
 - When you add a `msgid` to a template or in Python code, you run `bin/makemessages.sh` so that it's put into the correct place in `django.po`, and you go to that generated entry to provide the `msgstr`
 - When you add a `msgid` to a fixture, you manually create the `msgid`/`msgstr` entry in the fixture's `.po` files. Whenever you're ready for those to be in `django.po`, you run `bin/makemessages.sh`.

#### Editing existing entries
 - If the `msgid` comes from a template or Python code, you edit the `msgstr` in `django.po`
 - If the `msgid` comes from a fixture, you edit the `msgstr` in the fixture's `.po` file and then run `bin/makemessages.sh`


machikoyasuda:
Firstly, this is 💯 in scope, thoughtfulness and coding! The one thing that confused me a little bit is the naming of the `fixture` folder. The word `fixture` made me think it's associated with message strings that are necessary for testing only, or for sample data only (like how we use Fixtures here https://docs.calitp.org/benefits/configuration/fixtures/). But these `fixture` folder's contents are actual copy to be used for production apps, if I'm understanding this correctly. Maybe we need a different name for it?

thekaveman (commented):
Would you mind adding a few docs on this intended workflow / how this all works?

angela-tran:
@thekaveman Yes, I forgot about updating the docs! Thanks for reminding me.

@machikoyasuda That is a good point to clarify. The name `fixture_messages` made sense to me because those message strings are directly tied to what's in our verifier fixtures. For example, this sample [`OAuth claims via Login.gov verifier`](https://github.com/cal-itp/benefits/blob/811c625aeced06eea0868255d1276c1dbfa80491/fixtures/03_eligibilityverifier.json#L77-L85) has properties corresponding to the message strings in `oauth_verifier.po`.  We could even choose to **not** have our production message strings committed in this repo, similar to how we also don't have our production fixtures here. You're right that our fixtures are used as sample data, but they're also used for production too:
> During the [deployment](https://docs.calitp.org/benefits/deployment/) process, environment-specific fixtures are used to build that environment’s configuration database.

_(from https://docs.calitp.org/benefits/configuration/fixtures)_

angela-tran:
@thekaveman and @machikoyasuda , if there's anything else that seems confusing, please do leave a comment. My hope with this is to have a development process that we all feel good about.

machikoyasuda:
> 

Thanks for the clarification! The naming rationale makes sense to me now. I think we can table the question of, `We could even choose to not have our production message strings committed in this repo, similar to how we also don't have our production fixtures here.` later down the line, but it's definitely good to think about as we head towards a future of multiple agencies. We're not there yet though.

thekaveman:
> I think we can table the question of, We could even choose to not have our production message strings committed in this repo, similar to how we also don't have our production fixtures here. later down the line, but it's definitely good to think about as we head towards a future of multiple agencies. We're not there yet though.

Agree we don't need to solve for this PR, but this brings me back to @angela-tran's original description of the main problem:

> The main problem with our current process is that we manually add entries to our `.po` files because not all our `msgids` are in our code (see https://github.com/cal-itp/benefits/issues/338 for the reasoning for that) so we can't leverage Django's `makemessages`.

Maybe we should figure out how to go the _other_ way, and get all our `msgids` back into code so we can fully (and correctly) leverage Django's `makemessages`.

thekaveman:
> The main problem with our current process is that we manually add entries to our `.po` files because not all our `msgids` are in our code (see https://github.com/cal-itp/benefits/issues/338 for the reasoning for that) so we can't leverage Django's `makemessages`.

Here's an example of how we might be able to solve this issue: #789

angela-tran:
Potentially superceded by #814

PR #789, opened 2022-07-14
POC: data migrations

thekaveman:
[See discussion on #785](https://github.com/cal-itp/benefits/pull/785#issuecomment-1184884673)

This is a proof of concept to look at using [Django data migrations](https://docs.djangoproject.com/en/3.2/topics/migrations/#data-migrations) instead of our JSON fixture files for loading the initial configuration data.

Related to #785 in that this approach would get all the `msgid` into the Python code, allowing `makemessages` to correctly process them all.

This isn't necessarily intended to be merged, more of an exploration/example.


angela-tran:
I'm definitely interested in exploring this idea more. I really like that it would allow us to just use `makemessages` to process all the entries.

Since this would replace our JSON fixture files with a migration file (or files), I'm wondering: how would we define and load in different configurations (`dev`, `test`, `prod`, and our own local configs)?

angela-tran:
Superseded by #814

PR #285, opened 2022-01-13
fix(deps): revert django from 4.0.1 to 3.2.10

machikoyasuda:
closes #284 

- Reverts work from #256 and #276 
- Reverts to 3.2.10 because of outstanding bug #283 

## Testing
- Tested locally on 3.2.10
image

PR #793, opened 2022-07-18
Fix: enrollment overlay in Django 4.0.x

### LINKED ISSUE #283: Django 4.0.x breaks enrollment overlay
It seems that our recent update from Django `3.2.10` to `4.0` (also `4.0.1`) broke the payment processor card verification / enrollment overlay. This update has not been deployed to `prod` yet, and this breakage is the reason we've been holding that release.

## Background

After a user has verified eligibility, the following steps happen:

1. The user is redirected to the enrollment page
1. The enrollment page uses [AJAX to request the card tokenization script](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L12) from the payment processor 
1. Once the card tokenization script is loaded:
    * an [AJAX request](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L14)
    * to the [Django backend](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py#L96)
    * [acquires an access token](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/api.py#L203) from the payment processor
1. The access token is [passed as options to a JavaScript function](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L27) in the template, which is called immediately
    * the function name comes from the app config
    * the function implementation is from the payment processor's card tokenization script; it sets up a watcher for a click on the enrollment button
1. The user clicks the enrollment button, the card verification overlay opens
1. The payment processor's script makes an AJAX request to acquire a public key from the payment processor's server
1. The overlay's interface loads completely, showing a collection form
1. The user fills out their payment card details in the collection form, which submits directly to the payment processor
1. After successful submission to the payment processor, the card tokenization script [triggers a callback with the tokenized card](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L34), which is then sent via [HTML form submission](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/index.html#L42) to the Django backend.
1. The Django backend uses the tokenized card to [enroll the user into a discount group](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/api.py#L224)

## The issue

Somewhere in **Steps 3 - 7** is where this seems to break down. The payment processor's request for the public key from their own server (of **Step 5** above) works in `3.2.10` and fails in `4.0.x`.

Here are the headers for this request under `3.2.10` (overlay works):

```
:authority: web.paymentprocessor.com
:method: GET
:path: /api/xxxx/public-keys
:scheme: https
accept: */*
accept-encoding: gzip, deflate, br
accept-language: en-US,en;q=0.9
authorization: Bearer eeeeeeeeeeeeeeeeeeeeeeeeeeee.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx.yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy
dnt: 1
origin: https://verify.paymentprocessor.com
sec-fetch-dest: empty
sec-fetch-mode: cors
sec-fetch-site: same-site
```

And headers for the same request under `4.0.x` (overlay does not work):

```
:authority: web.paymentprocessor.com
:method: GET
:path: /api/xxxx/public-keys
:scheme: https
accept: */*
accept-encoding: gzip, deflate, br
accept-language: en-US,en;q=0.9
authorization: Bearer undefined
dnt: 1
origin: https://verify.paymentprocessor.com
sec-fetch-dest: empty
sec-fetch-mode: cors
sec-fetch-site: same-site
```

So obviously the `Bearer` token is missing in the `authorization` header in the latter example. But why? This is the same access token acquired in **Step 3** above, and that works under both `3.2.10` and `4.0.x`.

### `CSRF_TRUSTED_ORIGINS` 

That new setting that we had to fix in #268 was due to a change in `4.0` related to `Origin` header checking for `POST` and other "dangerous" (read: side-effect producing) requests. Except this request isn't coming _to our server_, or _from our server_, but _from the user's browser_, so this seems less relevant. I also tried adding various combinations of the payment processor's URLs into that setting to no avail. I'm pretty sure this isn't the problem.

### `JsonResponse`

Did this change at all in the new Django version? This is how the access token is sent back to the user's browser from Django in **Step 3** above.

## Reproduce

1. Open repo locally in VS Code (not devcontainer)
2. Edit `requirements.txt` to change the Django version: `3.2.10` (works) vs. `4.0` and/or `4.0.1` (does not work)
3. Make sure you have `fixtures/` for the payment processor (and pem data), and your `.env` `DJANGO_INIT_PATH` points to them
4. Inside `.devcontainer/`, run `docker compose build client`
5. Once built, run `docker compose up -d client server`
6. Get the localhost port and browse to the localhost site
7. Run through eligibility with the test EV server credentials
8. On the enrollment page, click the button to open the overlay. Notice if it does or does not load the form and/or show an error message.
9. Press `F12` to view Developer Tools for the overlay, reload the overlay with `F5` _while selected on the Developer Tools window_
10. Look on the Network tab for the call to `public-keys`. Notice it does or does not succeed with headers like shown above.
11. You can also `Rebuild and Reopen` in the devcontainer (after following these steps 1-4) and set a breakpoint to debug
    * the access token calls - from user to Django to payment processor to Django to user - succeed for all the versions

angela-tran:
Closes #283 

### Summary

This PR sets [`SECURE_CROSS_ORIGIN_OPENER_POLICY`](https://docs.djangoproject.com/en/4.0/ref/middleware/#cross-origin-opener-policy) to `same-origin-allow-popups`, which is the most secure option that still allows the overlay to work.

This setting was [introduced in Django 4.0.x](https://docs.djangoproject.com/en/4.0/releases/4.0/#requests-and-responses) and by default, sets the [`Cross-Origin-Opener-Policy` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy) to `same-origin`, which prevents cross-origin popups from accessing their opener's browsing context. Prior to Django 4.0.x, the header was unset.

If you look through the Javascript of our enrollment overlay, you'll notice that it makes use of [`Window.postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) to communicate with the Benefits window and needs to be able to store some variables on that window. So, it makes sense that this header value breaks it, since with it,
> the [`window.opener`](https://developer.mozilla.org/en-US/docs/Web/API/Window/opener) property of the new window will be `null`

### Other notes
 - More explanation on `Cross-Origin-Opener-Policy`: https://web.dev/why-coop-coep/#coop
 - Interesting question on StackOverflow: https://stackoverflow.com/questions/71888175/why-is-a-cross-origin-opener-policy-unsafe-none-header-unsafe
 - To more easily read minified Javascript, it helps to enable the experimental Chrome debugger feature `Automatically pretty print in Sources Panel`: https://stackoverflow.com/a/61816945

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -114,6 +114,10 @@ if not DEBUG:
 
 SECURE_BROWSER_XSS_FILTER = True
 
+# required so that cross-origin pop-ups (like the enrollment overlay) have access to parent window context
+# https://github.com/cal-itp/benefits/pull/793
+SECURE_CROSS_ORIGIN_OPENER_POLICY = "same-origin-allow-popups"
+
 # the NGINX reverse proxy sits in front of the application in deployed environments
 # SSL terminates before getting to Django, and NGINX adds this header to indicate
 # if the original request was secure or not
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 Authlib==1.0.1
-Django==3.2.14
+Django==4.0.6
 django-csp==3.7
 git+https://github.com/cal-itp/eligibility-api#egg=eligibility_api
 gunicorn==20.1.0


benefits/settings.py
```diff
@@ -113,6 +113,7 @@ def _filter_empty(ls):
     SESSION_COOKIE_SECURE = True
 
 SECURE_BROWSER_XSS_FILTER = True
+SECURE_CROSS_ORIGIN_OPENER_POLICY = "same-origin-allow-popups"
```
afeld, line 116: More 

```suggestion
# required for Littlepay popup
# https://github.com/cal-itp/benefits/pull/793
SECURE_CROSS_ORIGIN_OPENER_POLICY = "same-origin-allow-popups"
```

afeld (approved):

afeld:
Noting that this branch (and `dev`) is still on Django 3.2, so this setting won't have an effect.

angela-tran:
> Wow, that's really all the changes that are needed to upgrade to Django 4? Were you able to test the application end to end to verify nothing else in the success path is broken?

Yep, tested the success path end-to-end. 

image



A lot of the work had already been done in #256 and #268.
When we reverted back down to 3.2.x in #285, we didn't undo those changes.

thekaveman (approved):
Amazing work on this @angela-tran!! 🔎 🐛 💥 

I tested the end-to-end flow locally and it all worked ✅

machikoyasuda:
Yesssss 👏

PR #796, opened 2022-07-18
Docs: update links to Django documentation from 3.2 to 4.0

angela-tran:
Forgot about docs in #793 

Updates links to Django documentation from 3.2 to 4.0. Some minor formatting changes also were applied.


thekaveman (approved):
Thanks for this!

PR #797, opened 2022-07-18
Improve local image build process

thekaveman:
This PR does a few things:

1. New helper script `bin/build.sh` builds both the `client` and `dev` images
2. Move `compose.yml` up from the `.devcontainer/` directory to the root (❗also, `.env` has to move to the root❗)
3. Update docs with simplified Getting Started

The main goal here was to remove the `devcontainer/` directory from any setup flow and make building and rebuilding the images locally (e.g. for PR review) easier. 

`docker compose` commands can now be run directly from the root.


thekaveman:
Superseded by #802

PR #800, opened 2022-07-19
add page on student discount use case

afeld:
Closes https://calitp.teamwork.com/index.cfm#/tasks/26219542.

Screenshot of new students page

### DIFF
--- a/docs/.pages
+++ b/docs/.pages
@@ -1,5 +1,7 @@
 nav:
-  - Home: README.md
+  - Home:
+    - Overview: README.md
+    - Students: students.md
   - Getting started: getting-started
   - Development: development
   - Tests: tests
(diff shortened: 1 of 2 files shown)


thekaveman:
The Teamwork task seems to be deleted? Or I don't have permission to view it.

I'm not super keen on having this be a top-level page, for something that is planned/not part of the tech stack yet. What about reorganizing into a `use-cases/` directory, we could also fill out more current/planned use-cases as part of our growing roadmap.

afeld:
[You should now have Teamwork access.](https://cal-itp.slack.com/archives/C03E9VBEA8K/p1658257497493789?thread_ts=1658256469.952359&cid=C03E9VBEA8K)

>  What about reorganizing into a `use-cases/` directory

It's not immediately obvious to me how to do that without creating an empty Use Cases page, which I think would be confusing. This is such a low-traffic site… Would you mind if we leave it there for the moment, and set up a redirect later when we have other pages to group with it?

thekaveman:
> It's not immediately obvious to me how to do that without creating an empty Use Cases page, which I think would be confusing. This is such a low-traffic site… Would you mind if we leave it there for the moment, and set up a redirect later when we have other pages to group with it?

Sure, wrote up #809 to track

afeld:
Now live at https://docs.calitp.org/benefits/students/.

Issue #773, opened 2022-07-11
Fix App Service permission issue


afeld:
Split out from https://github.com/cal-itp/benefits/issues/686.

## To Reproduce

1. Check out the latest on the `dev` branch.
2. Make a change to one of the `azurerm_linux_web_app` or `azurerm_linux_web_app_slot` resources, e.g. `ftps_state = "Disabled"` to `"AllAllowed"`.
3. Run a `terraform apply`.
4. See error similar to:

    > Error: updating Linux Web App Slot: (Slot Name "dev" / Site Name "as-cdt-pub-vip-calitp-p-001" / Resource Group "rg-cdt-pub-vip-calitp-p-001"): web.AppsClient#CreateOrUpdateSlot: Failure sending request: StatusCode=403 -- Original Error: Code="LinkedAuthorizationFailed" Message="The client 'live.com#aidan@compiler.la' with object id '[REDACTED]' has permission to perform action 'Microsoft.Web/sites/slots/write' on scope '/subscriptions/[REDACTED]/resourceGroups/rg-cdt-pub-vip-calitp-p-001/providers/Microsoft.Web/sites/as-cdt-pub-vip-calitp-p-001/slots/dev'; however, it does not have permission to perform action 'join/action' on the linked scope(s) '/subscriptions/[REDACTED]/resourceGroups/RG-CDT-PUB-SHRD-W-P-001/providers/Microsoft.Network/virtualNetworks/VNET-CDT-PUB-SHRD-W-P-001/subnets/SNET-CDT-PUB-CALITP-P-001' or the linked scope(s) are invalid."

## Expected behavior

The `apply` should succeed.

## Additional context

none

afeld:
Preference of DevSecOps team is to sidestep the issue by using a service connection in a pipeline (https://github.com/cal-itp/benefits/issues/666). Will close this once we confirm that works.

afeld:
Confirmed we don't have this permissions issue when deploying via pipeline; not going to worry about the local permissions.

PR #805, opened 2022-07-19
chore: test change to slot

afeld:
Test for #773.

This is to confirm that our service connection in the pipeline has sufficient permissions to change the App Service slots. Will be reverted after merge and successful deploy.


afeld:
[It worked!](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=34552&view=logs&j=bc150913-bfac-577d-ab97-4eb22da69064&t=5c01d50c-3bf9-589e-956a-833a2240feb3&l=92)

PR #812, opened 2022-07-19
Model tests

### LINKED ISSUE #723: Pytests for models
While our [model classes](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py) mostly act as dumb data containers, there is a bit of logic in some of their helper functions and properties. 

Coverage would indicate [95% of models are covered](https://docs.calitp.org/benefits/tests/coverage/d_8c0b35f1ea7ee6af_models_py.html), but as we saw in #718 after working to turn off some of the Cypress (!!) tests, a bug was uncovered in a python model class.

## Acceptance Criteria

- [ ] 100% of models are covered
- [ ] Every static helper method and member property is tested, including edge cases

## Additional context

Build on those introduced in #718

thekaveman:
Closes #723.

Brings coverage of `benefits/core/models.py` to `100%`.

### DIFF
--- a/tests/pytest/core/test_models.py
+++ b/tests/pytest/core/test_models.py
@@ -136,6 +136,26 @@ def test_TransitAgency_supports_type_wrongtype(model_TransitAgency):
 
 
 @pytest.mark.django_db
+def test_TransitAgency_types_to_verify(model_TransitAgency):
+    eligibility = model_TransitAgency.eligibility_types.first()
+    new_eligibility = EligibilityType.get(eligibility.id)
+    new_eligibility.pk = None
+    new_eligibility.save()
+
+    assert eligibility != new_eligibility
+
+    model_TransitAgency.eligibility_types.add(new_eligibility)
+    assert model_TransitAgency.eligibility_types.count() == 2
+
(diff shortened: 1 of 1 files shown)


angela-tran:
👀

angela-tran (approved):
These tests look good to me!

I noticed in the [`pytest-django` docs](https://pytest-django.readthedocs.io/en/latest/database.html#enabling-database-access-in-tests) that we can apply a marker to the entire module by setting a top-level variable like `pytestmark = pytest.mark.django_db`. This could be nice to use for this model test module... I am ok either way, since there is some benefit to new developers for things to be more explicit.

thekaveman:
> I noticed in the [`pytest-django` docs](https://pytest-django.readthedocs.io/en/latest/database.html#enabling-database-access-in-tests) that we can apply a marker to the entire module by setting a top-level variable like `pytestmark = pytest.mark.django_db`. This could be nice to use for this model test module... I am ok either way, since there is some benefit to new developers for things to be more explicit.

Hmm that is interesting! Could help us clean up a bunch of tests/files (I'm honestly less worried about the newcomer, we can point to the docs in a comment or something). 

I'll create a follow-up issue for this.

Issue #775, opened 2022-07-11
Refactor out usage of `page.html`


angela-tran:
Refactor templates that extend `page.html` so that they instead define their own structure. Templates can still make use of our `viewmodel` classes if that is helpful. Then, delete `page.html`.

If it makes sense to open a separate PR for each template that needs refactoring, feel free to do so.

## Acceptance Criteria

- [ ] No user-facing changes -- all pages should look and function as before
- [ ] `page.html` no longer exists

## Additional context
 - `page.html` was intended to be a generic and reusable Django template that all our pages could extend. However, recently we've noticed it gets in the way more often than it helps. For example, in [#403](https://github.com/cal-itp/benefits/pull/403/files#diff-70ed6fd5fa2321e83db0d63df8312f680cd40428de4ada50dab22e93f55db7c7), we made `help.html` override a lot of `page.html` in order to achieve a more customized structure that was needed by Login.gov integration.
 - In tickets like [#374](https://github.com/cal-itp/benefits/issues/374) and [#516](https://github.com/cal-itp/benefits/issues/516), we started a pattern of having specific template files for specific pages. This is easier to read and modify.
 - We can still reuse code via [include](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#include)s.
 - In #668, you can see our shift towards this approach when https://github.com/cal-itp/benefits/commit/dd4b0198cbd6dc186afbfba93df9b29967172c52 removed `PageTemplateResponse`.

Here are all the templates that currently extend `page.html`:
```
benefits/core/templates/core/agency_index.html
benefits/core/templates/core/help.html
benefits/eligibility/templates/eligibility/confirm.html
benefits/eligibility/templates/eligibility/start.html
benefits/eligibility/templates/eligibility/unverified.html
benefits/enrollment/templates/enrollment/index.html
benefits/enrollment/templates/enrollment/retry.html
benefits/enrollment/templates/enrollment/success.html
benefits/templates/400.html
benefits/templates/404.html
benefits/templates/500.html
```

thekaveman:
I think this can either be closed and/or refactored into #961?

angela-tran:
@thekaveman Yep, I will go ahead and close this one.

PR #816, opened 2022-07-21
Fix: Django template formatting

### LINKED ISSUE #815: Fix Django template formatting/linting/pre-commit
Formatting for Django templates (HTML) is broken. Further, the bad formatting fails the `pre-commit` linter check, creating an endless loop where one cannot update/save/commit an HTML file:

html-formatting

## To Reproduce

1. Open a Django template (`core/help.html` in the example above)
2. Save the file to trigger a format (`prettier`?)
3. Try to commit the change
4. See pre-commit failure, file may be reformatted
5. Save file again, try to commit again, fail again

## Expected behavior

* [ ] We can edit/save/commit HTML files
* [ ] We get decent formatting on save on HTML files
* [ ] We have some linting checks locally and in `pre-commit`

## Additional context

Related to changes made in #761.

thekaveman:
Closes #815.

## What this PR does

* Swap `djhtml` for [`djlint`](https://github.com/Riverside-Healthcare/djlint) in `pre-commit` for HTML files
  * `prettier` handles CSS files
* [Integrate `djlint` with devcontainer](https://djlint.com/docs/integrations/#visual-studio-code) for linting and formatting
* Lint/format all our HTML files
  * Fix spacing (settings below)
  * Apply names to all `{% endblock %}` tags e.g. `{% endblock buttons %}`
  * Collapse inline tags e.g. `{% if %}...{% endif %}` inside an HTML `class` attribute
  * Expand block-level tags
  * Use `{% url %}` tag instead of hardcoding app URLs
  * Use `{% static %}` tag to load static files (images) instead of hardcoding paths

## `djlint` settings

To match our `prettier` config:

* Indent size 2
* Max line length 127
* Keep blank lines
* Use `.gitignore` (so e.g. local `coverage/` reports aren't processed)
* Ignore [rules](https://djlint.com/docs/linter/#rules):
  * `H006`: *`img` tag should have `height` and `width` attributes*
  * `H017`: *Tag should be self closing*
  * `H021`: *Inline styles should be avoided* (maybe fix with front-end cleanups)
  * `H025`: *Tag seems to be an orphan* (#775 may fix this)
  * `H031`: *Consider adding meta keywords* (we can look at doing this too)

### DIFF
--- a/benefits/enrollment/templates/enrollment/success.html
+++ b/benefits/enrollment/templates/enrollment/success.html
@@ -4,33 +4,29 @@
 {% load static %}
 
 {% block classes %}
-    {{ block.super |add:" enrollment-success"}}
-{% endblock %}
+  {{ block.super |add:" enrollment-success" }}
+{% endblock classes %}
 
 {% block paragraph_content %}
-    {% if "logged-out" in page.classes %}
-        <img class="success-image" src="/static/img/icon/happybus.svg">
-    {% else %}
-        <p>{% translate "enrollment.pages.success.p1" %}</p>
-        <p>{% translate "enrollment.pages.success.p2" %}</p>
-        <p>{% blocktranslate with help_link=help_link%}enrollment.pages.success.p3{{ help_link }}{% endblocktranslate %}</p>
-    {% endif %}
-{% endblock %}
+  {% if "logged-out" in page.classes %}
+    <img class="success-image" src="{% static "img/icon/happybus.svg" %}" alt="Bus icon with smiley face">
+  {% else %}
+    <p>{% translate "enrollment.pages.success.p1" %}</p>
+    <p>{% translate "enrollment.pages.success.p2" %}</p>
+    <p>{% blocktranslate with help_link=help_link%}enrollment.pages.success.p3{{ help_link }}{% endblocktranslate %}</p>
+  {% endif %}
+{% endblock paragraph_content %}
 
 {% block buttons %}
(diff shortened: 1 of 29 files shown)


benefits/enrollment/templates/enrollment/success.html
```diff
@@ -4,33 +4,29 @@
...
-    {% if "logged-out" in page.classes %}
-        <img class="success-image" src="/static/img/icon/happybus.svg">
-    {% else %}
-        <p>{% translate "enrollment.pages.success.p1" %}</p>
-        <p>{% translate "enrollment.pages.success.p2" %}</p>
-        <p>{% blocktranslate with help_link=help_link%}enrollment.pages.success.p3{{ help_link }}{% endblocktranslate %}</p>
-    {% endif %}
-{% endblock %}
+  {% if "logged-out" in page.classes %}
+    <img class="success-image" src="{% static "img/icon/happybus.svg" %}" alt="Bus icon with smiley face">
```
thekaveman, line 12: We should have a proper translated `alt` tag for this icon.

machikoyasuda (approved):

machikoyasuda:
Should make a ticket to fix this: `H006: img tag should have height and width attributes` and then re-add the rule. Adding width/height attribute to images is also something that the Google Lighthouse test tests for, and is good for web performance/page rendering speeds.

thekaveman:
@machikoyasuda can you please make tickets for the follow-ups related to the linter rules?

PR #846, opened 2022-08-02
Fix: sample client private key

### LINKED ISSUE #845: API-based eligibility verification with sample server fails to load PemData
When submitting the eligibility verification form, this error is shown:
```
ValueError at /eligibility/confirm

('Could not deserialize key data. The data may be in an incorrect format, it may be encrypted with an unsupported algorithm, or it may be an unsupported key type (e.g. EC curves with explicit parameters).', [_OpenSSLErrorWithText(code=503841036, lib=60, reason=524556, reason_text=b'error:1E08010C:DECODER routines::unsupported')])
```

## To Reproduce

1. Load sample data
2. Start the Benefits app
3. Choose the ABC transit agency
4. Choose "MST Courtesy Cardholder"
5. Enter values that match what the [sample eligibility-server](https://github.com/cal-itp/benefits/blob/dev/compose.yml#L39) expects (for example, [this record](https://github.com/cal-itp/eligibility-server/blob/main/data/server.json#L4)

## Expected behavior

The request to the server succeeds (and the user is eligible).

## Additional context

This appears to have broken after #814 was merged in. The problem is that the text in the [`Benefits client private key` PemData model](https://github.com/cal-itp/benefits/blob/601cf53f584a65ad6af30608fef2ecb6824b6e18/benefits/core/migrations/0002_sample_data.py#L37-L68) does not match the expected format.

From https://cryptography.io/en/latest/faq/#why-can-t-i-import-my-pem-file:
> If you are having trouble importing PEM files, make sure your file fits the following rules:
> ...
>  - all lines, except for the final one, must consist of exactly 64 characters.

In 9a0996f, the PEM data was [copied](https://github.com/cal-itp/benefits/commit/9a0996fad3d06017363b5f0e8b5ad079a464873f#diff-d5ffb1e04e2b938b352afe4ff3e975fbe3f78d57142922f2f4ea11d31b618c59R26-R52) over correctly, but somehow caa8524d inadvertently [deleted some characters](https://github.com/cal-itp/benefits/commit/caa8524d0200ee677b8894de698e6a9afc5c6ed8#diff-d5ffb1e04e2b938b352afe4ff3e975fbe3f78d57142922f2f4ea11d31b618c59L47-R52).

angela-tran:
Closes #845


thekaveman (approved):
🤷 😬

PR #468, opened 2022-04-21
Fix: configure test eligibility server

thekaveman:
The `server` Docker Compose service has been failing to start since https://github.com/cal-itp/eligibility-server/pull/58 introduced loading from a CSV file and expects an environment variable.

This PR provides the environment variable and a file to load with our sample data in the required format.

PR #847, opened 2022-08-02
Fix: configure and use latest test eligibility server

angela-tran:
This PR resurrects changes from #468 since the TODO has been resolved by cal-itp/eligibility-server#101.

I rebased the commit and made some fixes to account for all the work that's gone into `dev`.

### DIFF
--- a/.devcontainer/server/.env.server
+++ b/.devcontainer/server/.env.server
@@ -1,0 +1,1 @@
+IMPORT_FILE_PATH=.devcontainer/server/data.csv
--- a/.devcontainer/server/.env.server
+++ b/.devcontainer/server/.env.server
@@ -1,1 +1,1 @@
-IMPORT_FILE_PATH=.devcontainer/server/data.csv
+IMPORT_FILE_PATH=/.devcontainer/server/data.csv
--- a/.devcontainer/server/data.csv
+++ b/.devcontainer/server/data.csv
@@ -1,0 +1,4 @@
+A1234567;Garcia;['type1']
+B2345678;Hernandez;['type2']
+C3456789;Smith;[]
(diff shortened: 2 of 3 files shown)


thekaveman (approved):
Nice work! 🤖

thekaveman (commented):
Started up the devcontainer on this branch, loaded the sample data only.

I had assumed `A1234567` and `Garcia` would verify for `ABC` - but it does not.

In the Docker logs for `server`, I can see the call coming in and returning success:

```console
[04/Aug/2022 19:14:01] "GET /verify HTTP/1.1" 200 -
```

angela-tran:
@thekaveman Can you try with `B2345678` and `Hernandez` instead?

thekaveman:
@angela-tran Yes `Hernandez` worked

angela-tran:
> @angela-tran Yes `Hernandez` worked

Awesome, I think this means it's working as expected. [`Hernandez`](https://github.com/cal-itp/benefits/pull/847/files#diff-06e2c554bea6e7b973ea3bfcd5cf25df4c85c96778d60a576ce2ec3fa3376bbfR2) or [`Jones`](https://github.com/cal-itp/benefits/pull/847/files#diff-06e2c554bea6e7b973ea3bfcd5cf25df4c85c96778d60a576ce2ec3fa3376bbfR4) should be the only records that work with the sample "MST Courtesy Cardholder" (`type2`) discount option since the server checks the `type` when it [reads from its database](https://github.com/cal-itp/eligibility-server/blob/main/eligibility_server/verify.py#L96).

thekaveman:
Doh, of course - I see now that the sample `type2` is verified by our test server 👍

thekaveman (approved):

PR #849, opened 2022-08-04
Refactor: eligibility start view msgids

angela-tran:
A part of #776 

This PR cleans up how the `eligibility` `start` view function uses msgids. 

The view was hard-coding some oauth msgids when really those should be coming from the `EligibilityVerifier` model.    
The view also hard-codes some msgids for the links and bullet points, but because those are not configurable, we can at least add some logic to only show them if the verifier uses authentication as eligibility verification since that's what the links and bullet points are about.

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -95,8 +95,8 @@ def start(request):
             0,
             dict(
                 icon=viewmodels.Icon("idcardcheck", pgettext("image alt text", "core.icons.idcardcheck")),
-                heading=_("eligibility.pages.start.oauth.heading"),
-                details=_("eligibility.pages.start.oauth.details"),
+                heading=_(verifier.start_item_name),
+                details=_(verifier.start_item_description),
                 links=[
                     viewmodels.Button.link(
                         classes="btn-text btn-link",
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -83,18 +83,19 @@ def start(request):
     )
 
     if verifier.is_auth_required:
-        identity_item["links"] = [
-            viewmodels.Button.link(
-                classes="btn-text btn-link",
-                text=_("eligibility.pages.start.oauth.link_text"),
-                url=f"{reverse(ROUTE_HELP)}#login-gov",
-            ),
-        ]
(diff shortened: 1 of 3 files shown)


benefits/eligibility/views.py
```diff
@@ -75,58 +75,49 @@ def start(request):
...
-        ),
-    ]
-
-    if verifier.is_auth_required:
-        media.insert(
-            0,
-            dict(
-                icon=viewmodels.Icon("idcardcheck", pgettext("image alt text", "core.icons.idcardcheck")),
-                heading=_("eligibility.pages.start.oauth.heading"),
-                details=_("eligibility.pages.start.oauth.details"),
```
thekaveman, line 99: These `eligibility.pages.start.oauth.*` IDs need to go into the data migrations, right? They were deleted from the PO files.

thekaveman (dismissed):

benefits/eligibility/views.py
angela-tran: They should be deleted because they're duplicates of ones in the data migrations. (`eligibility.pages.start.oauth.heading` was a duplicate of `eligibility.pages.start.oauth.items[0].title` and `eligibility.pages.start.oauth.details` was a duplicate of `eligibility.pages.start.oauth.items[0].text`). 

The view was essentially ignoring the msgids configured on the `EligibilityVerifier` and using these hard-coded msgds.

angela-tran: Here are entries currently in `dev`:
https://github.com/cal-itp/benefits/blob/dev/benefits/locale/en/LC_MESSAGES/django.po#L570-L580
https://github.com/cal-itp/benefits/blob/dev/benefits/locale/en/LC_MESSAGES/django.po#L157-L167

thekaveman: Ah, my mistake! I misunderstood that these were dupes.

thekaveman (approved):
Ran it locally with sample data and production-like data, looks good 👍

benefits/eligibility/views.py
angela-tran: No worries! Our old approach made it easy for clutter to build up.

PR #852, opened 2022-08-05
Feat: upgrade to Python 3.10

### LINKED ISSUE #768: Upgrade to Python 3.10
We currently use Python 3.9 as defined in a number of places including [Docker images](https://github.com/cal-itp/benefits/blob/dev/Dockerfile#L1) and [GitHub Actions settings](https://github.com/cal-itp/benefits/blob/dev/.github/workflows/.python-version). Let's update to the more recent version 3.10.

## Acceptance Criteria

- [ ] Python 3.10 is used everywhere a version is mentioned

## Additional context

Our test suite should catch any obvious errors/differences between the versions, but ensure manual QA testing of the various flows works as expected.

thekaveman:
Closes #768 

This PR goes from a base image of `python:3.9-slim` to now using `python:3.10`.

Because we're not using the `-slim` variant, a number of tools like `curl`, `git`, and `ssh`, are already included - so we don't have to install them separately. The build process is a little simpler and faster as a result.


appcontainer/Dockerfile
```diff
@@ -25,7 +25,7 @@ RUN useradd --create-home --shell /bin/bash $USER && \
     # install server components
     apt-get update && \
     # install git for use when installing dependencies from VCS (https://pip.pypa.io/en/stable/topics/vcs-support/)
```
angela-tran, line 27: We can remove this comment now. It was just documenting that we need Git in the appcontainer so that pip can install `eligibility-api` from version control. (https://github.com/cal-itp/benefits/commit/55d0b177b8b58ea646099172c3f530b52b98f198)

appcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9-slim
+FROM python:3.10
```
angela-tran, line 1: Asking this just for my own knowledge/understanding: how did you determine that it's not necessary to use the `-slim` variant?

thekaveman: Both `-slim` and non`-slim` are built on the same base and both install Python and `pip` and are ready to go for Python code. 

The `-slim` variant just excludes some other tooling like `git`, `curl`, `ssh` etc. that takes up more time/space on the install. `-slim` is a subset of non`-slim`, so we don't lose anything by going up.

The original reason for `-slim` was more about keeping the image size as small as possible and the image contents as minimal as possible. As we've grown the project and had to install new tooling for e.g. supporting different dev platforms (`build-essential` in the base image) -- this has seemed to become less of a concern.

`-slim` felt like an unnecessary restriction that was forcing us to work around in different ways.

angela-tran: Oh I see, that makes a lot of sense! Thanks

angela-tran (approved):

thekaveman:
@angela-tran the UI tests were stuck for some reason. Force pushed.

angela-tran (approved):

PR #864, opened 2022-08-11
Fix: auth scope check when userinfo is present, but claim is not

### LINKED ISSUE #863: Improve checks in `oauth` `authorize` view
We ran into a case while deploying #862 where our `auth_provider` model is configured with a `scope`, but the response from the Auth server doesn't actually contain the scope (probably still being set up).

The `oauth` `authorize` function doesn't handle this case and raises an `AttributeError`
```
  File "/home/calitp/app/benefits/oauth/views.py", line 70, in authorize
    claim_flag = (userinfo.get(verifier_claim) if userinfo else "false").lower() == "true"
AttributeError: 'NoneType' object has no attribute 'lower'
```

Let's add a check and tests for this so that the logs for this are more helpful.

angela-tran:
Closes #863

### DIFF
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -70,7 +70,9 @@ def authorize(request):
         if userinfo:
             claim_value = userinfo.get(verifier_claim)
             # the claim comes back in userinfo like { "claim": "True" | "False" }
-            if claim_value is not None and claim_value.lower() == "true":
+            if claim_value is None:
+                logger.warning(f"userinfo did not contain: {verifier_claim}")
+            elif claim_value.lower() == "true":
                 # if userinfo contains our claim and the flag is true, store the *claim*
                 stored_claim = verifier_claim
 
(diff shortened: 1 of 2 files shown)


benefits/oauth/views.py
```diff
@@ -66,10 +66,12 @@ def authorize(request):
 
     if verifier_claim:
         userinfo = token.get("userinfo")
-        # the claim comes back in userinfo like { "claim": "True" | "False" }
-        claim_flag = (userinfo.get(verifier_claim) if userinfo else "false").lower() == "true"
-        # if userinfo contains our claim and the flag is true, store the *claim*
-        stored_claim = verifier_claim if claim_flag else None
+
+        if userinfo and userinfo.get(verifier_claim) is not None:
```
thekaveman, line 70: Maybe `userinfo.get(verifier_claim)` once and store in a variable along with `userinfo`.

Otherwise this is much cleaner 👍

thekaveman (approved):
Looks great, thanks for the fix and for the extra test coverage!

PR #873, opened 2022-08-23
Analytics - Send event when user successfully completes payment processor enrollment

### LINKED ISSUE #869: Create and send a real event for enrollment
Right now we send a page view event for the [`success`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/views.py#L138) view in `enrollment`. In Amplitude, we have a custom event for page view with path = `/enrollment/success` to indicate a successful enrollment.

We should send a dedicated event instead, for greater reliability and less chance of e.g. duplicate events or a missing page view.

## Acceptance Criteria

- [ ] We have a dedicated `completed enrollment` event, sent only when the user has successfully registered with the payment processor

machikoyasuda:
closes #869 

Sends `completed enrollment` event

### DIFF
--- a/benefits/enrollment/analytics.py
+++ b/benefits/enrollment/analytics.py
@@ -4,13 +4,15 @@ The enrollment application: analytics implementation.
 from benefits.core import analytics as core
 
 
-class CompletedEnrollment(core.Event):
-    """Analytics event representing a successfully completed payment processor enrollment."""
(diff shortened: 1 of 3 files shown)


thekaveman (approved):
This is great @machikoyasuda!

angela-tran (approved):
Very nice!

PR #876, opened 2022-08-25
chore(pipeline): wait for Terraform lock to be released

afeld:
Avoids race condition happening between pipeline runs.

### DIFF
--- a/terraform/azure-pipelines.yml
+++ b/terraform/azure-pipelines.yml
@@ -39,6 +39,9 @@ stages:
             inputs:
               provider: azurerm
               command: plan
+              # wait for lock to be released, in case being used by another pipeline run
+              # https://discuss.hashicorp.com/t/terraform-plan-wait-for-lock-to-be-released/6870/2
+              commandOptions: -lock-timeout=2m
               workingDirectory: "$(System.DefaultWorkingDirectory)/terraform"
               # service connection
               environmentServiceNameAzureRM: Production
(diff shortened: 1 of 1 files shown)


angela-tran (approved):

thekaveman (dismissed):
This looks great!

Should we have expected to see the pipeline succeed with this PR? Or not until the change makes it into the dev branch?

afeld:
Appears [the Terraform lock wasn't released from an operation on 8/22](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=35545&view=logs&j=bc150913-bfac-577d-ab97-4eb22da69064&s=c9bd906c-8296-5bac-0166-0b307c26c45c&t=0dc68013-2e89-56ea-c92f-d8381195fae6&l=37) for some reason. [Forced unlock](https://www.terraform.io/cli/commands/force-unlock) and re-ran the Job, which succeed.

In other words, this pull request seems to have "worked" since it waited two minutes for the lock to be released, but that never happened, so it exited. Why the lock wasn't released before is another question — let me know if you see the pipeline failing at other unexpected times and we can dig in.

angela-tran:
> Why the lock wasn't released before is another question — let me know if you see the pipeline failing at other unexpected times and we can dig in.

🤔 The Azure pipeline started failing again yesterday with the same error about not being able to acquire the lock.

For example, [#956](https://github.com/cal-itp/benefits/pull/956/checks?check_run_id=8564972026) failed ([Azure pipeline log](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=36454&view=logs&j=bc150913-bfac-577d-ab97-4eb22da69064&t=0dc68013-2e89-56ea-c92f-d8381195fae6))
image

Other PRs failing:
- https://github.com/cal-itp/benefits/pull/959/checks?check_run_id=8562566924
- https://github.com/cal-itp/benefits/pull/951/checks?check_run_id=8562291395

PR #881, opened 2022-08-26
Feat: agency public keys / remote keys

thekaveman:
This work is corollary to the work being done in cal-itp/eligibility-server#136.

## Background

The Eligibility Verification API uses asymmetric public-key cryptography to secure the verification messages. Read more about the [Eligibility Verification API specification](https://docs.calitp.org/eligibility-api/specification/).

For any given Eligibility Client and Server relationship, there must exist a total of 4 keys:

* The Client's private key (`client.key`) and corresponding public key (`client.pub`)
* The Server's private key (`server.key`) and corresponding public key (`server.pub`)

In order to compose a request to a Server, the Client signs the message with its own `client.key` and uses `server.pub` to encrypt the message.

In order to evaluate a request from a Client, the Server decrypts the message with its own `server.key` and verifies the signature with `client.pub`.

Similarly, to respond to a request, the Server signs the message with its `server.key` and encrypts the message with `client.pub`. To evaluate the Server's response, the Client decrypts the message with `client.key` and verifies the signature with `server.pub`.

TLDR; the Client needs `client.key` and `server.pub`, while the Server needs `server.key` and `client.pub`.

## Progress on simplifying key management

In the Eligibility Server, we are:

* Adding configuration to customize where the server loads its keys from
* Allowing the server to load keys from remote URLs (simple anonymous GET access)
* Publishing the server's public key to a public endpoint

In this PR, we are:

* Publishing each `TransitAgency`'s public key (the pair to its private key) to a public endpoint in Benefits
* Allowing Benefits to load keys from remote URLs (again, simple anonymous GET access)

With both of these PRs merged, we can align storage of each of the public keys to the same place that owns the corresponding private key.

## ⚠️ This PR updates the database schema ⚠️ 

Both local (beyond the samples) and deployed data migrations need to be updated with new fields.


benefits/core/migrations/0001_initial.py
```diff
@@ -93,8 +93,10 @@ class Migration(migrations.Migration):
             name="PemData",
             fields=[
                 ("id", models.AutoField(primary_key=True, serialize=False)),
-                ("text", models.TextField()),
                 ("label", models.TextField()),
+                ("text", models.TextField(null=True)),
+                ("remote_url", models.TextField(null=True)),
+                ("remote_loaded", models.BooleanField(default=False)),
```
thekaveman, line 99: Made the `text` property optional, and added a couple properties to load remote data

benefits/core/migrations/0001_initial.py
```diff
@@ -120,6 +122,10 @@ class Migration(migrations.Migration):
                     "private_key",
                     models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name="+", to="core.pemdata"),
                 ),
+                (
+                    "public_key",
+                    models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name="+", to="core.pemdata"),
+                ),
```
thekaveman, line 128: Another `PemData` field for the public key (corresponding to the private key)

benefits/core/migrations/0002_sample_data.py
```diff
@@ -74,6 +92,7 @@ def load_sample_data(app, *args, **kwargs):
 -----END CERTIFICATE-----
 """,
         label="Dummy certificate",
+        remote_loaded=True,
```
thekaveman, line 95: `remote_loaded=True` is a minor shortcut for these `PemData` that already contain their `text`

benefits/core/models.py
```diff
@@ -14,14 +16,29 @@ class PemData(models.Model):
...
+    @property
+    def data(self):
+        if self.remote_loaded:
+            return self.text
+        elif self.remote_url:
+            self.text = requests.get(self.remote_url).text
+
+        self.remote_loaded = True
+        self.save()
+        return self.text
```
thekaveman, line 40: New `PemData.data` property:

* Returns `text` if `remote_loaded` is `True`
* Sets `text` to the content from `remote_url` if there is one
* Sets `remote_loaded` to `True`
* Returns `text`

So basically the content is always (eventually) in `text`, with maybe a one-time load from `remote_url`.

benefits/core/models.py
```diff
@@ -105,7 +122,7 @@ def __str__(self):
     @property
     def public_key_data(self):
         """This Verifier's public key as a string."""
-        return self.public_key.text
+        return self.public_key.data
```
thekaveman, line 125: Use the new `data` property instead of `text` directly

benefits/core/models.py
```diff
@@ -195,13 +214,23 @@ def types_to_verify(self, eligibility_verifier):
...
     @property
     def index_url(self):
-        """Url to the TransitAgency's landing page."""
+        """Public-facing URL to the TransitAgency's landing page."""
         return reverse("core:agency_index", args=[self.slug])
 
+    @property
+    def public_key_url(self):
+        """Public-facing URL to the TransitAgency's public key."""
+        return reverse("core:agency_public_key", args=[self.slug])
```
thekaveman, line 223: Route back to this agency's public key

benefits/core/models.py
```diff
@@ -195,13 +214,23 @@ def types_to_verify(self, eligibility_verifier):
...
     @property
     def private_key_data(self):
         """This Agency's private key as a string."""
-        return self.private_key.text
+        return self.private_key.data
+
+    @property
+    def public_key_data(self):
+        """This Agency's public key as a string."""
+        return self.public_key.data
```
thekaveman, line 233: Use the `data` properties instead of `text`

benefits/core/urls.py
```diff
@@ -47,4 +47,5 @@ def to_url(self, agency):
     path("", views.index, name="index"),
     path("help", views.help, name="help"),
     path("<agency:agency>", views.agency_index, name="agency_index"),
+    path("<agency:agency>/publickey", views.agency_public_key, name="agency_public_key"),
```
thekaveman, line 50: New endpoint uses the `agency` URL pattern filter

benefits/core/views.py
```diff
@@ -71,6 +71,12 @@ def agency_index(request, agency):
     return TemplateResponse(request, TEMPLATE_AGENCY, context_dict)
 
 
+@pageview_decorator
+def agency_public_key(request, agency):
+    """View handler returns an agency's public key as plain text."""
+    return HttpResponse(agency.public_key_data, content_type="text/plain")
```
thekaveman, line 77: Simple view just returns the plain text PEM data

benefits/enrollment/api.py
```diff
@@ -143,13 +143,13 @@ def _cert_request(self, request_func):
...
+            cert.write(self.payment_processor.client_cert.data)
             cert.seek(0)
 
-            key.write(self.payment_processor.client_cert_private_key.text)
+            key.write(self.payment_processor.client_cert_private_key.data)
             key.seek(0)
 
-            ca.write(self.payment_processor.client_cert_root_ca.text)
+            ca.write(self.payment_processor.client_cert_root_ca.data)
             ca.seek(0)
```
thekaveman, line 153: Use the new `data` property rather than `text`

benefits/core/models.py
```diff
@@ -14,14 +16,29 @@ class PemData(models.Model):
...
+    remote_url = models.TextField(null=True)
+    # Flag if this PemData has been loaded from remote
+    remote_loaded = models.BooleanField(default=False)
 
     def __str__(self):
         return self.label
 
+    @property
+    def data(self):
+        if self.remote_loaded:
```
angela-tran, line 33: I could also see this being `if self.text:` and not having to maintain the `remote_loaded` field.

On the other hand, maybe it's better to have an explicit field? What do you think?

thekaveman: True, that is much simpler!

angela-tran (dismissed):

angela-tran (approved):
Nice!

thekaveman:
Marking this as `ON HOLD` until we have a chance to make the data migrations changes in `dev` at least.

Issue #877, opened 2022-08-25
Write Use Case page for Seniors / Login.gov


thekaveman:
## Acceptance Criteria

- [ ] A new Use Case page exists describing Seniors / Login.gov with similar formatting/content as the [Student use case](https://docs.calitp.org/benefits/use-cases/students/)
- [ ] Includes a Process Flow diagram (could be lifted/moved from https://docs.calitp.org/benefits/#enrollment-process)

## Additional context

We introduced the "Use Cases" concept in our docs based on work in #800 and #810.

See also #878

## What is the definition of done?

The page is published on https://docs.calitp.org/benefits/use-cases

thekaveman:
Closed by #889

PR #889, opened 2022-08-31
Docs: Use cases for Seniors

machikoyasuda:
#877 

This PR does two things:
1. Add a Seniors page for Use cases
2. Enables integrated navigation for better readability https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#__tabbed_6_1

### DIFF
--- a/docs/use-cases/README.md
+++ b/docs/use-cases/README.md
@@ -2,7 +2,7 @@
 title: Introduction
 ---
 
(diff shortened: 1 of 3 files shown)


thekaveman (approved):
This looks great, thank you for adding the Use case as well!

Love the new menu setting, so much cleaner ❤️

PR #897, opened 2022-09-06
feat(analytics): return Group ID with Returned Success

### LINKED ISSUE #890: Add group information to enrollment success event
#873 brought us an explicit event for enrollment success (and failure, and retry).

This ticket seeks to add an event property `payment_group` with the GUID of the Payment Processor group the user was enrolled into.

This will help us track down registration issues, verify users are getting into the correct groups, etc.

## Acceptance Criteria

- [ ] Every time the `enrollment returned` with `status = success` fires, an event property `payment_group` is sent with the event

## Additional context

The Group identifier is stored in the [`EligibilityType` model](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L45)

machikoyasuda:
closes #890 

Passes `eligibility.group_id` as `payment_group` into Amplitude for successful enrollment event

### DIFF
--- a/tests/pytest/enrollment/test_views.py
+++ b/tests/pytest/enrollment/test_views.py
@@ -113,7 +113,12 @@ def test_index_eligible_post_valid_form_failure(mocker, client, card_tokenize_fo
 
 @pytest.mark.django_db
 @pytest.mark.usefixtures("mocked_session_agency", "mocked_session_verifier", "mocked_session_eligibility")
-def test_index_eligible_post_valid_form_success(mocker, client, card_tokenize_form_data, mocked_analytics_module):
+def test_index_eligible_post_valid_form_success(
+    mocker,
+    client,
+    card_tokenize_form_data,
+    mocked_analytics_module,
+):
     mock_response = mocker.Mock()
     mock_response.success = True
     mocker.patch("benefits.enrollment.views.api.Client.enroll", return_value=mock_response)
@@ -121,10 +126,9 @@ def test_index_eligible_post_valid_form_success(mocker, client, card_tokenize_fo
     path = reverse(ROUTE_INDEX)
     response = client.post(path, card_tokenize_form_data)
 
-    mocked_analytics_module.returned_success.assert_called_once()
+    mocked_analytics_module.returned_success
     assert response.status_code == 200
     assert response.template_name == TEMPLATE_SUCCESS
-    mocked_analytics_module.returned_success.assert_called_once()
 
 
 @pytest.mark.django_db
--- a/tests/pytest/enrollment/test_views.py
+++ b/tests/pytest/enrollment/test_views.py
@@ -126,7 +126,7 @@ def test_index_eligible_post_valid_form_success(
     path = reverse(ROUTE_INDEX)
     response = client.post(path, card_tokenize_form_data)
 
-    mocked_analytics_module.returned_success
+    mocked_analytics_module.returned_success.assert_called_once()
     assert response.status_code == 200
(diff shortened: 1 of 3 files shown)


tests/pytest/enrollment/test_views.py
```diff
@@ -113,15 +113,20 @@ def test_index_eligible_post_valid_form_failure(mocker, client, card_tokenize_fo
...
     mocker.patch("benefits.enrollment.views.api.Client.enroll", return_value=mock_response)
 
     path = reverse(ROUTE_INDEX)
     response = client.post(path, card_tokenize_form_data)
 
-    mocked_analytics_module.returned_success.assert_called_once()
+    mocked_analytics_module.returned_success
     assert response.status_code == 200
     assert response.template_name == TEMPLATE_SUCCESS
     mocked_analytics_module.returned_success.assert_called_once()
```
machikoyasuda, line 132: Why is line 129/132 repeated? I wasn't sure.

I was _trying_ to change the line to something more specific like:

```
mocked_analytics_module.returned_success.assert_called_once_with(app_request_enrollment_success)
```

But I kept getting an error like this:

```python
E       Args:
E       assert (<WSGIRequest...ess'>, '1234') == (<WSGIRequest...nt/success'>,)
E         At index 0 diff: <WSGIRequest: POST '/enrollment/success'> != <WSGIRequest: POST '/enrollment/success'>
E         Left contains one more item: '1234'
E         Full diff:
E         - (<WSGIRequest: POST '/enrollment/success'>,)
E         + (<WSGIRequest: POST '/enrollment/success'>, '1234')
```

and I couldn't figure out how to mock a request that `==` exactly like `(<WSGIRequest: POST '/enrollment/success'>, '1234')`. I would want to make sure that the `1234` group ID is correctly getting passed.

machikoyasuda: cc @angela-tran @thekaveman

thekaveman: Hmm not sure why that is repeated either, definitely shouldn't be!

Regarding your question, here's an example of checking that a specific call is in the list of calls for the spied function: https://github.com/cal-itp/eligibility-server/blob/main/tests/test_keypair.py#L52

The `mocker` fixture has a helper to create a [Call object](https://docs.python.org/3/library/unittest.mock.html#call), and you can look if that Call is inside the list of all Calls on the spy:

```python
assert mocker.call("arg1", "arg2", "etc.") in spy.call_args_list
```

tests/pytest/enrollment/test_views.py
```diff
@@ -113,18 +113,20 @@ def test_index_eligible_post_valid_form_failure(mocker, client, card_tokenize_fo
...
     mocker.patch("benefits.enrollment.views.api.Client.enroll", return_value=mock_response)
 
     path = reverse(ROUTE_INDEX)
     response = client.post(path, card_tokenize_form_data)
 
-    mocked_analytics_module.returned_success.assert_called_once()
     assert response.status_code == 200
     assert response.template_name == TEMPLATE_SUCCESS
-    mocked_analytics_module.returned_success.assert_called_once()
+    mocked_analytics_module.returned_success.assert_called_once
```
thekaveman, line 128: You still want to call this like a function though: `.assert_called_once()`

thekaveman (changes requested):
Ah, I didn't notice this before:

The `ReturnedEnrollmentEvent` class needs to be updated too, to accept the new parameter (maybe with a default of `None`) and then make a call to `self.update_event_properties()` if there is a value.

thekaveman (approved):
Nice work! 🚀

PR #921, opened 2022-09-15
Refactor: use common base image

### LINKED ISSUE #879: Use the common base Docker image
This is a corollary to work done in cal-itp/eligibility-server#123.

We have a new repository [`cal-itp/docker-python-web`](https://github.com/cal-itp/docker-python-web) that publishes a Docker image to GitHub Container Registry at [`ghcr.io/cal-itp/docker-python-web`](https://github.com/cal-itp/docker-python-web/pkgs/container/docker-python-web).

This image contains a baseline `nginx` and `gunicorn` setup and configuration. We can replace the Benefits app container's base image with this image to simplify the repository and build process for Benefits.

## Acceptance Criteria

- [ ] the Benefits app container builds `FROM ghcr.io/cal-itp/docker-python-web`
- [ ] duplicate `nginx` and `gunicorn` setup in this repository is removed
- [ ] the app runs locally as a container `docker compose up client`
- [ ] the devcontainer starts up
- [ ] the app runs in DEBUG mode in the devcontainer with `F5`

## Additional context

See the [`docker-python-web` Docs](https://docs.calitp.org/docker-python-web/) for more information.

angela-tran:
Closes #879


thekaveman (commented):
Just one early comment - I think it actually makes more sense how you initially had it with multiple user switches before [`014c798`](https://github.com/cal-itp/benefits/pull/921/commits/014c798fb7a63afa8ed5957f3e95c66762adfd17)

Remember docker caches layers (each command) - usually we want commands that are less likely to change up near the top, and commands that are more likely near the bottom.

Moving the `gettext` install to the bottom means we likely can't take advantage of caching that layer.

appcontainer/Dockerfile
```diff
@@ -1,51 +1,20 @@
...
-    mkdir -p /var/log/nginx && \
-    chown -R $USER /var/log/nginx && \
-    touch /var/log/nginx/error.log && \
-    chown $USER /var/log/nginx/error.log && \
-    touch /var/run/nginx.pid && \
-    chown -R $USER /var/run/nginx.pid && \
-    # setup directories and permissions for Django and gunicorn
-    mkdir -p /home/$USER/app/config && \
-    mkdir -p /home/$USER/app/run && \
-    mkdir -p /home/$USER/app/static && \
```
thekaveman, line 23: Looks like we lost this creation of the `static` directory - it's not in the base. Not sure if it is entirely needed?

angela-tran: From my testing where I built from scratch, everything seems to still work even without creating `static`. 🤷

thekaveman: Yeah I guess even the output from the port issue we were discussing earlier showed that static files were being copied.

Let's just make sure they can be _read_ too when running the app container (i.e. not debug/devcontainer)

angela-tran: I ran `docker compose up client` and when I launched the app in my browser, the CSS and images all loaded ok. Are there any other static files I should check for?

thekaveman: Sounds good!

thekaveman (changes requested):
Can you please update the `bin/build.sh` helper with the `--pull` flag on `client` to ensure we always pull the latest base: https://github.com/cal-itp/benefits/blob/dev/bin/build.sh#L4

thekaveman:
Will fully review and test this tomorrow, but so far it looks really good!

thekaveman (dismissed):
I tested both in the devcontainer / debug mode, and in the app container / runtime mode - everything looks good!

I was able to get through eligibility verification with our test server (added minor update to our sample data following cal-itp/eligibility-server#148 and cal-itp/eligibility-server#151), and enrollment in the QA environment.

The build is so much faster 😍 🔥

thekaveman (approved):

angela-tran:
That's awesome that we can just point to eligibility-server's file!

PR #922, opened 2022-09-19
Fix: error page translations

### LINKED ISSUE #913: Bug: Spanish translations on all error pages are not showing up.
## To Reproduce

1. Go to benefits.calitp.org/
2. Click on Spanish
3. Go to benefits.calitp.org/asfakljsfasdf

image

Happening for 404 and 500

## Expected behavior

All the text should be in Spanish, not just the Footer and the Button.

## Screenshots

image

Both 

All

## Additional context

Fix translations on error pages (default arguments set once, need to use None and check for None instead)

angela-tran:
Closes #913 

Added unit tests.
Fixed the bug by using `gettext_lazy` so that it's ok to set default arguments.

### DIFF
--- a/benefits/core/viewmodels.py
+++ b/benefits/core/viewmodels.py
@@ -1,7 +1,7 @@
 """
 The core application: view model definitions for the root of the webapp.
 """
-from django.utils.translation import pgettext, gettext as _
+from django.utils.translation import pgettext, gettext_lazy as _
 from django.urls import reverse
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
This is great. Thanks for the simple fix and the tests!

BTW I had done some earlier work and finally created #925 as part of my review, it will be useful as we work on other error page tickets too.

PR #891, opened 2022-09-01
Feat: send eligibility type with eligibility events

thekaveman:
Closes #874 

* On `eligibility:start`, send the requested types
* On `eligibility:success`, send the verified types
* On `eligibility:error` or `eligibility:fail`, send the requested types

PR #924, opened 2022-09-19
Feat: track verifier in all analytic events

### LINKED ISSUE #875: Add verifier property to Events
Soon users will have multiple pathways to verify eligibility, and it would be useful to know how many users are taking which pathway(s) in our analytics. We can get ahead of this now by adding an event property with the user's selected verifier (if any).

Let's add this to the base Event definition so that we can later filter any event type by this property, similar to the Transit Agency information attached to each event.

Linking this Amplitude reference again: https://help.amplitude.com/hc/en-us/articles/115002380567-User-properties-and-event-properties

## Acceptance Criteria

- [ ] The base Event has an `eligibility_verifier` event property that gets its value from the user's `session.verifier()`
- [ ] The Amplitude Tracking Plan is updated to accept this event property
- [ ] The Amplitude --> Warehouse --> Metabase pipeline is updated to ingest this property

## Additional context

This task was broken out from #342 and is focused on adding the verifier information to _existing_ events. #342 is now focused on introducing _new_ events for some forthcoming UI.

angela-tran:
Closes #875

I noticed that `transit_agency` is also being sent as a user property, so I did the same with `eligibility_verifier`.

### DIFF
--- a/benefits/core/analytics.py
+++ b/benefits/core/analytics.py
@@ -43,8 +43,10 @@ class Event:
 
         agency = session.agency(request)
         agency_name = agency.long_name if agency else None
+        verifier = session.verifier(request)
+        verifier_name = verifier.name if verifier else None
 
-        self.update_event_properties(path=request.path, transit_agency=agency_name)
+        self.update_event_properties(path=request.path, transit_agency=agency_name, eligibility_verifier=verifier_name)
 
         uagent = request.headers.get("user-agent")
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Hmm yeah I'm still not sure if/how we'd use this property on the users - we don't have much info about them anyway, almost all of our charts are based on the events.

I think it doesn't hurt to add this one in both places. This PR also makes me think that we should be adding session eligibility to the base event too, rather than just where I added it in #891.

PR #925, opened 2022-09-19
Devcontainer: debug in non-debug mode

thekaveman:
Yes, really!

This PR adds a new VS Code *Run and Debug* configuration that:

* Launches the app with `DJANGO_DEBUG=False`
* Supports hitting breakpoints in the code

This is accomplished with a small override on the static files setting, which is only used for this debug config.

Useful for seeing and debugging error pages etc. more like how users see them.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,6 +14,18 @@
       "env": {
         "DJANGO_DEBUG": "true"
       }
+    },
+    {
+      "name": "Django: Benefits Client, Debug=False",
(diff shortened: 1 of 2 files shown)


angela-tran (dismissed):
Nice!

thekaveman:
Sorry, I cut out the boolean logic BS and just went with a straight settings override for simplicity.

angela-tran (approved):
That is more straightforward 👍

Issue #949, opened 2022-09-22
Agency Index: Update just the copy


machikoyasuda:
## Acceptance Criteria

- Add/remove change copy as necessary
- Do not change CSS

Issue #936, opened 2022-09-22
Make buttons use title-case


angela-tran:
## Acceptance Criteria
- [ ] All buttons are using title case

## Additional context
This is according to the design in Figma

PR #950, opened 2022-09-22
Feat/agency index  copy

machikoyasuda:


benefits/core/views.py
```diff
@@ -56,19 +56,15 @@ def agency_index(request, agency):
         return redirect(reverse(ROUTE_ELIGIBILITY))
 
     button = viewmodels.Button.primary(text=_("core.pages.index.continue"), url=reverse(ROUTE_ELIGIBILITY))
-    button.label = _("core.pages.agency_index.button.label")
 
     page = viewmodels.Page(
         title=_("core.pages.agency_index.title"),
-        content_title=_("core.pages.agency_index.content_title"),
+        content_title=_("core.pages.agency_index.mst_cc.content_title"),
```
thekaveman, line 62: Just so I'm following - this is where we'll need to dynamically pick the title in the future, right?

machikoyasuda: Yeah

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -513,10 +498,9 @@ msgid "core.pages.agency_index.title"
 msgstr "Introduction"
 
 #: benefits/core/views.py:63
-msgid "core.pages.agency_index.content_title"
+msgid "core.pages.agency_index.mst_cc.content_title"
```
angela-tran, line 501: I think this rename needs to be done in the Spanish .po file too.

Btw, you can run `./bin/makemessages.sh` to regenerate the .po files. It'll update line number references, add in missing entries, and remove any unused entries! 

... typing that out just made me realize that we don't have documentation on this! I created #954 to fix that.

angela-tran (changes requested):

benefits/locale/es/LC_MESSAGES/django.po
```diff
@@ -511,21 +496,19 @@ msgstr ""
...
-#: benefits/core/views.py:62
+#: benefits/core/views.py:61
 msgid "core.pages.agency_index.title"
 msgstr "Introducción"
 
-#: benefits/core/views.py:63
-msgid "core.pages.agency_index.content_title"
+#: benefits/core/views.py:62
+#, fuzzy
+#| msgid "core.pages.agency_index.content_title"
```
machikoyasuda, line 505: @angela-tran I ran the make messages command and it added this `fuzzy` line here. What does that mean?

angela-tran: Also, it only does fuzzy-guessing the first time that it generates the entry. Once you provide a `msgstr`, it won't try to guess the value again.

machikoyasuda:
Ready for re-review @angela-tran - Confirmed that the Spanish copy is working correctly

angela-tran (approved):
This looks good to me 💯

thekaveman:
> closes #949
>>#949:  _do not change CSS_
>
>
> closes #936

🤣

Issue #961, opened 2022-09-27
Build in support for new spacing and alignment system


angela-tran:
## Acceptance Criteria

- [ ] Devs can add elements to pages that adhere to new spacing and alignment guidelines from design

## Additional context
Look into using [Bootstrap grid system](https://getbootstrap.com/docs/5.1/layout/grid/) to achieve this. This may have us go back towards using `page.html` as a flexible and generic template for all pages. 

Alternatively, we would move towards having a specific template for each page that at least uses some sort of common system to follow the new spacing/alignment system.

angela-tran:
Adding notes on what @thekaveman , @machikoyasuda , and I discussed last week:

Instead of `page.html` having hard-coded blocks that conditionally show up, we will have a defined Bootstrap grid on either `page.html` or `base.html` which extending templates can fill in. If we define it on `base.html`, we can probably delete `page.html`.

PR #932, opened 2022-09-21
One column everything

machikoyasuda:
closes #933 

## What this PR does
- Make all views with the Page view model use the one-column layout by setting `noimage` to `True`.
- Deletes CSS style code for `eligibility/start` page design (shortened image on top)

## Deleted code
- `main-primary` class and css
- `one-column-image` class and css
- `with-image` (and `col-lg-6.image` class and css and jquery
- `noimage` from Page class and container template

## To test
- Test in Debug=False mode: Go through the whole flow in English/Spanish on Desktop and Mobile
- Ensure no regressions on /help
- Ensure no regressions on error pages: 404, 500
- http://localhost:11369/
- http://localhost:11369/eligibility/start
- http://localhost:11369/eligibility/
- http://localhost:11369/eligibility/confirm
- http://localhost:11369/enrollment/
- http://localhost:11369/enrollment/success

PR #964, opened 2022-09-27
Fix: sticky footer

angela-tran:
Related to #961 

This PR achieves a sticky footer, following an approach detailed by [CSS-tricks](https://css-tricks.com/couple-takes-sticky-footer/#aa-there-is-flexbox). This method is simple and effective at ensuring the footer actually sticks to the bottom.

(We also use it on Compiler's website.) 

### Additional context

This was needed because it appears that [a recent change](https://github.com/cal-itp/benefits/pull/932/files#diff-744df555aee38173a0b0868baa7befa720e6a28f80747523e74e724c126cd381L807-L814) caused a regression: a small amount of whitespace appeared under the footer. This only occurred for Desktop screen size.

| Before | After |
| --- | --- |
| before | after |

The inline comment that was removed and the commit message from https://github.com/cal-itp/benefits/pull/4/commits/f446bf6d8e1acffd78dc2c6b08b237fd467cbc52 are supporting evidence that the sticky footer implementation was relying on  `flex-shrink: unset;` to fix unexpected whitespace.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -1,3 +1,28 @@
+/* making the sticky footer */
+
+* {
+  box-sizing: border-box;
+}
+
+html,
+body {
+  height: 100%;
+  margin: 0;
+}
+
+body {
+  display: flex;
+  flex-direction: column;
+}
+
+main {
+  flex: 1 0 auto;
+}
+
+footer {
+  flex-shrink: 0;
+}
+
 /* HTML element overrides */
 
 body {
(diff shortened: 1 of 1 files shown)


benefits/static/css/styles.css
```diff
@@ -1,3 +1,28 @@
+/* making the sticky footer */
```
thekaveman, line 1: How much of the CSS below is actually for the sticky footer, vs. just part of our design/layout? I worry about a comment like this indicating all of it related? Maybe it is.

angela-tran: All of it up to line 25 is necessary for the sticky footer. Any ideas on how to make that more obvious? 🤔

thekaveman: Hmm maybe I am expecting it to come after the more generic HTML element overrides - any reason it moved up?

thekaveman: Maybe it's more about how we organize our CSS too (also, I thought we weren't worry about CSS until all the copy edits??)

Issue #376, opened 2022-03-30
Update Enrollment success page


thekaveman:
## Background

This ticket adds additional contextual information to the final success screen.

![image](https://user-images.githubusercontent.com/1783439/160731515-124338a6-f987-4dd7-a1e2-43236170705a.png)

## Tasks

* [x] Update existing entries for `enrollment.pages.success.p*` to match the new style, in English and Spanish files (these were missed in #328)
* [x] Update the existing language as needed in English and Spanish
* [x] Create text entries to split up the final paragraph at the "help page" link in English and Spanish
* [x] Create text entries for the final sub-heading "Please close your browser..." in English and Spanish
* [x] Update [`enrollment/success.html`](https://github.com/cal-itp/benefits/blob/dev/benefits/enrollment/templates/enrollment/success.html) to render this new content

PR #967, opened 2022-09-28
Refactor: enrollment connect card copy

### LINKED ISSUE #945: Update Enrollment Connect Card page with new copy
## Acceptance Criteria
- [ ] Enrollment Connect Card page has the latest copy

## Additional context
Keep in mind per-verifier dynamic content.
Don't worry about alignment for text or buttons for this ticket.

angela-tran:
Closes #945 

Made a minimal and temporary change to see Enrollment Connect Card copy. The change was to add a `media-list` include to `page.html`; of course the way it's actually done may change in #961.

How it looks as of this PR:
image


angela-tran:
D'oh, I just realized I need to extract out dynamic content as model properties.

benefits/enrollment/views.py
```diff
@@ -74,11 +74,24 @@ def index(request):
...
+
+        media = [confirmed_benefit_item, link_card_item]
+
         page = viewmodels.Page(
             title=_("enrollment.pages.index.title"),
             headline=_("enrollment.pages.index.headline"),
-            icon=viewmodels.Icon("idcardcheck", pgettext("image alt text", "core.icons.idcardcheck")),
-            paragraphs=[_("enrollment.pages.index.p[0]"), _("enrollment.pages.index.p[1]"), _("enrollment.pages.index.p[2]")],
+            paragraphs=[_("enrollment.pages.index.p[0]")],
             classes="text-lg-center no-image-mobile",
```
machikoyasuda, line 95: If the way everything is centered is bothering you, you can remove `text-lg-center` (and `no-image-mobile` cos it's not doing anything anmore) from here.

angela-tran:
Question regarding message key-naming:

Right now, our media items are represented as raw dicts with `icon`, `heading`, and `details` keys (see `eligibility/views` [`start`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L113-L117) for example). However, the message keys are not consistent with the dict keys. For example, the msgid for `heading` ends with `.title` and for `details` it ends with `.text`.

Recently, in [a discussion](https://compilerla.slack.com/archives/C02GQ7AJXQE/p1664400701663169?thread_ts=1664397539.650009&cid=C02GQ7AJXQE) about new copy for media items on the Enrollment success page, we talked about having message keys like:
 - `enrollment_expiry_header`
 - `enrollment_expiry_message`

So the question is: which one do we like the best? We should pick one and then rename the fields and the msgids to be the same.
 - `heading` + `details`?
 - `title` + `text`? (probably not this once since we use `title` to mean the value for `<title>` elements)
 - `header` + `message`?

benefits/enrollment/views.py
```diff
@@ -74,12 +76,25 @@ def index(request):
         tokenize_retry_form = forms.CardTokenizeFailForm(ROUTE_RETRY)
         tokenize_success_form = forms.CardTokenizeSuccessForm(auto_id=True, label_suffix="")
 
+        confirmed_benefit_item = dict(
+            icon=viewmodels.Icon("happybus", pgettext("image alt text", "core.icons.happybus")),
+            heading=_(verifier.enrollment_confirmed_header),
+            details=_(verifier.enrollment_confirmed_message),
```
angela-tran, line 82: See my [question](https://github.com/cal-itp/benefits/pull/967#issuecomment-1262963269) on the inconsistency between these dictionary keys vs. the message keys.

machikoyasuda:
> Question regarding message key-naming:
> 
> Right now, our media items are represented as raw dicts with `icon`, `heading`, and `details` keys (see `eligibility/views` [`start`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L113-L117) for example). However, the message keys are not consistent with the dict keys. For example, the msgid for `heading` ends with `.title` and for `details` it ends with `.text`.
> 
> Recently, in [a discussion](https://compilerla.slack.com/archives/C02GQ7AJXQE/p1664400701663169?thread_ts=1664397539.650009&cid=C02GQ7AJXQE) about new copy for media items on the Enrollment success page, we talked about having message keys like:
> 
>     * `enrollment_expiry_header`
> 
>     * `enrollment_expiry_message`
> 
> 
> So the question is: which one do we like the best? We should pick one and then rename the fields and the msgids to be the same.
> 
>     * `heading` + `details`?
> 
>     * `title` + `text`? (probably not this one since we use `title` to mean the value for `<title>` elements)
> 
>     * `header` + `message`?

Great question! I looked at Bootstrap's Card component https://getbootstrap.com/docs/5.1/components/card/ and specifically https://getbootstrap.com/docs/5.1/components/card/#kitchen-sink. It's very similar to this component we have, except vertical instead of horizontal. They use the naming: `card-img` (in our case, `card-icon`), `card-body`, which contains `card-title` and `card-text`. We could replace the term `card` with `media-item` and also align the copy key naming and the CSS class names along with it. I think `title` and `text` are good as long as we add `media-item` before it.

image

benefits/core/migrations/0001_initial.py
```diff
@@ -65,6 +65,8 @@ class Migration(migrations.Migration):
                 ("unverified_title", models.TextField()),
                 ("unverified_headline", models.TextField()),
                 ("unverified_blurb", models.TextField()),
+                ("enrollment_confirmed_header", models.TextField()),
+                ("enrollment_confirmed_message", models.TextField()),
```
thekaveman, line 69: This seems like the wrong key/property name -- this isn't the `enrollment_confirmed` page, it is the `enrollment_start` or `enrollment_connect` or something.

_Eligibility_ was confirmed, but enrollment has just started.

thekaveman (changes requested):

benefits/core/migrations/0001_initial.py
angela-tran: I see your point. I had it as `confirmed_benefit` originally, but anyways, I do think `enrollment_start` is clearer.

angela-tran:
@machikoyasuda That is a helpful reference point. I'm wondering if having a prefix would make sense for the Python side of this. Would the dict end up looking something like:
```python
 bank_card_item = dict(
        media_item_icon=viewmodels.Icon("bankcardcheck", pgettext("image alt text", "core.icons.bankcardcheck")),
        media_item_title=_("eligibility.pages.start.bankcard.media_item_title"),
        media_item_text=_("eligibility.pages.start.bankcard.media_item_text"),
    )
```
? Or did you have something different in mind?

I think @thekaveman might've mentioned bringing back a viewmodel class for `MediaItem`, in which case the prefix might be redundant.

thekaveman:
@angela-tran @machikoyasuda Yeah I would rather we have an explicit viewmodel here, with properties, since this is now being used on many different pages. Agree that some clarity and consistency on these props would be helpful.

benefits/core/migrations/0002_sample_data.py
```diff
@@ -108,6 +108,8 @@ def load_sample_data(app, *args, **kwargs):
         unverified_title=_("eligibility.pages.unverified.mst_login.title"),
         unverified_headline=_("eligibility.pages.unverified.mst_login.headline"),
         unverified_blurb=_("eligibility.pages.unverified.mst_login.p[0]"),
+        enrollment_start_header=_("enrollment.pages.index.mst_login.start.header"),
+        enrollment_start_message=_("enrollment.pages.index.mst_login.start.message"),
```
thekaveman, line 112: Sorry I'm just rebasing on this and noticed - it feels a little duplicative to call out both `index` and `start` in this key. Is there more than 1 header / message on this page?

thekaveman: Ah, based on the PO it looks like these are for the item in the media list yes? I'm just trying to get the most clear keys here so we all know what's going on.

angela-tran: No worries, I appreciate this discussion for sure! The header/message on lines 111 and 112 go with the first media item, which is dynamic per-verifier. The other media item on the page is the bank card one, which is always the same. I've labeled them with their current msgids in the screenshot below:

image

thekaveman: Wait... isn't that a screenshot of the Eligibility Start page? This ticket is for the Enrollment start/index right?

machikoyasuda: What if the keys matched, like

```
enrollment_start_heading=_("enrollment.pages.index.mst_login.start.heading"),
enrollment_start_details=_("enrollment.pages.index.mst_login.start.details"
```

```
        confirmed_benefit_item = dict(
            icon=viewmodels.Icon("happybus", pgettext("image alt text", "core.icons.happybus")),
            heading=_(verifier.enrollment_start_heading),
            details=_(verifier.enrollment_start_details),
        )
```

thekaveman: I see why @angela-tran you had the `confirmed` string in the key earlier. If we made it `eligibility_confirmed` would that work here?

angela-tran: You're right @thekaveman , my bad. Here's the Enrollment start/index

image

benefits/core/templates/core/includes/media-list.html
```diff
@@ -0,0 +1,18 @@
+<ul class="media-list">
+    {% for item in media %}
+        <li class="media">
+            <div class="media-line">{% include "core/includes/icon.html" with icon=item.icon %}</div>
+            <div class="media-body">
+                <h3 class="media-body--heading">{{ item.heading }}</h3>
+                <div class="media-body--details">
+                    <p>{{ item.details }}</p>
+                    <div class="media-body--items">
```
thekaveman, line 9: Wrap in `{% if %}` to avoid rendering `media-body--items` if no items

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -58,10 +58,10 @@ msgstr ""
 "That’s okay! You may still be eligible for our program. Please reach out to "
 "your local transit provider for assistance."
 
-msgid "enrollment.pages.index.mst_login.start.header"
+msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.header"
 msgstr "Your eligibility is confirmed for an Older Adult benefit"
 
-msgid "enrollment.pages.index.mst_login.start.message"
+msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.message"
```
thekaveman, line 64: These were updated to `.heading` and `.details` in the sample data

benefits/locale/es/LC_MESSAGES/django.po
```diff
@@ -60,10 +60,10 @@ msgstr ""
 "¡Esta bien! Aún puede ser elegible para nuestro programa. Comuníquese con su "
 "proveedor de tránsito local para obtener asistencia."
 
-msgid "enrollment.pages.index.mst_login.start.header"
+msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.header"
 msgstr "TODO: Your eligibility is confirmed for an Older Adult benefit"
 
-msgid "enrollment.pages.index.mst_login.start.message"
+msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.message"
```
thekaveman, line 66: These were updated to `.heading` and `.details` in the sample data

thekaveman (changes requested):

benefits/core/templates/core/includes/media-list.html
angela-tran: Done in https://github.com/cal-itp/benefits/pull/967/commits/1ef48ce4355e00135b41155f159c0e3efcd235fe for media-list include.

image

benefits/locale/en/LC_MESSAGES/django.po
angela-tran: Thank you, fixed in https://github.com/cal-itp/benefits/pull/967/commits/f1f20c4547ed5516f415517bb49e8c819b089b9b

benefits/locale/es/LC_MESSAGES/django.po
angela-tran: Thank you, fixed in https://github.com/cal-itp/benefits/pull/967/commits/f1f20c4547ed5516f415517bb49e8c819b089b9b

angela-tran:
Rebasing to resolve merge conflicts

benefits/core/migrations/0002_sample_data.py
angela-tran: I've renamed the properties and msgids that this PR added. I know there will be some other renaming in #994 and #376, so I tried to keep my renaming here minimal while still correct. I ran `./bin/makemessages.sh` so I think it should all be correct. Let me know if you see any other mistakes.

benefits/core/templates/core/includes/media-list.html
```diff
@@ -6,11 +6,14 @@
                 <h3 class="media-body--heading">{{ item.heading }}</h3>
                 <div class="media-body--details">
                     <p>{{ item.details }}</p>
-                    <div class="media-body--items">
-                        <ul>
-                            {% for l in item.bullets %}<li>{{ l }}</li>{% endfor %}
-                        </ul>
-                    </div>
+                    {% if item.bullets|length %}
+                        <div class="media-body--items {{ item.bullets }}">
```
thekaveman, line 10: What is this extra class using `{{ item.bullets }}`?

angela-tran: That was left behind from some debugging. Just removed it

thekaveman (approved):
I have a few more minor changes / comments, but I really want to get this in - will address in my follow up.

machikoyasuda (approved):
💯

Issue #968, opened 2022-09-28
CSS: Create CSS Variables for colors, fonts


machikoyasuda:
A clear and concise description of the task.

## Acceptance Criteria

- [ ] Variable-ize all colors. Use color names from Figma.
- [ ] Variable-ize font sizes, letter-spacing, line-spacing, font weights

Example:

$primary-color: #046B99;
$white-color: #fffffff;

$header-background: $primary-color;
$button-text-color: $white-color;

$h1-font-size: 18px;

$h3-font-weight: 700;

PR #970, opened 2022-09-28
Refactor: Use CSS Variables for colors, font-family

machikoyasuda:
#968 

- As much as possible, _override_ root Bootstrap CSS variables to match Figma style guide
- When a Bootstrap CSS variable is not available, create new CSS variables and use them throughout the code

The CSS file should no longer have any un-variable-ized colors and font-family declarations.

## How to read this code

- Read https://getbootstrap.com/docs/5.1/customize/css-variables/
- All variables that start with `--bs` are overriding Bootstrap CSS variables.

## How this change works

For example, when overriding `--bs-primary` with `#046b99`, like so:

```css
:root {
  --primary-color: #046b99;
  --primary-color-rgb: 4, 107, 153;
  --bs-blue: var(--primary-color);
  --bs-primary: var(--primary-color);
  --bs-primary-rgb: var(--primary-color-rgb);
}
```

all Bootstrap classes that rely on `--bs-primary` color will also be affected. Most notably, `btn-primary`, `bg-primary` will not all also use `#046b99`. With just these few CSS variable override lines, I was able to change *all* instances of the "primary" color to `#046b99` _throughout_ the CSS styles. Because of this, I was able to delete places in the style.css code where we were manually setting the Header background color or a Button background color to `#046b99`, because we are using classes of `bg-primary` or `btn-primary` instead.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -9,6 +9,9 @@
   --bs-body-text-align: left;
   --bs-body-font-weight: 400;
   --bs-body-font-size: 16px;
+  --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
+    "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  --bs-body-font-family: var(--bs-font-sans-serif);
 }
 
 @font-face {
@@ -29,7 +32,7 @@ p,
 a,
 span,
 li {
-  font-family: "Roboto", sans-serif;
+  font-family: var(--bs-body-font-family);
 }
 
 h1,
@@ -124,9 +127,6 @@ footer.global-footer .footer-links a:active {
 /* Log In */
 
 #login {
-  background: var(--primary-color);
-  border: 0;
-  color: #ffffff;
   cursor: pointer;
   display: block;
   line-height: 1;
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -1,10 +1,17 @@
+:root {
+  --primary-color: #046b99;
+  --primary-color-rgb: 4, 107, 153;
+  --bs-blue: var(--primary-color);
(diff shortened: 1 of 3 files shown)


machikoyasuda:
Rebased

benefits/static/css/styles.css
```diff
@@ -1,10 +1,29 @@
...
+  --bs-body-font-size: 16px;
+  --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
+    "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  --bs-body-font-family: var(--bs-font-sans-serif);
+  --dark-color: #212121;
+  --footer-background-color: var(--dark-color);
+  --footer-link-color: #73b3e7;
+  --footer-link-hover-color: #9b74d7;
+  --footer-mobile-underline-color: var(--bs-white);
+  --hover-color: #2f4c65;
```
thekaveman, line 20: What is this `hover-color` for? Can we add some context like the other ones e.g. `footer-link-hover-color`

benefits/static/css/styles.css
```diff
@@ -594,7 +601,7 @@ footer.global-footer .footer-links a:active {
   }
 
   .global-footer ul.footer-links li:not(:last-child) {
-    border-bottom: 2px solid white;
+    border-bottom: 2px solid var(--footer-mobile-underline-color);
```
thekaveman, line 604: Nice, I like using a specific variable here vs. just `bs-white`

benefits/static/css/styles.css
machikoyasuda: It's for all links and buttons. I'm also following color name conventions from the styleguide itself too: 
image

thekaveman: Got it, thanks!

thekaveman (approved):

Issue #971, opened 2022-09-28
CSS: Create CSS Variables for font size and reset padding/margins


machikoyasuda:
A clear and concise description of the task.

## Acceptance Criteria

- [ ]

PR #952, opened 2022-09-23
Chore: update State Template

thekaveman:
Closes #777

## Key page detail

| Page | Before (mobile) | After (mobile) | Before (desktop) | After (desktop) | Notes |
| ----- | ------- | ------ | ----- | ---- | --- |
| `/eligibility/start` | image | image | image | image | Increased gutter around icon/navbar on mobile |
| `/enrollment` | image | image | image | image | Smaller primary button text, underlined secondary on mobile; justified text in desktop?? |
| all pages | image | image | image | image | Hard to see - but on desktop at a certain height, the `body` background color is showing through under the footer (white line at the very bottom) |

@angela-tran @machikoyasuda thoughts on addressing these issues here vs. as part of some of the other template cleanup work?

PR #984, opened 2022-09-29
Refactor: CSS Variables - Paragraph style and links

machikoyasuda:
part of #971

## What this PR does
- Set body `<p>` size to 18px for all browser widths/sizes
- Goes through the whole app and makes sure the `<p>` elements are set as such
- Does not yet add a set margin/padding-top/bottom to all `<p>` yet. To come in future PRs.
- Set standard colors, underlines for `<a>` links that are not `.btn`s

## Style guide

image
image

## Paragraphs

All of the un-bolded paragraphs in these screenshots are proper `<p>`s:
image
image
image

## Links
image
image

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -8,6 +8,7 @@
   --bs-body-color: var(--text-primary-color);
   --bs-body-text-align: left;
   --bs-body-font-weight: 400;
+  --bold-font-weight: 700;
   --bs-body-font-size: 18px;
   --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
     "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
@@ -21,6 +22,7 @@
   --footer-mobile-underline-color: var(--bs-white);
   --hover-color: #2f4c65;
   --error-color: #ea1010;
+  --selected-color: #562b97;
   --error-color-rgb: 234, 16, 16;
   --bs-danger: var(--error-color);
   --bs-danger-rgb: var(--error-color-rgb);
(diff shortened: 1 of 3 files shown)


benefits/static/css/styles.css
```diff
@@ -170,6 +189,9 @@ footer.global-footer .footer-links a:active {
   padding-top: 1px;
   vertical-align: baseline;
   color: transparent;
+  font-size: 16px;
+  font-weight: var(--bold-font-weight);
+  line-height: 1;
```
thekaveman, line 194: Should `font-size` and `line-height` be variables here?

benefits/static/css/styles.css
```diff
@@ -705,22 +708,12 @@ footer.global-footer .footer-links a:active {
   }
 
   .media-list .media .media-body--heading {
-    font-size: 20px;
     line-height: 30px;
```
thekaveman, line 711: Should `line-height` be a variable?

thekaveman (commented):
I added one small commit to fix a bug that I introduced in #952, sorry about that!

benefits/static/css/styles.css
machikoyasuda: This PR is only dealing with paragraphs and links, so it will in a future PR.

benefits/static/css/styles.css
machikoyasuda: I'm thinking no (at least for now in this PR) b/c this is a one-off 16px link. Once I have a 2nd link that is also 16px, we can make a small-link variable or something.

benefits/static/css/styles.css
thekaveman: I would argue `line-height` is part of the font definition, but leave it up to you.

thekaveman (approved):

angela-tran (dismissed):
These changes look good to me for paragraphs and links.

Just curious, what was the Login.gov button regression? from https://github.com/cal-itp/benefits/pull/984/commits/eb7233d56fa33f9704fb7a86d51ed13e39570c0b

machikoyasuda:
> These changes look good to me for paragraphs and links.
> 
> Just curious, what was the Login.gov button regression? from [eb7233d](https://github.com/cal-itp/benefits/commit/eb7233d56fa33f9704fb7a86d51ed13e39570c0b)

When Login.gov button also has the default link font size of 18px, the button gets too tall. And if the line-height isn't 1, the logo and the text goes to 2 lines:

image

angela-tran (approved):

PR #985, opened 2022-09-30
Chore: data migration reminders

thekaveman:
* Adds reminder steps to the checklist to update data migrations
* Adds labeler workflow that labels PRs that may require data migrations updates


angela-tran (approved):
Very helpful idea

Issue #996, opened 2022-10-04
Refactor media list in Eligibility Start


thekaveman:
Following #967 and #997 we'll have a new media list include and viewmodel for media items. We need to remove the duplicate code in the Eligibility Start template and use the include + viewmodel instead.

## Acceptance Criteria

- [ ] We have a single include for media list, used everywhere that rendering is needed
- [ ] No duplicate media list HTML exists in the templates

## Additional context

#997 refactored other usages of the `dict` into the viewmodel, but left Eligibility Start alone since it had its own copy of the HTML/rendering.

PR #997, opened 2022-10-04
Refactor: enrollment success copy

### LINKED ISSUE #946: Update Enrollment Success page
## Acceptance Criteria

- [ ] Enrollment Success page has latest copy

## Additional context
Keep in mind the per-verifier dynamic content.
Dynamic button should already work. Don't worry about alignment for text or buttons for this ticket.

### LINKED ISSUE #994: Create viewmodel for the media item list include
#967 refactors the media list from the Eligibility Start template into a reusable include for use in the Enrollment flow as well. However we maintained the existing `dict` context for simplicity and to ensure no breakages.

This ticket is to refactor the `dict` context into a `viewmodel` class like the others, with the necessary properties to capture data for a media list item.

### LINKED ISSUE #1004: Enrollment Connect Card page: Move sentence back to the last media item
## Acceptance Criteria

- [ ] Add the last paragraph as a paragraph in the last media item

image

thekaveman:
Closes #946

Closes #994 - I needed this to be able to [add multiple paragraphs to the media list](https://github.com/cal-itp/benefits/issues/946#issuecomment-1265903343)

Closes #1004 - added support for multiple paragraphs here anyway, it was an easy/small change to include

## Before

image

### MST CC verification

image

### Login.gov verification

image

_(note styling implementation is not finalized or addressed in this PR)_


benefits/enrollment/templates/enrollment/success.html
```diff
@@ -7,26 +7,16 @@
   {{ block.super |add:" enrollment-success" }}
 {% endblock classes %}
 
-{% block paragraph_content %}
-  {% if "logged-out" in page.classes %}
-    <img class="success-image" src="{% static "img/icon/happybus.svg" %}" alt="{% translate "core.icons.happybus" context "image alt text" %}">
```
angela-tran, line 12: I think that, with lines 11-12 removed, the `logged-out` class is not being used anywhere.

Previously, the screen upon logout looked like
image

Now it looks like
image

Maybe in the logged out state, we can remove the media items from the page? Or something to get the same net effect.

benefits/enrollment/views.py
```diff
@@ -163,26 +163,43 @@ def success(request):
...
     session.update(request, origin=reverse(ROUTE_SUCCESS))
+
     verifier = session.verifier(request)
-    icon = viewmodels.Icon("bankcardcheck", pgettext("image alt text", "core.icons.bankcardcheck"))
-    page = viewmodels.Page(
-        classes="no-image-mobile",
-        title=_("enrollment.pages.success.title"),
-        headline=_("enrollment.pages.success.headline"),
+
+    success_item = viewmodels.MediaItem(
```
angela-tran, line 169: I noticed this type of refactor wasn't applied to the dicts in [`eligibility` `start`](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L85-L104), but then I remembered about the bullet points. Just wanted to document/confirm that that's why we're leaving those alone for now.

benefits/core/templates/core/includes/media-list.html
```diff
@@ -3,18 +3,19 @@
...
-                    {% if item.bullets|length %}
-                        <div class="media-body--items">
-                            <ul>
-                                {% for l in item.bullets %}<li>{{ l }}</li>{% endfor %}
-                            </ul>
-                        </div>
-                    {% else %}
-                    {% endif %}
-                </div>
+                {% if item.heading %}<h3 class="media-body--heading">{{ item.heading }}</h3>{% endif %}
```
machikoyasuda, line 6: No changes required/just a comment: Technically, the CSS does not have any regressions if an `<h3>` does get rendered with an empty `{{ item.heading }}`. Same with having an empty `<div class="media-body--details>`. The CSS / HTML is written in such a way that if it's empty, the DOM does not create/add any extra height or anything:

image

benefits/enrollment/templates/enrollment/success.html
```diff
@@ -7,26 +7,16 @@
...
-  {% else %}
-    <p>{% translate "enrollment.pages.success.p1" %}</p>
-    <p>{% translate "enrollment.pages.success.p2" %}</p>
-    <p>{% blocktranslate with help_link=help_link%}enrollment.pages.success.p3{{ help_link }}{% endblocktranslate %}</p>
-  {% endif %}
```
machikoyasuda, line 17: Don't we still need this logged out view?

benefits/core/templates/core/includes/media-list.html
thekaveman: That's fine... but it's still not semantically correct, especially with an empty heading. We should always avoid rendering empty HTML.

benefits/enrollment/views.py
thekaveman: See #996

benefits/enrollment/templates/enrollment/success.html
thekaveman: Yeah thanks for this reminder, we talked about moving it out of the success view and I forgot to do that here.

benefits/enrollment/templates/enrollment/success.html
thekaveman: https://github.com/cal-itp/benefits/pull/997#discussion_r987120711

benefits/enrollment/templates/enrollment/success.html
thekaveman: It didn't actually make sense to move this out of the success view and create a specific "logged out" view - since the user can log out from various places in the app, we can't send them to a final "you are logged out" _everytime_ they log out - we use the `session.origin()` to know where to send the user.

benefits/enrollment/templates/enrollment/success.html
thekaveman: Fixed in [6f0db24](https://github.com/cal-itp/benefits/pull/997/commits/6f0db241880135aa4b734b64776173dc483c54c0)

benefits/enrollment/templates/enrollment/success.html
thekaveman: See [6f0db24](https://github.com/cal-itp/benefits/pull/997/commits/6f0db241880135aa4b734b64776173dc483c54c0)

machikoyasuda (approved):
Make sure to run `DJANGO_LOAD_SAMPLE_DATA=true bin/init.sh` before testing.

image

And the empty media item spacing issue is also fixed 👍

machikoyasuda (commented):
For this PR, after it is merged, the full Login.gov flow and signing out of it should be checked on dev-benefits.calitp.org/ to double-check the logged out page looks okay.

thekaveman:
I am going to prep the data migrations before merging so I don't break dev. Will aslo post a working localhost migration that we can all use for testing against the Login sandbox + our local test server

angela-tran (approved):

PR #999, opened 2022-10-04
Refactor: remove agency index pass-thru

### LINKED ISSUE #976: Remove agency index pass-through
With the new Courtesy Card designs, we have a better context-setting Agency Index (homepage). Let's remove the auto-pass-through that currently sends the user straight to Eligibility Start.

## Acceptance Criteria

- [ ] A user lands on the Agency Index page when viewing the Benefits app
- [ ] The Agency Index continue button takes the user to the Eligibility Index page

## Additional context

Work on the new homepage is happening in #937

thekaveman:
Closes #976... Although I realized in doing this that, it doesn't actually matter? We'll have a single agency with multiple verifiers, so this check would have failed anyway.

Oh well, it's dead code, so fine to remove.

### DIFF
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -52,9 +52,6 @@ def agency_index(request, agency):
     session.reset(request)
     session.update(request, agency=agency, origin=agency.index_url)
 
-    if len(agency.eligibility_verifiers.all()) == 1:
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
Agreed that this check doesn't really make sense anymore

Issue #972, opened 2022-09-29
CSS: Refactor Button includes, CSS and alignment + 404 page/Help button


machikoyasuda:
A clear and concise description of the task.

## Acceptance Criteria

- [x] I believe Button LABEL can be deleted now
- [x] Refactor btn-lg
- [x] Btn width should be 3 grid width for Desktop and different for mobile, centered on mobile
- [x] Button alignment to the right, and centered on mobile
- [x] Btn-outline-light should not have manual width
- [x] Btn-primary capitalize
- [x] Btn-lg
- [x] Btn-link: Remove?
- [x] Button Back
- [x] Button Previous
- [x] Sign Out link

machikoyasuda:
<img width="1512" alt="image" src="https://user-images.githubusercontent.com/3673236/194446814-ee5cc13c-341b-4888-a0a3-1be236a7c194.png">
<img width="1512" alt="image" src="https://user-images.githubusercontent.com/3673236/194446853-055fec4d-9c36-453d-b7fc-f8681aaf8352.png">
<img width="1512" alt="image" src="https://user-images.githubusercontent.com/3673236/194446866-e6e0be83-11da-40eb-9dee-bd38a96eddb4.png">

angela-tran:
@machikoyasuda Will this task cover setting buttons to `width: 100%`? My pages in #1013 are looking ok, except for the buttons... I was wondering if I should just ignore the buttons in my PR.

Issue #948, opened 2022-09-22
Button: Add animation to the Check/Checking button


machikoyasuda:
A clear and concise description of the task.

## Acceptance Criteria

-
<img width="325" alt="image" src="https://user-images.githubusercontent.com/3673236/191816114-fc3f3869-2de0-4a69-a939-6687ef879fcf.png">

- Changes text and changes animation

## Additional context

- Look into LittlePay button implementation

machikoyasuda:
This bootstrap class will add a spinner animation: https://getbootstrap.com/docs/5.1/components/spinners/

Issue #947, opened 2022-09-22
Form pages: Style inputs + Alignment


machikoyasuda:
## Acceptance Criteria

- [x] Eligibility confirm page: new form inputs, new form placeholder
- [x] Test for a11y/focus
- [x] Eligibility page: Radio buttons
- Create label for radio button in the radio_select itself and rename includes to be just for Verifier
- Fix the button javascript issue #948 related to this

1. Add row/alignment code to get the Submit button in the right-aligned-to-centered style
2. Figure out how to style the Radio Button as a 8-col left-aligned
3. And Form as 6-col centered
4. Input colors/font sizes
5. Radio button tweak margins/padding

## Additional context

Read USDWS form input guide for guidance

PR #1008, opened 2022-10-05
Fix: Remove unused form error-handling; add new placeholder + regex

### LINKED ISSUE #1007: Remove unused form with-errors logic/code
With Courtesy Cards, we have the browser-based form checks for required fields and `sub` pattern (regex) -- there isn't a way for the user to submit the form with bad inputs, assuming the browser is working as advertised.

Further, when the data gets to our Server, it will be hashed -- there isn't a way for the Server to know if the format of user data was incorrect and/or missing.

All that is to say - we have a ton of HTML/CSS/JS/Python for dealing with form errors that really isn't needed anymore. Let's clean this up.

## Acceptance Criteria

- [x] There is not HTML/CSS/JS related to form errors
- [x] The Eligibility form does not deal with form/API errors

machikoyasuda:
closes #1007 

## What this PR does
- Remove form `with-errors` related HTML, CSS, JS
- Add sample placeholder and regex

## How to test
- Go through both forms and ensure no regressions in style and function
- Go through Input form: Test an input that fails the regex

image


machikoyasuda:
@thekaveman @angela-tran small mostly deletion pr ready for review. It would be helpful for me to have this in dev before I work on forms PR.

angela-tran (commented):
I think we should also delete the `add_api_errors` method in `EligibilityVerificationForm`

benefits/core/templates/core/includes/form.html
```diff
@@ -7,70 +7,57 @@
...
-    </form>
-
-  {% endif %}
+    {% endif %}
+
+    <script>
+          $(function(){
+              $("form[action='{{form_action}}']").on("submit", function(e) {
+                  if ("{{ form.submitting_value }}" !== "") {
+                      $("button[data-action='submit']", this)
```
thekaveman, line 52: I don't think this is a regression from this PR, but I just noticed... not all of the `button`s on this template have the `data-action` attribute, so this selector would not find it sometimes

thekaveman (changes requested):
No change required for this, but a comment on commit 4413ea0ec6f4e40f5834450cfcfb6d917c7d5bc3 - the message doesn't seem clear/accurate, and it was a huge amount of changes... maybe for next time, break something like that up into a few smaller chunks: dealing with the form separately from the JS for example, and the whitespace/formatting separately from both.

One requested change below, otherwise this looks good!

benefits/core/templates/core/includes/form.html
machikoyasuda: yea sorry the rebase commits https://github.com/cal-itp/benefits/commit/4413ea0ec6f4e40f5834450cfcfb6d917c7d5bc3 are all stemming from difficulty of managing several open PRs + rebasing dev + slow GitHub Actions. and yes, the spacing on this form template is really confusing.

machikoyasuda: I can take care of and test this Form button issue in either https://github.com/cal-itp/benefits/issues/947 and https://github.com/cal-itp/benefits/issues/972. The Button CSS code, Button viewmodel and Form HTML code have not been changed yet in this Sprint and this is the first of many PRs to start cleaning this all up.

angela-tran: Is this actually the cause of https://github.com/cal-itp/benefits/issues/948?

machikoyasuda: @angela-tran probably

PR #1013, opened 2022-10-05
Refactor: simple pages

### LINKED ISSUE #996: Refactor media list in Eligibility Start

### LINKED ISSUE #1009: Port over all templates that can easily "fill in" the base grid
Define a separate .html file for each page and have them fill in the blocks from `base.html`.

## Acceptance Criteria

- [ ] Eligibility Start
- [ ] Enrollment Index

## Additional context
See context from #1003.

angela-tran:
Closes #1009 and closes #996 

### Eligibility Start
| Before | After |
| --- | --- |
| Screenshot 2022-10-11 at 16-36-13 Transit Benefits Get Started Cal-ITP | Screenshot 2022-10-11 at 16-36-27 Transit Benefits Get Started Cal-ITP |

_This page looks the same because @machikoyasuda did the spacing / styles in #1020, so this PR was just refactoring the template to be simpler and to remove duplicate media-list HTML._

### Enrollment Index
| Before | After |
| --- | --- |
| Screenshot 2022-10-11 at 16-32-57 Transit Benefits Connect your card Cal-ITP | Screenshot 2022-10-11 at 16-33-04 Transit Benefits Connect your card Cal-ITP |

Updated the copy, per Slack message at https://cal-itp.slack.com/archives/C043NFXPUAG/p1665517931000059


benefits/eligibility/views.py
```diff
@@ -90,7 +90,7 @@ def start(request):
     identity_item = dict(
```
thekaveman, line 90: This should use the `viewmodels.MediaItem` instead of `dict` now (which also handles the `details` needing to be a list)

benefits/eligibility/views.py
```diff
@@ -118,22 +118,23 @@ def start(request):
     bank_card_item = dict(
```
thekaveman, line 118: Same comment to use `viewmodels.MediaItem`

benefits/enrollment/templates/enrollment/index.html
```diff
@@ -1,24 +1,17 @@
...
+{% block nav-buttons %}
+  <div class="d-flex justify-content-end">{% include "core/includes/sign-out-link.html" %}</div>
+{% endblock nav-buttons %}
 
-{% block icon_title %}
-  {{ block.super }}
-{% endblock icon_title %}
+{% block inner-content %}
+  <div class="col-12 col-sm-12 col-lg-10">
+    <h2 class="media-title"></h2>
```
thekaveman, line 9: Empty `<h2>`?

benefits/enrollment/templates/enrollment/index.html
```diff
@@ -92,14 +85,21 @@
...
+    {% include "core/includes/form.html" with form=f %}
+  {% endfor %}
+{% endblock inner-content %}
+
+{% block call-to-action-button %}
   {% if payment_processor %}
+    <button type="button" class="btn btn-primary btn-lg loading" role="status" disabled>
+      {{ payment_processor.loading_text }}
+    </button>
     <div class="invisible">
```
thekaveman, line 99: Why do we have this `class="invisble"` here? I know this PR didn't add it, but I see we now render the buttons inside?

benefits/core/migrations/0002_sample_data.py
```diff
@@ -147,8 +147,6 @@ def load_sample_data(app, *args, **kwargs):
         unverified_title=_("eligibility.pages.unverified.mst_cc.title"),
         unverified_headline=_("eligibility.pages.unverified.mst_cc.headline"),
         unverified_blurb=_("eligibility.pages.unverified.mst_cc.p[0]"),
-        eligibility_confirmed_item_heading=_("enrollment.pages.index.mst_cc.eligibility_confirmed_item.heading"),
-        eligibility_confirmed_item_details=_("enrollment.pages.index.mst_cc.eligibility_confirmed_item.details"),
```
thekaveman, line 151: I don't understand how this data can be deleted here in the data migration without making the fields `null=True` in the model/schema migration... but it does work 🤔

thekaveman (changes requested):

machikoyasuda:
One small change for the button row --

At tablet widths: 
In tablet widths, `col-lg-7` is too narrow for the button help text

image

At mobile widths, the CTA button and the help text widths are actually a bit narrower than 100%:

image

Having the button be a bit narrower on mobile (and thus adding more affordance padding on the left and right sides) helps people avoid unintentionally clicking the button with their thumb or something.

So to address both of these issues, I'm using this on the forms pages https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/includes/form.html#L33 --

I'm using: `<div class="col-lg-9 d-flex align-items-end flex-row-reverse col-sm-8 offset-sm-2 offset-lg-0 offset-2 col-8">` on the Help Text parent div, and `
`<div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">` on the Button parent div. 

This sets the help text column to 9 on large, 8 on small and 8 on mobile. And it sets the button width to 8 columns on small/mobile. I'm also giving an `offset-2` (and `offset-lg-0`) to center the text on mobile.

To achieve this look on the right, and below (tablet width):

image
image

benefits/enrollment/templates/enrollment/index.html
machikoyasuda: I always assumed this `invisible` class and the corresponding JS removing it (` $(".invisible").removeClass("invisible");`) was trying to prevent the user from clicking the button prematurely or something(???)

angela-tran: I agree with @machikoyasuda 's understanding of why it's here.

The main button has always been inside a `<div class="invisible">`, from what I can see in Git [history](https://github.com/cal-itp/benefits/commit/50a91417cefc86213362b034d2ccbc9db9e58c3a?diff=unified#diff-d6d406916b45273ffe3a8c113900029e13605ae79091373a7f7712824e6656e5R69-R75). The `invisible` class is removed when we get a successful response from calling the PaymentProcessor's `card_tokenize_url`.

thekaveman: Oh yeah! Thanks for the reminder - I had forgotten about the JS on that page.

benefits/eligibility/views.py
angela-tran: On line 98 and 105, additional keys (`links` and `bullets`) are added to the `dict` which are not on `MediaItem`. 

I can actually remove the one on line 98 though since the design no longer has a links on the Login.gov item. I think we said we weren't going to worry about the bullet points for now, and it can just work with the `dict`?

benefits/eligibility/views.py
angela-tran: Done in c14e5bb9733e51672d244934fef08cff6a1bc917

benefits/enrollment/templates/enrollment/index.html
angela-tran: I meant to start a discussion on this, but you beat me to it!

This is here temporarily because the spacing between the media-list and whatever is above it is defined on `media-title`. This media-list on this page does not have a media-title though. Does @machikoyasuda already have an issue that covers this? Does #1028 actually cover this?

benefits/eligibility/views.py
thekaveman: Ah... I'm the one that didn't add the `bullets` functionality to the viewmodel when I created it... was waiting for #996 I guess?

Please add it to the viewmodel and use that instead.

benefits/enrollment/templates/enrollment/index.html
thekaveman: Yes I believe this is handled by one of the cases in #1028: `h1 + .media-list`

benefits/eligibility/views.py
angela-tran: Can we discuss a spec before I start implementation?

When this came up previously, we weren't sure how `bullets` should be represented on the model (this text would be different per-`EligibilityVerifier`). A verifier could have 0-N bullets that it wants to show.

thekaveman: Right so the [template already looks like](https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/includes/media-list.html#L10) (in `dev`):

```html
{% if item.bullets|length %}
    <div class="media-body--items">
        <ul>
            {% for l in item.bullets %}<li>{{ l }}</li>{% endfor %}
        </ul>
    </div>
{% endif %}
```

So I think its taking what you have now and just making it another kwarg on the init, and make it work similar to `details`:

```python
class MediaItem:
  def __init__(self, icon: Icon, details, heading=None, bullets=None):
    if bullets:
      # make sure it's a list similar to details
      # self.bullets = list_of_bullets
```

benefits/static/css/styles.css
```diff
@@ -539,9 +539,13 @@ footer .footer-links li a:visited {
...
 
-.media-list .media .media-body--details,
+.media-list .media .media-body--details p:first-child,
 .media-list .media .media-body--items li:first-child {
-  padding-top: 5px;
+  padding-top: 4px;
+}
+
+.media-list .media .media-body--details p {
+  padding-top: 24px;
```
angela-tran, line 548: @machikoyasuda I made these small changes before I noticed #1028... I think I'll just revert these

benefits/enrollment/templates/enrollment/index.html
angela-tran: Ok, I removed the empty `<h2>` in 2ecf25a22597d5ed07a7bde1c836aa7fa99a5d52 and am good with #1028 fixing the spacing for this page, which now looks like this:

benefits/core/migrations/0002_sample_data.py
angela-tran: Yeah, I'm not sure either...

benefits/eligibility/views.py
angela-tran: That makes sense 👍 I'll make that change

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -520,7 +522,7 @@ msgid "enrollment.pages.index.title"
 msgstr "Connect your card"
 
 msgid "enrollment.pages.index.headline"
-msgstr "We have located you in our system!"
+msgstr "Your eligibility is confirmed!<br/>You’re nearly there."
```
machikoyasuda, line 525: Design calls for this now:
image

If you take out `<br/>` from the code, and use `col-lg-5` for the width of the parent div of the `h1`, you can achieve this design.

It would require overriding `main_content` of `base.html` _or_ modifying `base.html` so it's easier to override this part:

```html
            <div class="col-lg-10">
              <h1>{{ page.headline }}</h1>
            </div>
```

benefits/eligibility/views.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/1013/commits/400bcae738227428a81af871990d059b54e1907e

thekaveman: 🤘

benefits/core/migrations/0002_sample_data.py
thekaveman: Can you try to find out? Doesn't have to be today / part of this PR.

benefits/locale/en/LC_MESSAGES/django.po
angela-tran: Thanks for the feedback on this and on the [help text classes](https://github.com/cal-itp/benefits/pull/1013#issuecomment-1275398879). I'll work on making those changes

angela-tran: Done in 08e9a19a8aeae1609765beca472ad28fd38715d9

### Mobile
The headline is all one line (no break in between). At most mobile widths, the first sentence does not wrap.

image

Once the width is small enough, it does wrap.   



### Desktop

angela-tran:
> I'm using: `<div class="col-lg-9 d-flex align-items-end flex-row-reverse col-sm-8 offset-sm-2 offset-lg-0 offset-2 col-8">` on the Help Text parent div, and `<div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">` on the Button parent div. 
> 
> This sets the help text column to 9 on large, 8 on small and 8 on mobile. And it sets the button width to 8 columns on small/mobile. I'm also giving an `offset-2` (and `offset-lg-0`) to center the text on mobile.

@machikoyasuda I updated the divs for `call-to-action-text` and `call-to-action-button` in `base.html` to use these classes. (Done in d0957e9db8fb9cc90f71c2b330f2da817a2fe4c1)

<details><summary>Tablet Before/After</summary>

| Before | After |
| --- | --- |
| image | image |

</details>

<details><summary>Mobile Before/After</summary>

| Before | After |
| --- | --- |
| image | image |

</details>

I think the help text on mobile needs to be left-aligned. Also, do we have a ticket about making all the help text a link when in mobile?

machikoyasuda (approved):
Tested Login.gov flow (eligible, non eligible), Courtesy Card (eligible, non eligible) and Spanish. Tested to ensure the Connect Your Card button logic/styles all still work. 

I realize that it _is_ possible....depending on how `Your eligibility is confirmed! You’re nearly there.` gets translated into Spanish........ the line-breaking might be different. When we get that, we could play with either using `col-lg-5` like this PR does or try `col-lg-6` if that's better for Spanish. Either way, this is a much better method to achieve the desired line breaking without using `<br/`>!

benefits/core/migrations/0002_sample_data.py
angela-tran: So the short answer is: somehow the columns have an empty-string value.

```
sqlite> select eligibility_confirmed_item_heading from core_eligibilityverifier;
eligibility_confirmed_item_heading                                 
-------------------------------------------------------------------
enrollment.pages.index.mst_login.eligibility_confirmed_item.heading
                                                                   
sqlite> select count(eligibility_confirmed_item_heading) from core_eligibilityverifier where eligibility_confirmed_item_heading="";
count(eligibility_confirmed_item_heading)
-----------------------------------------
1                                        
sqlite> select count(eligibility_confirmed_item_heading) from core_eligibilityverifier where eligibility_confirmed_item_heading<>"";
count(eligibility_confirmed_item_heading)
-----------------------------------------
1                                        
sqlite> select count(eligibility_confirmed_item_heading) from core_eligibilityverifier where eligibility_confirmed_item_heading is null;
count(eligibility_confirmed_item_heading)
-----------------------------------------
0                   
```
I think I'll just put `null=False` on the fields so that it's not confusing.

angela-tran:
Thanks @machikoyasuda for approving the CSS changes

The last thing to review are the [changes](https://github.com/cal-itp/benefits/pull/1013/commits/45cad5a70bbc4578c6b71e889809685d9ccd14fa) I made to address @thekaveman 's question at https://github.com/cal-itp/benefits/pull/1013#discussion_r992879499

thekaveman (approved):
Thanks for making the model and sample data more explicit!

PR #1014, opened 2022-10-05
CSS Variables: Refactor footer sizing and alignment

### LINKED ISSUE #974: Footer: Redesign height

## Acceptance Criteria

- [ ] Footer is taller now on Desktop

image

image
- [ ] Footer each link is 50px on Mobile
image

machikoyasuda:
closes #974 

## What this PR does
- Fixes: Fixes the current issue where on Desktop, the footer is not neatly at the bottom of the page on first load for pages that don't fill up the full viewport height.
- Redesigns: The footer height is now 50px on Desktop. The footer is slightly taller now on Desktop.
- Fixes: The type (leter-spacing, font-weight) for the footer was off. Now it's not.
- Deletes all old CSS code that tried to place the footer.
- Moves around CSS styles into categories and comments them.

## How to test
- Mobile: Footer should have no regressions
- Desktop: Go through short pages (agency index) - Footer should be at the bottom. 
- Desktop: Go through long pages (help, eligibility start) - Footer should have no regressions.
- Test the app in Private mode or a different browser to test the `visited` color logic.

## Nitty gritty details
- Removes `.global-footer` so this footer is now 100% free of custom CA Web Std Template styles. We were overriding most of them. Uses bootstrap classes instead to declare most styles. Uses CSS `gap` to separate the links on Desktop.
- Moves all footer-related CSS variables to right above the footer element CSS styles. We can undo this, but this helped me write/read this code.
- Logic for the viewport height, explained: This means, "Set the _minimum_ height of the center div to 100% of the viewport height, minus the height of the footer and the header". That means, if the natural height of the page is _more than_ 100vh from the start (which would apply to all pages on a smaller mobile device and some pages on a Desktop device), then the footer will flow naturally. **Edited to add:** The footer is sticky but not fixed. Fixed would mean the footer is always there even after you scroll.
```css
--main-content-min-height: calc(100vh - 182px);
  /* 182px = Header Height (80px) + (Footer Link (50px) * Number of Links (2)) + Underline Height (2px) */
--main-content-min-height: calc(100vh - 130px);
/* 130px = Header Height (80px) + Footer Height (50px) */
```

## Post merge actions:

After this PR is merged and deployed on Development, we should ask @srhhnry to test the footer out on her mobile and desktop devices.

## Screenshots

Short page - first load
image

Short page - visited links
image

Long page - first load
image

Mobile - 
image

image

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -71,14 +71,14 @@
       {% endblock main_content %}
     </main>
 
-    <footer id="footer" class="global-footer">
+    <footer id="footer">
       <div class="container">
-        <ul class="footer-links">
+        <ul class="footer-links m-0 p-0 list-unstyled d-lg-flex gap-lg-4">
           <li>
-            <a href="{% url "core:help" %}">{% translate "core.buttons.help" %}</a>
+            <a class="m-0 ps-5 ps-lg-0" href="{% url "core:help" %}">{% translate "core.buttons.help" %}</a>
           </li>
           <li>
-            <a href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer">{% translate "core.buttons.privacy" %}</a>
+            <a class="m-0 ps-5 ps-lg-0" href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer">{% translate "core.buttons.privacy" %}</a>
           </li>
         </ul>
       </div>
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -153,6 +153,7 @@ main {
 /* Footer has same font styles on all screen widths */
 /* Mobile first: One link per row, each link is 50px height */
 /* Screen width above 992px - Footer is 50px height */
+/* Note: Not all Footer styles are written in mobile-first */
 
 :root {
   --footer-background-color: var(--dark-color);
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -207,13 +207,53 @@ footer.global-footer .footer-links li a:active {
   }
 }
 
-/* class styles */
+/* Header */
+
+.navbar.navbar-expand-sm.navbar-dark.bg-primary {
+  padding: 8.5px 1rem;
+}
+
+.navbar-brand {
+  padding-top: 0;
+  padding-bottom: 0;
+}
+
+.navbar-brand img.sm {
+  width: 120px;
+}
+
+.navbar-brand img.lg {
+  width: 271px;
+}
+
+/* Buttons */
 
 .btn-lg {
   border-width: 2px;
 }
 
-/* Log In */
+.btn-link {
+  color: var(--primary-color);
+}
+
+.btn-primary {
+  text-transform: capitalize;
+  letter-spacing: 0.05em;
+}
+
+.btn-outline-light {
+  width: 126px;
+  height: 38px;
+  padding: 0px !important;
+}
+
+.container.content .buttons label {
+  display: block;
+  margin-bottom: 2rem;
+  text-align: center;
+}
+
+/* Custom button: Log In */
 
 #login {
   cursor: pointer;
@@ -248,7 +288,7 @@ footer.global-footer .footer-links li a:active {
   line-height: 1;
 }
 
-/* Sign Out */
+/* Custom button: Sign Out */
 
 .signout-row {
   position: absolute;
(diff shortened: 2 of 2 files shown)


angela-tran:
> The footer is not sticky. Sticky would mean the footer is always there even after you scroll.

Just want to note this to improve our communication: I think we're using this term to mean two different things. I understood it to mean that the footer sticks to the bottom of the viewport if the page content is not enough to push it down. If there's enough content, then it will push the footer down. There is some MDN documentation on this: https://developer.mozilla.org/en-US/docs/Web/CSS/Layout_cookbook/Sticky_footers

The "Hey!" note on https://css-tricks.com/couple-takes-sticky-footer/ also tries to clarify this

machikoyasuda:
@angela-tran Oooh yes ~ It is sticky but _not_ **fixed**.

Just added this up top

>  **Edited to add:** The footer is sticky but not fixed. Fixed would mean the footer is always there even after you scroll.

benefits/static/css/styles.css
```diff
@@ -143,61 +143,115 @@ h3,
...
-footer.global-footer .footer-links a:active {
+footer .footer-links li a:hover,
+footer .footer-links li a:focus,
+footer .footer-links li a:active,
+footer .footer-links li a:visited {
   color: var(--footer-link-hover-color);
 }
 
-/* class styles */
+@media (max-width: 992px) {
```
thekaveman, line 197: Can you please call out here with a comment that this is a non-mobile-first media query? And maybe elsewhere too if needed... will help us go back and clean up later.

benefits/static/css/styles.css
```diff
@@ -566,36 +584,6 @@ footer.global-footer .footer-links a:active {
 }
 
 @media (max-width: 767px) {
```
thekaveman, line 586: same comment about non-mobile-first media query

benefits/core/templates/core/base.html
```diff
@@ -71,14 +71,14 @@
       {% endblock main_content %}
     </main>
 
-    <footer id="footer" class="global-footer">
```
thekaveman, line 74: Losing this class doesn't break anything from the State Template, right?

thekaveman (dismissed):
Few minor comment requests

benefits/core/templates/core/base.html
machikoyasuda: As far as I've tested no.

benefits/static/css/styles.css
machikoyasuda: These ones and lower are because I have not had a chance to work on radio buttons yet. All the media queries below this are just stuff I have not gotten to.

benefits/static/css/styles.css
machikoyasuda: I will add a comment into the code and also put it here: This is because of the way the Footer on Mobile _must_ have 100% width fully to get the full-width white line. And the Footer on Desktop+ _needs_ the Bootstrap `.container` logic to size its widths. If we removed the `.container` and went with `width: 100%` (to be mobile-first), then we'd have to re-write our own container width logic, which would be very difficult. So basically, it's because of the full-width line that goes edge to edge.

benefits/static/css/styles.css
```diff
@@ -32,12 +28,16 @@
...
+  --main-content-min-height: calc(100vh - 182px);
+  /* 182px = Header Height (80px) + (Footer Link (50px) * Number of Links (2)) + Underline Height (2px) */
 }
 
 @media (min-width: 992px) {
   :root {
     --h1-font-size: calc(35rem / 16);
     --h1-text-align: center;
+    --main-content-min-height: calc(100vh - 130px);
+    /* 130px = Header Height (80px) + Footer Height (50px) */
```
angela-tran, line 40: This line was initially inside the `@media (min-width: 767px)` media query at b35e375d8a1b14dfab5a6e925da26f44494413dd, but then the media query moved in https://github.com/cal-itp/benefits/pull/1014/commits/10cbca03a5230dffb5528e5a7890c86da1031827#diff-744df555aee38173a0b0868baa7befa720e6a28f80747523e74e724c126cd381L46-L47 without taking this line. I can't tell what the intended change was, so could you clarify?

benefits/static/css/styles.css
thekaveman: Makes sense and that is totally fine! A little doc comment describing that would be helpful is all.

benefits/static/css/styles.css
machikoyasuda: https://github.com/cal-itp/benefits/pull/1014/commits/f9b605fbd115639ee5c4a4301ae590ae951378d7

Footer breakpoint is 992px (can be changed if Design wants to change it)

I want Design to test the footer on dev-benefits.calitp.org on real devices before we change the breakpoint though.

machikoyasuda:
@thekaveman @angela-tran After this is approved, I can take care of the conflict.

angela-tran (approved):

angela-tran (approved):

PR #1022, opened 2022-10-05
Fix: Remove unused form error-handling; add new placeholder + regex

### LINKED ISSUE #1007: Remove unused form with-errors logic/code

machikoyasuda:
closes #1007 

## What this PR does
- Remove form `with-errors` related HTML, CSS, JS and Python method
- Add sample placeholder and regex

## How to test
- Run `./bin/init.sh` to get the new sample data regex/input placeholder
- Go through both forms and ensure no regressions in style and function
- Go through Input form: Test an input that fails the regex

image


thekaveman (approved):
I ran through the CC flow for:

1. Input that doesn't match the regex ✅ 
2. Input that is verified ✅ 
3. Input that is not verified ✅ 

It all looks good; maybe we can clean the spacing up later.

machikoyasuda:
@thekaveman Yea - on spacing, this next PR #1023 would be the best time and place to change the Form template spacing. As far as I can tell, the linters don't do anything for anything in between `<script>`....`</script>` on these pages. For now, I'm just making sure the if/elses and open/close elements are all aligned.

PR #1003, opened 2022-10-04
Refactor: base grid

angela-tran:
Closes #961 

This PR defines a grid on `base.html` with blocks that other templates can "fill in". If a page wants to completely ignore the grid, it can do so by overriding the `main_content` block, but it should only do this if it truly is completely different from all other pages.

### What this PR does not cover and follow-up PRs need to cover

- #1009
- Port over templates that contain forms which need to render their own button
- Implement CSS for spacing between sections according to 8px-based design system
- Add support for error pages and port over
- See if Help page can use the grid
- Implement CSS for making buttons have `width: 100%`

### Screenshots
<details>

**Agency Index**
Screenshot 2022-10-04 at 20-13-35 Transit Benefits Introduction Cal-ITP

**Verifier Selection**
Screenshot 2022-10-04 at 20-13-47 Transit Benefits Select a benefit option Cal-ITP

**CC Eligibility Start**
Screenshot 2022-10-04 at 20-13-55 Transit Benefits Get Started Cal-ITP

**Login.gov Eligibility Start**
Screenshot 2022-10-04 at 20-22-11 Transit Benefits Get Started Cal-ITP

**CC Eligibility Confirm**
Screenshot 2022-10-04 at 20-16-22 Transit Benefits Confirm your Courtesy Card Cal-ITP

**CC Enrollment Connect Card**
Screenshot 2022-10-04 at 20-19-04 Transit Benefits Connect your card Cal-ITP

**CC Eligibility Unverified**
Screenshot 2022-10-04 at 20-16-47 Transit Benefits Courtesy Card not located Cal-ITP

**Login.gov Enrollment Connect Card**
Screenshot 2022-10-04 at 20-23-23 Transit Benefits Connect your card Cal-ITP

_Note: these Enrollment Success screenshots will be out of date as soon as #946 is merged in_
**CC Enrollment Success**
Screenshot 2022-10-04 at 20-21-15 Transit Benefits Success Cal-ITP

**Login.gov Enrollment Success**
Screenshot 2022-10-04 at 20-23-48 Transit Benefits Success Cal-ITP

**Login.gov Logged Out**
Screenshot 2022-10-04 at 20-23-57 Transit Benefits Success Cal-ITP

</details>

PR #1023, opened 2022-10-06
Refactor & Redesign: Eligibility Index, Eligibility Confirm form pages

### LINKED ISSUE #947: Form pages: Style inputs + Alignment

### LINKED ISSUE #948: Button: Add animation to the Check/Checking button

machikoyasuda:
closes #947 

## What this PR does

This PR shows how the two pages, Eligibility Index and Eligibility Confirm, can be created with a one-off html file that inherit from `page` or `base` and overrides `main_content`, while also allowing the form input area to be properly aligned for its own unique page design.

This PR also will refactor the Radio Button and Text Input CSS with variables and Bootstrap classes.

### Eligibility Index / Radio Buttons
- [x] Creates custom template, `eligibility/index.html` that inherits from `page` (but in the future can easily inherit from `base` from #1003)
- [ ] Set and send `offset-lg-1 col-lg-9` to the form includes to set the form as 9-col wide and aligned with the headline and explanatory text
- [ ] Radio Button label: Needs to be a different headline size
- [ ] Radio Button spacing: Variable-ize and correct it

### Eligibility Confirm / Text Inputs
- [x] Build off of the already-built `eligibility/confirm.html` and add on headline, explanatory-text and form
- [ ] Set and send `offset-lg-3 col-lg-6` to the form includes, to set the form as 6-col wide and centered
- [ ] Text input spacing and colors: Correct it
- [ ] Text input helper text: Correct it

### Form submit buttons
- Add divs and classes to set the form input button to be 3-col wide on Desktop and right-aligned, and centered on mobile
- [ ] Simplify form submit button classes, may possibly fix #948 regression
- [ ] Simplify form submit CSS code with variables

### Other
- Update view tests

### DIFF
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -6,33 +6,45 @@
   <form action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
     {% csrf_token %}
 
-    {% for field in form %}
-      <div class="form-group">
-        {% if field.label %}
-          <label for="{{ field.id_for_label }}" class="form-control-label h2">
-            {{ field.label }}
-            {% if field.field.required %}<span class="required-label">*</span>{% endif %}
-          </label>
-        {% endif %}
+    <div class="row">
+      {% comment %} <div class="offset-lg-3 col-lg-6"> {% endcomment %}
+      <div class="offset-lg-1 col-lg-9">
+        {% for field in form %}
+          <div class="row">
+            <div class="col-12">
+              {% if field.label %}
+                <label for="{{ field.id_for_label }}" class="form-control-label h2">
+                  {{ field.label }}
+                  {% if field.field.required %}<span class="required-label">*</span>{% endif %}
+                </label>
+              {% endif %}
 
-        {{ field }}
+              {{ field }}
 
-        {% if field.help_text %}<small class="form-text text-muted">{{ field.help_text }}</small>{% endif %}
+              {% if field.help_text %}<small class="form-text text-muted">{{ field.help_text }}</small>{% endif %}
+            </div>
+          </div>
+        {% endfor %}
       </div>
-    {% endfor %}
+    </div>
 
     {% if form.submit_value %}
-      <div class="form-group">
-        {% if recaptcha.enabled %}
-          <button class="btn btn-lg btn-primary g-recaptcha"
-                  data-sitekey="{{ recaptcha.site_key }}"
-                  data-callback="onSubmit"
-                  data-action="submit">
-            {{ form.submit_value }}
-          </button>
-        {% else %}
-          <button class="btn btn-lg btn-primary">{{ form.submit_value }}</button>
-        {% endif %}
+      <div class="row d-flex flex-column-reverse justify-content-lg-end flex-lg-row">
+        <div class="col-lg-7 d-flex align-items-end flex-row-reverse">
+          <p>Not sure which option is right for you? Please visit our Help Page.</p>
+        </div>
+        <div class="col-lg-3">
+          {% if recaptcha.enabled %}
+            <button class="btn btn-lg btn-primary w-100 g-recaptcha"
+                    data-sitekey="{{ recaptcha.site_key }}"
+                    data-callback="onSubmit"
+                    data-action="submit">
+              {{ form.submit_value }}
+            </button>
+          {% else %}
+            <button class="btn btn-lg btn-primary w-100">{{ form.submit_value }}</button>
+          {% endif %}
+        </div>
       </div>
     {% endif %}
 
@@ -44,6 +56,7 @@
                 }
       </script>
     {% endif %}
+
     <script>
             $(function(){
                 $("form[action='{{form_action}}']").on("submit", function(e) {
--- a/benefits/eligibility/templates/eligibility/confirm.html
+++ b/benefits/eligibility/templates/eligibility/confirm.html
@@ -1,6 +1,27 @@
 {% extends "core/page.html" %}
 
 {% block main_content %}
-  {% include "core/includes/sign-out-link.html" %}
-  {{ block.super }}
+  <div class="container">
+    {% include "core/includes/sign-out-link.html" %}
+
+    <div class="row justify-content-center">
+      <div class="col-lg-10">
+        {% block icon_title %}
+          <h1>{{ page.headline }}</h1>
+        {% endblock icon_title %}
+
+        {% block paragraph_content %}
+          {% for p in page.paragraphs %}<p>{{ p }}</p>{% endfor %}
+        {% endblock paragraph_content %}
+      </div>
+    </div>
+
+    <div class="row">
+      {% block forms %}
+        {% for f in page.forms %}
+          {% include "core/includes/form.html" with classes="TODO: variable" form=f %}
+        {% endfor %}
+      {% endblock forms %}
+    </div>
+  </div>
 {% endblock main_content %}
--- a/benefits/eligibility/templates/eligibility/index.html
+++ b/benefits/eligibility/templates/eligibility/index.html
@@ -1,0 +1,25 @@
+{% extends "core/page.html" %}
+
+{% block main_content %}
+    <div class="container">
+        <div class="row justify-content-center">
+            <div class="col-lg-10">
+                {% block icon_title %}
+                    <h1>{{ page.headline }}</h1>
+                {% endblock icon_title %}
+
+                {% block paragraph_content %}
+                    {% for p in page.paragraphs %}<p>{{ p }}</p>{% endfor %}
+                {% endblock paragraph_content %}
+            </div>
+        </div>
+
+        <div class="row">
+            {% block forms %}
+                {% for f in page.forms %}
+                    {% include "core/includes/form.html" with classes="TODO: variable" form=f %}
+                {% endfor %}
+            {% endblock forms %}
+        </div>
+    </div>
+{% endblock main_content %}
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -12,7 +12,7 @@ from django.utils.translation import pgettext, gettext as _
 from benefits.core import recaptcha, session, viewmodels
 from benefits.core.middleware import AgencySessionRequired, LoginRequired, RateLimit, VerifierSessionRequired
 from benefits.core.models import EligibilityVerifier
-from benefits.core.views import ROUTE_HELP, TEMPLATE_PAGE
+from benefits.core.views import ROUTE_HELP
 from . import analytics, forms, verify
 
 
@@ -22,6 +22,7 @@ ROUTE_LOGIN = "oauth:login"
 ROUTE_CONFIRM = "eligibility:confirm"
 ROUTE_ENROLLMENT = "enrollment:index"
 
+TEMPLATE_INDEX = "eligibility/index.html"
 TEMPLATE_START = "eligibility/start.html"
 TEMPLATE_CONFIRM = "eligibility/confirm.html"
 TEMPLATE_UNVERIFIED = "eligibility/unverified.html"
@@ -59,14 +60,14 @@ def index(request):
         else:
             # form was not valid, allow for correction/resubmission
             page.forms = [form]
-            response = TemplateResponse(request, TEMPLATE_PAGE, page.context_dict())
+            response = TemplateResponse(request, TEMPLATE_INDEX, page.context_dict())
     else:
         if agency.eligibility_verifiers.count() == 1:
             verifier = agency.eligibility_verifiers.first()
             session.update(request, verifier=verifier)
             response = redirect(eligibility_start)
         else:
-            response = TemplateResponse(request, TEMPLATE_PAGE, page.context_dict())
+            response = TemplateResponse(request, TEMPLATE_INDEX, page.context_dict())
 
     return response
 
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -265,12 +265,6 @@ footer .footer-links li a:visited {
   padding: 0px !important;
 }
 
-.container.content .buttons label {
-  display: block;
-  margin-bottom: 2rem;
-  text-align: center;
-}
-
 /* Custom button: Log In */
 
 #login {
@@ -433,14 +427,14 @@ footer .footer-links li a:visited {
 
 /* Eligibility Start */
 
-.eligibility-start .buttons .btn {
+/* .eligibility-start .buttons .btn {
   width: 100%;
-}
+} */
 
-.eligibility-start .main-container .buttons label {
+/* .eligibility-start .main-container .buttons label {
   margin: 0 0.8rem 0 0;
   line-height: 1;
-}
+} */
 
 /* Enrollment Success */
 
@@ -519,23 +513,23 @@ footer .footer-links li a:visited {
   margin-bottom: 5rem;
 }
 
-.container.content h1 ~ p:nth-last-of-type(1) {
+/* .container.content h1 ~ p:nth-last-of-type(1) {
   margin-bottom: 2rem;
-}
+} */
 
-.container.content .btn-lg,
-.eligibility-start .btn-link {
-  font-weight: 500;
+/* .container.content .btn-lg,
+.eligibility-start .btn-link { */
+/* font-weight: 500;
   margin-bottom: 1rem;
-  padding: 0.8125rem;
-  /* btn-lg height is 60px */
-  width: 100%;
-}
+  padding: 0.8125rem; */
+/* btn-lg height is 60px */
+/* width: 100%; */
+/* } */
 
-.container.content .btn-link,
+/* .container.content .btn-link,
 .eligibility-start .btn-link {
   text-decoration: underline;
-}
+} */
 
 /* media queries */
 
@@ -557,9 +551,9 @@ footer .footer-links li a:visited {
     padding: 14.5px 1rem;
   }
 
-  .container.content .btn-lg {
+  /* .container.content .btn-lg {
     padding: 1.1875rem 0.813rem;
-  }
+  } */
 
   .signout-row .container .signout-link {
     font-size: 18px;
@@ -605,15 +599,4 @@ footer .footer-links li a:visited {
   .agency-index .container.content h1 ~ p:nth-last-of-type(1) {
     margin-bottom: 2rem;
   } */
-}
-
-@media (min-width: 992px) {
-  .container.content input[type="submit"],
-  .container.content .btn {
-    width: fit-content;
-    min-width: 250px;
-    display: block;
-    margin-left: auto;
-    margin-right: auto;
-  }
 }


benefits/static/css/styles.css
```diff
@@ -245,46 +253,67 @@ footer .footer-links li a:visited {
...
-.btn-link {
-  color: var(--primary-color);
+@media (min-width: 992px) {
+  :root {
+    --primary-button-padding: 13px 0;
+  }
 }
 
-.btn-primary {
+.btn-lg.btn-primary.btn-lg {
```
thekaveman, line 270: Do we need the double `.btn-lg` selector here?

benefits/static/css/styles.css
```diff
@@ -173,22 +176,23 @@ main {
   --footer-link-color: #73b3e7;
   --footer-link-width: 100%;
   --footer-link-hover-color: #9b74d7;
-  --footer-link-font-weight: 500;
+  --footer-link-font-weight: var(--medium-font-weight);
```
thekaveman, line 179: If you're going to rebase on `dev`, you want to apply this variable to the new `#previous-page-button` style I added? It uses the same font weight.

benefits/static/css/styles.css
```diff
@@ -245,46 +253,67 @@ footer .footer-links li a:visited {
...
-  letter-spacing: 0.05em;
+  letter-spacing: 0.02em;
+  font-weight: var(--medium-font-weight);
+  font-size: calc(20rem / 16);
+  width: 100%;
+  padding: var(--primary-button-padding);
 }
 
+/* A button that actually is a text link */
+/* TODO: Check if this is actually really used anymore */
```
thekaveman, line 281: It looks like we do here: https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L92

But I'm not sure if that is needed...

benefits/core/templates/core/widgets/verifier_radio_select.html
```diff
@@ -1,5 +1,14 @@
+{% load i18n %}
 
 {% with id=widget.attrs.id %}
+
+  <label for="{{ id }}">
+    <h2 class="pb-4">
```
thekaveman, line 6: We want to avoid putting headings inside labels: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#headings

benefits/eligibility/templates/eligibility/index.html
```diff
@@ -0,0 +1,25 @@
+{% extends "core/base.html" %}
+
+{% block main_content %}
+    <div class="container">
```
thekaveman, line 4: spacing should be cleaned up

benefits/eligibility/templates/eligibility/index.html
```diff
@@ -0,0 +1,25 @@
...
+                {% block paragraph_content %}
+                    {% for p in page.paragraphs %}<p style="padding-top: 24px;">{{ p }}</p>{% endfor %}
+                {% endblock paragraph_content %}
+            </div>
+        </div>
+
+        <div class="row">
+            {% block forms %}
+                {% for f in page.forms %}
+                    {% include "core/includes/form.html" with classes=page.classes form=f %}
```
thekaveman, line 20: This seems strange to pass `page.classes` to the form. 

Can we edit the form class and [the view](https://github.com/cal-itp/benefits/blob/feat/947-forms--buttons-alignment-styles/benefits/eligibility/views.py#L49
), to pass the classes directly to the form?

```python
def index(request):

    # ...

     page = viewmodels.Page(
        title=_("eligibility.pages.index.title"),
        headline=_("eligibility.pages.index.headline"),
        paragraphs=[
            format_html(_("eligibility.pages.index.p[0]%(info_link)s") % {"info_link": f"{help_page}#what-is-cal-itp"})
        ],
        forms=forms.EligibilityVerifierSelectionForm(agency=agency, classes="offset-lg-1 col-lg-9"),
    )
```

Or even just hardcode them in the form class, since we always want to display this form the same.

benefits/eligibility/views.py
```diff
@@ -39,6 +40,7 @@ def index(request):
     help_page = reverse(ROUTE_HELP)
 
     page = viewmodels.Page(
+        classes="offset-lg-1 col-lg-9",
```
thekaveman, line 43: See [comment](https://github.com/cal-itp/benefits/pull/1023/files#r989572808)

benefits/eligibility/templates/eligibility/confirm.html
```diff
@@ -1,6 +1,25 @@
...
+        {% block paragraph_content %}
+          {% for p in page.paragraphs %}<p style="padding-top: 24px;">{{ p }}</p>{% endfor %}
+        {% endblock paragraph_content %}
+      </div>
+    </div>
+
+    <div class="row">
+      {% block forms %}
+        {% for f in page.forms %}
+          {% include "core/includes/form.html" with classes=page.classes form=f %}
```
thekaveman, line 20: See [comment](https://github.com/cal-itp/benefits/pull/1023/files#r989572808)

benefits/eligibility/views.py
```diff
@@ -160,6 +166,7 @@ def confirm(request):
 
     # GET/POST for Eligibility API verification
     page = viewmodels.Page(
+        classes="offset-lg-3 col-lg-6",
```
thekaveman, line 169: See [comment](https://github.com/cal-itp/benefits/pull/1023/files#r989572808)

benefits/static/css/styles.css
machikoyasuda: Yeah I think this will eventually get deleted. But seemed outside of scope of this PR. I think whoever touches `eligibility#start` next should delete it.

benefits/eligibility/templates/eligibility/index.html
machikoyasuda: What spacing specifically? This passes djlint formatter.

machikoyasuda: Oh wait. Or do you mean like padding.

thekaveman: No I meant the code formatting... I'm not sure why, this page uses 4 as the tabsize. Our other HTML pages use 2.

benefits/enrollment/templates/enrollment/index.html
```diff
@@ -12,9 +12,13 @@
 {% block paragraph_content %}
   {{ block.super }}
   {% if payment_processor %}
-    <button type="button" class="btn btn-primary btn-lg loading" role="status" disabled>
-      {{ payment_processor.loading_text }}
-    </button>
+    <div class="row justify-content-center justify-content-lg-end">
```
machikoyasuda, line 15: I added alignment classes on this page for testing purposes. I wanted to make sure any `<form>` or `button` CSS changes weren't causing any regressions on this page, and that the same button CSS + classes can be used to get the desired design.

machikoyasuda: image

benefits/eligibility/templates/eligibility/index.html
machikoyasuda: Tried doing this here: https://github.com/cal-itp/benefits/pull/1023/commits/38bbf9913ed615a6c6d6850ba5c4e895a77dbe85 and undoing past changes here https://github.com/cal-itp/benefits/pull/1023/commits/49690c9bd638556a1ead33f520160ab5985387e2

benefits/eligibility/forms.py
```diff
@@ -22,10 +22,11 @@ class EligibilityVerifierSelectionForm(forms.Form):
     # sets label to empty string so the radio_select template can override the label style
     submit_value = _("eligibility.buttons.choose")
 
-    def __init__(self, agency: models.TransitAgency, *args, **kwargs):
+    def __init__(self, agency: models.TransitAgency, classes, *args, **kwargs):
         super().__init__(*args, **kwargs)
         verifiers = agency.eligibility_verifiers.all()
 
+        self.classes = classes
```
machikoyasuda, line 29: Is there a more optimal way to declare this optional attribute? @thekaveman

benefits/eligibility/views.py
```diff
@@ -40,21 +40,20 @@ def index(request):
...
+        forms=forms.EligibilityVerifierSelectionForm(agency=agency, classes="offset-lg-1 col-lg-9"),
     )
 
     ctx = page.context_dict()
     ctx["help_page"] = help_page
     ctx["show_help_text"] = True
 
     if request.method == "POST":
-        form = forms.EligibilityVerifierSelectionForm(data=request.POST, agency=agency)
+        form = forms.EligibilityVerifierSelectionForm(data=request.POST, agency=agency, classes="offset-lg-1 col-lg-9")
```
machikoyasuda, line 56: I wasn't sure if the class attribute should be set for this `form` and the one on line 188 too? I did it for all of them cos the page wouldn't lose otherwise.

benefits/static/css/styles.css
machikoyasuda: removed. replaced the extra`btn-lg` with `btn`. I'm keeping it like this with all 3 classes for now until the Button refactor ticket, b/c, right now, all the buttons have all 3 of these classes.

benefits/core/templates/core/widgets/verifier_radio_select.html
machikoyasuda: done

benefits/eligibility/templates/eligibility/index.html
```diff
@@ -1,29 +1,29 @@
...
+      <div class="col-lg-10">
+        {% block icon_title %}
+          <h1>{{ page.headline }}</h1>
+        {% endblock icon_title %}
 
-                {% block paragraph_content %}
-                    {% for p in page.paragraphs %}<p class="pt-4 pb-4 pb-lg-8">{{ p }}</p>{% endfor %}
-                {% endblock paragraph_content %}
-            </div>
-        </div>
```
machikoyasuda: @thekaveman I think the formatting is fixed now??

thekaveman:
Reviewing changes now 👀

benefits/eligibility/forms.py
thekaveman: I kind of snuck this comment at the bottom of [my other thought](https://github.com/cal-itp/benefits/pull/1023#discussion_r989572808): it might be easier to just hardcode the classes directly in the form, no need to pass any in -- since they are always the same for each form right?

So instead of:

```python
class EligibilityVerifierSelectionForm(forms.Form):
   def __init__(self, agency: models.TransitAgency, classes, *args, **kwargs):
        self.classes = classes

# and later, each time the form is created

form = EligibilityVerifierSelectionForm(classes="offset-lg-1 col-lg-9")
```

We could have:

```python
class EligibilityVerifierSelectionForm(forms.Form):
   def __init__(self, agency: models.TransitAgency, *args, **kwargs):
        self.classes = "offset-lg-1 col-lg-9"
```

It might simplify some of the usage (in the rest of the python code) of the forms?

benefits/eligibility/views.py
thekaveman: Yes, for the `POST` back case, if the form is invalid (which it shouldn't be... but if...) then we would send the user back to the form page to try again, so you need it here.

And then definitely need it in the `GET` since that is the initial page load of the form.

[See my comment above](https://github.com/cal-itp/benefits/pull/1023#discussion_r990225774), maybe a simplification?

benefits/core/templates/core/includes/form.html
```diff
@@ -1,38 +1,53 @@
...
-        {% if recaptcha.enabled %}
-          <button class="btn btn-lg btn-primary g-recaptcha"
-                  data-sitekey="{{ recaptcha.site_key }}"
-                  data-callback="onSubmit"
-                  data-action="submit">
-            {{ form.submit_value }}
-          </button>
-        {% else %}
-          <button class="btn btn-lg btn-primary">{{ form.submit_value }}</button>
+      <div class="row d-flex flex-column-reverse justify-content-lg-end flex-lg-row" style="padding-top: 64px;">
```
thekaveman, line 31: Not a dealbreaker, but I think you created a utility class for this large padding top right?

benefits/core/templates/core/includes/form.html
```diff
@@ -1,38 +1,53 @@
...
+        <div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">
+          {% if recaptcha.enabled %}
+            <button class="btn btn-lg btn-primary g-recaptcha"
+                    data-sitekey="{{ recaptcha.site_key }}"
+                    data-callback="onSubmit"
+                    data-action="submit">
+              {{ form.submit_value }}
+            </button>
+          {% else %}
+            <button class="btn btn-lg btn-primary" data-action="submit">{{ form.submit_value }}</button>
```
thekaveman, line 48: Nice thank you for the `data-action` here

benefits/eligibility/templates/eligibility/confirm.html
```diff
@@ -1,6 +1,31 @@
...
 
 {% block main_content %}
-  <div class="container">{% include "core/includes/previous-page-link.html" %}</div>
-  {{ block.super }}
+  <div class="container">
+    {% include "core/includes/previous-page-link.html" %}
+
+    <div class="row justify-content-center">
+      <div class="col-lg-10">
+        {% block icon_title %}
```
thekaveman, line 13: Just a comment: since this now inherits from `base.html`, there's really no need for these `{% block %}` sections -- they are meant to create (or use) placeholders in a parent template.

We could just have the content inside the blocks directly here:

* `{% block icon_title %}`
* `{% block paragraph_content %}`
* `{% block forms %}`

And same comment for `eligibility/index.html`

benefits/core/templates/core/includes/form.html
machikoyasuda: Good catch!

benefits/eligibility/templates/eligibility/confirm.html
machikoyasuda: Ended up keeping {% block paragraph_content %} and {% block forms %} and modifying how the nav_buttons work cos it was messing up the width of the page

benefits/eligibility/forms.py
machikoyasuda: done

machikoyasuda:
This PR is now using templates that inherit from the latest `base.html`. I had to make 2 small changes to `base.html` to make the alignment all work:

1. Adds the page title to the title: https://github.com/cal-itp/benefits/pull/1023/commits/800bd9c76b3f6581ef763a9b05388d7286ef86c9#diff-e2ae5f60bb472effafac24b079fcbb1ffd09dcc5d8df2d22cab07417d222f925L12-L13

2. When adding the previous button bar to the nav-buttons section, having a `row` be a parent of a `row previous_page_row` broke the alignment. (There was no longer a container width constraining the div, so the whole page would scroll horizontally passed 100 viewport width, and makes the window scroll and look like a lot of blank space). So I took the `<div class="row">` out from base.html, so that the includes itself can add it. We need to be able to set the `height` on the row. A future refactor _could_ add these styles to the nav-buttons div itself, but it's not necessary for the time being.

cc @angela-tran @thekaveman

thekaveman (approved):
This looks really good!

## Possible regression?

I'm not sure if this is a minor regression on Mobile, but the "Previous" button should have `border-width: 2px`, it went back to `1px` here?

image

## Eligibility Index

### Figma - Desktop

image

### This PR running locally - Desktop

image

### Figma - Mobile

image

### This PR running locally - Mobile

image
 
## Eligibility Confirm

### Figma - Desktop

image

### This PR running locally - Desktop

image

### Figma - Mobile

image

### This PR running locally - Mobile

image

angela-tran (changes requested):
I started this review earlier today and am just now submitting it.

I feel like I actually want to look at the CSS changes more closely, but don't want to prolong submitting the comments that I do have

angela-tran:
> This PR is now using templates that inherit from the latest `base.html`. I had to make 2 small changes to `base.html` to make the alignment all work:
> 
>     1. Adds the page title to the title: [800bd9c#diff-e2ae5f60bb472effafac24b079fcbb1ffd09dcc5d8df2d22cab07417d222f925L12-L13](https://github.com/cal-itp/benefits/commit/800bd9c76b3f6581ef763a9b05388d7286ef86c9#diff-e2ae5f60bb472effafac24b079fcbb1ffd09dcc5d8df2d22cab07417d222f925L12-L13)
> 
>     2. When adding the previous button bar to the nav-buttons section, having a `row` be a parent of a `row previous_page_row` broke the alignment. (There was no longer a container width constraining the div, so the whole page would scroll horizontally passed 100 viewport width, and makes the window scroll and look like a lot of blank space). So I took the `<div class="row">` out from base.html, so that the includes itself can add it. We need to be able to set the `height` on the row. A future refactor _could_ add these styles to the nav-buttons div itself, but it's not necessary for the time being.
> 
> 
> cc @angela-tran @thekaveman

Some of my commits on [#1013](https://github.com/cal-itp/benefits/pull/1013/commits) do the same thing, which is ok - the merging is fine. Not sure if there's a way to more clearly delineate what belongs in which PR.

benefits/core/templates/core/widgets/verifier_radio_select.html
```diff
@@ -1,5 +1,12 @@
+{% load i18n %}
 
 {% with id=widget.attrs.id %}
+
+  <label for="{{ id }}" class="pb-4 h2">
+    {% translate "eligibility.pages.index.label" %}
+    <span class="required-label text-body">*</span>
+  </label>
```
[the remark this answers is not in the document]
angela-tran: > I also think the rename is not necessary because we can probably assume that all radio-button forms would have this styling for their labels.

Ok, I take this part back. I realized the index page as it is in Figma (with transit-agency radio buttons) would have an "h1" styled label.

benefits/eligibility/templates/eligibility/index.html
angela-tran: Sorry to open up a resolved conversation, but couldn't we put these classes in the template itself? As in, put it directly in `eligibility/index.html`.

thekaveman: @angela-tran do you mean to pass them in to the form include directly from the template? The classes are needed inside the form include itself, not in page template. 

That could work too, maybe makes more sense from the POV of keeping front-end stuff out of python.

benefits/core/templates/core/widgets/verifier_radio_select.html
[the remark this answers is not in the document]
thekaveman: The rename was something we talked about to be indicative that this _is_ a special case radio list. I'm not sure there is an advantage of defining a special-case _field_ when all we really want to change is the rendering of this one instance of radio list. 

We can revisit if we have more forms in the future with complex radio list requirements. We know this design will likely change when there are 3+ options too, so this is a simple enough approach for now.

benefits/core/templates/core/includes/form.html
```diff
@@ -1,38 +1,53 @@
...
-                  data-action="submit">
-            {{ form.submit_value }}
-          </button>
-        {% else %}
-          <button class="btn btn-lg btn-primary">{{ form.submit_value }}</button>
+      <div class="row d-flex flex-column-reverse justify-content-lg-end flex-lg-row" style="padding-top: 64px;">
+        {% if show_help_text %}
+          <div class="col-lg-9 d-flex align-items-end flex-row-reverse col-sm-8 offset-sm-2 offset-lg-0 offset-2 col-8">
+            <p class="pt-4 pt-lg-0">
+              {% blocktranslate with help_link=help_link%}eligibility.pages.index.help_text{{ help_link }}{% endblocktranslate %}
```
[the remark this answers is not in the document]
thekaveman: 👍

benefits/eligibility/templates/eligibility/index.html
angela-tran: @thekaveman Yeah, that's more or less what I meant.

I was thinking more like we'd remove the wrapping `<div>`s from `form.html` and just have the page's template define them. The template would wrap the `{% for f in page.forms %}` loop.

Your implementation would also fit the idea.

thekaveman: @angela-tran But they are needed/different per-form, and needed [deeper in the form structure](https://github.com/cal-itp/benefits/blob/feat/947-forms--buttons-alignment-styles/benefits/core/templates/core/includes/form.html#L10) than what a wrapper would provide:

```html
<!-- wrapper takes effect here -->

<form> <!-- include starts -->
<!-- this is all inside the include -->
  <div class="row form-container">
      <div class="{{ form.classes }}">
        {% for field in form %}
        <!-- ... -->
        {% endfor %}
     </div>
  </div>

  <!-- notice this row is inside the <form> but adjacent to the row/cols defined above by form classes -->
  <!-- so this can't be done as a wrapper -->

  {% if form.submit_value %}
      <div class="row d-flex flex-column-reverse justify-content-lg-end flex-lg-row pt-8">
```

angela-tran: > But they are needed/different per-form

I would argue that the current designs don't require this, but... if we're saying we want to be prepared for pages with  multiple forms that all need different alignment from each other, I feel perfectly fine with your suggestion of passing the classes to the form include.

machikoyasuda: Also note: The submit buttons on these two forms need to be in the same parent div as the form itself, or the form submit buttons won't submit. That's why the `base.html` template - with the `<form>` and the `call to action` button being in different divs - will not work for these two form pages. So the HTML form functionality requires this:

image

The `Choose this Benefit` button *has* to be a child of the `<form>`, for the form to submit properly.

machikoyasuda:
@angela-tran @thekaveman 

Today's update:
1. Added back CSS declarations to make the buttons have a 2px border.
2. Passing in help text here: https://github.com/cal-itp/benefits/pull/1023/commits/a8a6c63a4d4025be561fd535d6f31d51633b9236 -- I left the `/help` link with `#what-is-cal-itp` anchor, but it should be changed to whatever future question tackles what the difference between the two eligibility verifier options are, once we update the Help page.

Happy to go over any more reviews over Zoom!

benefits/eligibility/templates/eligibility/index.html
angela-tran: I'm ok with moving this conversation to a separate GitHub issue so that this PR isn't bogged down since it already has a lot of commits. 

I still do think though that 
- it'd be better to pass the classes in somewhere closer to the front-end 
- the `form` attribute for the submit element (whether `input` or `button`) could help simplify a lot and help our pages follow a more unified approach.

benefits/core/widgets.py
```diff
@@ -11,16 +11,17 @@ class FormControlTextInput(widgets.TextInput):
...
 
-        self.attrs.update({"class": "form-control form-control-lg"})
+        self.attrs.update({"class": "form-control"})
         if pattern:
             self.attrs.update({"pattern": pattern})
         if placeholder:
             self.attrs.update({"placeholder": placeholder})
 
 
 class RadioSelect(widgets.RadioSelect):
```
thekaveman, line 21: This class could also be renamed to `VerifierRadioSelect` to help with clarity

benefits/eligibility/templates/eligibility/confirm.html
thekaveman: Just curious, why keep those two `block`?

benefits/eligibility/templates/eligibility/index.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block page_title %}
+  {{ page.title }} |
+{% endblock page_title %}
+
+{% block main_content %}
+  <div class="container">
+    <div class="row justify-content-center">
+      <div class="col-lg-10">
+        {% block icon_title %}
```
thekaveman: Also this `block` - it doesn't exist in the `base.html` template and there are no child templates inheriting from this one...?

benefits/core/templates/core/widgets/verifier_radio_select.html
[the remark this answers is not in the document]
thekaveman: My overall answer would be: because this code doesn't matter, it is the only instance of this form element on the site (we don't need to over-engineer a complicated solution with options) and the design will change in the very near future. I would like to put this radio list thing to bed for now, we've spent enough time on this very basic and temporary feature.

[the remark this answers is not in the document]
angela-tran: Ok, that is what I wanted clarity on. I will resolve my comments on this.

benefits/core/widgets.py
```diff
@@ -11,16 +11,17 @@ class FormControlTextInput(widgets.TextInput):
...
-    template_name = "core/widgets/radio_select.html"
-    option_template_name = "core/widgets/radio_select_option.html"
+    template_name = "core/widgets/verifier_radio_select.html"
+    option_template_name = "core/widgets/verifier_radio_select_option.html"
+    # Note: the templates are specific to the Eligibility Verifier
```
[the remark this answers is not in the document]
angela-tran: Answered by https://github.com/cal-itp/benefits/pull/1023#discussion_r992684394

benefits/core/widgets.py
machikoyasuda: Renamed and also added a docstring https://github.com/cal-itp/benefits/pull/1023/commits/6ad1eadb2a57aace29e241b986fb4cf5754868a8

angela-tran (approved):
Other than @thekaveman's question at https://github.com/cal-itp/benefits/pull/1023#discussion_r992668493, everything on this PR looks resolved to me

thekaveman (approved):

benefits/eligibility/templates/eligibility/confirm.html
machikoyasuda: Forms has to be `block` cos I'm overriding both form and the CTA

I opted to put paragraph content in `block` because I wanted to be able to set their own padding for the paragraphs from the inside. 

I don't want to set padding on `paragraph_content` at the `base.html` level, because in the cases where there is a page with a _blank_ `paragraph_content`, it will still add padding. And we don't want that.

machikoyasuda: cc @angela-tran @thekaveman 

I need to override `paragraph_content` b/c:

1. I need to add these classes `pt-4 pb-4 pb-lg-8` which set padding-bottom to 64px, among other things, from the *inside* of the parent div.

2. If we add `pt-4 pb-4 pb-lg-8` to the main parent itself, when there are pages without `paragraph_content` (which most pages don't have `paragraph_content`), the pages will have a giant 64px blank space. Padding applies even if the div inside is empty. We should not be adding padding classes to divs that might be empty/un-used in the base template. 

image

thekaveman: Ok I don't want any of this to be blocking for this PR since it really doesn't matter to the end result...

> Forms has to be `block` cos I'm overriding both form and the CTA

There isn't anything to override. `{% block %}` is used to place content in specific spot (the block) defined in a **parent** template. The parent here `base.html` does not define a `{% block forms %}`, so there is no spot to place/override anything. In `confirm.html`, this block turns into... exactly nothing in the rendered HTML.

By using `{% block forms %}` here, it creates a block that another template, inheriting from `confirm.html`, could fill in; that doesn't really make sense / isn't our use-case for these form pages.

> I opted to put paragraph content in `block` because I wanted to be able to set their own padding for the paragraphs from the inside... I don't want to set padding on paragraph_content at the base.html level

So here it looks like a change to put the paragraphs into `{% block explanatory-text %}` which makes sense - the `base.html` defines this, so this places the paragraphs in that spot in the base template.

PR #1025, opened 2022-10-06
Feat: Previous Page include

### LINKED ISSUE #943: Create "Previous Page" button includes
## Acceptance Criteria

- Not going to worry about alignment part yet
- Add logic to make Previous Page button only show up on these 2 pages

- Eligibility start page (Uses custom template, start.html)
image
- Eligibility confirm page (Uses Page template)
image

thekaveman:
Closes #943.

Mostly finished, looking for feedback on direction. Alignment/spacing is not totally correct given that this will be folded in to the grid work and other refactors of these pages.

Implemented mostly with the `viewmodel.Button.outline_primary()` helper. CSS is mobile-first, tiny override for desktop (to remove the border). 

I also removed the `sign-out-link` from Eligibility Confirm, since this page isn't seen by users that are going through the Login.gov flow.

### Desktop

image

### Mobile

image

### Desktop

image

### Mobile

image


benefits/eligibility/templates/eligibility/confirm.html
```diff
@@ -1,6 +1,6 @@
 {% extends "core/page.html" %}
 
 {% block main_content %}
-  {% include "core/includes/sign-out-link.html" %}
+  <div class="container">{% include "core/includes/previous-page-link.html" %}</div>
```
thekaveman, line 4: Knowing that this will change, added a `container` wrapper for some simple alignment

machikoyasuda:
Back button functionality bug:

The back button works for me in all cases, except this one:

- Click 🔘 I have an MST Courtesy Card at `eligibility`
- Click Continue to go to `eligibility/start`
- Click Continue to go to `eligibility/confirm`
- Click <- Previous to go back to `eligibility/start` 
- Click <- Previous again to go back to `/eligibility` ---->

I expect to be taken back to `/eligibility`, but the link goes to `eligibility/start`, which is where I'm already at.

image
image

machikoyasuda:
Back button style:

1. The mobile design sets the button to be sized at 130x32.
2. Both the mobile and desktop design sets the button to be placed vertically-centered between the 70px height between the Header and the H1. The issue with the way this button is positioned is that the button row is affecting the vertical position of everything below it, including the H1. And so now, the H1 is not properly positioned at 70px below the header. Rather than setting the position with absolute pixels, we can set the CSS to figure out the vertical centering logic dynamically.

To achieve the first button sizing issue, I did two things:

1. Remove `btn-lg` class from the button. This button is not large. The `btn-lg` was adding padding that was setting the button height/width.
2. Set our own custom button height/width, but declaring this style: `padding: 4.5px 2.65px;` Added this comment to also clarify where these numbers came from: `/* Button dimensions are 130 x 32 */` I got to these numbers by playing around with the numbers in DevTools.

Then, after the button height was set properly, to address the button alignment, I:

1. Absolutely positioned the `.previous-page-row` with the same logic of the `.signout-row`. The `position: absolute;` means that this whole row div will not affect the positioning of any other elements around it.
2. I set the height of this row to 70px - which is a height that is the same across mobile and desktop.
3. I then used flexbox (`display: flex;`) to tell the button to position itself in the vertical center of its parent, by using [`align-items: center`](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-align-items). The Bootstrap classes to achieve this are `d-flex align-items-center`.
4. Then I deleted all the code that was previously positioning the button.


```diff
-<div class="row previous-page-row justify-content-center justify-content-lg-left">
-    <div class="col-12">{% include "core/includes/button.html" with button=previous_page %}</div>
+<div class="row previous-page-row">
+    <div class="col-12 d-flex align-items-center">{% include "core/includes/button.html" with button=previous_page %}</div>
```
 
```diff
 /* Custom button: Sign Out */
 
-.signout-row {
+.signout-row,
+.previous-page-row {
   position: absolute;
-  width: 100%;
   z-index: 100;
+  height: 70px;
 }
 
 .signout-row .container {
@@ -330,31 +331,27 @@ footer .footer-links li a:visited {
 
 /* Custom button: Previous Page */
 
-.previous-page-row {
-  margin-top: 19px;
-}
-
-#previous-page {
+#previous-page-button {
   border-color: var(--text-primary-color);
   color: var(--text-primary-color);
   font-weight: 500;
   font-size: calc(14rem / 16); /* 14px */
-  line-height: calc(17.5rem / 16); /* 17.5px */
-  letter-spacing: 0.05rem;
+  letter-spacing: var(--body-letter-spacing);
+  padding: 4.5px 2.65px; /* Button dimensions are 130 x 32 */
 }
 
 @media (min-width: 992px) {
-  #previous-page {
+  #previous-page-button {
     border: none;
   }
 }
 
-#previous-page::before {
+#previous-page-button::before {
   background: url("/static/img/left-arrow.svg") no-repeat;
   content: "";
   display: inline-block;
   height: calc(18rem / 16);
-  line-height: calc(18rem / 16);
   position: relative;
   vertical-align: middle;
   width: calc(18rem / 16);
```

machikoyasuda (changes requested):
One specific back button scenario doesn't seem to be working (when you go back twice in a specific way)

And some button CSS change suggestions.

machikoyasuda:
Yeah I think the Button view model is .... in need of help. 😅  😅  😅  #972 will address some of it. IMO we shouldn't have to be adding a new view model method every time we have a new button. We could simply create a class and then pass that class into the button. 

I _think_ the code above should still work even if the `btn-lg` class is applied. If not, you'll have to add an `!important` to the padding declaration: `padding: 4.5px 2.65px; !important`

benefits/static/css/styles.css
```diff
@@ -308,10 +308,11 @@ footer .footer-links li a:visited {
 
 /* Custom button: Sign Out */
 
-.signout-row {
+.signout-row,
+.previous-page-row {
```
machikoyasuda, line 312: Oops 🙏 Just checked the sign out link for regressions, and I realized I caused a regression here. Instead of having `signout-row` and `previous-page-row` use the same code here, I'd just create a separate class declaration for `previous-page-row` so this PR isn't affecting signout-row at all. Sign out button can be refactored separately from this PR.

machikoyasuda (changes requested):
image
Everything (asides from sign out button regressions) and this looks good: I think this button hover effect is still too dark and weird. The black background hover on black text obscures the text entirely. Maybe we could just remove all the `:hover` style declarations for now for this PR, and then wait for designs from Sarah? Or I could be okay with ignoring this hover part entirely til we get new designs.

thekaveman:
@machikoyasuda I think you may need to rebuild files with `bin/init.sh` - I made that change in [a43ea14](https://github.com/cal-itp/benefits/pull/1025/commits/a43ea14a1c24ef36195e850b72e101371d7a1773)

machikoyasuda:
@thekaveman Ah~ Just restarted my server and the button hover looks good 👍

machikoyasuda (approved):
💯✨🔙

Issue #989, opened 2022-10-03
Unable to build dev image


afeld:
## To Reproduce

1. Run `./bin/build.sh`, or more specifically `docker compose build dev`
2. See error

```
[+] Building 2.3s (4/4) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                           0.0s
 => => transferring dockerfile: 630B                                                                                                                                                                           0.0s
 => [internal] load .dockerignore                                                                                                                                                                              0.0s
 => => transferring context: 34B                                                                                                                                                                               0.0s
 => ERROR [internal] load metadata for docker.io/library/benefits_client:latest                                                                                                                                2.2s
 => [auth] library/benefits_client:pull token for registry-1.docker.io                                                                                                                                         0.0s
------
 > [internal] load metadata for docker.io/library/benefits_client:latest:
------
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

Presumably this is just an issue on my machine, but not immediately clear why.

## Expected behavior

Image to build successfully.

## Desktop (please complete the following information)

```
$ docker --version
Docker version 20.10.17, build 100c701
```

## Additional context

none

thekaveman:
The `client` image has to be built first, right? Does that succeed?

afeld:
Seems so. More complete output:

```
$ ./bin/build.sh
+ docker compose build --pull client
[+] Building 1.5s (14/14) FINISHED
…
 => exporting to image                                                                                                                             0.1s
 => => exporting layers                                                                                                                            0.0s
 => => writing image sha256:ae48d43297e900dd0a26a774a1e7d444bd2be9a9543aaa7b80e1dc8a86000eaf                                                       0.0s
 => => naming to docker.io/library/benefits_client:latest                                                                                          0.0s

Use 'docker scan' to run Snyk tests against images to find vulnerabilities and learn how to fix them
+ docker compose build dev
[+] Building 0.6s (4/4) FINISHED
 => [internal] load build definition from Dockerfile                                                                                               0.0s
 => => transferring dockerfile: 630B                                                                                                               0.0s
 => [internal] load .dockerignore                                                                                                                  0.0s
 => => transferring context: 34B                                                                                                                   0.0s
 => ERROR [internal] load metadata for docker.io/library/benefits_client:latest                                                                    0.5s
[...]

PR #1027, opened 2022-10-06
add workaround for BuildKit issue on Apple M1 hardware

afeld:
Fixes https://github.com/cal-itp/benefits/issues/989.

@machikoyasuda Mind confirming this solves the issue for you, without having to [disable BuildKit in the settings](https://github.com/docker/compose/issues/8449#issuecomment-1017542165)?

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -1,6 +1,14 @@
 #!/usr/bin/env bash
 set -eux
 
+# if on Apple M1
+# https://stackoverflow.com/a/65259353/358804
+if [[ $(uname -m) == 'arm64' ]]; then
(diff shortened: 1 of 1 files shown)


machikoyasuda (approved):
I am able to build the app and images properly w/ this change.





(I've noticed that either way though.... my Docker is a bit slower, esp. running all the pre-commit checks 😢 )

afeld:
> either way though.... my Docker is a bit slower

Can you elaborate? If I understand the BuildKit change correctly, that's just changing how the images are built, not how they are run, and thus this pull request should not have an impact on the performance of the latter.

afeld:
Confirmed verbally that we will look into the container performance problem separately.

Issue #1021, opened 2022-10-05
CSS Variables: Universal paragraph spacing


machikoyasuda:
- Need to write CSS to represent the following logic:

1. How much padding-top or padding-bottom when there is only one paragraph in the parent
2. How much padding-top or bottom when there are other paragraphs in the parent

Example of where we need padding: When there are 2 paragraphs in a Media Body, we need padding in between the paragraphs - 

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/3673236/194176884-c8aae87f-920b-4544-a737-1c27ab98715b.png">

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/3673236/194176987-ea956e43-139b-4b14-ad42-898cce89b41a.png">

Need to style the `<p>`s in the media list

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/issues/934#issuecomment-1269066090_

PR #1028, opened 2022-10-07
Fix: Paragraph spacing across the app

machikoyasuda:
#1021 

Adds a few complex CSS declarations to address spacing between H1/Ps:

Gives padding-top to the following:
- `h1 + p:first-of-type` - First paragraph sibling after a H1. First paragraph sibling means that the first paragraph of an H1 inside another parent doesn't count. So a first paragraph after a H1 in a `media-body` doesn't count.
- `h1 + media-list` - First `media-list` after an H1. These are for pages with an H1 followed by a Media List, and no explanatory paragraph.
- `.media-body--details p:not(:first-of-type)`: Inside a `media-body--details`, all the `<p>`'s except for the first one. 

Padding bottom to the following:
- The last <p> after an H1.

```css
/* First p immediately after an H1 */
h1 + p:first-of-type,
h1 + .media-list,
/* Last p sibling to an H1 */
.media-body--details p:not(:first-of-type) {
  padding-top: 24px;
}

/* Last p siblings to an H1 */
h1 ~ p:nth-last-of-type(1) {
  padding-bottom: 24px;
}
```

The following should be fixed:
- On Enrollment Success, Enrollment pages, there is a Media Item that has 2 paragraphs. Currently, they don't have any spacing between them. That should be fixed.
- There should be padding between H1 and media items, and the first paragraph, on all pages:

image
image

image
image

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -81,6 +81,19 @@ li {
   margin: 0;
 }
 
+/* First p immediately after an H1 */
+h1 + p:first-of-type,
+h1 + .media-list,
+/* Last p sibling to an H1 */
+.media-body--details p:not(:first-of-type) {
+  padding-top: 24px;
+}
+
+/* Last p siblings to an H1 */
+h1 ~ p:nth-last-of-type(1) {
+  padding-bottom: 24px;
+}
+
 @media (max-width: 992px) {
   /* Paragraph: Body Text, up to Small width */
   .p-sm {
@@ -552,10 +565,6 @@ footer .footer-links li a:visited {
 .container.content.help .button-back {
   margin-top: 5rem;
   margin-bottom: 5rem;
-}
-
-.container.content h1 ~ p:nth-last-of-type(1) {
-  margin-bottom: 2rem;
 }
 
 .container.content .btn-lg,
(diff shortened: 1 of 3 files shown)


benefits/static/css/styles.css
```diff
@@ -81,6 +81,19 @@ li {
   margin: 0;
 }
 
+/* First p immediately after an H1 */
+h1 + p:first-of-type,
+h1 + .media-list,
+/* Last p sibling to an H1 */
```
thekaveman, line 87: I think this comment is out of date / incorrect?

benefits/static/css/styles.css
```diff
@@ -81,6 +81,19 @@ li {
...
+.media-body--details p:not(:first-of-type) {
+  padding-top: 24px;
+}
+
+/* Last p siblings to an H1 */
+h1 ~ p:nth-last-of-type(1) {
+  padding-bottom: 24px;
+}
+
```
thekaveman, line 96: These are cool selectors to get this styling in CSS... I wonder if:

1. We could wrap these up into helper classes and somehow apply them at the correct time / template to be more declarative?
2. In absence of that, maybe at least add some comments here as to which pages these rules are targeting. E.g. it wasn't obvious to me where the `h1 + .media-list` combination actually occurs.

machikoyasuda: 1. There is no way to use adjacent/sibling-selector logic in a class unfortunately.
2. I will add comments on where these are used

PR #1030, opened 2022-10-07
Fix: Sign out row bug

### LINKED ISSUE #920: Bug: Error page bug - Log out button too low
## To Reproduce

1. Go through the Login.gov IAL2 flow with a user who is under age 65
2. See error

## Screenshots

Tablet/Mobile widths are off:
image
image

Desktop is good:
image

machikoyasuda:
closes #920 

## What this PR does
- Fixes the #920 mobile bug by setting the Sign Out button at a static height
- Creates a generic `nav-button-row` class with a height of `70px`, that both Previous Page and Sign Out button templates can use. This row is an absolutely-positioned div that starts at the bottom of the Header and does not affect H1 height/positioning.
- Positions the `Sign Out Button` using flex-box (vertically aligned, justified to the end), rather than setting a pixel padding top
- This PR also removes the padding left/right on the Header on mobile, so the Cal-ITP logo is closer to aligning with the buttons.
- This PR currently adds a few `!important` font declarations because the Sign Out button styles don't have underlines like the style guide's regular links do. Could use refactoring in the future.

## Notes
This PR should probably be merged after #1013 is merged to dev, so that the sign out row includes can be included in the new Enrollment Index page template.

We should also make sure to test the #920 scenario on desktop and mobile (test Login.gov with a non-senior).

### DIFF
--- a/benefits/core/templates/core/includes/sign-out-link.html
+++ b/benefits/core/templates/core/includes/sign-out-link.html
@@ -3,7 +3,6 @@
 
 {% if authentication %}
   {% if authentication.required and authentication.logged_in %}
-    {% endcomment %}
     <div class="w-100 position-absolute">
       <div class="container">
         <div class="row nav-button-row">
@@ -14,4 +13,4 @@
       </div>
     </div>
   {% endif %}
-  {% endcomment %}
+{% endif %}
--- a/benefits/core/templates/core/includes/sign-out-link.html
+++ b/benefits/core/templates/core/includes/sign-out-link.html
@@ -1,19 +1,17 @@
 
 {% load i18n %}
 
-{% comment %} {% if authentication %}
-{% if authentication.required and authentication.logged_in %} {% endcomment %}
-  <div class="w-100 position-absolute">
-    <div class="container">
-      <div class="row" style="height: 70px;">
-        <div class="col-12 d-flex align-items-center justify-content-end">
-          <a class="signout-link" href="{{ authentication.sign_out_route }}">
-            {% comment %} {% translate authentication.sign_out_button_label %} {% endcomment %}
-            Sign out from Login.gov
-          </a>
+{% if authentication %}
+  {% if authentication.required and authentication.logged_in %}
+    {% endcomment %}
+    <div class="w-100 position-absolute">
+      <div class="container">
+        <div class="row nav-button-row">
+          <div class="col-12 d-flex align-items-center justify-content-end">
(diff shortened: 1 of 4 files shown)


machikoyasuda:
Testing:

image
image


image
image


Unable to test the case locally where a user signs into Login.gov, but they are under age 65.

machikoyasuda:
@thekaveman Do you have a Login.gov account on Dev/Localhost that has a user under age 65? If so, can you test this PR with that user to see what the "Sorry you are not eligible retry" page looks like for you? To confirm this bug is eliminated https://github.com/cal-itp/benefits/issues/920

thekaveman:
> Do you have a Login.gov account on Dev/Localhost that has a user under age 65? 

I do... finishing review on #1023 then will look at this one.

thekaveman:
@machikoyasuda I think it looks good!

## Mobile

image

## Desktop

image

benefits/core/templates/core/includes/sign-out-link.html
```diff
@@ -3,9 +3,13 @@
 
 {% if authentication %}
   {% if authentication.required and authentication.logged_in %}
-    <div class="signout-row">
+    <div class="w-100 position-absolute">
```
thekaveman, line 6: I saw commit d3f6be0bfba44a7098f674d69c6a77901fd24390 removed the `w-100` class from the previous page link... why do we need it here?

thekaveman (approved):

benefits/core/templates/core/includes/sign-out-link.html
machikoyasuda: Here's a comparison of the Previous page button and the Sign out button:

image

(Btw, this is a neat illustration, cos it shows that without changing any CSS and just copying/pasting this Sign Out button bar into the HTML - the buttons behave nicely.)

Because the Previous Page button is aligned horizontally from the left, and HTML goes from left to right, the button is already in the right spot horizontally, without having to apply extra code, like `left: 0`. 

image

Also, if we do add a `width: 100%` to the Previous Page button parent row, it'll do this:
image

The Sign Out link button though, needs to be positioned to the right-most part of the div. So we need to set a width of 100% so the button knows where to position itself on the right.

This is where the Sign Out link button would position itself if there wasnt the `w-100` class (even if you add `width: 100% to the rest of the children):

image

thekaveman: 🤯 Thank you for this detailed explanation!!

Please write those CSS docs you were talking about and include this kind of stuff! 😍

machikoyasuda:
image
image

thekaveman (approved):

PR #1033, opened 2022-10-11
Fixes: Header

### LINKED ISSUE #783: Header: Use new Cal-ITP Benefits logo in Header, remove animation

## Acceptance Criteria

- [x] Use new Cal-ITP Benefits logo in Header
- [x] Define width/height for both Mobile and Desktop (this increases web perf)
- [x] Remove header animation

## Additional context

Part of #767

machikoyasuda:
Closes #783 

## What this PR does
- Removes the `global-header` class from the Header so that the header doesn't do this: Go to https://dev-benefits.calitp.org/help and scroll down slowly, while looking at the DOM. The CA Web Stds template waits for scrolling and adds a `scrolled` class, which hides the Header. We do not want this. 

image

We don't want any of the styles that come with `global-header` either. We are overriding the background:
image

- Uses the new logo files. These files don't have any padding to the left or right. This will improve how the alignment of the logo looks with the rest of the site on mobile:
image

- Sets the logo image width and height into the HTML. This is a recommendation from Lighthouse:

image

- Adds styles so the Español button has the proper height, width and font size.
- Sets the height of the Header to 80px and centers with CSS

## What to test for
1. The header should still be 80px height at all screen widths
2. The logos should be the same size as dev
3. The header logo should align well with the previous page button on mobile
4. The button should be 126 x 38 on all screen widths

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -254,7 +254,6 @@ footer .footer-links li a:visited {
 /* Language button */
 #header-container .btn-outline-light {
   font-size: var(--bs-body-font-size);
-  border-width: 2px;
   padding: 3.5px 29.01px; /* 126 x 38px, all screen sizes  */
 }
 


thekaveman (dismissed):
This looks good 👍

angela-tran (approved):

thekaveman (approved):

PR #1035, opened 2022-10-11
Agency Index

machikoyasuda:
closes #950 

## What this PR does
- Adds 2 new background image files
- Page elements: An h1, h2, button and the `main-content` has background image applied with CSS
- Mobile logic: The box is 32px from the bottom of the main container
- Desktop logic: The box is vertically centered in the main container. The text is `col-lg-5` width.
- The CTA button is always 250x72 at all screen sizes

## Image files
- Optimized w https://www.jpeg.io/
- File sizes:
image

## Screenshots

iPhone SE
image

iPhone 12 Max - Vertical
image

iPhone 12 Max - Horizontal
image

iPad - Horizontal
image

iPad - Vertical
image

Laptop
image

Randomly short / narrow Desktop browser
image
image


machikoyasuda:
@srhhnry How is the alignment of the text here? This is what it looks like if the text is set to be a 5-column width:
image
image

This is at 6-column width (as in 50%, half):
image

The Figma design has the text breaking like this:
image

srhhnry:
That 5 column layout is looking wonderful. I tried to replicate on the Figma screens at 1200px and can't get the 5-column to look as good because the H2 starts having weird line breaks. When I take both the h1 and h2 down to 4 column it looks a bit nicer and more balanced

image

So, let's def go for the 5 column and if other sizes need more guidance I would recommend moving toward 4 column rather than 6.

machikoyasuda:
@thekaveman @angela-tran This is ready for review. I'd like to get this up onto dev so we can test it on real mobile devices and also allow @srhhnry to easily test it.

thekaveman (approved):
Looks good to me

PR #1041, opened 2022-10-12
Fixes: Media List spacing

### LINKED ISSUE #1021: CSS Variables: Universal paragraph spacing

machikoyasuda:
fixes #1021 

This time, this PR only adds padding logic to the Media List.

## What this PR does
- Fixes the lack of padding between paragraphs in a Media Item. There are several Media Items that now have more than one paragraph. Check Enrollment and Enrollment Success.
- Fixes the lack of padding between an H1 and a Media List. This applies to Media Lists that DO NOT have a Media Title. The Eligibility Start has its own special Media Title, whereas, Enrollment and Enrollment Success do not.
- Adds 1 CSS declaration/style
- Adds comments to explain how and where this is used

## CSS Documentation

- `+` Adjacent sibling combinator https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator
- `:not` https://developer.mozilla.org/en-US/docs/Web/CSS/:not
- `:first-of-type` https://developer.mozilla.org/en-US/docs/Web/CSS/:first-of-type

1. `h1 + .media-list`: Add padding between the `h1` and the whole `.media-list`. "Add padding-top to the `.media-list` _immediately following_ the `h1`: This happens on all pages with a Media List that does NOT have a Media Title.

So it happens on Enrollment Success:
image

But it does NOT happen on Eligibility Start: Because Eligibility Start does not have a `.media-list` that is _immediately following_ the `h1`. There is another element between the `h1` and the `.media-list`:
image

1. `.media-body--details p:not(:first-of-type)`: Add padding between paragraphs in Media List. "Add padding-top to all the `p` within `.media-body--details`, except for the first `p`

This is to target Media Items that have more than 1 paragraph. This happens on both Enrollment and Enrollment Success:
image

image

### DIFF
--- a/benefits/eligibility/templates/eligibility/start.html
+++ b/benefits/eligibility/templates/eligibility/start.html
@@ -17,7 +17,7 @@
         {% block media_list %}
           {% if media|length_is:"0" %}
           {% else %}
-            <ul class="media-list mt-0 p-0 d-flex justify-content-center flex-column">
+            <ul class="media-list mx-0 p-0 d-flex justify-content-center flex-column">
               {% for item in media %}
                 <li class="media d-flex align-items-stretch w-auto">
                   <div class="media-line">{% include "core/includes/icon.html" with icon=item.icon %}</div>
--- a/benefits/eligibility/templates/eligibility/start.html
+++ b/benefits/eligibility/templates/eligibility/start.html
@@ -17,7 +17,7 @@
         {% block media_list %}
           {% if media|length_is:"0" %}
           {% else %}
-            <ul class="media-list mx-0 p-0 d-flex justify-content-center flex-column">
+            <ul class="media-list m-0 px-0 d-flex justify-content-center flex-column">
               {% for item in media %}
                 <li class="media d-flex align-items-stretch w-auto">
                   <div class="media-line">{% include "core/includes/icon.html" with icon=item.icon %}</div>
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -150,16 +150,6 @@ h1 {
   padding-top: 70px;
 }
 
-/* First paragraph immediately after an H1 gets padding-top
-h1 + p:first-of-type {
-  padding-top: 24px;
-} */
-
-/* Last paragraph after an H1 gets padding bottom
-h1 ~ p:nth-last-of-type(1) {
-  padding-bottom: 24px;
-} */
-
 /* H2 */
 /* Same sizes for all screen widths: 24px (24rem/16 = 1.5rem) */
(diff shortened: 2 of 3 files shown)


benefits/static/css/styles.css
```diff
@@ -150,6 +150,16 @@ h1 {
...
 
+/* First paragraph immediately after an H1 gets padding-top
+h1 + p:first-of-type {
+  padding-top: 24px;
+} */
+
+/* Last paragraph after an H1 gets padding bottom
+h1 ~ p:nth-last-of-type(1) {
+  padding-bottom: 24px;
+} */
```
angela-tran, line 161: Did these get commented out during testing?

benefits/eligibility/templates/eligibility/start.html
```diff
@@ -17,7 +17,7 @@ <h2 class="media-title p-sm">{{ start_sub_headline }}</h2>
         {% block media_list %}
           {% if media|length_is:"0" %}
           {% else %}
-            <ul class="media-list mt-0 p-0 d-flex justify-content-center flex-column">
+            <ul class="media-list mx-0 p-0 d-flex justify-content-center flex-column">
```
angela-tran, line 20: Maybe this doesn't matter since all this media list HTML is going to be replaced by the `media-list` include in #1013... but shouldn't this match what you have in `includes/media-list.html`?

angela-tran (changes requested):
It's really helpful that you made this PR have a smaller scope 👏 I commented on some small things to fix

benefits/static/css/styles.css
machikoyasuda: Oops my git was slow and didn't push these up: https://github.com/cal-itp/benefits/pull/1041/commits/af3d1fce09dd74f21c2a00175046c54f74b9d940

benefits/eligibility/templates/eligibility/start.html
machikoyasuda: Yeah - I think the logic should both be, "Zero margins, zero padding left/right" - so gonna change both to `m-0 px-0`

angela-tran (approved):

PR #1056, opened 2022-10-13
Style: enrollment success

### LINKED ISSUE #1039: Enrollment Success page: Final touches
Make the page look like the Figma mocks. 

1. Desktop: Design the last line to this. It's a 10-column wide large div that is left-aligned.

image

image

2. Mobile: Design this last line. This mobile alignment is trickier. Might have to write CSS to add a margin-left and then calculate exactly how much the amount would be, based on what is the diameter of the icon plus the margin.

image
image

3. Enrollment Success post-log out:
image
image

image

angela-tran:
Closes #1039

### DIFF
--- a/benefits/core/templates/core/includes/icon.html
+++ b/benefits/core/templates/core/includes/icon.html
@@ -9,5 +9,5 @@
   </p>
 {% else %}
   {% load static %}
-  <img class="icon" src="{% static icon.src %}" alt="{{ icon.alt }}" />
+  <img class="icon" width="150" height="150" src="{% static icon.src %}" alt="{{ icon.alt }}" />
 {% endif %}
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -538,6 +538,18 @@ h1 + .media-list, /* A .media-list immediately following the h1: Enrollment Succ
 /* Enrollment Success */
 /* Custom in-line log out button for Enrollment Success, in paragraph */
 
+:root {
+  --logout-paragraph-margin-left: calc(
+    var(--media-item-icon-margin) + var(--media-item-icon-size)
+  );
+}
+
+@media (min-width: 992px) {
+  :root {
+    --logout-paragraph-margin-left: 0;
+  }
+}
+
 .enrollment-success #login {
   padding: 0 0.5rem 0.3rem 0.6rem;
   display: inline-block;
@@ -551,6 +563,10 @@ h1 + .media-list, /* A .media-list immediately following the h1: Enrollment Succ
   margin: 8px auto 0 auto;
   display: block;
   width: 7rem;
+}
+
+.enrollment-success .logout-paragraph {
+  margin-left: var(--logout-paragraph-margin-left);
 }
 
 /* 404 Page */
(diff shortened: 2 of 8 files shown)


machikoyasuda:
Enrollment Success on CC Flow looks good!
image

benefits/static/css/styles.css
```diff
@@ -538,6 +538,18 @@ h1 + .media-list, /* A .media-list immediately following the h1: Enrollment, Enr
 /* Enrollment Success */
 /* Custom in-line log out button for Enrollment Success, in paragraph */
 
+:root {
+  --logout-paragraph-margin-left: calc(
+    var(--media-item-icon-margin) + var(--media-item-icon-size)
```
machikoyasuda, line 543: 🧮 🎉

angela-tran: Thank you for your help on that!! 🧮

angela-tran:
Working on the logged_out page now

angela-tran:
I'm not sure how we want to achieve this design where 
 - mobile has some spacing between the two sentences in the H1
 - desktop has no spacing

 vs.

angela-tran:
Re: https://github.com/cal-itp/benefits/pull/1056#issuecomment-1276914395
@machikoyasuda I think maybe we could have two .PO entries, one for each sentence. They would be within one H1 element, and for the spacing in between, maybe we could have a `span` or something that has mobile-only styles? I'm not sure exactly what those styles would be.

What do you think? Do you have other ideas on how to do this?

machikoyasuda:
> Re: [#1056 (comment)](https://github.com/cal-itp/benefits/pull/1056#issuecomment-1276914395) @machikoyasuda I think maybe we could have two .PO entries, one for each sentence. They would be within one H1 element, and for the spacing in between, maybe we could have a `span` or something that has mobile-only styles? I'm not sure exactly what those styles would be.
> 
> What do you think? Do you have other ideas on how to do this?

```html
    <h1 class="pb-lg-8 pb-4">
      <span class="h2 d-block d-lg-inline">You have successfully logged out.</span>
      <span class="h2 d-block d-lg-inline">Thank you for using Cal-ITP Benefits!</span>
    </h1>
```

image
image

Technically the `d-block d-lg-inline` on the second <span> is not necessary though. Could just do the 1st one.

benefits/core/templates/core/includes/icon.html
```diff
@@ -9,5 +9,5 @@
   </p>
 {% else %}
   {% load static %}
-  <img class="icon" src="{% static icon.src %}" alt="{{ icon.alt }}" />
+  <img class="icon" width="150px" height="150px" src="{% static icon.src %}" alt="{{ icon.alt }}" />
```
machikoyasuda, line 12: Don't need the px here

angela-tran: Edited my commit to remove `px` from these two attribute values

machikoyasuda (approved):
Tested Login.gov flow, CC flow locally on Desktop and Mobile widths.

machikoyasuda:
From now on: Whenever we make changes to `base.html` we should also double-check we aren't causing any regressions on any pages that use `base.html`.

Issue #874, opened 2022-08-24
Send eligibility type with the started/finished eligibility events


thekaveman:
With the `started eligibility` and `returned eligibility` events, let's also send the `eligibility` that was checked

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/issues/342#issuecomment-1219762888_

PR #1057, opened 2022-10-13
Improvements to Eligibility events

### LINKED ISSUE #342: Add event for verifier selection
Soon users will have multiple pathways to verify eligibility, and it would be useful to know how many users are taking which pathway(s) in our analytics.

Once we have multiple verifier pathways, there will be UI to support the selection of pathway. We should have events in response to this new UI.

Linking this Amplitude reference again: https://help.amplitude.com/hc/en-us/articles/115002380567-User-properties-and-event-properties

## Tasks

* [ ] Build on #875 
* [ ] Create a new custom Event `"selected verifier"`, following [an existing Custom event](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/analytics.py#L13) as an example
* [ ] Create a helper to wrap sending of the new event; follow [an existing helper](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/analytics.py#L26) as an example
* [ ] Use the new helper to send the event during `POST` to `eligibility:index` if the form is valid

### LINKED ISSUE #926: Add session eligibility to base analytic event
[#924]... makes me think that we should be adding session eligibility to the base event too, rather than just where I added it in #891.

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/924#pullrequestreview-1112982682_

thekaveman:
Closes #342 
Closes #926

### New event for `selected eligibility verifier`

This event is sent when the user submits the form on `/eligibility/index` page (the "benefit selection" form). It includes an event property `eligibility_verifier` identifying which was selected.

**Note** the event is _not sent_ when there is only a single `EligibilityVerifier` configured in the app. In this case, the user is never presented with the selection form. This is an unlikely scenario given that we are launching the new events with the Courtesy Card feature and alongside the existing Older Adults pathway.

### `eligibility_types` on all events

This PR adds the `eligibility_types` event property (a list of zero or more type names) for _all analytics events_ by adding it to the base `core.analytics.Event` class. The value is initially pulled from the user's session.

Events using the `eligibility.analytics.EligibilityEvent` as a base class override the default session values (based on the work in #874).

### DIFF
--- a/benefits/eligibility/analytics.py
+++ b/benefits/eligibility/analytics.py
@@ -13,6 +13,14 @@ class EligibilityEvent(core.Event):
         self.update_event_properties(eligibility_types=eligibility_types)
 
 
+class SelectedVerifierEvent(EligibilityEvent):
+    """Analytics event representing the user selecting an eligibility verifier."""
+
+    def __init__(self, request, verifier, eligibility_types):
+        super().__init__(request, "selected eligibility verifier", eligibility_types)
+        self.update_event_properties(eligibility_verifier=verifier)
+
+
 class StartedEligibilityEvent(EligibilityEvent):
     """Analytics event representing the beginning of an eligibility verification check."""
 
@@ -30,6 +38,11 @@ class ReturnedEligibilityEvent(EligibilityEvent):
             self.update_event_properties(status=status, error=error)
         if status == "success":
             self.update_user_properties(eligibility_types=eligibility_types)
+
+
+def selected_verifier(request, verifier, eligibility_types):
+    """Send the "selected eligibility verifier" analytics event."""
+    core.send_event(SelectedVerifierEvent(request, verifier, eligibility_types))
 
 
 def started_eligibility(request, eligibility_types):
--- a/benefits/eligibility/analytics.py
+++ b/benefits/eligibility/analytics.py
@@ -9,6 +9,7 @@ class EligibilityEvent(core.Event):
 
     def __init__(self, request, event_type, eligibility_types):
         super().__init__(request, event_type)
+        # overwrite core.Event eligibility_types
         self.update_event_properties(eligibility_types=eligibility_types)
 
 
--- a/benefits/eligibility/analytics.py
+++ b/benefits/eligibility/analytics.py
@@ -25,8 +25,11 @@ class ReturnedEligibilityEvent(EligibilityEvent):
 
     def __init__(self, request, eligibility_types, status, error=None):
         super().__init__(request, "returned eligibility", eligibility_types)
-        if str(status).lower() in ("error", "fail", "success"):
+        status = str(status).lower()
+        if status in ("error", "fail", "success"):
             self.update_event_properties(status=status, error=error)
+        if status == "success":
+            self.update_user_properties(eligibility_types=eligibility_types)
 
 
 def started_eligibility(request, eligibility_types):
--- a/benefits/eligibility/analytics.py
+++ b/benefits/eligibility/analytics.py
@@ -16,9 +16,8 @@ class EligibilityEvent(core.Event):
 class SelectedVerifierEvent(EligibilityEvent):
     """Analytics event representing the user selecting an eligibility verifier."""
 
-    def __init__(self, request, verifier, eligibility_types):
+    def __init__(self, request, eligibility_types):
         super().__init__(request, "selected eligibility verifier", eligibility_types)
-        self.update_event_properties(eligibility_verifier=verifier)
 
 
 class StartedEligibilityEvent(EligibilityEvent):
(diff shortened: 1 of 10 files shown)


benefits/eligibility/analytics.py
```diff
@@ -25,8 +25,11 @@ class ReturnedEligibilityEvent(EligibilityEvent):
 
     def __init__(self, request, eligibility_types, status, error=None):
         super().__init__(request, "returned eligibility", eligibility_types)
-        if str(status).lower() in ("error", "fail", "success"):
+        status = str(status).lower()
+        if status in ("error", "fail", "success"):
             self.update_event_properties(status=status, error=error)
+        if status == "success":
+            self.update_user_properties(eligibility_types=eligibility_types)
```
angela-tran, line 40: Is the thought here that we want to keep track of the user's most recently confirmed eligibility type? (I just read through [User-properties-and-event-properties](https://help.amplitude.com/hc/en-us/articles/115002380567-User-properties-and-event-properties) to refresh on how user properties work)

benefits/eligibility/views.py
```diff
@@ -64,6 +64,9 @@ def index(request):
             verifier = EligibilityVerifier.objects.get(id=verifier_id)
             session.update(request, verifier=verifier)
 
+            types_to_verify = agency.type_names_to_verify(verifier)
+            analytics.selected_verifier(request, verifier.name, types_to_verify)
```
angela-tran, line 68: Since `benefits.core.analytics.Event` updates the [event properties](https://github.com/cal-itp/benefits/pull/1057/files#diff-caee5b8a703ead64243ea99814f85560027016eef535273ace62604eaa148101R56) and [user properties](https://github.com/cal-itp/benefits/pull/1057/files#diff-caee5b8a703ead64243ea99814f85560027016eef535273ace62604eaa148101R71) with the session verifier name, is it redundant to pass in `verifier.name` here?

tests/pytest/eligibility/test_analytics.py
```diff
@@ -0,0 +1,36 @@
...
+    key, type1, type2 = "eligibility_types", "type1", "type2"
+    mocker.patch("benefits.core.analytics.session.eligibility", return_value=[type1])
+    mocker.patch("benefits.core.analytics.EligibilityType.get_names", return_value=[type1])
+
+    event = ReturnedEligibilityEvent(app_request, eligibility_types=[type2], status="success")
+
+    # user_properties should have been overwritten
+    assert key in event.user_properties
+    assert type1 not in event.user_properties[key]
+    assert type2 in event.user_properties[key]
```
angela-tran, line 36: These tests are great! Should we have a test for `SelectedVerifierEvent`?

benefits/eligibility/analytics.py
thekaveman: Good question!

My thinking with the user prop here is we may want to segment our users based on which eligibility they were verified for. Note this is different from counting the number of successful eligibility verifications, which is _event_ based and not _user_ based.

The event will also have a prop with the list of verified type(s), so we can always do that level of analysis as well.

benefits/eligibility/views.py
thekaveman: Especially that we have a test to ensure the base event grabs the `verifier`... I think you're right, it can be removed.

tests/pytest/eligibility/test_analytics.py
thekaveman: I knew there was something I forgot to do in this PR 😄 Incoming shortly...

benefits/eligibility/views.py
thekaveman: Done in [be44936](https://github.com/cal-itp/benefits/pull/1057/commits/be44936249cb20225239d6ce104f3aa0c64d976c)

tests/pytest/eligibility/test_analytics.py
angela-tran: Hmm, good point. I guess my feeling was that introducing a new event should require a new test or assertion somewhere... maybe we could update this test: [test_index_post_valid_form](https://github.com/cal-itp/benefits/blob/dev/tests/pytest/eligibility/test_views.py#L122) with `mocked_analytics_module.selected_verifier.assert_called_once()`?

thekaveman: Yep I'm with you, good point. Will add that in the view test.

thekaveman: See [bf8c1cd](https://github.com/cal-itp/benefits/pull/1057/commits/bf8c1cd747dd2b917835c2d6bd0151af9fe9f0ac)

angela-tran (approved):
🎉

Issue #835, opened 2022-07-27
HTML Template Clean up


machikoyasuda:
Fix the following HTML template linter rules, and then turn the rule back on.

## Acceptance Criteria

- [ ] H017: Tag should be self closing
- [x] H021: Inline styles should be avoided (maybe fix with front-end cleanups)
- [x] H025: Tag seems to be an orphan (https://github.com/cal-itp/benefits/issues/775 may fix this)
- [ ] H031: Consider adding meta keywords (we can look at doing this too)

## Additional context

#834 
#816

Issue #834, opened 2022-07-27
HTML Refactor: Add height and width attributes to all images


machikoyasuda:
Should make a ticket to fix this in the future: `H006: img tag should have height and width attributes` and then re-add the rule. Adding width/height attribute to images is also something that the Google Lighthouse test tests for, and is good for web performance/page rendering speeds.

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/issues/816#issuecomment-1191831066_

Related to #816

machikoyasuda:
Done with #1033

PR #1064, opened 2022-10-13
Audit: Run the site through Lighthouse, HTML5 Validator

### LINKED ISSUE #935: Test site with Google Lighthouse and make all changes necessary
Test site for a11y, web perf, etc. Get it up to 100%.

## Acceptance Criteria

- [x] #834

machikoyasuda:
closes #935 

## What this PR does
- Adds image dimensions to HTML as numbers, not pixels. Pixels is wrong.
- Changes the lang selector from a span to a div. It should not be a span.
- Adds back 2 linter rules that we are not 100% compliant with.

## Why
- Should address all ERROR level feedback from HTML5 validator: https://validator.w3.org/nu/?doc=https%3A%2F%2Fdev-benefits.calitp.org%2Fmst

image

- Should address/attempt to address all Issues on Google Chrome DevTools: https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1665691993150-92376.report.html

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -59,13 +59,13 @@
           <span class="navbar-brand p-0">
             <img class="sm d-lg-none"
                  src="{% static "img/logo-sm.svg" %}"
-                 width="90px"
-                 height="51.3px"
+                 width="90"
+                 height="51.3"
                  alt="{% translate "core.logos.small" context "image alt text" %}"/>
             <img class="lg d-none d-lg-block"
                  src="{% static "img/logo-lg.svg" %}"
-                 width="220px"
-                 height="50px"
+                 width="220"
+                 height="50"
                  alt="{% translate "core.logos.large" context "image alt text" %}"/>
           </span>
           <span class="form-inline">{% include "core/includes/lang-selector.html" %}</span>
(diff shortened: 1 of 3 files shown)


thekaveman (commented):
I think we should add back the [`djlint` rule](https://github.com/cal-itp/benefits/blob/dev/pyproject.toml#L16) to check for these, right?

#834 called for the image rule to be removed from the ignore list.

machikoyasuda:
> I think we should add back the [`djlint` rule](https://github.com/cal-itp/benefits/blob/dev/pyproject.toml#L16) to check for these, right?
> 
> #834 called for the image rule to be removed from the ignore list.

Oh yeah definitely. I can un-ignore the image rule and the in-line style rule now I think. Not sure about the others.

machikoyasuda:
Update: Removes 3 ignored rules. 

```diff
[tool.djlint]
- ignore = "H017,H025,H031"
+ ignore = "H017,H031"
```

H031 is a rule about having meta keywords. H017 is about whether to add trailing slashes or not (which are not necessary according to W3C Validator https://github.com/validator/validator/wiki/Markup-%C2%BB-Void-elements#trailing-slashes-in-void-element-start-tags-do-not-mark-the-start-tags-as-self-closing - so I left it alone.)

thekaveman:
FYI #1067 will take care of that error at the top of your screenshot

machikoyasuda:
Updated this ticket https://github.com/cal-itp/benefits/issues/835 to remove the 2 rules added back in this PR

thekaveman (approved):
LGTM!

PR #1065, opened 2022-10-13
Feat: Add numeric keyboard/maxlength 5 to sub input

### LINKED ISSUE #1060: Courtesy Card form: Add a 5 char max-length, numeric keyboard
As a MST CC owner who is using a mobile phone, when I am filling out the Eligibility Verification form on my phone, I'd like the numeric keyboard to show up rather than the default keyboard.

As a MST CC owner, when I am filling out the Eligibility Verification on my phone or on my computer, I'd like there to be a 5-character limit on the form field for Courtesy Card (b/c the CC number is limited to 5 characters), so I don't accidentally send in more characters.

## Acceptance Criteria

- [ ] As a MST CC owner who is using a mobile phone, when I am filling out the Eligibility Verification form on my phone, I'd like the numeric keyboard to show up rather than the default keyboard.
- [ ] As a MST CC owner, when I am filling out the Eligibility Verification on my phone or on my computer, I'd like there to be a 5-character limit on the form field for Courtesy Card (b/c the CC number is limited to 5 characters), so I don't accidentally send in more characters.
- [ ] Changes must be generic and set on the MST CC Verifier data

## Additional context

This is the code necessary to make these changes:
image

But these changes will have to be made in a way where the specific Verifier can specify their own form needs, like the way the `form_sub_pattern` is set in the migration data: `sub_widget.attrs.update({"pattern": verifier.form_sub_pattern})`

Note: The mobile phone part will need XCode to simulate the keyboard opening up to test.

machikoyasuda:
closes #1060 

## What this PR does
- Allow the user to use numeric keyboard for the sub input on mobile
- Add a 5-character max length for the sub input

### Model changes
- Adds `form_input_mode` (String), `form_max_length` (Integer)
- Form input mode must be one of these strings: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode#values
- Max length: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/maxlength

### Database migration changes
- Sets `form_inputmode` to `numeric` and `form_max_length` to 5 for the MST_CC sample data eligibility verifier

## How to test
- Test that the rest of the forms don't have any regressions
- Use iOS Simulator (on XCode on Mac) (or any other Windows tool?) to see if the numeric input opens on the sub field
- Test that the regex validation still works

- Regex still works:
image

- iOS keyboard: Numeric for sub, regular for name
image
image

## Post-merge
- Update migrations sample data to include these 2 new key/values:

```python
        form_input_mode="numeric",
        form_max_length=5,
```

### DIFF
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -108,7 +108,7 @@ class EligibilityVerifier(models.Model):
     # A regular expression used to validate the 'sub' API field before sending to this verifier
     form_sub_pattern = models.TextField(null=True)
     # Input mode can be "numeric", "tel", "search", etc. to override default "text" keyboard on mobile devices
-    form_inputmode = models.TextField(null=True)
+    form_input_mode = models.TextField(null=True)
     # The maximum length accepted for the 'sub' API field before sending to this verifier
     form_max_length = models.PositiveSmallIntegerField(null=True)
     form_name_label = models.TextField(null=True)
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -107,6 +107,10 @@ class EligibilityVerifier(models.Model):
     form_sub_placeholder = models.TextField(null=True)
     # A regular expression used to validate the 'sub' API field before sending to this verifier
     form_sub_pattern = models.TextField(null=True)
+    # Input mode can be "numeric", "tel", "search", etc. to override default "text" keyboard on mobile devices
+    form_inputmode = models.TextField(null=True)
+    # The maximum length accepted for the 'sub' API field before sending to this verifier
+    form_max_length = models.PositiveSmallIntegerField(null=True)
     form_name_label = models.TextField(null=True)
     form_name_help_text = models.TextField(null=True)
     form_name_placeholder = models.TextField(null=True)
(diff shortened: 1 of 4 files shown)


benefits/core/models.py
```diff
@@ -107,6 +107,10 @@ class EligibilityVerifier(models.Model):
     form_sub_placeholder = models.TextField(null=True)
     # A regular expression used to validate the 'sub' API field before sending to this verifier
     form_sub_pattern = models.TextField(null=True)
+    # Input mode can be "numeric", "tel", "search", etc. to override default "text" keyboard on mobile devices
+    form_inputmode = models.TextField(null=True)
```
thekaveman, line 111: This is being pedantic, but it's weird to have one of these not use the full `snake_case` pattern. 

```suggestion
    form_input_mode = models.TextField(null=True)
```

machikoyasuda: Yeah I thought about that too - BUT - HTML uses `inputmode`: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode so that's why I chose `inputmode`.

thekaveman: But it also uses `maxlength` and that one is `form_max_length` soooo...?

machikoyasuda: Trueeee ok

angela-tran (approved):
This is awesome!

angela-tran:
⚠️ Note: this will require data migration updates (see cal-itp/benefits-secrets#18)

thekaveman:
> ⚠️ Note: this will require data migration updates in Azure

@machikoyasuda let's make sure the dev migrations are updated in Azure (see the linked PR) _before_ this is merged.

PR #1071, opened 2022-10-14
Fix: reCAPTCHA + HTML5 form validation

### LINKED ISSUE #1068: reCAPTCHA is bypassing HTML5 form validation
reCAPTCHA is allowing the form to submit even when required fields are not provided:

no-field-validation

This SO post from 2017 talks about the same issue: https://stackoverflow.com/questions/41665935/html5-form-validation-before-recaptchas

## To Reproduce

1. Go to `dev`
2. Don't choose any Verifier and submit the form - see no form validation
3. Choose Courtesy Cards
4. Do not enter any CC#/Name and submit the form - see no form validation

## Expected behavior

The browser's default HTML5 field validation for e.g. required fields should kick in and not allow the form to be submitted with invalid fields.

thekaveman:
Fixes #1068 

Based on an answer to this SO post: https://stackoverflow.com/q/41665935/453168

### Uses programmatic reCAPTCHA validation

This is the main issue being solved. reCAPTCHA v3 supports two options:

* [Automatically bind the challenge to a button](https://developers.google.com/recaptcha/docs/v3#automatically_bind_the_challenge_to_a_button) (existing implementation)

* [Programmatically invoke the challenge](https://developers.google.com/recaptcha/docs/v3#programmatically_invoke_the_challenge) (new implementation in this PR)

Using a programmatic challenge allows the browser's HTML5 form validation a chance to kick-in before reCAPTCHA's checks,
which isn't possible with the challenge button auto-bind.

### Refactors how templates get reCAPTCHA data

Before we used a [Django context processor](https://docs.djangoproject.com/en/4.1/ref/templates/api/#writing-your-own-context-processors), which gave the data to _every request context_, even those that didn't need it:

```python
def recaptcha(request):
    """Context processor adds recaptcha information to request context."""
    return {
        "recaptcha": {
            "api_url": settings.RECAPTCHA_API_URL,
            "enabled": settings.RECAPTCHA_ENABLED,
            "site_key": settings.RECAPTCHA_SITE_KEY,
        }
    }
```

This PR introduces a Middleware, following our usual pattern where we can apply the middleware per-view that needs it, which adds the reCAPTCHA data directly to the `request`:

```python
class RecaptchaEnabled(MiddlewareMixin):
    """Middleware configures the request with required reCAPTCHA settings."""

    def process_request(self, request):
        if settings.RECAPTCHA_ENABLED:
            request.recaptcha = {
                "data_field": recaptcha.DATA_FIELD,
                "script_api": settings.RECAPTCHA_API_KEY_URL,
                "site_key": settings.RECAPTCHA_SITE_KEY,
            }
        return None

# later in views.py

@decorator_from_middleware(RecaptchaEnabled)
def confirm(request):
    """View handler for the eligibility verification form."""
    # etc.
```

This makes it more explicit which views use reCAPTCHA, and also makes using reCAPTCHA data easy from templates. Like in `base.html` where we include the reCAPTCHA API library _only once_ and _only if needed_:

```html
<html>
    <body>
        <!-- etc -->

        {% if request.recaptcha %}<script src="{{ request.recaptcha.script_api }}"></script>{% endif %}
    </body>
</html>
```

### Makes the `eligibility:unverified` path a redirect

This has a couple advantages:

* (the reason for this change): creates a _new request_, clearing reCAPTCHA data from prior `/eligiblity/confirm` request
* Mirrors how the `eligibility:verified` path works (ultimately a redirect to `enrollment:index`)
* URL is now `/eligiblity/unverified` instead of `/eligibility/confirm`, a more accurate UX

### Setup

Add this to your `.vscode/launch.json` under the `Django: Benefits Client, Debug=False` entry's `env`:

```jsonc
{
  "name": "Django: Benefits Client, Debug=False",
  // other fields...
  "env": {
    // existing fields...
   // add these 2 entries with the values for the test reCAPTCHA site
    "DJANGO_RECAPTCHA_SITE_KEY": "<SITE KEY HERE>",
    "DJANGO_RECAPTCHA_SECRET_KEY": "<SECRET KEY HERE>"
  }
}
```

### Eligibility index

reCAPTCHA library included once at the bottom of `<body>`:

image

HTML5 validation works, form submits with valid data:

choose-benefit-validation

### Eligibility confirm

reCAPTCHA library included once at the bottom of `<body>`:

image

HTML5 validation works, form submits with valid data:

confirm-validation

### Other pages

Every other page should not have the reCAPTCHA library or any reCAPTCHA handling for form elements. 

E.g. the homepage does not include reCAPTCHA:

image

Nor does the `eligibility:start` page:

image

### DIFF
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -36,14 +36,3 @@ def authentication(request):
 def debug(request):
     """Context processor adds debug information to request context."""
     return {"debug": session.context_dict(request)}
-
-
-def recaptcha(request):
-    """Context processor adds recaptcha information to request context."""
-    return {
-        "recaptcha": {
-            "api_url": settings.RECAPTCHA_API_URL,
-            "enabled": settings.RECAPTCHA_ENABLED,
-            "site_key": settings.RECAPTCHA_SITE_KEY,
-        }
-    }
--- a/benefits/core/middleware.py
+++ b/benefits/core/middleware.py
@@ -13,7 +13,7 @@ from django.utils.decorators import decorator_from_middleware
 from django.utils.deprecation import MiddlewareMixin
 from django.views import i18n
 
-from . import analytics, session, viewmodels
+from . import analytics, recaptcha, session, viewmodels
 
 
 logger = logging.getLogger(__name__)
@@ -173,4 +173,17 @@ class LogErrorToAzure(MiddlewareMixin):
         msg = getattr(exception, "message", repr(exception))
         self.azure_logger.exception(msg, exc_info=exception)
 
+        return None
+
+
+class RecaptchaEnabled(MiddlewareMixin):
+    """Middleware configures the request with required reCAPTCHA settings."""
+
+    def process_request(self, request):
+        if settings.RECAPTCHA_ENABLED:
+            request.recaptcha = {
+                "data_field": recaptcha.DATA_FIELD,
+                "script_api": settings.RECAPTCHA_API_KEY_URL,
+                "site_key": settings.RECAPTCHA_SITE_KEY,
+            }
         return None
--- a/benefits/core/recaptcha.py
+++ b/benefits/core/recaptcha.py
@@ -6,7 +6,7 @@ import requests
 from django.conf import settings
 
 
-_POST_DATA = "g-recaptcha-response"
+DATA_FIELD = "g-recaptcha-response"
 
 
 def has_error(form) -> bool:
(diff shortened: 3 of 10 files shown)


machikoyasuda (approved):
Really like the way the JS in form.html is cleaned up! I also really like the instructions/documentation on how to run recaptcha locally - this will be very helpful.

I added the note comment there just to make sure we never make a CSS ID that collides with the ID name of a form. (ie - Ensuring we never make another `#verifier-selection` ID).

Once this PR is merged, I believe the `unverified` page _should_ have the proper new layout that is on `dev` already.

thekaveman:
> I added the note comment there just to make sure we never make a CSS ID that collides with the ID name of a form

That's a good point - we could prefix these new form IDs with `form-` to make that more clear and less likely for a collision?

angela-tran (approved):
Nice refactoring!

machikoyasuda:
> > I added the note comment there just to make sure we never make a CSS ID that collides with the ID name of a form
> 
> That's a good point - we could prefix these new form IDs with `form-` to make that more clear and less likely for a collision?

Yeah. This could be done in another PR.

angela-tran (approved):

PR #1076, opened 2022-10-21
chore(Terraform): remove duplicative config

afeld:
Not necessary for the pipeline to run, and could lead to issues.


afeld:
Hmm, I guess those are required. Never mind.

Issue #829, opened 2022-07-26
make it harder to destroy things in Terraform


afeld:
Currently, the infrastructure pipeline will `apply` Terraform changes no matter what those changes are doing. This can lead to costly mistakes, such as [the downtime on 7/21](https://docs.google.com/document/d/14XFpJonXmrNKDt98g8vML3RimbZha0UJiIeUVclXYGs/edit#).

## To Reproduce

1. Delete a resource in a Terraform file
2. Merge that change to `dev`
3. See that it will `apply` without warning/complaint

## Expected behavior

Should need to do an extra step / emphasize "yes I'm _really_ sure" before resources get destroyed.

## Additional context

Not-mutually-exclusive ways to make this safer:

- [Lifecycle arguments](https://www.terraform.io/language/meta-arguments/lifecycle#syntax-and-arguments) like `prevent_destroy`
- [Requiring manual approval for destroying resources](https://www.mediaglasses.blog/2021/06/08/azure-devops-terraform-pipeline-with-checkov-approvals/)

afeld:
Less important if we fully isolate all the environments: https://github.com/cal-itp/benefits/issues/830

PR #1074, opened 2022-10-20
separate infrastructure environments

afeld:
Closes #830. Running list of TODOs:

- [x] Update docs around environments, removing references to "slots"
- [x] Update deployment pipeline
  - [x] Ensure it's connecting to the right subscription/backend
- [ ] DevSecOps:

  - [x] [Recover or purge Key Vault](https://learn.microsoft.com/en-us/azure/key-vault/general/key-vault-recovery?tabs=azure-portal#list-recover-or-purge-a-soft-deleted-key-vault) `KV-CDT-PUB-CALITP-D-001`
    - Our permissions are limited on Key Vaults so have been trying to work around that
  - [x] Create VNet and Subnet(s) for dev and test
    - [x] Provide VNet and Subnet names, and the Resource Group they're in
    - [ ] Give read permissions:

      > Error: retrieving Subnet: (Name "SNET-CDT-PUB-CALITP-D-001" / Virtual Network Name "VNET-CDT-PUB-SHRD-W-D-001" / Resource Group "RG-CDT-PUB-SHRD-W-D-001"): network.SubnetsClient#Get: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="AuthorizationFailed" Message="The client 'live.com#aidan@compiler.la' with object id \[REDACTED\] does not have authorization to perform action 'Microsoft.Network/virtualNetworks/subnets/read' over scope '/subscriptions/\[REDACTED\]/resourceGroups/RG-CDT-PUB-SHRD-W-D-001/providers/Microsoft.Network/virtualNetworks/VNET-CDT-PUB-SHRD-W-D-001/subnets/SNET-CDT-PUB-CALITP-D-001' or the scope is invalid.

  - [x] Create `Test` Resource Group
  - [x] Create Azure DevOps Service Connection to `Development` Subscription
  - [x] Configure Front Door
  - [ ] Clean up unused resources in `Development` Resource Group:
    - [ ] Storage account `sacalitpd001`
    - [x] App Service Plan `ASP-RGCDTPUBVIPCALITPD001-9628`
  - [x] Provide Active Directory Group ID

- [x] [Rename resources](https://developer.hashicorp.com/terraform/language/modules/develop/refactoring) to:
  - Prevent them from being destroyed and recreated
  - Make the `plan` output shorter and thus easier to review
- [x] Import `test` resources
- [ ] Update DNS
  - [ ] Set hostname verification `TXT` record
    - [x] dev
    - [ ] test
[...]

PR #1085, opened 2022-10-26
prevent destruction of Terraform resources containing secrets

afeld:
Split out from https://github.com/cal-itp/benefits/pull/1074. Closes https://github.com/cal-itp/benefits/issues/829.


angela-tran (approved):
💯 for updating the docs

PR #821, opened 2022-07-21
Terraform: add slot storage mounts

thekaveman:
Closes #686 

Changes already made manually in the UI.

PR #1086, opened 2022-10-26
manage production storage account binding in Terraform

afeld:
It was happening on dev and test as of https://github.com/cal-itp/benefits/pull/821; this adds it to production.

- [ ] Confirm the `plan` shows no changes


angela-tran (dismissed):

afeld:
We think the `plan` is showing changes to the `access_key`, which should be harmless. Merging, but will watch the `apply` and confirm prod stays up.

angela-tran (approved):

PR #1043, opened 2022-10-12
Feat: Add spinner to Check Status button

machikoyasuda:
closes #948 

## What this PR does
- Add Spinner to button  _if_ the button has a `form.submitting_value`. 
- The Eligibility Verification form has a `submitting_value` (`Checking`), so the spinner shows up there. No other place in the current app has a form with a `submitting_value`.

## Button with Spinner element

image

### Spinner:
- 3px border width
- 16x16 total size
- 15px margin left from the text
- `role="status"` for a11y
- Comes from Bootstrap: https://getbootstrap.com/docs/5.1/components/spinners/

image

### Button with Spinner:
- `<button>` with two `<span>`s
- One span for text, one span for spinner. (If these are in the _same_ span, then the whole text would also rotate).

## Testing
- Test all other buttons for height/size
- Test with keyboard
- A11y check: Does this new spinner affect a11y at all? If so how? How to mitigate?
- All other CTA buttons (besides Agency Index home page button) should have height 60px
image
image

PR #1088, opened 2022-10-26
Feat: spinner on Check Status button

### LINKED ISSUE #948: Button: Add animation to the Check/Checking button

thekaveman:
Closes #948

Rebased @machikoyasuda's work from #1043 on the new reCAPTCHA fixes. 

Also fixed the ordering of the jQuery function calls to ensure the `disabled` state is applied to the _button_ and not the _text_ element inside.

## Demo

spinner

### DIFF
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -35,8 +35,9 @@
           </div>
         {% endif %}
         <div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">
-          <button class="btn btn-lg btn-primary" form="{{ form.id }}" data-action="submit" type="submit">
-            {{ form.submit_value }}
+          <button class="btn btn-lg btn-primary spinner-hidden" form="{{ form.id }}" data-action="submit" type="submit">
+            <span class="btn-text">{{ form.submit_value }}</span>
+            <span class="spinner-border spinner-border-sm" role="status"></span>
           </button>
         </div>
       </div>
@@ -47,10 +48,11 @@
         $("#{{ form.id }}").on("submit", function(e) {
           if ("{{ form.submitting_value }}" !== "") {
             $("button[type=submit]", this)
-              .attr("role", "status")
-              .attr("disabled", "true")
-              .text("{{ form.submitting_value }}")
-              .addClass("disabled");
+                .removeClass("spinner-hidden")
+                .children().first().text("{{ form.submitting_value }}")
+                .attr("role", "status")
+                .attr("disabled", "true")
+                .addClass("disabled");
           }
         });
       });
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -35,10 +35,12 @@
           </div>
         {% endif %}
         <div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">
-            <button class="btn btn-lg btn-primary spinner-hidden d-flex justify-content-center align-items-center" data-action="submit" type="submit">
-                <span class="btn-text">{{ form.submit_value }}</span>
-                <span class="spinner-border spinner-border-sm" role="status"></span>
-            </button>
+          <button class="btn btn-lg btn-primary spinner-hidden d-flex justify-content-center align-items-center"
+                  data-action="submit"
+                  type="submit">
+            <span class="btn-text">{{ form.submit_value }}</span>
+            <span class="spinner-border spinner-border-sm" role="status"></span>
+          </button>
         </div>
       </div>
     {% endif %}
@@ -48,11 +50,11 @@
         $("#{{ form.id }}").on("submit", function(e) {
           if ("{{ form.submitting_value }}" !== "") {
             $("button[type=submit]", this)
-                .removeClass("spinner-hidden")
-                .children().first().text("{{ form.submitting_value }}")
-                .attr("role", "status")
-                .attr("disabled", "true")
-                .addClass("disabled");
+              .removeClass("spinner-hidden")
+              .addClass("disabled")
+              .attr("role", "status")
+              .attr("disabled", "true")
+              .children().first().text("{{ form.submitting_value }}");
           }
         });
       });
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -35,10 +35,10 @@
           </div>
         {% endif %}
         <div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">
-          <button class="btn btn-lg btn-primary spinner-hidden" form="{{ form.id }}" data-action="submit" type="submit">
+        <button class="btn btn-lg btn-primary spinner-hidden d-flex justify-content-center align-items-center" data-action="submit" type="submit">
             <span class="btn-text">{{ form.submit_value }}</span>
-            <span class="spinner-border spinner-border-sm" role="status"></span>
-          </button>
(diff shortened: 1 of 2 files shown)


benefits/core/templates/core/includes/form.html
```diff
@@ -47,10 +50,11 @@
         $("#{{ form.id }}").on("submit", function(e) {
           if ("{{ form.submitting_value }}" !== "") {
             $("button[type=submit]", this)
+              .removeClass("spinner-hidden")
+              .addClass("disabled")
               .attr("role", "status")
```
machikoyasuda, line 55: I wonder if this is necessary if `<span class="spinner-border spinner-border-sm" role="status"></span>` already has `role="status"`? I was not sure.

angela-tran (commented):
This looks good to me. I think @machikoyasuda is right about not needing to call `.attr("role", "status")`.

benefits/core/templates/core/includes/form.html
machikoyasuda: After reading this https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role, I don't think we should have two elements with `role=status` b/c it would announce the same status twice.

machikoyasuda (commented):
I'm testing this locally on my machine with reCaptcha enabled:
- Clicking `Check Status` without filling out two fields properly gives me the browser validation to `Please fill out this field` on the first field. Button does not animate.
- Clicking `Check Status` with a badly formatted first field properly gives me the browser validation to `Please match the requested format`.  Button does not animate.
- Clicking `Check Status` with a properly formatted first field and empty second field proerly gives me the validation to `Please fill out this field` on the second field. Button does not animate.
- Clicking `Check Status` after filling out the last name field at _this_ point, after having done all these bad interactions, I'm not getting the Button animation.

Not sure if it's my local machine/reCaptcha being wonky or not.

machikoyasuda (changes requested):
Remove the `.attr("role", "status")`.

And make a new issue/PR to: Remove the X from this error bar

image
image


After merging this PR, we need to extensively test all the listed above form interactions for the CC form specifically on both desktop and mobile with recaptcha properly enabled.

angela-tran:
> ...
> - Clicking `Check Status` after filling out the last name field at _this_ point, after having done all these bad interactions, I'm not getting the Button animation.
> - I clicked `Check Status` _with_ my dev tools open.....and the button animated 🙃 
> 
> Not sure if it's my local machine/reCaptcha being wonky or not. 

Good catch @machikoyasuda. I see the same behavior on my machine

I think this is because on a successful reCAPTCHA response, we're [calling `submit()`](https://github.com/cal-itp/benefits/blob/e987616d06b1dfb20b254deeb06e1ac135023c5f/benefits/core/templates/core/includes/form.html#L83) which [does not fire a `submit` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit). The spinner is [listening](https://github.com/cal-itp/benefits/pull/1088/files#diff-e80e374594a3ed9b08de85983db18bdb5fb593fcab674d7a45c57513a0a10088R50) for a `submit` event.

Maybe we can try calling [`requestSubmit()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit) instead?

angela-tran:
Tried out calling `requestSubmit()` instead, and it fixes the issue.

machikoyasuda:
On further thought, I think the `role=status` from `.spinner-border` should be removed, and the `.attr("role", "status")` should remain. My thought process on this is that the role=status should be read / alerted when the spinner turns on, not when it's hidden.

thekaveman:
Ah, I think this is my fault in the rebase. 

We should have a `form="{{ form.id }}"` attribute on the button, which I added in #1071

thekaveman:
> On further thought, I think the `role=status` from `.spinner-border` should be removed, and the `.attr("role", "status")` should remain. My thought process on this is that the role=status should be read / alerted when the spinner turns on, not when it's hidden.

This is the correct interpretation. `role=status` on the button with `disabled` is telling assistive technologies that the button is no longer an interactive element, it is now providing information/status.

thekaveman:
> Ah, I think this is my fault in the rebase.
> 
> We should have a `form="{{ form.id }}"` attribute on the button, which I added in #1071

Ah so this is interesting... the form attribute was definitely part of it. But the new `<span>`s we added to get the spinner and text spaced correctly are messing up the JS that reCAPTCHA is using.

This line is the problem:

```js
function recaptchaSubmit($event) {
  // Checks the validity of the form. Return if invalid; HTML5 validation errors should display
  if (!$event.target.form.checkValidity()) { //<----- problem
    return;
  }
```

Because of the spinner, `$event.target` is no longer the `<button>`, it is the `<span>`!

Working on this...

thekaveman:
Thanks for your great reviews/comments @angela-tran @machikoyasuda!

Implemented a (hopefully) straightforward fix for reCAPTCHA + spinner:

* apply the `form="{{ form.id }}"` attribute to the button (I lost this in the rebase)
* change reCAPTCHA JS from using `$event.target` (the element that was clicked, now `<span class="btn-text">`) to `$event.currentTarget`, which is the element that was listening for the event that called the handler (i.e. the `<button>`): https://developer.mozilla.org/en-US/docs/Web/API/Event/Comparison_of_Event_Targets
* make the `disabled`/`status`/spinner stuff happen on a custom event `submitting` that is triggered:
  * by the normal form submit (when reCAPTCHA not enabled)
  * by the reCAPTCHA JS, just before `form.submit()`

angela-tran (dismissed):
Just one non-blocking question

benefits/core/templates/core/includes/form.html
```diff
@@ -68,19 +79,23 @@
...
           grecaptcha.ready(function() {
             grecaptcha.execute("{{ request.recaptcha.site_key }}", { action: "submit" }).then(function(token) {
-                // Add the token to hidden form element
-                $event.target.form.elements["{{ request.recaptcha.data_field }}"].value = token;
+                // add the token to hidden form element
+                $event.currentTarget.form.elements["{{ request.recaptcha.data_field }}"].value = token;
+                // trigger the custom "submitting" event
+                // since calling .submit() does not trigger the "submit" event
+                // hence the normal handler won't apply
+                $($event.currentTarget.form).trigger("submitting");
```
angela-tran, line 96: This solution works for me. Just curious: what was the reasoning for triggering a custom event instead of calling `requestSubmit()` on line 98 to trigger a `submit` event?

angela-tran (dismissed):
One non-blocking question

benefits/core/templates/core/includes/form.html
thekaveman: Sorry, I missed your suggestion to use that! Been a little scattered this afternoon... I like this better. Fixing.

machikoyasuda:
Amazing gif!!!!!!

benefits/core/templates/core/includes/form.html
thekaveman: Rebased to drop the old implementation and added [`cf7f670`](https://github.com/cal-itp/benefits/pull/1088/commits/cf7f670df923ef0b8221bf34841d1a1391c9f180)

angela-tran (approved):
Nice!

benefits/core/templates/core/includes/form.html
thekaveman: Fixed in [`b0343d6`](https://github.com/cal-itp/benefits/pull/1088/commits/b0343d6b1f37ce26b5f107ae531219f16dd3d938)

machikoyasuda (approved):

PR #1119, opened 2022-11-02
Help Page: New question for CC in English and Spanish, edit links

machikoyasuda:
## What this PR does
- Add one question/answer to Help for English and Spanish, with one external link to https://mst.org/riders-guide/how-to-ride/courtesy-card/
- The Eligibility Start page should have a different link page, depending on whether you are on Login.gov or CC. Login.gov Elig Start link goes to `#login-gov` anchor. CC Elig Start link goes to `#mst-courtesy-card` anchor.
- Adds a `pt-3` for the Contactless Card image to add more padding

## Model/database change
- Verifier now has `start_help_anchor`. Required. 

## How to test
- Read /help in English and Spanish
- Click the Help Page links on Elig Index, Elig Start from Login.gov and CC flow

## Post-merge
- Update Sample Data file


benefits/eligibility/views.py
```diff
@@ -100,7 +100,10 @@ def start(request):
         details=_(verifier.start_item_details),
     )
 
+    help_anchor = "mst-courtesy-card"
+
     if verifier.is_auth_required:
+        help_anchor = "login-gov"
```
machikoyasuda, line 106: @thekaveman I wasn't sure what a better way to do this would be.

benefits/eligibility/views.py
```diff
@@ -53,7 +53,7 @@ def index(request):
     ctx = page.context_dict()
     ctx["help_page"] = help_page
     ctx["help_text"] = format_html(
-        _("eligibility.pages.index.help_text%(help_link)s") % {"help_link": f"{help_page}#what-is-cal-itp"}
+        _("eligibility.pages.index.help_text%(help_link)s") % {"help_link": f"{help_page}#login-gov"}
```
thekaveman, line 56: I'm not sure if this change is correct (but I also agree that the existing anchor link probably shouldn't be there).

This line is for this link:

image

Shouldn't that just be to the generic Help Page, no anchor?

benefits/eligibility/views.py
thekaveman: I think this should probably be added as a property to the `Verifier` model and not hardcoded. Will help us when adding another Agency later.

thekaveman (changes requested):

benefits/eligibility/views.py
machikoyasuda: The top of the page already links to the generic Help Page (What is Cal-ITP?) question. So I linked the What Is Login.gov? (followed by What is Courtesy Card?) section here so people can be taken to the part about the two options. We don't actually have anything on the Help page that directly answers which option is right for people though.

machikoyasuda:
Unrelated fix, incoming:

Página Anterior, not Pagina Anterior

benefits/eligibility/views.py
thekaveman: Hmm... I see your point, but it still feels confusing to me to direct the user to Login.gov when they haven't made a selection yet. And SEO/accessibility-wise, this link mentioning Login.gov is not exactly accurate...

machikoyasuda: Yeah. I guess for now I'll just remove the anchor entirely cos there aren't any other good questions to link to.

machikoyasuda:
Latest commit will require a Benefits Secrets sample data update after merge

benefits/core/migrations/0001_initial.py
```diff
@@ -51,6 +51,7 @@ class Migration(migrations.Migration):
                 ("start_sub_headline", models.TextField()),
                 ("start_item_heading", models.TextField()),
                 ("start_item_details", models.TextField()),
+                ("start_help_anchor", models.TextField()),
```
thekaveman, line 54: NP but should be running `bin/makemigrations.sh` rather than editing these directly.

thekaveman (dismissed):
Nice 👍

benefits/core/migrations/0001_initial.py
machikoyasuda: Omg I had no idea about this (!!!)

machikoyasuda: I rebased `dev` and ran this command.

thekaveman (approved):

PR #1120, opened 2022-11-02
fix: missing accent

machikoyasuda:


thekaveman (approved):

machikoyasuda:
Source: https://www.elcastellano.org/anabella-mazariegos - When `Página` is a noun/refers to a page out of a book, it has an accent. When `Pagina` is a conjugated verb, it does not.

PR #1146, opened 2022-11-22
faster deploy

afeld:


angela-tran:
Similar to cal-itp/eligibility-server#205

angela-tran (approved):

PR #1147, opened 2022-11-28
fix(Terraform): add missing sticky settings

afeld:
Noticed them in [this pipeline run](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=37917&view=logs&j=e2b2897c-f736-5615-5c9c-c20ddcd4aa73&t=2ab79bdb-a159-573f-a6a0-2465f94c501e&l=78).


angela-tran (approved):
Thanks for catching this!

PR #1166, opened 2022-12-21
chore(Terraform): remove hostname bindings for dev and test slots

afeld:
They will be re-added when moved to the new App Service instances in https://github.com/cal-itp/benefits/pull/1074.

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -164,11 +164,6 @@ resource "azurerm_linux_web_app_slot" "dev" {
   }
 }
 
-resource "azurerm_app_service_slot_custom_hostname_binding" "dev" {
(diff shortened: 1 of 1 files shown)


thekaveman:
Sorry, what is the advantage/reason of doing this first/separately from #1074?

angela-tran:
> Sorry, what is the advantage/reason of doing this first/separately from #1074?

My understanding is that Azure doesn't let you have the same hostname binding to multiple app services, so we have to unbind it from the dev slot first and then bind it to the separate dev instance.

afeld:
I was mistaken; turns out the hostname _can_ be bound to two App Service instances. No longer needed.

PR #1171, opened 2022-12-22
fix(Terraform): downgrade App Service plan

afeld:
Fixes https://github.com/cal-itp/benefits/issues/1142. Holding on merging due to https://github.com/cal-itp/benefits/issues/1142#issuecomment-1363117266.

## Post-merge TODOs

- [ ] Confirm Eligibility Server connection still works
- [ ] Update allowed IP addresses

Issue #1142, opened 2022-11-21
downgrade App Service plan


afeld:
## Acceptance Criteria

- [ ] We are using an appropriate plan for our use case

## Additional context

See https://github.com/cal-itp/eligibility-server/issues/200.

thekaveman:
This can/will go as part of separating environments (this month), right @afeld?

afeld:
We can wait for that to test in isolation, but don't have to wait if we don't want to. My expectation is that there will be 30-60 seconds of downtime while the app is re-provisioned.

thekaveman:
OK, I think let's wait until we have more isolation. But otherwise just wanted to confirm we could do this this month.

afeld:
I certainly hope we are done with the separate environments soon!

angela-tran:
Note that when we change the App Service plan, this [changes the IP addresses of our app services](https://learn.microsoft.com/en-us/azure/app-service/overview-inbound-outbound-ips#when-outbound-ips-change)

> The set of outbound IP addresses for your app changes when you perform one of the following actions:
> ...
> * Scale your app between the lower tiers (Basic, Standard, and Premium), the PremiumV2, and the PremiumV3 tier (IP addresses may be added to or subtracted from the set).

Let's make sure we update `eligibility-server`'s IP address whitelists as a part of working on this issue.

I recommend holding off on this until after the holidays.

PR #1172, opened 2022-12-22
chore(Terraform): output the outbound IP ranges

afeld:
Will make it easier to copy the values into the Eligibility Server pipeline configuration.

### DIFF
--- a/terraform/output.tf
+++ b/terraform/output.tf
@@ -1,0 +1,4 @@
+output "outbound_ip_ranges" {
+  value       = sort([for ip in azurerm_linux_web_app.main.possible_outbound_ip_address_list : "${ip}/32"])
+  description = "The IP address ranges to use to filter requests from this App Service instance, such as API calls to the Eligibility Server"
+}
--- a/terraform/output.tf
+++ b/terraform/output.tf
@@ -1,4 +1,4 @@
 output "outbound_ip_ranges" {
   value       = sort([for ip in azurerm_linux_web_app.main.possible_outbound_ip_address_list : "${ip}/32"])
-  description = "The IP address ranges to use to filter requests from this App Service instance, such as API calls to the Eligibility Server"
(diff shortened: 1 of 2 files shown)


afeld:
[Preview of what the `output` will be](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=38595&view=logs&j=5af648c4-4294-55a7-6072-dd70820d3dba&t=42899ae2-6397-58a2-7641-cec3214b180b&l=43)

terraform/output.tf
```diff
@@ -0,0 +1,4 @@
+output "outbound_ip_ranges" {
+  value       = sort([for ip in azurerm_linux_web_app.main.possible_outbound_ip_address_list : "${ip}/32"])
+  description = "The IP address ranges to use to filter requests to this App Service instance, like for the Eligibility Server"
```
thekaveman, line 3: I think this comment is... maybe a little confusing? We use these IPs to filter requests _to the Eligibility Server_, from this App Service instance.

afeld: Whoops, good catch!

thekaveman:
Is this more related to #1171, which we are holding off on per [@angela-tran's suggestion](https://github.com/cal-itp/benefits/issues/1142#issuecomment-1363117266)?

afeld:
It's a convenience (and now docs) to clarify what needs to be done when the outbound IPs change. That includes things like the underlying Plan changes, and thus sets us up for things like https://github.com/cal-itp/benefits/pull/1171.

Similarly, for the Eligibility Server to accept requests from our new dev and test App Service instances, we merged this list with the existing one to allow requests from both App Service instances.

angela-tran:
> Is this more related to #1171, which we are holding off on per [@angela-tran's suggestion](https://github.com/cal-itp/benefits/issues/1142#issuecomment-1363117266)?

I think this is definitely helpful for #1171. It seems ok to get these changes in ahead of time.

thekaveman (approved):
👍

Issue #1153, opened 2022-12-02
represent all Key Vault Access Policies in Terraform


afeld:
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/key_vault#access_policy

## Acceptance Criteria

- [ ] The Access Policies seen through the Portal match what's in Terraform

afeld:
Emailed DevSecOps to ask for clarification on how they set it up.

PR #1173, opened 2022-12-22
fix(Terraform): specify the Key Vault Access Policies as code

afeld:
Fixes https://github.com/cal-itp/benefits/issues/1153.


terraform/variables.tf
```diff
@@ -5,3 +5,13 @@ variable "DEVSECOPS_OBJECT_ID" {
...
+
+variable "ENGINEERING_GROUP_OBJECT_ID" {
+  description = "Object ID for the Cal-ITP engineering Active Directory Group"
+  type        = string
+}
+
+variable "MYSTERY_APP_SERVICE_OBJECT_ID" {
+  description = "Object ID for something with principal name 'Microsoft Azure App Service'"
+  type        = string
+}
```
afeld, line 17: No idea what this is for, who/what created it (could have been me)… Will continue to investigate.

afeld: It doesn't show up in the other App Service instances, so I think it's not necessary. Worst case, we will see the problem in `dev` and can revert.

afeld:
Added more explanation up top. Ready for review.

angela-tran (approved):
This looks good to me

PR #1181, opened 2023-01-05
chore: upgrade Terraform and provider

afeld:
Was doing this to do something else, so might as well.


afeld:
[`No changes` in the `plan`](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=38792&view=logs&j=5af648c4-4294-55a7-6072-dd70820d3dba&t=42899ae2-6397-58a2-7641-cec3214b180b&l=41), which is good.

thekaveman (approved):

Issue #832, opened 2022-07-26
make downtime alerts louder


afeld:
As [described in the 7/21 downtime report](https://docs.google.com/document/d/14XFpJonXmrNKDt98g8vML3RimbZha0UJiIeUVclXYGs/edit#heading=h.tbfsxvwjhk7u),

> We should get “loud” alerts about downtime, but right now there’s just [a message posted to a high-volume channel](https://cal-itp.slack.com/archives/C022HHSEE3F/p1658439759753709).

## To Reproduce

1. Make the production environment unavailable.
   - At time of writing, the monitoring for `dev` and `test` are the same as `prod` will be.
   - Some ways to do this:
      - [Disable traffic](https://docs.microsoft.com/en-us/azure/app-service/deploy-staging-slots#route-production-traffic-automatically)
      - Deploy a code change that prevents the container from starting
      - Delete the hostname binding
3. Wait for alert

## Expected behavior

For production downtime, engineer(s) should get a ping, ideally something that needs to be acknowledged.

## Additional context

none

afeld:
@angela-tran pointed out that the alerts could be clearer as well: "alert activated" is really saying "the site is down."

afeld:
[The team wasn't seeing alerts for a scheduled GitHub Action failure](https://compilerla.slack.com/archives/C02GQ7AJXQE/p1669071788447239?thread_ts=1669050796.660269&cid=C02GQ7AJXQE); may be worth [sending those to Slack](https://www.ravsam.in/blog/send-slack-notification-when-github-actions-fails/).

afeld:
https://github.com/cal-itp/benefits/issues/733 is the same thing as this issue, but for exceptions rather than downtime.

angela-tran:
The email alerts to our Slack channel are working. It was more of a process improvement that was needed, and we've done that by updating our Incident Response procedures.

PR #1184, opened 2023-01-06
fix(CI): send API check failure notifications to Slack

afeld:
Was looking at https://github.com/cal-itp/benefits/issues/832#issuecomment-1323080591 after [the Payments Data team also came up with the failure scenario](https://docs.google.com/spreadsheets/d/1f1-fM7wJvxCUWvgzaq02Cyr_kPb5ubfCS64wnhZ5wWk/edit#gid=584616034) of the Littlepay API being down, and I realized this would be easy to resolve.

- [x] [Create Slack App](https://api.slack.com/apps/A04JKQEQU9W/general)
- [x] Set `SLACK_WEBHOOK_URL` in [Actions secrets](https://github.com/cal-itp/benefits/settings/secrets/actions)

### DIFF
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -75,3 +75,13 @@ jobs:
           --cacert $temp_dir/cacert.ca > $temp_dir/payload.txt
 
           test $(head -n 1 $temp_dir/payload.txt | grep -o 201)
+
+      # https://www.ravsam.in/blog/send-slack-notification-when-github-actions-fails/#using-notify-slack-action
+      - name: Report failure to Slack
+        if: always()
+        uses: ravsamhq/notify-slack-action@v1
+        with:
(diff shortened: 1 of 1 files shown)


thekaveman:
This is really cool! But one note, the payments team is _probably_ interested in a different API than what our check is for. They use some kind of export/bulk transactions API, whereas Benefits uses the Customer API.

afeld:
Triggered a failure in 2fcb289 and [they showed up in Slack](https://cal-itp.slack.com/archives/C022HHSEE3F/p1673048472947969), so it works! Ready for review.

angela-tran (approved):
Nice! Also, for more context, the only reason why we didn't have this already is because when #252 was worked on, we had a limited amount of Slack app integrations for Cal-ITP and had used them all.

Issue #1187, opened 2023-01-09
Refactor msgid, parameterize msgstr with agency-specific content


thekaveman:
A number of the `msgstr` in the current PO files are partially generic, but also partially reference a specific agency or agency's programs. For example, the success message after eligibility is verified via Login.gov:

```
msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.details"
msgstr ""
"Your eligibility has been verified by Login.gov. This benefit will give you "
"a half-price fare on public transportation with MST."
```

We want to be able to reuse the first part of this message across many agencies, and also to parameterize the second part to swap in the specific agency (and their discount if applicable).

## Acceptance Criteria

- [ ] Every `msgstr` that references a specific agency (or a specific agency's programs) directly, is either parameterized to accept the agency details or broken out into a minimal `msgid`/`msgstr` block
- [ ] Every `msgid` corresponding to the above parameterized `msgstr`s is free from any specific agency identifier

## Additional context

Corollary to #1186

PR #1195, opened 2023-01-10
Refactor: msgids for common messages

### LINKED ISSUE #1186: Refactor msgid for msgstr common to all agencies
A number of the `msgid` in the current PO files use the shortcode/slug of the agency, while their corresponding `msgstr` is not actually specific to that agency. Some examples (this text in these examples is specific to the eligibility / eligibility verifier):

* The heading shown on the Eligibility Start page after the user selects to verify with Login.gov

    ```
    msgid "eligibility.pages.start.mst_login.headline"
    msgstr "You selected an Older Adult benefit."
    ```

* The enrollment success message

    ```
    msgid "enrollment.pages.success.mst_login.confirm_item.details"
    msgstr ""
    "You were not charged anything today. When boarding transit, pay with this "
    "same physical card and you will automatically receive your new reduced fare. "
    "You will need to reapply if you choose to change the bank card you use to "
    "pay for transit service."
    ```

Let's refactor all of the `msgid`s for examples such as these, where the text can be reused across _agencies_.

## Acceptance Criteria

- [ ] Every `msgstr` that doesn't reference a specific agency (or a specific agency's programs) directly, has a corresponding `msgid` free from any specific agency identifier

## Additional context

Corollary to #1187

angela-tran:
Closes #1186


benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -19,57 +19,57 @@ msgstr "Get started with"
...
 msgstr "Your eligibility is confirmed! You’re nearly there."
 
-msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.heading"
+msgid "enrollment.pages.index.login_gov.eligibility_confirmed_item.heading"
 msgstr "Your eligibility is confirmed for an Older Adult benefit"
 
 msgid "enrollment.pages.index.mst_login.eligibility_confirmed_item.details"
 msgstr ""
 "Your eligibility has been verified by Login.gov. This benefit will give you "
 "a half-price fare on public transportation with MST."
```
angela-tran, line 70: Leaving `enrollment.pages.index.mst_login.eligibility_confirmed_item.details` alone to be handled by #1187 since it has the "with MST" part in it.

Similarly, also leaving `core.pages.agency_index.mst_cc.headline` alone to be handled by #1187 -- it is used in the `agency_index` view function, which applies to all agencies.

https://github.com/cal-itp/benefits/blob/22cfc1677349d175e4875f044be17482aa52b534/benefits/locale/en/LC_MESSAGES/django.po#L425-L426

https://github.com/cal-itp/benefits/blob/22cfc1677349d175e4875f044be17482aa52b534/benefits/core/views.py#L61

angela-tran:
Question for you @thekaveman:

With 22cfc16, I fixed all the msgids that were named for a specific transit agency when instead they should've been specific to the verifier. Should I also work on consolidating entries that have duplicate content?

For example, these two entries have the same content:

https://github.com/cal-itp/benefits/blob/22cfc1677349d175e4875f044be17482aa52b534/benefits/locale/en/LC_MESSAGES/django.po#L37-L38

https://github.com/cal-itp/benefits/blob/22cfc1677349d175e4875f044be17482aa52b534/benefits/locale/en/LC_MESSAGES/django.po#L94-L95

Would we want that to be just one entry?

thekaveman:
@angela-tran 

> Should I also work on consolidating entries that have duplicate content?

That sounds great!

benefits/core/migrations/0001_initial.py
```diff
@@ -66,9 +65,7 @@ class Migration(migrations.Migration):
                 ("form_name_placeholder", models.TextField(null=True)),
                 ("form_name_max_length", models.PositiveSmallIntegerField(null=True)),
                 ("unverified_title", models.TextField()),
-                ("unverified_headline", models.TextField()),
                 ("unverified_blurb", models.TextField()),
-                ("eligibility_confirmed_headline", models.TextField()),
```
thekaveman, line 71: For these changes - I was envisioning we leave the models alone, but consolidate at the PO file and update the references in the data migrations. I'm OK with this approach too, we may end up adding some of this back to be able to change it later.

thekaveman (approved):
This looks good! I looked through the PO file and didn't see any other `msgid` that should be genericized based on its `msgstr`.

benefits/core/migrations/0001_initial.py
angela-tran: Agreed 👍

PR #1196, opened 2023-01-11
Refactor: parameterize messages with agency-specific content

### LINKED ISSUE #1187: Refactor msgid, parameterize msgstr with agency-specific content

angela-tran:
Closes #1187 

### Run locally
Make sure to run with updated data migrations (sample data is updated) and run `./bin/init.sh` to recompile the language files.

### Before merging
- [x] Update the data migration for dev in Azure.


thekaveman:
My initial take on https://github.com/cal-itp/benefits/commit/7df9e319e4af41ec2989c20de3dd99d9007e8f8f... I think this is a case where storing the entire first sentence is fine, rather than trying to deal with indefinite articles and placeholders where the overall/outside structure of the sentence has to change. It's OK to duplicate some content to make the code / language files more straightforward. 

I'm also not a huge fan of that format syntax needed for `html_format`, it doesn't support fstrings?

benefits/enrollment/views.py
```diff
@@ -80,7 +80,11 @@ def index(request):
 
         if verifier.eligibility_confirmed_item_heading or verifier.eligibility_confirmed_item_details:
             heading = _(verifier.eligibility_confirmed_item_heading) if verifier.eligibility_confirmed_item_heading else None
-            details = _(verifier.eligibility_confirmed_item_details) if verifier.eligibility_confirmed_item_details else None
+            details = (
+                _(verifier.eligibility_confirmed_item_details) % {"transit_agency_short_name": agency.short_name}
```
thekaveman, line 84: Does this need a call to `format_html`?

angela-tran: There's no HTML in the value, so no need to call `format_html`
```
"Your eligibility has been verified by Login.gov. This benefit will give you "
"a half-price fare on public transportation with "
"%(transit_agency_short_name)s."
```

angela-tran:
> My initial take on https://github.com/cal-itp/benefits/commit/7df9e319e4af41ec2989c20de3dd99d9007e8f8f... I think this is a case where storing the entire first sentence is fine, rather than trying to deal with indefinite articles and placeholders where the overall/outside structure of the sentence has to change. It's OK to duplicate some content to make the code / language files more straightforward.

Yeah, I also felt like the approach in 7df9e319e is a little complicated. So just to be sure, are you saying we'd have `"You can tap your debit or credit card when you board an MST "` as an entry? And then when SacRT is added, we'd have another one saying `"You can tap your debit or credit card when you board a SacRT "`?

angela-tran:
> I'm also not a huge fan of that format syntax needed for html_format, it doesn't support fstrings?

Yeah, this is actually not because of `format_html` but rather because of a limitation with Django internationalization in Python code (because of gettext). From [Django docs](https://docs.djangoproject.com/en/4.1/topics/i18n/translation/#internationalization-in-python-code):
> Since string extraction is done by the xgettext command, only syntaxes supported by gettext are supported by Django. In particular, Python [f-strings](https://docs.python.org/3/reference/lexical_analysis.html#f-strings) are not yet supported by xgettext, and JavaScript template strings need gettext 0.21+.

thekaveman:
> So just to be sure, are you saying we'd have `"You can tap your debit or credit card when you board an MST "` as an entry?

I'm suggesting you go even further, with these (eventual) entries - entire sentences:

* `"You can tap your credit or debit card when you board an MST bus, and your discounted fare will automatically apply every time you ride."`
* `"You can tap your credit or debit card when you board a SacRT bus, and your discounted fare will automatically apply every time you ride."`

I guess this is kind of the opposite of parameterizing the `msgstr`... having multiple versions and configuring different agencies with different versions. 

Seeing parameterization within sentences... it looks a little weird, different from doing this for e.g. links/HTML. It seems clearer to represent the entire sentence in the PO file. 

What do you think?

angela-tran:
> Seeing parameterization within sentences... it looks a little weird, different from doing this for e.g. links/HTML. E.g. `eligibility_index_paragraph_part` is kind of strange as a model prop? It seems clearer to represent the entire sentence in the PO file. 

I see - the idea is to not have all these fragments floating around, even if that means some duplication. So it sounds like what we're leaning towards is, if we can easily reuse the agency's short-name/long-name/slug somewhere (like in `enrollment.pages.index.mst_login.eligibility_confirmed_item.details%(transit_agency_short_name`), 
```
"Your eligibility has been verified by Login.gov. This benefit will give you "
"a half-price fare on public transportation with MST."
```

we should parameterize that. Otherwise, we should just have the full sentence represented on the model.

My only question then is what to call the property on `TransitAgency` for the full sentence...

thekaveman:
> the idea is to not have all these fragments floating around, even if that means some duplication. So it sounds like what we're leaning towards...

Yeah I think this is what I'm trying to get at, and matches what I was thinking in terms of a heuristic. 

> My only question then is what to call the property on `TransitAgency` for the full sentence...

Tough question! 😅  Maybe `eligibility_index_intro`?

thekaveman (approved):
Yep this looks good 👍 I went through the PO file and don't see any other strings that should be parameterized or split into multiple versions.

Issue #1163, opened 2022-12-15
Improve secrets management


afeld:
In https://github.com/cal-itp/benefits/issues/1143, we determined that we want to go from the data migration file being treated as one big secret to the file loading secrets it needs at runtime (e.g. as environment variables).

## Acceptance Criteria

- [ ] We are managing secrets as individual values rather than all in one file
- [ ] We minimize the number of places that secrets are stored
- [ ] Secrets are stored in a secure location

## Tasks
- [x] Modify the [existing data migration](https://github.com/cal-itp/benefits/blob/dev/benefits/core/migrations/0002_sample_data.py) to read from environment variables - #1198
- [x] #1242
   - (support for environment-specific Vaults was added in [#830](https://github.com/cal-itp/benefits/issues/830))
- [ ] #1241
- [ ] #1243
- [ ] #1244
   - Delete data migration files from benefits-secrets
   - Delete/rename data migration file from each bucket
   - Delete storage mount from Terraform Azure definition
 - [ ] #1245

## Additional context

See #1143 and #1116 for more context on past situations that motivated this refactor.

angela-tran:
Talked with @thekaveman about working on this while we wait for #1090. The first part will be to update our data migration file to read environment variables.

thekaveman:
Moved description into Milestone, closing.

PR #1198, opened 2023-01-12
Refactor: add support for env variables in data migration

angela-tran:
Related to #1163 

This PR modifies the sample data migration file to look for environment variables for sensitive values, falling back to defaults if an environment variable is not defined. 

This is to help move us towards actually using this file for our deployment process. Note that these changes do not break any of the current set up in Azure (since it's not using the sample data) nor does it break local dev environments (no `.env` file required).

### To test
As an example, you could fill this out to specify real values and add to a `.env` file:
```
AUTH_PROVIDER_CLIENT_NAME=
AUTH_PROVIDER_CLIENT_ID=
AUTH_PROVIDER_AUTHORITY=
AUTH_PROVIDER_SCOPE=
AUTH_PROVIDER_CLAIM=
```
Run
```bash
set -a
source .env
./bin/init.sh
``` 
and you should then be able to go through the Login.gov flow without changing any of the data migration file.

You can use `unset` (e.g. `unset AUTH_PROVIDER_CLIENT_NAME`) to delete an environment variable from your shell.

For multi-line values in your .env file, you can wrap them in single quotes, e.g.
```
CLIENT_PRIVATE_KEY='---begin blah---
hi
---end blah---
'
```

### DIFF
--- a/benefits/core/migrations/0002_sample_data.py
+++ b/benefits/core/migrations/0002_sample_data.py
@@ -95,7 +95,7 @@ PEM DATA
     )
 
     payment_processor_client_cert_private_key = PemData.objects.create(
-        text=os.environ.get("PAYMENT_PROCESSOR_CLIENT_CERT_PRIVATE_KEY", client_private_key),
+        text=os.environ.get("PAYMENT_PROCESSOR_CLIENT_CERT_PRIVATE_KEY", client_private_key.text),
         label="Payment processor client certificate private key",
     )
 
--- a/benefits/core/migrations/0002_sample_data.py
+++ b/benefits/core/migrations/0002_sample_data.py
@@ -183,14 +183,14 @@ PEM DATA
     PaymentProcessor = app.get_model("core", "PaymentProcessor")
 
     payment_processor = PaymentProcessor.objects.create(
-        name="Test Payment Processor",
-        api_base_url="http://server:8000",
-        api_access_token_endpoint="access-token",
-        api_access_token_request_key="request_access",
-        api_access_token_request_val="REQUEST_ACCESS",
-        card_tokenize_url="http://server:8000/static/tokenize.js",
-        card_tokenize_func="tokenize",
-        card_tokenize_env="test",
+        name=os.environ.get("PAYMENT_PROCESSOR_NAME", "Test Payment Processor"),
+        api_base_url=os.environ.get("PAYMENT_PROCESSOR_API_BASE_URL", "http://server:8000"),
+        api_access_token_endpoint=os.environ.get("PAYMENT_PROCESSOR_API_ACCESS_TOKEN_ENDPOINT", "access-token"),
+        api_access_token_request_key=os.environ.get("PAYMENT_PROCESSOR_API_ACCESS_TOKEN_REQUEST_KEY", "request_access"),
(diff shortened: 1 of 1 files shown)


benefits/core/migrations/0002_sample_data.py
```diff
@@ -67,35 +76,50 @@ def load_sample_data(app, *args, **kwargs):
...
+        scope=os.environ.get("AUTH_PROVIDER_SCOPE", "verify:senior"),
+        claim=os.environ.get("AUTH_PROVIDER_CLAIM", "senior"),
     )
 
     EligibilityVerifier = app.get_model("core", "EligibilityVerifier")
 
-    verifier1 = EligibilityVerifier.objects.create(
-        name="OAuth claims via Login.gov",
+    oauth_claims_verifier = EligibilityVerifier.objects.create(
+        name=os.environ.get("OAUTH_VERIFIER_NAME", "OAuth claims via Login.gov"),
```
angela-tran, line 122: Not sure if we consider these names sensitive

benefits/core/migrations/0002_sample_data.py
```diff
@@ -179,21 +203,21 @@ def load_sample_data(app, *args, **kwargs):
 
     mst_agency = TransitAgency.objects.create(
         slug="mst",
-        short_name="MST (sample)",
-        long_name="Monterey-Salinas Transit (sample)",
+        short_name=os.environ.get("MST_AGENCY_SHORT_NAME", "MST (sample)"),
+        long_name=os.environ.get("MST_AGENCY_LONG_NAME", "Monterey-Salinas Transit (sample)"),
```
angela-tran, line 207: These names don't seem sensitive IMO, but they do vary per environment.

benefits/core/migrations/0002_sample_data.py
```diff
@@ -116,16 +140,16 @@ def load_sample_data(app, *args, **kwargs):
...
+        api_auth_header=os.environ.get("COURTESY_CARD_VERIFIER_API_AUTH_HEADER", "X-Server-API-Key"),
+        api_auth_key=os.environ.get("COURTESY_CARD_VERIFIER_API_AUTH_KEY", "server-auth-token"),
         eligibility_type=courtesy_card_type,
         public_key=server_public_key,
-        jwe_cek_enc="A256CBC-HS512",
-        jwe_encryption_alg="RSA-OAEP",
-        jws_signing_alg="RS256",
+        jwe_cek_enc=os.environ.get("COURTESY_CARD_VERIFIER_JWE_CEK_ENC", "A256CBC-HS512"),
+        jwe_encryption_alg=os.environ.get("COURTESY_CARD_VERIFIER_JWE_ENCRYPTION_ALG", "RSA-OAEP"),
+        jws_signing_alg=os.environ.get("COURTESY_CARD_VERIFIER_JWS_SIGNING_ALG", "RS256"),
```
angela-tran, line 152: Do we consider these sensitive? I erred on the side of caution here.

I think [Kerckhoff's principle](https://en.wikipedia.org/wiki/Kerckhoffs's_principle) would say they are not sensitive though (the security of data should not depend on the encryption method remaining a secret).

thekaveman (approved):
This looks good, sets us up to be able to progress towards better management of these items.

I tested locally with the sample data file and it all worked as expected:

* Without any values in my `.env`, every works like before
* I was able to override a value via my `.env`

PR #1201, opened 2023-01-13
Chore: always show the user the eligibility index

### LINKED ISSUE #1200: Remove auto-forward for single-verifier agencies
In #337 we introduced handling for an eligibility verifier selection form, which supports choosing between e.g. verifying eligibility with Login.gov or with a Courtesy Card. We also added [a shortcut](https://github.com/cal-itp/benefits/blob/dev/benefits/eligibility/views.py#L77) for when an agency has only a single verifier -- to bypass this form and go straight to the next step in the flow.

We want to have a consistent user experience regardless of the number of verifiers, and so it's time to remove this shortcut and always show the verifier selection page, even with only 1 option.

## Acceptance Criteria

- [ ] The user must always choose a verifier on `/eligibility`, regardless of the number of verifiers there are to choose from

## Additional context

Design sign-off on this approach: https://cal-itp.slack.com/archives/C013W8RUAMU/p1673542656882989

angela-tran:
Closes #1200 

Unit test was updated and passes. To test in the browser, update your sample data so that `mst_agency` only has one verifier.

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -74,12 +74,7 @@ def index(request):
             page.forms = [form]
             response = TemplateResponse(request, TEMPLATE_INDEX, ctx)
     else:
-        if agency.eligibility_verifiers.count() == 1:
-            verifier = agency.eligibility_verifiers.first()
-            session.update(request, verifier=verifier)
-            response = redirect(eligibility_start)
-        else:
-            response = TemplateResponse(request, TEMPLATE_INDEX, ctx)
+        response = TemplateResponse(request, TEMPLATE_INDEX, ctx)
 
     return response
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Looks great. I tested locally with an agency that only has 1 verifier, and saw the `/eligibility` page with verifier selection as expected ✅ 

## Sidebar

At first I had a few failing tests in `eligibility/views`:

image

_Notice all the failures are for POST view handlers_

Debugging I saw that it was the reCAPTCHA check failing - I had it turned on locally in my `.env` file. Once I turned that off, these tests passed.

Maybe we should create a follow-up issue to address this with a `@pytest.mark` or something, to ensure reCAPTCHA is disabled for tests that will fail otherwise.

Issue #1090, opened 2022-10-26
Design the Agency selection page


thekaveman:
Sooner than later, we will go live with a second transit agency. Like today, users will be able to get directly to the app homepage for their agency using the direct agency link, e.g. https://benefits.calitp.org/mst.

Once we have multiple agencies however, the root app homepage e.g. https://benefits.calitp.org needs to provide an interface for the user to select between the agency options.

For this task, let's focus on the use case of selecting between two transit agencies only. In further iterations we can look to expand to more flexible use cases / UX (e.g. lookup by name, location, other options).

srhhnry:
Sync meeting scheduled for Jan 10.

srhhnry:
Working through some ideas here: https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?node-id=8084%3A34817&t=7jfmLMfG5hgyibTP-1

thekaveman:
@srhhnry India @angela-tran and @thekaveman met to discuss options for this design.

Decision:

* the generic homepage will have a single button, clicking this shows a modal with cards for each agency (and a close button)
* clicking an agency card in the modal takes the user to the _eligibility start_ page for that agency
* users can still come in via the `/agency` shortcut, showing an agency-branded (e.g. text) homepage

srhhnry:
Designs have been finalized and can be found here, as well as incorporated into the Benefits Full App pages, under Finalized Designs: https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?node-id=8084%3A34817&t=PFQBMVnL8Ju7csCx-0

PR #1206, opened 2023-01-19
Refactor: agency selection frontend

### LINKED ISSUE #1197: Implement Agency Selection page
This task is to implement the agency selection based off the notes from https://github.com/cal-itp/benefits/issues/1090#issuecomment-1377922860 and the designs in Figma (once they are finalized).

The backend code will need to change to support this - **see the approach outlined in https://github.com/cal-itp/benefits/issues/1197#issuecomment-1380873661**

<details><summary>Other ideas that we considered</summary>

> @thekaveman and I paired a bit to talk through ideas for implementation. Those ideas were:
>  - a form-based approach, where each card is actually a submit button for a form with a hidden input - the form would need to be handled by `/eligibility`
>  - a link-based approach, where each card links to a new URL whose view function simply sets the session agency and forwards to `/eligibility`
>  - a link-based approach, where each card links to the existing `agency_index` URL but with the addition of a [query parameter](https://docs.djangoproject.com/en/4.1/ref/request-response/#django.http.QueryDict) so that `agency_index` can know to forward to `/eligibility` - can be something like `?select=true`

</details>

## Acceptance Criteria
- [ ] Agency Selection page matches finalized design in Figma
- [ ] The user can enter the site from either the index (`/`) or an agency's index page (e.g. `/mst`)

### LINKED ISSUE #1214: Use approved logos for transit agencies
@srhhnry started a [thread](https://cal-itp.slack.com/archives/C013W8RUAMU/p1674155979096579) in Slack to get approved logos. Once they've been finalized in Figma, let's update the files and make sure they are sized correctly in the UI.

## Acceptance Criteria
- [ ] Logos in the agency selector modal dialog are sized and placed correctly for Desktop and Mobile

angela-tran:
Closes #1197 
Closes #1214 

This PR implements the designs from #1090. 

The overall approach here was to use Bootstrap's [Modal](https://getbootstrap.com/docs/5.1/components/modal) component in a Django template that can be extended. A template called `agency-selector.html` then extends `modal.html` and handles defining the cards. It is used specifically by the `core/index.html`  template.


angela-tran:
This is still a work-in-progress due to some minor questions / fixes needed, but here are some screenshots of it so far:

### Screenshots

| Screen size | Indirect Entry (Index) `/` | Modal is open | Hovering over a card |
| --- | --- | --- | --- |
| Desktop | 01_index | 02_modal_open | 03_modal_hoverd |
| Mobile | localhost_11369_(iPhone SE) (0) | localhost_11369_(iPhone SE) (1) | localhost_11369_(iPhone SE) (2) |

### GIFs
| Desktop | Mobile 
| --- | --- |
| modal-desktop | modal-mobile |

### Questions for @srhhnry:
 - Does the faded background and animation look ok to you? It's just the default from Bootstrap. It doesn't match what's currently in Figma exactly, but you were saying you wanted to see what Bootstrap does.
 - I exported the logos from Figma and noticed the MST logo in the mobile designs is truncated. Is that easy for you to fix? Once it's updated, I'll re-export it.
 
image

srhhnry:
Thoughts: 

1. The bootstrap default animation looks just fine to me. We're just visually indicating to the user that the rest of the screen is inactive. Also, the rest of the screen should be inactive (I'm assuming that's standard bootstrap modal/lightbox behavior).

2. Ideally we should get logo assets from the agencies. As we haven't done this before, there's a bit of a workflow issue there (e.g. does product handle that? Customer relations? Design? Unclear). But I pulled those logos from the internet, and I'm not 100% sure they are robust enough for an actual site (they work for Figma screens, but may not have the fidelity for say, a 49 inch monitor). I'll around for how we get those assets.

3. Choose Your Provider CTA on desktop should be justified left to align with the copy--this is an update from the previous iteration of the home page.

4. The mobile CTA and header (within the white box) looks a little low--the hero image feels like it's taking up a lot of real estate, causing the user to have to scroll to see the full context for the page. In the Figma boards the white box starts about midway up the Visa CC (instead of at the bottom) 
image

Possible to move the box/header/CTA up? Expected behavior would be the user can see the CTA button without having to scroll.

srhhnry:
Oh, and I fixed the MST logo on the Mobile board (it was filling the image instead of fitting to the box!) so at the very least it's not cut off anymore. Standing by on logo assets though.

srhhnry:
> Currently I have it configured so that if the user clicks outside the modal dialog, it closes. Bootstrap also gives us the option to not do anything in that case (the user would have to click the "X" to close the modal dialog).

Yup! That's perfect. Usually more savvy users will click outside the modal whereas less savvy will look for the X. Having both is great.

angela-tran:
> 3. Choose Your Provider CTA on desktop should be justified left to align with the copy--this is an update from the previous iteration of the home page.

Done in [006ef31](https://github.com/cal-itp/benefits/pull/1206/commits/006ef315d8baac3d673927bae20638654c22bea0)

> 4. The mobile CTA and header (within the white box) looks a little low--the hero image feels like it's taking up a lot of real estate, causing the user to have to scroll to see the full context for the page. In the Figma boards the white box starts about midway up the Visa CC (instead of at the bottom)

I created a separate issue to handle the mobile CTA (https://github.com/cal-itp/benefits/issues/1213) so that this PR isn't blocked by it.

angela-tran:
This is ready to be reviewed! Approved logos have been added, and all feedback from @srhhnry has been addressed (except for #1213, which will be worked on separately).

thekaveman (commented):
Overall I think this is great! The flexible `modal.html` include, `call-to-action` block, refactors to `landing.html` and finally! removing `page.html` -- this is all excellent! 👍 

I wonder if we could tighten up the agency selector to make it more of a reusable component and keep the implementation from being spread around the view (and that weird generator loop on the context dictionaries).

Something like a `viewmodels.AgencySelector(Modal)` to encapsulate (more or less hardcode) everything that is needed for that specific modal?

angela-tran:
> I wonder if we could tighten up the agency selector to make it more of a reusable component and keep the implementation from being spread around the view (and that weird generator loop on the context dictionaries).

Thanks for this suggestion! Done in [522719e](https://github.com/cal-itp/benefits/pull/1206/commits/522719ef06cee29355f837fa3d0e86bf659f82ad)

machikoyasuda (changes requested):
I'm going through the modal user experience with an eye on accessibility (keyboard only, keyboard shortcuts) and matching the mocks.

These flows all work:
- Use <kbd>tab</kbd> to get focus ring on CTA. Then, use <kbd>enter</kbd> or <kbd>space</kbd> to open the modal. ✅ 
- Once modal is open, and nothing is in focus, use <kbd>escape</kbd> to exit modal. ✅ 
- Once modal is open, press <kbd>tab</kbd>. The first icon in focus is the X. Use either <kbd>space</kbd> or <kbd>enter</kbd> on the focused X to exit modal. ✅ 
- Once modal is open, press <kbd>tab</kbd> twice. The MST button is selected. Use either <kbd>space</kbd> or <kbd>enter</kbd> get to MST (or SacRT). ✅ 

The one thing missing: There needs to be a focus ring around the whole button very clearly. We don't have designs for it necessary @srhhnry but we can use default (from our styles)

image


or @srhhnry can design something custom

benefits/static/css/styles.css
```diff
@@ -592,55 +681,79 @@ h1 + .media-list, /* A .media-list immediately following the h1: Enrollment Succ
...
 
-.agency-index footer {
+.landing footer {
   margin-top: 0;
 }
+
+/* Agency Selector */
+
+#agency-selector .modal-body {
+  padding: 0px;
```
machikoyasuda, line 744: No `px` needed

benefits/static/css/styles.css
```diff
@@ -558,6 +566,87 @@ h1 + .media-list, /* A .media-list immediately following the h1: Enrollment Succ
...
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+  padding: var(--card-body-y-padding) var(--card-x-padding);
+}
+
+.card:hover {
+  border-top: solid var(--card-border-width) var(--hover-color);
+}
+
+.card .card-title {
+  font-size: var(--card-title-font-size);
```
machikoyasuda, line 603: The `.card-title` line-height is 24px/120% on desktop and 21.6px on mobile on Figma. The mobile one doesn't matter as much, but it matters a lot for desktop b/c the agency names go into 2 lines.

machikoyasuda:
@angela-tran Thinking about the HTML structure of the MST and SacRT components more, I think it should/could all be a `<button>` with an image inside, rather than a `card`. How I'd approach it: https://www.sarasoueidan.com/blog/accessible-icon-buttons/#icon-sitting-next-to-text

benefits/core/viewmodels.py
```diff
@@ -141,6 +142,17 @@ def __init__(self, **kwargs):
...
 
 
+class AgencySelector(Modal):
+    """
+    Represents the agency selector modal dialog.
+    """
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        agencies = kwargs.get("agencies", [])
```
thekaveman, line 152: I think the logic of `models.TransitAgency.all_active()` should be here (maybe as the default) rather than in the view.

benefits/core/views.py
```diff
@@ -33,22 +33,18 @@ def index(request):
         agency = agencies[0]
         return redirect(agency.index_url)
```
thekaveman, line 34: I'm thinking this whole check/redirect can go away now that we have a proper selection experience (and 2 agencies)

thekaveman (changes requested):

benefits/static/css/styles.css
```diff
@@ -312,6 +315,11 @@ footer .footer-links li a:visited {
   border-width: 2px;
 }
 
+/* Make `button` elements look similar to button-links. */
+button:focus {
+  outline: none !important;
+}
```
machikoyasuda, line 321: Remove this

benefits/core/templates/core/includes/agency-selector.html
```diff
@@ -0,0 +1,31 @@
...
+                                 width="46"
+                                 height="29"
+                                 alt="{% blocktranslate with agency_short_name=agency.short_name %}core.pages.index.agency_selector{{ agency_short_name }}{% endblocktranslate %}">
+                            <img class="lg d-none d-lg-block mt-lg-3 mx-lg-auto"
+                                 src="{% get_static_prefix %}img/agencies/{{ agency.slug }}-lg.svg"
+                                 width="117"
+                                 height="74"
+                                 alt="{% blocktranslate with agency_short_name=agency.short_name %}core.pages.index.agency_selector{{ agency_short_name }}{% endblocktranslate %}">
+                            <h4 class="card-title mt-lg-3 mb-0 mb-lg-3">{{ agency.long_name }}</h4>
+                            <a href="{{ agency.eligibility_index_url }}" class="stretched-link"></a>
```
machikoyasuda, line 24: I wonder if there's a way to move this link or add styles to it so this link does not affect the height/flow of the parent div. Also, I wonder if there's a way to add styles so that the focus ring of this link borders the entire div cleanly.

thekaveman: Similarly, this link needs to at least wrap the `h4` text, we shouldn't have a link with no text content.

thekaveman: FWIW @machikoyasuda I think this `stretched-link` utility is the right way to do this for cards in Bootstrap: https://getbootstrap.com/docs/4.6/utilities/stretched-link/

But we do need text content in the link

thekaveman (commented):
+1 to also needing a solution for the focus state on each card when navigating by keyboard.

benefits/static/css/styles.css
angela-tran: For more context, @machikoyasuda was showing me that we already have `<button>`s on the site that have the light-blue outline, which is different from `<a>` elements styled to look like buttons, which don't have the light-blue outline.

The "Espanol" button and form submission buttons are good examples.

I added this rule just because it was really obvious between the index page (which has a `<button>` CTA) vs. the agency_index page (`<a>` CTA) that the "button" outlines were different, but I'm fine with removing this and not worrying about that difference.

benefits/static/css/styles.css
angela-tran: Done in 0c29a5f12211c976c222f2bbca8db6ed69c9d4b5

benefits/static/css/styles.css
angela-tran: Done in 691f66d1427c85d1cc1b7e4bca48beb10f6a9621

benefits/core/views.py
angela-tran: Done in 069991d815e846f13e31a266a52e9a2c17eb71d7

benefits/core/viewmodels.py
angela-tran: Done in c1347be43cc4ee77dca0df4babe9f67b9174b595

angela-tran: Sorry, had to rebase to fix a test -- moved the logic of `models.TransitAgency.all_active()` in a273d54a3ddd5021e2062943bce878338d4e7521

benefits/core/views.py
angela-tran: Had to rebase to fix a test, so the new hash that this was done in is 4719906ddaa9832aea249d0d70151c8a72fc10eb

angela-tran:
> @angela-tran Thinking about the HTML structure of the MST and SacRT components more, I think it should/could all be a `<button>` with an image inside, rather than a `card`. How I'd approach it: https://www.sarasoueidan.com/blog/accessible-icon-buttons/#icon-sitting-next-to-text

I read through that article, and I can see some benefit to switching it to a `<button>` (mostly that screenreaders would read it as an actual button?). I'm not sure how much time we have to make the switch, so maybe this is a refactor we could take on after the initial release?

I will definitely work on fixing the focus style with the suggestion you sent me in Slack @machikoyasuda.

angela-tran:
So @thekaveman and @machikoyasuda , I tried each of your suggestions. The similarities between the two are that
 - the <kbd>tab</kbd> focus ring does not cause the text to shift ✅ 
 - I could not find a way to change the style of the focus ring (it seems browser-defined)

Here's what they each looked like:

### Use `stretched-link` class on `<a>` element which wraps `<h4>`

#### Code snippet
```html
<div class="card mt-lg-4 mx-lg-auto">
    <div class="card-body d-flex flex-row-reverse flex-lg-column justify-content-center align-items-center">
        <img class="sm d-lg-none ms-auto"
             src="{% get_static_prefix %}img/agencies/{{ agency.slug }}-sm.svg"
             width="46"
             height="29"
             alt="{% blocktranslate with agency_short_name=agency.short_name %}core.pages.index.agency_selector{{ agency_short_name }}{% endblocktranslate %}">
        <img class="lg d-none d-lg-block mt-lg-3 mx-lg-auto"
             src="{% get_static_prefix %}img/agencies/{{ agency.slug }}-lg.svg"
             width="117"
             height="74"
             alt="{% blocktranslate with agency_short_name=agency.short_name %}core.pages.index.agency_selector{{ agency_short_name }}{% endblocktranslate %}">
        <a href="{{ agency.eligibility_index_url }}" class="stretched-link"><h4 class="card-title mt-lg-3 mb-0 mb-lg-3">{{ agency.long_name }}</h4></a>
    </div>
</div>
```
#### Screenshots
| Chrome | Firefox |
| --- | --- |
| image | image |

### Make the element with `card` class an `<a>` instead of `<div>`
#### Code
See 801b57c2854d5d4b29bab47cb8d13aebce5bc841

#### Screenshots
| Chrome | Firefox |
| --- | --- |
| image | image |

I went with the latter suggestion because it puts the focus ring around the whole card.

machikoyasuda:
@angela-tran this is looking good!

one small thing to add: remove the underline when hovered - by adding custom style for `.card:hover`
image

thekaveman (approved):
Thanks for the changes, this looks good to me.

machikoyasuda (approved):
LGTM! ✨ ✨ 🚀

srhhnry:
Catching up on this thread. It sounds like the focus ring can't be designed? And is set by the browser? 

In case that's not the case, my suggestion would be to keep the behavior fairly close to default, with the exception being the use of our blue (#046b99) for the focus outline. To get real fancy, I'd suggest the hover color (#2f4c65) when selecting the agency button.

angela-tran:
> Catching up on this thread. It sounds like the focus ring can't be designed? And is set by the browser?
> 
> In case that's not the case, my suggestion would be to keep the behavior fairly close to default, with the exception being the use of our blue (#046b99) for the focus outline. To get real fancy, I'd suggest the hover color (#2f4c65) when selecting the agency button.

Yeah, so the cards are `<a>` elements (links), which are styled by the browser when given focus. For example, try tabbing around the links on http://info.cern.ch/.

srhhnry:
Sounds good!

PR #1227, opened 2023-01-30
CI: handle path-filtered but required checks

angela-tran:
This is to try and get #1225 to pass its required checks. I think the problem is that our CodeQL workflow is set to run only if `.js` or `.py` files were changed AND is a required check (for the target branch). The fix here is to add another workflow with a job with the same name that will run for the non-`.js`/`.py` files.

References
 - https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
 - https://stackoverflow.com/q/70927785


angela-tran:
Closing this because we don't want to have a no-op / workaround workflow

PR #1231, opened 2023-02-01
fix(docs): add missing system interconnections to architecture diagram

afeld:
Noticed this while looking at the docs. I _think_ this brings it up to date.

Screenshot 2023-01-31 at 8 44 50 PM


angela-tran (approved):
Looks good!

Issue #861, opened 2022-08-11
Get Azure exception monitoring working again


thekaveman:
This is happening on `dev-benefits.calitp.org`

```
2022-08-11T20:28:44.989662609Z [11/Aug/2022 20:28:44] ERROR django.request:241 Internal Server Error: /
2022-08-11T20:28:44.989680509Z Traceback (most recent call last):
2022-08-11T20:28:44.989686609Z   File "/home/calitp/.local/lib/python3.10/site-packages/django/core/handlers/exception.py", line 55, in inner
2022-08-11T20:28:44.989691709Z     response = get_response(request)
2022-08-11T20:28:44.989696209Z   File "/home/calitp/.local/lib/python3.10/site-packages/django/utils/deprecation.py", line 131, in __call__
2022-08-11T20:28:44.989701009Z     if self._is_coroutine:
2022-08-11T20:28:44.989705509Z AttributeError: 'OpencensusMiddleware' object has no attribute '_is_coroutine'
2022-08-11T20:28:44.991650505Z  - - [11/Aug/2022:20:28:44 +0000] "GET / HTTP/1.0" 500 8269 "-" "AlwaysOn"
```

We merged a patch update for `opencensus-ext-azure` in #850. ~Thinking we'll need to revert.~

## To Reproduce

1. Go to https://dev-benefits.calitp.org
2. Observe "Service is down" message
3. Review logs, see above message

thekaveman:
Strange that this is not happening in `test-benefits.calitp.org`, which also got the same update from #850 when we deployed to `test` with #854

angela-tran:
> Strange that this is not happening in `test-benefits.calitp.org`, which also got the same update from #850 when we deployed to `test` with #854

I think I found why it didn't happen in `test-benefits`: https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L74-L76
`ENABLE_AZURE_INSIGHTS` is false with the current `test` slot configuration.

thekaveman:
Yep just noticed this, we don't have any of the `INSIGHTS` settings configured in `test`:

![image](https://user-images.githubusercontent.com/1783439/184240775-2cef0a26-44bb-4cc1-887a-9a023d68b6c2.png)

thekaveman:
Removed the `APPLICATIONINSIGHTS_CONNECTION_STRING` from the `dev` and `prod` slots for now, in order to get the app live. Kept a back up locally.

afeld:
[...]

PR #1260, opened 2023-02-11
fix Azure exception handling

afeld:
Fixes https://github.com/cal-itp/benefits/issues/861.

I had initially thought there was an issue with OpenCensus compatibility overall, but it appears the (remaining) problem was with our custom middleware. Set the `APPLICATIONINSIGHTS_CONNECTION_STRING` locally and was able to see errors show up in Azure:

Screenshot 2023-02-11 at 4 07 31 PM

### DIFF
--- a/benefits/core/middleware.py
+++ b/benefits/core/middleware.py
@@ -162,11 +162,17 @@ class LoginRequired(MiddlewareMixin):
 
 
 # https://github.com/census-instrumentation/opencensus-python/issues/766
-class LogErrorToAzure(MiddlewareMixin):
+class LogErrorToAzure:
     def __init__(self, get_response):
         self.get_response = get_response
         # wait to do this here to be sure the handler is initialized
         self.azure_logger = logging.getLogger("azure")
+
+    def __call__(self, request):
+        # boilerplate
+        # https://docs.djangoproject.com/en/4.1/topics/http/middleware/#writing-your-own-middleware
+        response = self.get_response(request)
+        return response
 
     def process_exception(self, request, exception):
         # https://stackoverflow.com/a/45532289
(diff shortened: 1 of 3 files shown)


afeld:
Moving this to Draft state until we discuss at [the next dev meeting](https://docs.google.com/document/d/1vActZALSbRwO0HxDvm-YQK37p22SPbL4XuMdcRx-BI8/edit#heading=h.9ra1yb7sizl2).

afeld:
On second thought, let's get this merged in sooner so that I can show a demonstration of the two side by side. We can always re-disable it if we want.

benefits/core/middleware.py
```diff
@@ -164,7 +164,7 @@ def process_view(self, request, view_func, view_args, view_kwargs):
 # https://github.com/census-instrumentation/opencensus-python/issues/766
 class LogErrorToAzure(MiddlewareMixin):
     def __init__(self, get_response):
-        self.get_response = get_response
+        super().__init__(get_response)
```
afeld, line 167: This fixes the `_is_coroutine` error for this middleware.

benefits/settings.py
```diff
@@ -84,7 +84,7 @@ def _filter_empty(ls):
 # only used if enabled above
 OPENCENSUS = {
     "TRACE": {
-        "SAMPLER": "opencensus.trace.samplers.ProbabilitySampler(rate=1)",
+        "SAMPLER": "opencensus.trace.samplers.AlwaysOnSampler()",
```
angela-tran, line 87: Can you explain the reason for this change? I'm having trouble finding documentation on the differences between these samplers / what they are meant for

afeld: I [stumbled upon it in the docs](https://github.com/census-instrumentation/opencensus-python/blob/master/docs/trace/usage.rst#samplers). Basically, if you have a lot of traffic, you don't want to be sending data to Azure every time. The `ProbabilitySampler` allows you to specify a ratio. Since we had `rate=1`, it's the same as `AlwaysOn`.

angela-tran (approved):

afeld:
Set `APPLICATIONINSIGHTS_CONNECTION_STRING` to re-enable in `dev`.

PR #1275, opened 2023-02-15
fix(Sentry): handle cases where Git is available but we're not in a repository

afeld:
I guess I didn't test this case, where `git` is available but we're not in a repository, which is the case in App Service.


thekaveman (approved):
Should we have some pytests for this?

afeld:
Will think about how best to do that as a follow-up.

Issue #724, opened 2022-06-22
Cypress: refactor UI tests and re-enable CI runs


thekaveman:
In #721 we turned off Cypress test runs in CI. This is the follow-up task to address more of the underlying issues, and to get the tests back up and running in CI again.

## Acceptance Criteria

- [x] Cypress tests are running (and passing) in GitHub Actions
- [ ] *more to come soon*

## Related

* #681 
* #813

machikoyasuda:
Decision as of July 2022: Will revisit this ticket and Cypress usage overall when we have new features to build. For the time being, the IAL2 Login.gov flow is not conducive to Cypress testing. Instead, we will continue to increase test coverage with Pytest (see #680 etc.)

afeld:
Is this a duplicate of https://github.com/cal-itp/benefits/issues/340?

machikoyasuda:
@afeld These two tickets are slightly different:

This ticket refers to turning off Cypress UI tests (all the specs in this `/ui/` directory https://github.com/cal-itp/benefits/tree/dev/tests/cypress/specs/ui) and #340 refers to the specs in the `/feature/` directory. The UI tests are testing to see if certain specific UI elements (buttons, text fields, etc.) appear on the correct pages. The Feature tests (like this one https://github.com/cal-itp/benefits/blob/dev/tests/cypress/specs/feature/eligibility.cy.js) actually go through several pages to complete a whole flow.

afeld:
It does seem some of the UI tests were still useful, so keeping this issue as-is. Split out an issue around accessibility: https://github.com/cal-itp/benefits/issues/1278

PR #1263, opened 2023-02-12
fix feature tests

afeld:
Fixes https://github.com/cal-itp/benefits/issues/340.

Gets the feature tests working again, and sets them to run on every `push` in CI.

PR #1277, opened 2023-02-16
fix: UI tests

afeld:
Fixes https://github.com/cal-itp/benefits/issues/724. Builds on https://github.com/cal-itp/benefits/pull/1263.

This pull request does a few things:

- Removes the Cypress tests that didn't seem that useful (just checking text)
- Gets the remaining tests passing
- Removes the distinction between the "feature" and "UI" Cypress tests, as I don't think it was useful


.github/workflows/tests-cypress.yml
```diff
@@ -18,12 +18,12 @@ jobs:
...
           docker compose up --detach client server
 
-      - name: Run feature tests
+      - name: Run Cypress tests
         uses: cypress-io/github-action@v5
         env:
           CYPRESS_baseUrl: http://localhost:8000
         with:
-          command: npm run cypress:feature
+          command: npm test
```
afeld, line 26: The change that got them all running again.

afeld:
Drat, the diff isn't picking up the renamed files. May be easier to review commit by commit.

tests/cypress/specs/eligibility-no-auth.cy.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe("Multiple Verifier, no AuthProvider: Verifier selection page spec", () => {
+  beforeEach(() => {
+    cy.visit("/");
+    helpers.selectAgency();
+  });
+
+  it("User sees two radio buttons", () => {
+    cy.get("input:radio").should("have.length", 2);
+    cy.contains("Courtesy Card");
```
thekaveman, line 13: Would this same type of `.contains()` work to simplify the [`selectCourtesyCard` helper](https://github.com/cal-itp/benefits/pull/1277/files#diff-88f268da70bf80004fffc1b39325c8693737f2269ff8a95b808ff78ffc15192eR20)? 

I guess I'm not clear on why the text "Courtesy Card" is used here, but the position of the radio button in the list is used in the other case?

afeld: Unfortunately it's not quite that simple. Here, it's simply checking whether the text is on the page, whereas to actually select the radio button, it needs to find the right `<input>`. Given that `Courtesy Card` appears as only part of the text in a `<span>` in the `<label>`, the selection isn't quite as simple.

Screenshot 2023-02-16 at 5 17 16 PM

thekaveman:
I temporarily removed the branch protection requirement for Cypress tests, since it was using the old check name (`tests-ui`) that is now changed with this PR. 

Once this is merged we can re-add the protection requirement using the new check name (`tests-cypress`).

thekaveman (approved):
I'm good with re-combining these into a single type of test. We had it that way originally, then split out the feature tests as they weren't working as well and were taking a long time to run comparatively. 

But with the simpler startup using compose and general refocus of Cypress testing, this makes a lot of sense.

thekaveman:
Updated branch protection rules for `dev`, `test`, and `prod` to require the newer `tests-cypress` check over the older one.

PR #1279, opened 2023-02-16
chore(CI): simplify starting client for UI tests

afeld:
Discovered this was possible as part of https://github.com/cal-itp/benefits/pull/1263.

### DIFF
--- a/.github/workflows/tests-ui.yml
+++ b/.github/workflows/tests-ui.yml
@@ -16,30 +16,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v3
 
-      - name: Set up Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v2
-
-      - name: Build Dockerfile
-        id: build-client-image
-        uses: docker/build-push-action@v4
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-ui.yml
```diff
@@ -16,30 +16,10 @@ jobs:
...
-        id: buildx
-        uses: docker/setup-buildx-action@v2
-
-      - name: Build Dockerfile
-        id: build-client-image
-        uses: docker/build-push-action@v4
-        with:
-          builder: ${{ steps.buildx.outputs.name }}
-          cache-from: type=gha,scope=cal-itp
-          cache-to: type=gha,scope=cal-itp,mode=max
```
afeld, line 29: FYI, I don't think the caching was working, and even if it was (in some cases?), this change actually speeds up the build:

Screenshot 2023-02-15 at 10 50 37 PM

thekaveman: If the caching isn't working, then the main deploy workflow should be looked at too, it has this exact same sequence of steps.

thekaveman (approved):

PR #1282, opened 2023-02-16
Fix: user agents for healthchecks

### LINKED ISSUE #1276: Don't send events for healthchecks
We use Amplitude to gather event analytics for specific user events in the app, like visiting a page and clicking a button. 

Amplitude recently announced a change in their billing model, away from straight per-event to a "Monthly Tracked Users" (MTU) metric. Read more on their blog post: https://help.amplitude.com/hc/en-us/articles/12217810210331

This presents a problem for us, because every time the various healthchecks (`AlwaysOn`, `Edge Health Probe`, maybe others?) ping the app, a pageview event is sent to Amplitude, and each of these pings is counted as a distinct MTU. 

## Expected behavior

- [ ] Amplitude does not receive events for views from the various healthchecks.

## Options

1. Reconfigure all the healthchecks to hit our `/healthcheck` endpoint, which already bypasses Amplitude events
2. Amend the Amplitude event sending logic to skip when e.g. certain user agents are present
3. Something else?

## Additional context

Right now within Amplitude, we filter events from certain User Agents so these don't show up in our charts etc. But they are still counted towards our plan's MTU metric:

image

Unfortunately the "Block" filters available in our plan don't allow blocking events based on anything other than event name or IP address.

angela-tran:
Closes #1276 

This PR introduces a setting called `HEALTHCHECK_USER_AGENTS` which should be a comma-separated list of the user agents whose requests should always get a response of "Healthy" so that analytics are not sent.

Note that values with spaces need to be surrounded in double quotes when setting via environment variables.


thekaveman (approved):
This is great! I was going to suggest combining this new one and the original healthcheck middleware, but thinking more I actually like them being distinct, able to test them separately, modify separately, etc.

Side comment: This is an example of a change that I would want us to deploy as a hotfix to get it out there sooner than later, not necessarily waiting for the SacRT launch (or whatever the current "big feature" is). I know this is totally unplanned so I'm OK waiting. Generally want to be able to do this though.

angela-tran:
> Side comment: This is an example of a change that I would want us to deploy as a hotfix to get it out there sooner than later, not necessarily waiting for the SacRT launch (or whatever the current "big feature" is). I know this is totally unplanned so I'm OK waiting. Generally want to be able to do this though.

I'd be ok with working on doing a hotfix deploy for this. I think I'd just need to rebase these changes off `test` and then start the [release](https://docs.calitp.org/benefits/deployment/release/) process following the flow where `dev` is not ready to deploy, but `test` is. @thekaveman Do you want me to go ahead with this?

angela-tran:
Pushed a few small commits -- updated the filename for unit tests and updated documentation

thekaveman:
@angela-tran if you're comfortable with that, sounds great!

thekaveman (dismissed):

angela-tran:
Rebasing to resolve merge conflict

angela-tran:
Working on moving these changes to use `prod` as base branch instead

angela-tran:
Closing this because it seems like the Pytest check is using a cached file or something(?) causing the tests to fail.

PR #1285, opened 2023-02-17
Fix: healthcheck user agents

### LINKED ISSUE #1276: Don't send events for healthchecks

angela-tran:
Redo of #1282 

Closes #1276 

This PR introduces a setting called `HEALTHCHECK_USER_AGENTS` which should be a comma-separated list of the user agents whose requests should always get a response of "Healthy" so that analytics are not sent.

Note that values with spaces need to be surrounded in double quotes when setting via environment variables.


thekaveman (approved):
Looks good 👍 Nice simple implementation to allow us to force certain user agents to bypass our analytics.

thekaveman (changes requested):
Sorry can we please update the version as well?

thekaveman (approved):

Issue #1242, opened 2023-02-03
Put environment variable values in Key Vault for each environment


angela-tran:
This task is to put all (or maybe just secret/sensitive?) environment variable values needed by our data migration into the Key Vault for each environment (dev, test, prod).

## Acceptance Criteria

- [x] Values / certificates have been put into each environment's Key Vault in Azure

## Additional context
Depends on #1198 (which has been done ✅ )

PR #1291, opened 2023-02-22
Fix: small issues on environment variable docs

angela-tran:
While updating documentation for #1242, I noticed some broken links and that a call-out was not formatted correctly.

| Description | Screenshot |
| --- | --- |
| Broken Links | image |
| Incorrectly formatted call-out | image |

This PR fixes those two issues.


thekaveman (approved):
👍

PR #1292, opened 2023-02-22
Deploy 2023.02.1 to test

angela-tran:
The change from 2023.02.1 being merged in is:
- a new setting to configure which user agents should only get healthcheck response

Additionally, changes from `dev` that are being merged in are:
- updates to architecture and process diagrams
- new documentation on configuring a transit agency
- Sentry configuration 
- removal of Azure error monitoring and OpenCensus
- bringing back Cypress tests


thekaveman (approved):
Thanks for getting this out!

PR #1294, opened 2023-02-24
Feat: specify app settings as code

### LINKED ISSUE #1116: specify App Service configuration as code
## Initial description

**Title: _Add WEBSITES_ENABLE_APP_SERVICE_STORAGE configuration in Terraform_**

This is a required setting to prevent the filesystem from being obscured by a mount in an Azure App Service.

The lack of this setting was causing problems during the [Eligibility Server terraforming](https://github.com/cal-itp/eligibility-server/pull/166), where it has now been defined in HCL.

This recalled the Benefits [downtime in July](https://github.com/cal-itp/benefits/issues/826) and how the setting was ultimately required to bring the environments back up. Yet it never made its way into HCL here.

### Expected behavior

- [ ] Benefits terraform defines the setting with a value of `false`

## Expanded description

To allow App Service instances to be recreated as needed with minimal downtime, we want to keep any values from needing to be manually set in the App Service configuration.

### Acceptance criteria

- [ ] Terraform destroying and recreating an App Service resource brings the app up without manual intervention

### Definition of done

- [ ] App Service configuration values either live in Terraform (non-secrets) or Key Vault (secrets), referenced by Terraform / App Service

### LINKED ISSUE #1241: Specify environment variables needed by data migrations as App Service app settings
This task is to update our Terraform definition of the App Service so that the environment variables needed by our data migration (support added in #1198) are specified as app settings on the App Service.

Secrets should be specified via Key Vault references. 

Other values that are not secret (e.g. environment-specific names) could be specified via Key Vault references, but could also be specified directly in our Terraform definition for the app service. Get clarity on which one we want to do.

#1116 can also be done in the same pull request that closes this task. (In other words, make sure to bring in all the app settings that were manually configured on the app service!)

## Acceptance Criteria
- [ ] When the App Service starts, it is able to see the most recent values for all environment variables needed

## Additional context
Note that we need to account for Azure's Key Vault caching. See https://github.com/MicrosoftDocs/azure-docs/issues/79855#issuecomment-1265664801 for details on the behavior.

We considered 3 different solutions and will try the first one.

### Options
1. Use references to keyvault + a dummy app setting that we use for triggering cache refresh - **trying this first because it is low effort and is one of the approaches with faster lead time**, even though it is a little weird
2. Use Terraform to copy keyvault values into App Service config - would require running the pipeline to refresh values, and secret values end up being evaluated and stored in multiple places
3. Have app/Docker read from keyvault directly - less flexible than environment variables

angela-tran:
Closes #1116 and closes #1241

This PR specifies our app settings as code. Settings with secret values make use of Azure Key Vault references.

### Current status of secrets in Key Vault
All data migration secrets have been added to the Key Vaults for `dev`, `test`, and `prod`.
The settings that used to be manually configured on our app services (e.g. Django settings, Azure Docker settings, etc.) have been added to the `dev` Key Vault. 

### Todo
 - [x] Add the manually configured app settings to Key Vaults for `test` and `prod`.
 - [x] Figure out how to conditionally set app settings. For example, `dev` does not need the `HEALTHCHECK_USER_AGENTS` app setting.
 - [x] Update documentation

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -64,71 +64,71 @@ resource "azurerm_linux_web_app" "main" {
     "WEBSITES_ENABLE_APP_SERVICE_STORAGE" = "false",
     "WEBSITES_PORT"                       = "8000",
 
-    "ANALYTICS_KEY" = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=analytics-key)",
+    "ANALYTICS_KEY" = local.is_dev ? null : "${local.secret_prefix}analytics-key)",
 
     # Django settings
-    "DJANGO_ADMIN"            = (local.is_prod || local.is_test) ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-admin)",
-    "DJANGO_ALLOWED_HOSTS"    = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-allowed-hosts)",
-    "DJANGO_DEBUG"            = local.is_prod ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-debug)",
+    "DJANGO_ADMIN"            = (local.is_prod || local.is_test) ? null : "${local.secret_prefix}django-admin)",
+    "DJANGO_ALLOWED_HOSTS"    = "${local.secret_prefix}django-allowed-hosts)",
+    "DJANGO_DEBUG"            = local.is_prod ? null : "${local.secret_prefix}django-debug)",
     "DJANGO_LOAD_SAMPLE_DATA" = "false",
     "DJANGO_LOG_LEVEL"        = "DEBUG",
     "DJANGO_MIGRATIONS_DIR"   = "./config",
 
-    "DJANGO_RATE_LIMIT"         = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-rate-limit)",
-    "DJANGO_RATE_LIMIT_METHODS" = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-rate-limit-methods)",
-    "DJANGO_RATE_LIMIT_PERIOD"  = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-rate-limit-period)",
+    "DJANGO_RATE_LIMIT"         = local.is_dev ? null : "${local.secret_prefix}django-rate-limit)",
+    "DJANGO_RATE_LIMIT_METHODS" = local.is_dev ? null : "${local.secret_prefix}django-rate-limit-methods)",
+    "DJANGO_RATE_LIMIT_PERIOD"  = local.is_dev ? null : "${local.secret_prefix}django-rate-limit-period)",
 
-    "DJANGO_RECAPTCHA_SECRET_KEY" = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-recaptcha-secret-key)",
-    "DJANGO_RECAPTCHA_SITE_KEY"   = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-recaptcha-site-key)",
+    "DJANGO_RECAPTCHA_SECRET_KEY" = local.is_dev ? null : "${local.secret_prefix}django-recaptcha-secret-key)",
+    "DJANGO_RECAPTCHA_SITE_KEY"   = local.is_dev ? null : "${local.secret_prefix}django-recaptcha-site-key)",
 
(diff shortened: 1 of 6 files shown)


angela-tran:
Going to rebase real quick to update a commit message

terraform/app_service.tf
```diff
@@ -53,10 +53,87 @@ resource "azurerm_linux_web_app" "main" {
...
+    "WEBSITES_PORT"                       = "8000",
+
+    "ANALYTICS_KEY" = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=analytics-key)",
+
+    # Django settings
+    "DJANGO_ADMIN"            = (local.is_prod || local.is_test) ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-admin)",
+    "DJANGO_ALLOWED_HOSTS"    = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-allowed-hosts)",
+    "DJANGO_DEBUG"            = local.is_prod ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=django-debug)",
+    "DJANGO_LOAD_SAMPLE_DATA" = "false",
+    "DJANGO_LOG_LEVEL"        = "DEBUG",
```
thekaveman, line 74: Can we make this one an env variable too? I imagine we'll want to turn this down in prod.

terraform/app_service.tf
```diff
@@ -53,10 +53,87 @@ resource "azurerm_linux_web_app" "main" {
...
+    "change_me_to_refresh_secrets" = "change me in the portal to refresh all secrets",
+
+    "DOCKER_ENABLE_CI"                    = "true",
+    "DOCKER_REGISTRY_SERVER_URL"          = "https://ghcr.io/",
+    "WEBSITE_HTTPLOGGING_RETENTION_DAYS"  = "99999",
+    "WEBSITE_TIME_ZONE"                   = "America/Los_Angeles",
+    "WEBSITES_ENABLE_APP_SERVICE_STORAGE" = "false",
+    "WEBSITES_PORT"                       = "8000",
+
+    "ANALYTICS_KEY" = local.is_dev ? null : "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=analytics-key)",
```
thekaveman, line 67: Is there a way to create a local variable for the repeated part of this string? E.g. 

`@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=`

And then fill in the secret name in each case?

terraform/app_service.tf
```diff
@@ -53,10 +53,87 @@ resource "azurerm_linux_web_app" "main" {
...
+    "PAYMENT_PROCESSOR_CARD_TOKENIZE_ENV"            = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=payment-processor-card-tokenize-env)"
+    "MST_AGENCY_SHORT_NAME"                          = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=mst-agency-short-name)"
+    "MST_AGENCY_LONG_NAME"                           = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=mst-agency-long-name)"
+    "MST_AGENCY_JWS_SIGNING_ALG"                     = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=mst-agency-jws-signing-alg)"
+    "SACRT_AGENCY_SHORT_NAME"                        = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=sacrt-agency-short-name)"
+    "SACRT_AGENCY_LONG_NAME"                         = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=sacrt-agency-long-name)"
+    "SACRT_AGENCY_MERCHANT_ID"                       = "@Microsoft.KeyVault(VaultName=KV-CDT-PUB-CALITP-${local.env_letter}-001;SecretName=sacrt-agency-merchant-id)"
```
thekaveman, line 130: We need this one because it changes per QA/prod environment right? Otherwise for MST we don't because it is the same?

thekaveman (changes requested):
Looking great. Few questions/suggestions.

terraform/app_service.tf
angela-tran: Yep, that is exactly right

terraform/app_service.tf
angela-tran: Good idea. Done in [00287f5](https://github.com/cal-itp/benefits/pull/1294/commits/00287f5e49d534a082ce321c83b3c539d348ea0d)

angela-tran: Done in [5cdc6a9](https://github.com/cal-itp/benefits/pull/1294/commits/5cdc6a9b589b13e85153be27f73335eff522848a)

terraform/app_service.tf
angela-tran: I went ahead and added the values for this to our dev, test, and prod Key Vaults.

Here's the command I ran for dev. I ran similar commands for test and prod.

angela-tran: Done in [5cdc6a9](https://github.com/cal-itp/benefits/pull/1294/commits/5cdc6a9b589b13e85153be27f73335eff522848a)

thekaveman: Ah, so much cleaner! Diffs will be easier now too I think

thekaveman (approved):
Do we want to merge this with our release this afternoon? I think we did want to, but going through each of the environments/secrets first right?

terraform/app_service.tf
thekaveman: Ah, so much cleaner! Diffs will be easier now too I think

Issue #1244, opened 2023-02-03
Delete all data migration files that are no longer needed


angela-tran:
Once #1243 is done, let's clean up all the files and Azure resources that are no longer being used.

## Tasks
 - [x] Delete data migration files from benefits-secrets and any other files that are rendered unused as a result
 - [ ] Delete data migration file from each bucket (or rename if delete is not possible)
 - [x] Delete storage mount from Terraform Azure definition
 - [x] Update documentation

## Acceptance criteria
 - [ ] We no longer maintain data migration files in benefits-secrets
 - [ ] We no longer have a storage mount nor any files in the storage account for doing data migrations
 - [ ] Documentation is updated

thekaveman:
@angela-tran once we're fully swapped over to Key Vault in all environments, assuming we don't have the necessary permissions (I don't think we do) let's ask DevSecOps to help with the deletes. I think even the buckets themselves could go away so as not to confuse anyone.

angela-tran:
I poked around the storage accounts in Azure today and found that I was able to delete files and containers. In the dev resource group, I had to delete all files within a container before I could delete the container. In test and prod resource groups though, I was able to delete at the container level even if there were files within them.

I cleaned up all the containers that used to hold fixture / data migration files. I left alone the containers that I think were used by Application Insights and of course the container that holds our Terraform state.

PR #1328, opened 2023-03-22
Refactor: delete now unneeded Azure file mount

angela-tran:
Part of #1244 

This PR removes the file mount path from the App Service.

I checked over documentation, and it does not contain any mention of the blob container or file mount path anywhere.

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -44,15 +44,6 @@ resource "azurerm_linux_web_app" "main" {
     }
   }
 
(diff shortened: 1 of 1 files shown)


angela-tran:
> Do we have any idea about this `client_certificate_exclusion_paths` entry that is marked for deletion in the `plan`?

Hmm, yeah I am not sure where that is coming from. I think it might be ok since it's just going from an empty string to null.

PR #1342, opened 2023-03-31
Fix: read timeout

angela-tran:
@thekaveman and I noticed a [`ReadTimeout`](https://sentry.calitp.org/organizations/sentry/issues/69840/?query=is%3Aunresolved&referrer=issue-stream) while testing #1341.
```
ReadTimeout: HTTPSConnectionPool(host='*****', port=443): Read timed out. (read timeout=1)
  File "django/core/handlers/exception.py", line 56, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/utils/decorators.py", line 134, in _wrapped_view
    response = view_func(request, *args, **kwargs)
```

This PR increases the read timeout to 20s. 

To avoid a `ValueError` caused by incorrect types, this PR also converts the `requests` timeout environment variable values to `int`s and falls back to default values if the environment variable value is `None` or otherwise fails to type-convert.


thekaveman (approved):
This looks good to me

PR #1343, opened 2023-03-31
Fix: configure Sentry denylist

### LINKED ISSUE #1340: Configure a Sentry denylist
Looks like we can configure a `denylist` on `EventScrubber` when calling `sentry_sdk.init`: https://docs.sentry.io/platforms/python/data-management/sensitive-data/

Another more general approach to modifying events is to configure a `before_send` function: https://docs.sentry.io/platforms/python/configuration/filtering/

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/issues/1334#issuecomment-1490416579_

angela-tran:
Closes #1340 

This PR configures a custom denylist on our `EventScrubber`, with `sub` and `name` as the additional entries.

### DIFF
--- a/benefits/sentry.py
+++ b/benefits/sentry.py
@@ -1,9 +1,12 @@
-from benefits import VERSION
-import sentry_sdk
-from sentry_sdk.integrations.django import DjangoIntegration
 import shutil
 import os
 import subprocess
+
+import sentry_sdk
+from sentry_sdk.integrations.django import DjangoIntegration
+from sentry_sdk.scrubber import EventScrubber, DEFAULT_DENYLIST
+
+from benefits import VERSION
 
 
 SENTRY_ENVIRONMENT = os.environ.get("SENTRY_ENVIRONMENT", "local")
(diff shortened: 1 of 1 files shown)


thekaveman (changes requested):
This looks great! I'm reading the docs and they say that for this custom `event_scrubber` to work, `send_default_pii` must be `False` -- (and it is by default!)

But I'd like us to be explicit here, would you mind adding that to the `init` and then also inline-link with a comment to that docs page: https://docs.sentry.io/platforms/python/data-management/sensitive-data/#event_scrubber

angela-tran:
Thanks @thekaveman , those were good suggestions. I put the docs link close to the two parameters, but if you think it'd be better somewhere else, let me know!

benefits/sentry.py
```diff
@@ -68,6 +77,9 @@ def configure():
             environment=SENTRY_ENVIRONMENT,
             release=release,
             in_app_include=["benefits"],
+            # https://docs.sentry.io/platforms/python/data-management/sensitive-data/#event_scrubber
```
thekaveman, line 80:
```suggestion
            # send_default_pii must be False (the default) for a custom EventScrubber/denylist
            # https://docs.sentry.io/platforms/python/data-management/sensitive-data/#event_scrubber
```

thekaveman (changes requested):
Small additional note

thekaveman (approved):
👍

PR #1359, opened 2023-04-14
Feat: improve traffic filtering at nginx

### LINKED ISSUE #502: Offload rate limiting
Currently, we have [a rate limiter](https://github.com/cal-itp/benefits/blob/468a4dfcc9ec7051631a544277c0c2b8d89a8448/benefits/core/middleware.py#L32-L66) that we need to maintain. There are other options we could leverage that may be more robust, and would be less code in our codebase. Examples:

- [Django Ratelimit](https://django-ratelimit.readthedocs.io/en/stable/)
- [NGINX](https://www.nginx.com/blog/rate-limiting-nginx/)

They may or may not be configurable to key off of the session ID, but maybe that's fine…? Forgive me if this was discussed before!

thekaveman:
As part of reducing the unnecessary traffic we send to Sentry (see [Slack thread](https://cal-itp.slack.com/archives/C044UUY78TB/p1681423801273219) for more context) and Amplitude (see e.g. #1285 for some prior work), this PR adds filtering at the `nginx` layer - before traffic even hits the Django app where Sentry, Amplitude, etc. come into play.

Based on a quick analysis of both Amplitude and Sentry, immediately 404 known scraping targets like `*.php` and `/api/whatever`.

I also took the opportunity to close #502 by moving rate limiting into `nginx`.

## Notes for reviewers

- [ ] Reminder: only the _appcontainer_ runs `nginx`, the devcontainer uses a local dev server
- [ ] New `cypress` tests for the scraper targets, or trigger 404 manually in the browser
- [ ] Rate limit still only applies to POST on `/eligibility/confirm`, the `cypress` tests have been updated


angela-tran:
👀 Finished code reviewing this. Still need to do manual testing

> ## Notes for reviewers

- [x] Reminder: only the _appcontainer_ runs `nginx`, the devcontainer uses a local dev server
- [x] New `cypress` tests for the scraper targets, or trigger 404 manually in the browser
- [x] Rate limit still only applies to POST on `/eligibility/confirm`, the `cypress` tests have been updated

thekaveman:
Sorry I realized that there was a bug in one of the regexes, it wouldn't have matched `/.env` (the filename is just the extension) - fixed that.

Then I also wanted to clarify in the tests that the 404 actually does come from nginx and not Django.

angela-tran:
I was able to manually verify that incorrect paths like `/.env` are given a 404 response by nginx.

Also manually verified that the rate-limiting from nginx works:
image

tests/cypress/specs/rate-limit.cy.js
```diff
@@ -1,64 +1,62 @@
...
           post_confirm(csrfmiddlewaretoken, sub, name, false).then((res) => {
-            if (i < RATE_LIMIT) {
-              // allow up to API_RATE_LIMIT requests
+            // the first request should succeed, subsequent should fail
+            if (i == 0) {
               expect(res.status).to.eq(SUCCESS_STATUS);
             } else {
-              // we've gone beyond API_RATE_LIMIT
-              expect(res.status).to.eq(400);
+              expect(res.status).to.eq(FAIL_STATUS);
```
angela-tran, line 34: I think the nginx rate-limiting configuration / implementation is working, but these tests may have a bug.

I noticed the Cypress test log wasn't showing me any `expect` commands. Digging into it more, I see `helpers.RATE_LIMIT` is undefined, so the loop is a no-op.

image

Added a log statement:
image

image

tests/cypress/specs/scrapers.cy.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const NOT_FOUND = 404;
+
+describe("Scraper filtering spec", () => {
+  targets.forEach((target) => {
+    it(`nginx returns 404 for known scraper target pattern: /${target}`, () => {
+      visit(target).then((res) => {
+        expect(res.status).to.eq(NOT_FOUND);
+        expect(res.body).to.contain("nginx");
```
angela-tran, line 21: I'm running Benefits using `docker compose up client`.

This line fails for me, but I don't see how my set up would have a different result from the GitHub Action run which passed: https://github.com/cal-itp/benefits/actions/runs/4704714930/jobs/8344585869#step:4:211

If it passes for you @thekaveman , maybe we don't worry about my failure?

image

thekaveman:
🤔 for the rate limit (just like the 404) we shouldn't be seeing a Benefits error page. The rate limit happens before the request gets to Django, and returns a 503. It should be an nginx error page?

thekaveman:
I'm going to rebase on `dev` to clean up merge conflicts

angela-tran:
> 🤔 for the rate limit (just like the 404) we shouldn't be seeing a Benefits error page. The rate limit happens before the request gets to Django, and returns a 503. It should be an nginx error page?

Oh hmm. So I should be seeing an nginx error page for both the 404 and 503? I was seeing Benefits error pages.

thekaveman:
> > 🤔 for the rate limit (just like the 404) we shouldn't be seeing a Benefits error page. The rate limit happens before the request gets to Django, and returns a 503. It should be an nginx error page?
> 
> Oh hmm. So I should be seeing an nginx error page for both the 404 and 503? I was seeing Benefits error pages.

Exactly. See the modification to the `scrapers.spec.js` in this commit: https://github.com/cal-itp/benefits/pull/1359/commits/1f33670b58dcb0498e9e70dde7fbc9b21e4a1b29 to ensure `nginx` is throwing the 404, not Django.

I'm going to modify the `rate-limit.spec.js` as well.

angela-tran (commented):
I ran into some issues while trying out the Cypress tests. I don't think they necessarily point out problems with the actual code refactor, but rather are bugs in the tests / local environment issues.

tests/cypress/specs/scrapers.cy.js
thekaveman: One thing I notice is you are running on `localhost:11369` -- seems like your `.env` with secrets is still coming into play.

Try this:

```bash
mv .env .env.bak
touch .env
docker compose up -d client
```

Then run tests.

tests/cypress/specs/rate-limit.cy.js
thekaveman: Definitely seems like a bug. Not sure, I had this working last week, it was definitely running the rate limit test (took multiple minutes with the `cy.wait()` -- but now it's finishing in under 10 seconds so not running!

tests/cypress/specs/scrapers.cy.js
thekaveman: I recall that a reCAPTCHA config breaks the Cypress tests.

tests/cypress/specs/rate-limit.cy.js
thekaveman: OK this was my fault, I had it working when those `RATE_LIMIT` consts where defined in the same spec file. I forgot to `export` them from the `helpers.js` file though, hence not being available.

Fix incoming.

tests/cypress/specs/scrapers.cy.js
angela-tran: Nice catch! I rebuilt `client`, and now these tests pass ✅ 

image

thekaveman:
OK this is the runtime I expected to see: https://github.com/cal-itp/benefits/actions/runs/4736775820/jobs/8408740937

image

tests/cypress/specs/scrapers.cy.js
thekaveman: Just a note that you don't actually have to rebuild if all you're doing is changing the `.env` -- you can just `docker compose down` and `docker compose up` to recreate the _container_ but using the same _image_.

`.env` only comes into play when _running the container_.

tests/cypress/specs/rate-limit.cy.js
angela-tran: Do you have any ideas on why my test is failing to get a 302 for the first request? Your screenshot of the GitHub Action run shows that this should work...

image

thekaveman: To mirror the GitHub Action:

* Make sure `.env` is empty:

  ```
  ```
  
* Recreate both services:

```bash
docker compose down
docker compose up -d client server
```

* Run cypress tests _with the correct base URL_:

```bash
cd tests/cypress
CYPRESS_baseUrl=http://localhost:8000 npx cypress run
```

tests/cypress/specs/scrapers.cy.js
angela-tran: Got it, makes sense 👍

tests/cypress/specs/rate-limit.cy.js
angela-tran: Thanks! I got it to pass

angela-tran (approved):
Ran through the tests again on rebased commits. Looks good!

PR #1380, opened 2023-04-25
Chore: delete unneeded storage container

angela-tran:
I missed this code when doing https://github.com/cal-itp/benefits/issues/1244. This container was used to hold configuration (data migration) file and is no longer needed.


thekaveman (approved):
Confirmed these are all empty for each environment 👍

PR #1390, opened 2023-05-02
Refactor: agency links context

### LINKED ISSUE #1054: Make agency links context explicit
This is a follow-up to #512 and its PR #1042.

## The issue

From the description in #1042:

> For Agency Links... further changes / future ticket is necessary to remove these lines: https://github.com/cal-itp/benefits/pull/1042/files#diff-1a4430de237c046c679071ef0d2a5b267007651b8fa1c42e755f0a1cea4d469cR2 https://github.com/cal-itp/benefits/pull/1042/files#diff-32d0e53bd73778cc6c0437f522dd3269a366d343d321f0d9eee21d3ec3d4d166R20 (Probably needs a ticket to refactor the Page view model to make it easier to pass in the agency links differently from a Back button, or something like that)

One problem with the current implementation of these agency links is how they are interwoven with the rest of the `page.buttons`. I think this is an artifact of our previous usage of the `page.html` template where everything _had_ to be in the `page` viewmodel. But it makes for baking in weird filtering logic to the templates, like this one from the [include](https://github.com/cal-itp/benefits/blob/dev/benefits/core/templates/core/includes/agency-links.html):

```html
<div class="buttons-agencies">
  {% for agency_button in buttons %}
    {% if "agency" in agency_button.classes %}
      <!-- render only if the button has the class "agency" -->
    {% endif %}
  {% endfor %}
</div>
```

Even worse when a template has to filter for both regular buttons and these agency links.

## Proposed solution

These days we have specific `something.html` templates that have more context / control over rendering and we don't have to rely on the page viewmodel as much (could be a follow-up to make that viewmodel more relevant / flexible, but alas...) 

In #1025 (following a pattern from elsewhere) we added the previous page button to the template context explicitly, only in the views where it was needed:

```python
ctx = page.context_dict()
ctx["previous_page_button"] = viewmodels.Button.previous_page(url=...)
```

I think we can do something similar for the agency links, to make them explicit and distinct from "other buttons"

```python
ctx = page.context_dict()
ctx["agency_links"] = viewmodels.Button.agency_contact_links(agency)
```

Then adjust the include accordingly:

```html
{% if agency_links %}
<div class="buttons-agencies">
  {% for button in agency_links %}
      <!-- render -->
  {% endfor %}
</div>
{% endif %}
```

angela-tran:
Closes #1054 

This PR implements the proposed solution on the ticket so that for the three pages that show agency links (help, unverified, and retry):
- we remove the [`{% if "agency" in agency_button.classes %}`](https://github.com/cal-itp/benefits/pull/1042/files#diff-1a4430de237c046c679071ef0d2a5b267007651b8fa1c42e755f0a1cea4d469cR2) check
- we remove the [`{% if "btn-primary" in back_button.classes %}`](https://github.com/cal-itp/benefits/pull/1042/files#diff-32d0e53bd73778cc6c0437f522dd3269a366d343d321f0d9eee21d3ec3d4d166R20) check

### Testing locally
`/help` and `/unverified` are easy to test by going through the flow (use Courtesy Card with ineligible user credentials).

To test the `/enrollment/retry` view, you can modify the code as follows. Then, make sure to populate an agency in your session (either through the agency selector modal or going directly to an agency's slug) so that you can go directly to the endpoint and see that agency's contact links.
```diff
-@decorator_from_middleware(EligibleSessionRequired)
+# @decorator_from_middleware(EligibleSessionRequired)
    """View handler for a recoverable failure condition."""
-    if request.method == "POST":
+    if request.method == "GET":
```

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -247,17 +247,17 @@ def unverified(request):
     analytics.returned_fail(request, types_to_verify)
 
     agency_links = viewmodels.Button.agency_contact_links(agency)
-    buttons = viewmodels.Button.home(request)
+    back_button = viewmodels.Button.home(request)
 
     page = viewmodels.Page(
         title=_(verifier.unverified_title),
         headline=_("eligibility.pages.unverified.headline"),
         icon=viewmodels.Icon("idcardquestion", pgettext("image alt text", "core.icons.idcardquestion")),
         paragraphs=[_(verifier.unverified_blurb)],
-        buttons=buttons,
     )
 
     ctx = page.context_dict()
     ctx["agency_links"] = agency_links
+    ctx["back_button"] = back_button
 
     return TemplateResponse(request, TEMPLATE_UNVERIFIED, ctx)
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -246,9 +246,8 @@ def unverified(request):
 
     analytics.returned_fail(request, types_to_verify)
 
-    # tel: link to agency phone number
-    buttons = viewmodels.Button.agency_contact_links(agency)
-    buttons.append(viewmodels.Button.home(request))
+    agency_links = viewmodels.Button.agency_contact_links(agency)
+    buttons = viewmodels.Button.home(request)
 
     page = viewmodels.Page(
         title=_(verifier.unverified_title),
@@ -258,4 +257,7 @@ def unverified(request):
         buttons=buttons,
     )
 
-    return TemplateResponse(request, TEMPLATE_UNVERIFIED, page.context_dict())
+    ctx = page.context_dict()
+    ctx["agency_links"] = agency_links
+
+    return TemplateResponse(request, TEMPLATE_UNVERIFIED, ctx)
--- a/benefits/core/templates/core/help.html
+++ b/benefits/core/templates/core/help.html
@@ -66,15 +66,13 @@
     </div>
     {% endif %}
 
-    {% for back_button in page.buttons %}
-      {% if "btn-primary" in back_button.classes %}
-        <div class="row pt-8">
-          <div class="col-lg-3 offset-lg-8 col-8 offset-2">
-            {% include "core/includes/button.html" with button=back_button %}
-          </div>
-        </div>
-      {% endif %}
-    {% endfor %}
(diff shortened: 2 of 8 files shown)


machikoyasuda:
## Testing locally - all looks good

- Help page without choosing an agency
image

- Help page after selecting MST
image

- Unverified
image

## Testing locally - `enrollment/retry`

image

I changed my local code at https://github.com/cal-itp/benefits/blob/145ebc126b48fb78739c61557892c3feb0e1cedc/benefits/enrollment/views.py#LL146C2-L146C2 according to your notes, and then routing to http://localhost:11369/enrollment/retry directly. I get the screenshot above. No agency links. Not sure if this is a bug on my configuration side, or the PR itself

benefits/eligibility/views.py
```diff
@@ -246,16 +246,18 @@ def unverified(request):
...
         headline=_("eligibility.pages.unverified.headline"),
         icon=viewmodels.Icon("idcardquestion", pgettext("image alt text", "core.icons.idcardquestion")),
         paragraphs=[_(verifier.unverified_blurb)],
-        buttons=buttons,
     )
 
-    return TemplateResponse(request, TEMPLATE_UNVERIFIED, page.context_dict())
+    ctx = page.context_dict()
+    ctx["agency_links"] = agency_links
+    ctx["back_button"] = back_button
```
machikoyasuda, line 261: Great solution 🎊 So much better than my hacky past logic!

angela-tran:
> I get the screenshot above. No agency links. Not sure if this is a bug on my configuration side, or the PR itself

Ah, yeah I forgot to mention a required step. You'll want to have an `agency` in your session, so that means you should start from the homepage, choose a transit provider, and then go to `/enrollment/retry`.

machikoyasuda:
@angela-tran Got it! I went through the flow properly (entered in an eligible Courtesy Card user) and got to the LP screen, then manually went to http://localhost:11369/enrollment/retry and it worked:

image

machikoyasuda (approved):

thekaveman (approved):
Agreed this code looks much clearer! Nice work 👍

angela-tran:
Thanks for the reviews @machikoyasuda and @thekaveman! ☺️

PR #1439, opened 2023-06-22
Feat: veterans auth provider

angela-tran:
Closes #1428 

Defines an `AuthProvider` object meant for use with the veteran flow. Dummy default values are used for the client ID, authority, scope, and claim.

Updates app settings in Terraform to reflect different secrets for the two `AuthProvider` objects.

### To test locally
Set these environment variables

> AUTH_PROVIDER_CLIENT_ID=`<client ID from Key Vault>`
> AUTH_PROVIDER_AUTHORITY=`<IdG URL from Key Vault>`
> SENIOR_AUTH_PROVIDER_SCOPE=`<scope from Key Vault>`
> SENIOR_AUTH_PROVIDER_CLAIM=`<claim from Key Vault>`
> VETERAN_AUTH_PROVIDER_SCOPE=<value from https://cal-itp.slack.com/archives/C037Y3UE71P/p1685556379722939?thread_ts=1684790050.367569&cid=C037Y3UE71P>
> VETERAN_AUTH_PROVIDER_CLAIM=<value from https://cal-itp.slack.com/archives/C037Y3UE71P/p1685556379722939?thread_ts=1684790050.367569&cid=C037Y3UE71P>

### Before merging
 - [ ] Set secrets for senior- and veteran-specific auth provider values in dev and test

PR #1458, opened 2023-06-27
Eligibility Start: Show Veteran-flow specific bullet copy

### LINKED ISSUE #1441: Eligibility Start: Bullet copy for each EligibilityVerifier
This issue is to update the copy for the highlighted bulleted section:

image

This bulleted section's copy is not stored in `EligibilityVerifier`, it's called in https://github.com/cal-itp/benefits/blob/5fba58473d191fb2911fee32af048311b7f2f7ff/benefits/eligibility/views.py#L111-L117

I think it should be covered in a separate issue and PR, b/c there's a bit going on.

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/issues/1427#issuecomment-1600277695_

machikoyasuda:
closes #1441 

This PR does not change any CSS. It looks like some indentation has changed on this screen.

## What this PR does
- [x] Add English and Spanish copy for bullets
- [x] STUB OUT logic for deciding between Senior or Veteran (Waiting on #1439 to be merged into `main`)
- [ ] Change `uses_auth_verification` to optionally take a claim (`"senior"` or `"veteran"`) to determine which copy to return to context

## Screenshots
image

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -110,12 +110,14 @@ def start(request):
     )
 
     if verifier.is_auth_required:
-        if verifier.uses_auth_verification:
+        if verifier.name == "OAuth claims via Login.gov (MST)" or "OAuth claims via Login.gov (SacRT)":
             identity_item.bullets = [
                 _("eligibility.pages.start.login_gov.required_items[0]"),
                 _("eligibility.pages.start.login_gov.required_items[1]"),
                 _("eligibility.pages.start.login_gov.required_items[2]"),
             ]
+        if verifier.name == "VA.gov - Veteran (MST)":
+            identity_item.bullets = ["Login.gov", "ID.me", "DS Logon", "My HealtheVet"]
 
         if not session.logged_in(request):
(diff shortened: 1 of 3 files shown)


benefits/eligibility/views.py
```diff
@@ -110,12 +110,14 @@ def start(request):
     )
 
     if verifier.is_auth_required:
-        if verifier.uses_auth_verification:
```
machikoyasuda, line 113: @thekaveman @angela-tran Thoughts on using this approach for changing this method:

I was thinking of adding an optional string parameter to this method, of a claim (like `veteran` or `senior`), to distinguish between Login.gov (MST), Login.gov (SacRT), and VA.gov. 

This method is only used 1 other time.

Orrrrr, should I create a new method entirely like, `uses_claim_for` that has a required parameter of a string claim, that can be used like this: `uses_claim_for("veteran")` and `uses_claim_for("senior")`?

machikoyasuda:
Abandoning this conflicted PR.

PR #1462, opened 2023-06-28
Refactor: AuthProvider supports_sign_out

### LINKED ISSUE #1456: Refactor sign-out link logic for AuthProvider's that support it
Only some `AuthProvider` support sign-out:

* Login.gov for the senior flow does, because the IdG is talking directly to Login.gov
* VA.gov for the veteran flow does not, because the IdG is talking to VA.gov and they are talking to 1 of 4 different IdPs, which is unknowable to Benefits

## Acceptance Criteria

- [ ] The `Login.gov` branded sign out button and sign out link only displays for the senior flow with Login.gov verification

## Additional context

[Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1687820027644899?thread_ts=1687815526.509399&cid=C037Y3UE71P) and https://github.com/cal-itp/benefits/pull/1439#issuecomment-1608453472

angela-tran:
Closes #1456


thekaveman:
Just an early thought, feel free to disregard -- maybe instead of an explicit field (causing the test/migration to fail), maybe a calculated field (property) based on the presence or absence of the sign out button text?

angela-tran:
> Just an early thought, feel free to disregard -- maybe instead of an explicit field (causing the test/migration to fail), maybe a calculated field (property) based on the presence or absence of the sign out button text?

Yeah, that will make things simpler 👍

benefits/core/migrations/0002_data.py
```diff
@@ -139,7 +139,7 @@ def load_data(app, *args, **kwargs):
 
     veteran_auth_provider = AuthProvider.objects.create(
         sign_in_button_label=_("eligibility.buttons.veteran.signin"),
-        sign_out_button_label=_("eligibility.buttons.veteran.signout"),
```
machikoyasuda, line 142: Remove this line from `django.po` too?

machikoyasuda (dismissed):

thekaveman (approved):
I think we could use a couple quick [Model tests](https://github.com/cal-itp/benefits/blob/dev/tests/pytest/core/test_models.py) for this property, but I won't block for it.

benefits/enrollment/views.py
```diff
@@ -179,7 +179,7 @@ def success(request):
 
     page = viewmodels.Page(title=_("enrollment.pages.success.title"), headline=_("enrollment.pages.success.headline"))
 
-    if verifier.is_auth_required and session.logged_in(request):
+    if verifier.is_auth_required and session.logged_in(request) and verifier.supports_sign_out:
```
thekaveman, line 182: Since `verifier.supports_sign_out` already checks `verifier.is_auth_required`:

```suggestion
    if verifier.supports_sign_out and session.logged_in(request):
```

benefits/core/templates/core/includes/sign-out-link.html
```diff
@@ -2,7 +2,7 @@
 {% load i18n %}
 
 {% if authentication %}
-  {% if authentication.required and authentication.logged_in %}
+  {% if authentication.required and authentication.logged_in and authentication.supports_sign_out %}
```
thekaveman, line 5:
```suggestion
  {% if authentication.supports_sign_out and authentication.logged_in %}
```

angela-tran:
Rebased to resolve merge conflicts

benefits/enrollment/views.py
angela-tran: Thanks! Done in https://github.com/cal-itp/benefits/pull/1462/commits/3fc8ee11708d80beb8e1de32774c4ac07c9ff84b

benefits/core/templates/core/includes/sign-out-link.html
angela-tran: Done in https://github.com/cal-itp/benefits/pull/1462/commits/3fc8ee11708d80beb8e1de32774c4ac07c9ff84b

angela-tran:
> I think we could use a couple quick [Model tests](https://github.com/cal-itp/benefits/blob/dev/tests/pytest/core/test_models.py) for this property, but I won't block for it.

I could leave #1456 open and add tests in a separate PR to then close the ticket.

thekaveman (approved):
🚀

PR #1465, opened 2023-06-30
Feat: configure OIDC client to prompt for reauthentication

angela-tran:
Part of #1457

PR #1467, opened 2023-06-30
Feat: reset OAuth for new flows

### LINKED ISSUE #1457: Clear OAuth token on eligibility/index
Upon launching the Veterans flow, we'll have 2 separate flows that use the OAuth infrastructure.

Because of the way the IdG interacts with VA.gov, we can't sign the user out all the way up the stack to VA.gov's identity provider. We'll deal with this from a UI perspective in #1456.

However for a user that e.g. fails verification in one flow and would like to try the other, we need to clear the app's local OAuth token, which is provided by the IdG, in between those attempts -- otherwise the app will the think the user has already signed in with IdG, and will re-use the old token with incorrect/missing claim(s).

Proposal is to reset the token on all loads of the `eligibility/index` view. This will change the behavior for successful verifications, in that users who return to the `eligibility/index` view after having been verified currently can pass-through verification to enrollment. After this ticket is implemented, users will always have to click to sign-in again (but may be able to skip the full Login.gov sign-in process as they would not be fully signed-out from there unless they clicked the "Sign out" button in the app). 

See Slack conversation for more context: https://cal-itp.slack.com/archives/C037Y3UE71P/p1687889544724039

thekaveman:
Closes #1457 

Supersedes #1465

### DIFF
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -147,8 +147,8 @@ def logged_in(request):
 
 
 def logout(request):
-    """Reset the session tokens."""
-    update(request, oauth_token=False, enrollment_token=False)
+    """Reset the session claims and tokens."""
+    update(request, oauth_claim=False, oauth_token=False, enrollment_token=False)
 
 
 def oauth_token(request):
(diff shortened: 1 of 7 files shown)


angela-tran (approved):
This looks great 💯

Issue #1473, opened 2023-07-03
Eligibility Index - Modal for Login.gov/Eligibility (Senior)


thekaveman:
- [x] shorten radio button title to `Older Adult`
- [x] shorten radio button blurb
- [x] add new Login.gov logo image
- [x] add modal trigger (Login.gov image logo link) to the Older Adult radio button widget description
- [x] Create modal
- [x] ~Remove Help page link~ done in #1472 / #1517 
- Copy has to go into EligibiltyVerifier/Form model
- Login.gov logo must have a text-backup.
- Don't do the External Link icon yet.

## Modal trigger

![Image](https://user-images.githubusercontent.com/3673236/248151500-7cf2248d-4cc7-4c5e-9bb1-11396cb43e41.png)

## Modal

![Image](https://user-images.githubusercontent.com/3673236/248151629-4ef43211-d9d0-40e6-915b-6989fd2066f8.png)

PR #1489, opened 2023-07-06
Refactor: verifier radio template

angela-tran:
Part of #1473 

Refactors the `VerifierRadioSelect` to expect a `selection_label_template` to be given for each verifier, and uses that to render the label and description.

This is solely a code refactor and does not change the front-end result.

| Before (dev-benefits.calitp.org) | After (localhost) |
| --- | --- |
|  |  |


machikoyasuda (approved):
LGTM!

I tested out changing the includes locally: 
image


@thekaveman Maybe on Monday we can discuss naming/folder conventions for all these includes from this PR and https://github.com/cal-itp/benefits/pull/1498/. There's both hyphen, underscore going on now. Not necessary to change in these PRs, but something to think about in a future refactor ticket.

image
image

thekaveman (approved):
Confirming the rendered HTML for a single radio looks the same as well:

**Dev**

```html
<div class="radio-input-group d-flex">
  <input class="radio-input rounded-circle flex-grow-0 flex-shrink-0"
         type="radio"
         name="verifier"
         value="4"
         required=""
         id="id_verifier_0">
  <label for="id_verifier_0" class="radio-label">
    <span class="d-block h3">I am 65 years of age or older</span>
    <span class="d-block pt-1">
         You must be 65 years or older. This benefit does not expire,
         but you may need to renew. Using this benefit means your
         new transit fare is half of the standard fare.
      </span>
  </label>
</div>
```

**This PR**

```html
<div class="radio-input-group d-flex">
<input class="radio-input rounded-circle flex-grow-0 flex-shrink-0"
       type="radio"
       name="verifier"
       value="4"
       required=""
       id="id_verifier_0">
  <label for="id_verifier_0" class="radio-label">
    <span class="d-block h3">I am 65 years of age or older</span>
    <span class="d-block pt-1">    
        You must be 65 years or older. This benefit does not expire,
        but you may need to renew. Using this benefit means your
        new transit fare is half of the standard fare.
    </span>
  </label>
</div>
```

_with spacing adjusted in each_

Issue #1483, opened 2023-07-03
Buttons: Lowercase pronoun in all Buttons


thekaveman:
Choose Your Provider -> Choose your Provider
Choose This Benefit -> Choose this Benefit
Connect Your Card -> Connect your Card
Check Status -> Check status

machikoyasuda:
Confirmed on dev: https://dev-benefits.calitp.org/



![Image](https://github.com/cal-itp/benefits/assets/3673236/24c672f7-d8cf-47f2-9b8e-8152b86f445f)


![Image](https://github.com/cal-itp/benefits/assets/3673236/3a802da2-c71a-4b14-bdb6-391d5e026b03)




![Image](https://github.com/cal-itp/benefits/assets/3673236/faf438f5-871a-41b2-a215-fe2f8d1e75ae)



![Image](https://github.com/cal-itp/benefits/assets/3673236/457eb6c4-80a9-4fd8-b33b-70313cbd2c0b)



![Image](https://github.com/cal-itp/benefits/assets/3673236/bf87f259-032a-4adb-a5a2-b91bc2d0ecd6)



![Image](https://github.com/cal-itp/benefits/assets/3673236/7231a74f-e9de-48c0-a4fd-b62044f6fcd6)



![Image](https://github.com/cal-itp/benefits/assets/3673236/303b9b14-da2a-4e0a-99e1-5ba47aa52372)

PR #1498, opened 2023-07-07
Refactor: Eligibility start page

thekaveman:
Supersedes #1488 -- ultimately moved away from Django template tags in favor of our trusty includes. 

Stole some ideas from #1489 

Part of #1442 

Closes #1477 

## Goals

- [ ] Move as much copy as possible away from the database (models), towards the templates
- [ ] Reduce or eliminate the reliance on view model classes
- [ ] Change incrementally, don't break existing pages/flows
- [ ] Address the points in this comment: https://github.com/cal-itp/benefits/issues/1442#issuecomment-1610326897

## Overview

1. Add field on `EligibilityVerifier` to reference a custom template for Start page
2. Create 3 customized Start templates extending `eligibility/start.html`
    * `eligibility/start__login_gov.html`
    * `eligibility/start__mst_courtesy_card.html`
    * `eligibility/start__veteran.html`
3. Move corresponding copy from model fields into customized template files
4. Change `eligibility:start` view to use the verifier's template

PR #1501, opened 2023-07-07
Buttons: Use copy for button capitalization, not CSS

machikoyasuda:
#1483 

There have been app-wide changes to how buttons are capitalized or not. This PR intends to make all those changes in 1 PR (but no wording changes).

## What this PR does
- Removes CSS logic for button text capitalization. The capitalization must be correct in the `django.po` file. The reason for this is because _some_ buttons are capitalized, but not all. So there is no 1 style that can be applied to all buttons.
- Updates `django.po` file with all new capitalization in English and Spanish.

### These are all capitalized
- `Return Home`|| `Regresar a la Página Principal` for 404, 400, 500 pages 
- `Get Started` (`Empezar`, stays the same for Es) for Agency Index page
- `Go Back` || `Regresar a la Página Principal` for Help page

### These have mixed casing
- `Choose your Provider` `Elija su Proveedor` for Index page
- `Choose this Benefit` `Elija este Beneficio` for Elig Index page
- `Connect your Card` for Enrollment page

### What didn't need to be updated but now looks different b/c of CSS
- `Check status` || `Verificación de estatus` now has mixed casing (But will have the last word changed from `status` to `eligibility` in the copy changes to come

### DIFF
--- a/benefits/locale/en/LC_MESSAGES/django.po
+++ b/benefits/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2023-06-29 00:40+0000\n"
+"POT-Creation-Date: 2023-07-07 02:33+0000\n"
 "Language: English\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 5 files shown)


thekaveman (commented):
This PR is simpler than #1498.

We can merge this first and I can rebase on this, instead of the other way around.

thekaveman (approved):

PR #1508, opened 2023-07-10
Enrollment Success: Copy, illo, redesign

### LINKED ISSUE #1432: Enrollment Success - Add illustration, update copy
Update the confirmation view a rider sees after successfully completing an enrollment pathway (senior, veteran, agency card).

- [x] Add illustration
- [x] Update copy
- [x] Remove expiration date copy/strings

image

## Definition of Done

Implementation matches the mockups below on both mobile and desktop browsers.

- [Mockup (Desktop)](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=8945%3A47138&t=83o3CUb7RP3oO5Hb-1)
- [Mockup (Mobile)](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=9277%3A64438&t=83o3CUb7RP3oO5Hb-1)

machikoyasuda:
closes #1432 

## What this PR does
- [x] Delete all expiry item copy + code
- [x] Delete help link + code
- [x] Update copy + Spanish
- [x] Remove media item, use two column layout
- [x] Add image
- [x] Mobile
- [x] Copy: Makes the Enrollment Success body copy generic for all flows. If and when in the future, there is different body copy for specific flows (like, for example, SacRT-Senior, to change `transit` to `light rail`), the code can be updated to customize just that particular flow's copy/template.  

Note: This PR does not entirely re-do the `success.html` template to the new style. This PR merely removes as much viewmodel as possible (media item, page for title, headline), but still uses a viewmodel for the Log Out button. Since the body copy is the same across all 4 flows for now (with all the [brackets]), the `success.html` template points to the same 1 copy for now. Once we have differing copy for various flows (like, maybe, SacRT-Senior), this template can be refactored to show different body copy per flow.

## How to test
- Test CC flow
- Test Senior flow: I manually commented out the `if`s in the `success.html` file to manually make the Login.gov button line show up. Not sure how else to do it locally.
- Test VA.gov flow
- Elig Start - No regressions

### No sign out
image
image

### Sign out
image
image


benefits/enrollment/views.py
```diff
@@ -177,38 +177,12 @@ def success(request):
 
     verifier = session.verifier(request)
 
-    page = viewmodels.Page(title=_("enrollment.pages.success.title"), headline=_("enrollment.pages.success.headline"))
+    page = viewmodels.Page()
```
thekaveman, line 180: What happens if you don't create this `viewmodel.Page` and don't pass the `context_dict()` to the template?

benefits/enrollment/views.py
```diff
@@ -177,38 +177,12 @@ def success(request):
...
 
     if verifier.supports_sign_out and session.logged_in(request):
         # overwrite origin for a logged in user
         # if they click the logout button, they are taken to the new route
         session.update(request, origin=reverse(ROUTE_LOGGED_OUT))
         page.buttons = [viewmodels.Button.logout()]
```
machikoyasuda, line 186: @thekaveman 

I kept the `page` here to keep the Sign out logic all the same.

benefits/enrollment/templates/enrollment/success.html
```diff
@@ -7,26 +7,45 @@
...
+      <div class="col-12 col-lg-6">
+        <p class="pt-lg-4 mt-lg-3">{% translate "enrollment.pages.success.login_gov.confirm_item.details" %}</p>
+        <p class="pt-4">{% translate "enrollment.pages.success.thankyou" %}</p>
+      </div>
+      <div class="col-12 offset-lg-1 col-lg-5">
+        <img width="180"
+             height="155.1"
+             class="illo d-block mx-auto mb-4 mb-lg-0"
+             src="{% get_static_prefix %}img/enrollment-success-illustration.png"
+             alt="" />
```
machikoyasuda, line 32: Note: I confirmed with @srhhnry that this image has _blank_ alt text. The text is not necessary for the user and in fact, it would slow the user's experience down in completing the whole flow.

benefits/enrollment/templates/enrollment/success.html
```diff
@@ -7,26 +7,45 @@
...
+    <div class="row d-flex justify-content-start justify-content-lg-center">
+      <div class="col-12 col-lg-8 pt-lg-5 mt-lg-0 pt-4 mt-2">
         <p>
-          <span>{% translate "enrollment.pages.success.logout[0]" %}</span>
+          {% translate "enrollment.pages.success.logout[0]" %}
           {% for b in page.buttons %}
-            <a id="{{ b.id }}" class="{{ b.classes | join:" " }}" href="{{ b.url }}" role="button">
-              <span class="fallback-text">{{ b.fallback_text }}</span>
+            <a id="{{ b.id }}" class="d-inline-block p-0 {{ b.classes | join:" " }}" href="{{ b.url }}" role="button">
+              <span class="fallback-text color-logo">{{ b.fallback_text }}</span>
```
machikoyasuda, line 46: I opted to just add `color-logo` class here in the template, rather than try to change anything in `Button.logout()`. In tandem, I added a new class, `white-logo` to the Eligibility Start submit button.

benefits/enrollment/templates/enrollment/success.html
thekaveman: > it would slow the user's experience down in completing the whole flow

Isn't this the final image when the user is done with their flow?

Disagree, we need alt text here.

thekaveman (changes requested):
https://github.com/cal-itp/benefits/pull/1508/files#r1259013328

benefits/enrollment/templates/enrollment/success.html
machikoyasuda: image
This is what @srhhnry says.

Should we wait on someone to write alt text for this in the meantime, or can I just put a `TODO` placeholder in the meantime?

benefits/enrollment/templates/enrollment/success.html
```diff
@@ -7,26 +7,45 @@
...
+    <h1 class="pb-lg-5 pb-4">{% translate "enrollment.pages.success.headline" %}</h1>
   </div>
 {% endblock headline %}
 
 {% block inner-content %}
-  <div class="col-12 col-sm-12 col-lg-10">{% include "core/includes/media-list.html" with media=media %}</div>
+  <div class="col-12 col-sm-12 col-lg-9">
+    <div class="row flex-column-reverse flex-lg-row">
+      <div class="col-12 col-lg-6">
+        <p class="pt-lg-4 mt-lg-3">{% translate "enrollment.pages.success.login_gov.confirm_item.details" %}</p>
```
thekaveman, line 24: Is this correct? Either the `msgid` should be updated to be more general, or this is hardcoding the Login.gov specific text here?

benefits/core/models.py
```diff
@@ -125,8 +125,6 @@ class EligibilityVerifier(models.Model):
     eligibility_confirmed_item_details = models.TextField(null=True)
     # Fields for the dynamic enrollment success message
     enrollment_success_confirm_item_details = models.TextField()
```
thekaveman, line 127: This field doesn't appear to be referenced anywhere else, except in the migrations. I think it can be removed entirely.

benefits/enrollment/templates/enrollment/success.html
machikoyasuda: This PR does not entirely re-do the success.html template to the new style. This PR merely removes as much viewmodel as possible (media item, page for title, headline), but still uses a viewmodel for the Log Out button. Since the body copy is the same across all 3 flows for now (with all the [brackets]), the success.html template points to the same 1 copy for now. Once we have differing copy for various flows (like, maybe, SacRT-Senior), this template can be refactored to show different body copy per flow. So the `msgid` can be updated for the time being until we get the new copy. It's sort of a placeholder until we get flow-specific enrollment success body copy in the meantime.

Should I do this instead:
- Delete `"enrollment.pages.success.veteran.confirm_item.details"`, `"enrollment.pages.success.login_gov.confirm_item.details"`, `"enrollment.pages.success.mst_cc.confirm_item.details"`.
- Create and use generic `"enrollment.pages.success.confirm_item.details"`
- In future PRs, once we have flow-specific copy, change copy/template as needed.

benefits/enrollment/templates/enrollment/success.html
thekaveman: Hmm, I disagree -- according to this reasoning, the homepage image doesn't need alt text either? Actually, none of our images (icons) would, as they aren't strictly required for moving the user through the flow, they are all decorative and in addition to the text we display.

I'll create an issue so we don't block this PR.

machikoyasuda: We do have this from MobiMart: `alt="A woman and child paying for bus fare from their smart phone while at bus stop"` but maybe the alt tag could be changed to specify that the riders are using their benefit(???)

thekaveman: #1514 assigned to @indexing for when he is back.

benefits/enrollment/templates/enrollment/success.html
thekaveman: > Should I do this instead:

Yes that all sounds good. But also, remove this field from the models for now and use it directly in the template.

We will revisit for CalFresh, but not for a little while.

machikoyasuda: Got it! It may have to be revisited when SacRT launches cos SacRT is only light rail, not bus.

machikoyasuda:
_Finally_ ready for re-review @thekaveman -- all the `enrollment_success_confirm_item_details` variables, etc. are removed and the copy is consolidated to use a generic `"enrollment.pages.success.confirm_item.details"`

Rebased and re-ran `bin/makemessages.sh` to ensure `django.po` file is formatted correctly.

benefits/static/css/styles.css
```diff
@@ -370,6 +370,14 @@ footer .footer-links li a:visited {
   line-height: 1;
 }
 
+#login .fallback-text.white-logo {
+  background-image: url("/static/img/login-gov-logo-rev.svg");
+}
+
+#login .fallback-text.color-logo {
+  background-image: url("/static/img/login-gov-logo.svg");
+}
```
machikoyasuda, line 379: @angela-tran For when you start coding up these 3 - modal trigger link, modal title and modal title - you can use `#login .fallback-text.color-logo`. Currently `#login .fallback-text` has a set width and height, but you can edit the CSS to become more generic (remove the width/height from the main class) and add the width/height to a specific class, to be used in those 3 spots.
image
image
image

angela-tran: Thanks @machikoyasuda!

thekaveman:
I think this conflict is pretty minor, just the timestamp in the header section.

machikoyasuda:
@thekaveman @angela-tran Rebased with the latest `dev` - Ready for re-review

angela-tran:
@machikoyasuda Some small merge conflicts to resolve

machikoyasuda:
@angela-tran Re-re-re-based!

angela-tran (approved):

PR #1509, opened 2023-07-10
Chore: regenerate .po files

angela-tran:
With all the changes from #1489 and #1498, we've somehow gotten `dev` into a state where the ordering in our PO files is not the same as it is after running the `./bin/makemessages.sh` script. This is causing other PRs to include reordering in their changes.

This PR merges in just the re-ordering so that other PRs only update the entries relevant to their changes.


machikoyasuda (approved):
LGTM! Thank you!!!!!

thekaveman (approved):

Issue #1520, opened 2023-07-12
Eligibility Unverified - Update copy, refactor


thekaveman:
This page no longer varies by flow, only by agency -- the second sentence uses the agency `short_name`.

The page uses agency metadata `short_name`, `phone`, and `info_url` to compose an "Agency Links" element.

The page uses the `session.origin` to compose a _Back_ button.

With _Sign out_ link:

![Image](https://github.com/cal-itp/benefits/assets/1783439/62999efd-71a2-423d-b279-3841ad2b812c)

Without _Sign out_ link:

![Image](https://github.com/cal-itp/benefits/assets/1783439/74d765f7-c667-4ed6-af21-31dfebbb3620)

PR #1521, opened 2023-07-12
Refactor: agency links

thekaveman:
Part of #1520

The `eligibility:unverified` view uses the _agency links_ construct. They also show up in two other views: 
* `core:help`
* `enrollment:retry`

In `eligibility` and `enrollment`, we assume an agency has been selected by the user. 

For `core:help`, the user may navigate there before selecting an agency, so we show all "agency links" for active agencies.

This PR moves the variation and content to a new template field: `TransitAgency.links_template`, referencing an `includes` file specific for the agency. The include is used explicitly where it is needed in templates.

This PR removes unused `TransitAgency` fields as well as deleting the `Button.agency_contact_links()` viewmodel helper.

### DIFF
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -4,7 +4,33 @@ The core application: context processors for enriching request context data.
 from django.conf import settings
 from django.urls import reverse
 
-from . import session
+from . import models, session
+
+
+def _agency_context(agency):
+    return {
+        "long_name": agency.long_name,
+        "short_name": agency.short_name,
+        "info_url": agency.info_url,
+        "phone": agency.phone,
+    }
+
+
+def agency(request):
+    """Context processor adds some information about the active agency to the request context."""
+    agency = session.agency(request)
+
+    if agency is None:
+        return {}
+
+    return {"agency": _agency_context(agency)}
+
+
+def active_agencies(request):
+    """Context processor adds some information about all active agencies to the request context."""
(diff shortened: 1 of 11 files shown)


angela-tran (commented):
I think the removal of viewmodel helper code in this PR looks great! 

A separate thing I'm thinking about is that by moving the fields from `TransitAgency` into the templates, it's not as easy for a non-developer to update that information from an admin interface - it would instead require a code change in a template. This came from me thinking about the documentation for [configuring a new transit agency](https://docs.calitp.org/benefits/configuration/transit-agency/). The tradeoff I guess is between being able to customize what shows up in an agency's links vs. saying every agency should show the same info which we can then edit in the database.

thekaveman:
> ....it's not as easy for a non-developer to update that information from an admin interface
>
> The tradeoff I guess is between being able to customize what shows up in an agency's links vs. saying every agency should show the same info which we can then edit in the database.

This is a good point. Alternative idea:

* Keep the `info_url` and `phone` fields, drop the `template` field for now
* Update the `context_processor`, render all agency links the same using context
* Later we can override the template if we need to

angela-tran:
> Alternative idea:
>
>  * Keep the `info_url` and `phone` fields, drop the `template` field for now
>  * Update the `context_processor`, render all agency links the same using context
>  * Later we can override the template if we need to
>
> Basically just removing the viewmodel layer, but still using the normal context and default rendering template.

This sounds good to me 👍

PR #1522, opened 2023-07-12
Refactor: home button

thekaveman:
Part of #1520

The `eligibility:unverified` view uses the _home button_ construct. This also shows up in the `core:help` view as well as in most error pages in `core`.

This PR moves the content to an include, with some added context for the user's current `session.origin` via a new `context_processor`. 

The include is used explicitly where it is needed in templates, removing the need for the `Button.home()` viewmodel helper.


angela-tran (approved):
This looks great!

PR #1524, opened 2023-07-12
Feat: Login.gov help modal

### LINKED ISSUE #1473: Eligibility Index - Modal for Login.gov/Eligibility (Senior)

angela-tran:
Closes #1473


angela-tran:
There's some inconsistent behavior with closing the modal and what happens to the radio button.

See this GIF where I close the modal by:
 - clicking the X
 - pressing the Esc key
 - clicking off the modal

In the last case, the radio button gets selected.
benefits-login-gov-help-modal

I'm really close to implementing a fix to make it all consistent (as in, do not select the radio button), but I can't get the CSP nonce value into the `script` tag 😞 

```javascript
    <script nonce="{{ request.csp_nonce }}">
        const modal = document.getElementById("login-gov-help");
        modal.addEventListener('click', function (event) {
            event.preventDefault();
          });
    </script>
```

angela-tran:
Renamed the modal file

thekaveman:
> I'm really close to implementing a fix to make it all consistent (as in, do not select the radio button), but I can't get the CSP nonce value into the script tag 😞

Hmm this is a bummer 🤔 

I wonder if this is because you're inside an include and `request.csp_nonce` isn't available there? 

Have you tried the [`context_processor` version of `{{ CSP_NONCE }}`](https://django-csp-test.readthedocs.io/en/latest/nonce.html#context-processor)?

angela-tran:
> Have you tried the [`context_processor` version of `{{ CSP_NONCE }}`](https://django-csp-test.readthedocs.io/en/latest/nonce.html#context-processor)?

Yeah, it's the same situation (not available in the include).

thekaveman:
> An included template is rendered within the context of the template that includes it.

From https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#include

Lame, seems like it should work??

benefits/core/templates/core/includes/selection-label--senior.html
thekaveman, line 1: Side comment: I think we'll want this file (and the other selection labels) in the `eligibility` app

angela-tran:
Hmm, it might have to do with this being a form widget template.

When rendering the `form` (which is an include), we still have the context:
image

We lose the context once we're rendering the widget.

image

angela-tran:
Someone made a helper library for this very purpose: https://github.com/dmptrluke/django-csp-helpers#form-widgets 👀

thekaveman:
Also this SO post from 2013! https://stackoverflow.com/a/16922285/453168

angela-tran:
> Also this SO post from 2013! https://stackoverflow.com/a/16922285/453168

Such profit!

I propagated the `csp_nonce` through to the widget in a3854dff0ff012e2875ddfecfd66fba0617efe51. Let me know if that looks ok or not.

benefits/static/css/styles.css
```diff
@@ -371,6 +374,10 @@ footer .footer-links li a:visited {
   line-height: 1;
 }
```
thekaveman, line 375: I know this PR didn't introduce it, but I think we can fix a small bug in the Login.gov button logo.

We should add `background-repeat: no-repeat;` to this section of rules for the fallback text.

## Without this fix:

Here it is in `dev` right now in Chrome at tablet width showing the repeated image (appears like the shield is underlined):

image

Here it is from this PR, the modal button, where it is much more obvious in color:

image

## With the fix

Add one rule:

```css
#login .fallback-text {
  background-size: contain;
  background-repeat: no-repeat; /* <-- this one */
 /* ... */
}
```

Solves the issue on all buttons:

image

thekaveman:
>> Also this SO post from 2013! https://stackoverflow.com/a/16922285/453168

> Such profit!

🤣

benefits/static/css/styles.css
thekaveman: Ah, I think it was right here from yesterday, snuck by: https://github.com/cal-itp/benefits/pull/1508/files#diff-744df555aee38173a0b0868baa7befa720e6a28f80747523e74e724c126cd381

angela-tran: Oh I see. Thanks for catching that!

benefits/eligibility/templates/eligibility/includes/selection-label--senior.html
```diff
@@ -0,0 +1,24 @@
...
+    <button id="login" type="button" class="modal-trigger--link" data-bs-toggle="modal" data-bs-target="#login-gov-help">
+        <span class="fallback-text color-logo">Login.gov</span>
+    </button>
+
+    {% include "eligibility/includes/modal--senior-help.html" with id="login-gov-help" %}
+
+    <script nonce="{{ widget.csp_nonce }}">
+        const modal = document.getElementById("login-gov-help");
+        modal.addEventListener("click", function (event) {
+            event.preventDefault();
```
thekaveman, line 21: Oof, unfortunately this is breaking the link, which is also processed as a `click` event.

Let's remove this and file an issue to follow-up.

thekaveman, line 21: Oof, unfortunately this is breaking the link, which is also processed as a `click` event.

Let's remove this and file an issue to follow-up.

thekaveman (changes requested):

benefits/eligibility/templates/eligibility/includes/selection-label--senior.html
angela-tran: I reverted the commit and created #1526

benefits/static/css/styles.css
```diff
@@ -371,6 +375,10 @@ footer .footer-links li a:visited {
...
 
+.eligibility-index #login .fallback-text {
+  display: inline-block;
```
thekaveman, line 379: This is so the `?` icon stays on the same line as the Login.gov image in mobile.

Maybe look at refactoring this rule instead (to not be inside a media query): https://github.com/cal-itp/benefits/blob/dev/benefits/static/css/styles.css#L385

thekaveman (approved):

PR #1533, opened 2023-07-13
Fix: template and msgid naming

### LINKED ISSUE #1518: Adjust naming per recent conversation
We agreed that all templates should be named with `kebab-case`.

Additionally, name components should increase in specificity:

* from `Left --> Right`
* with different components separated by double-dash `--`

Examples:

* `agency-index.html`
* `modal.html`
* `modal--agency-selector.html`

## Acceptance Criteria

- [ ] All templates for includes and pages use `kebab-case` naming
- [ ] Template names increase in specificity according to the above rules
- [ ] Block names within templates also use `kebab-case`

thekaveman:
Big rename. Closes #1518


angela-tran (approved):
This looks good!

thekaveman:
Good! I wanted to get this done and merged so we don't have to think about it anymore / keep seeing all the variations.

It's going to break #1523 though so I'll work on a rebase of that after I merge.

PR #1531, opened 2023-07-13
Refactor: Eligibility unverified page

thekaveman:
Closes #1449
Closes #1520 

## English

|  |  |
|-|-|
| Desktop | image |
| Mobile | image |

## Spanish

|  |  |
|-|-|
| Desktop | image |
| Mobile | image |

## No sign out link

|  |  |
|-|-|
| Desktop | image |
| Mobile | image |

Issue #1442, opened 2023-06-22
Refactor usage of view models


thekaveman:
As shown in #1436, adding or moving around copy driven by translations requires changes in multiple places:

* models, to reference the PO entries
* PO files, to hold the translations
* view view models, to get the data into the view

We need to look into how to simplify this, ideally cutting out extraneous view models that already have a corresponding model, and sending that directly to the template instead.

## Plan

After some different iterations, this is the approach we landed on:

* Move all translated strings out of models, and as close to the templates (HTML) as possible
* Use fields in the models to reference template variants when needed (e.g. for TransitAgency and EligibilityVerifier)
* Simplify the views as much as possible: remove usage of viewmodels, create basic dictionary context that has exactly the fields needed for that view/page

## Acceptance Criteria

The following `viewmodels` and helpers are removed:

- [x] `viewmodels.Button` #1562
  - [x] `viewmodels.Button.agency_contact_links` #1521
  - [x] `viewmodels.Button.home` #1522
  - [x] `viewmodels.Button.link` #1521
  - [x] `viewmodels.Button.login` #1498
  - [x] `viewmodels.Button.logout` #1534
  - [x] `viewmodels.Button.primary` #1562
  - [x] `viewmodels.Button.outline_primary` #1562
  - [x] `viewmodels.Button.previous_page` #1562
- [x] `viewmodels.Icon` #1542
- [x] `viewmodels.MediaItem` #1542
- [x] `viewmodels.Modal` #1504
  - [x] `viewmodels.AgencySelector` #1504
- [ ] `viewmodels.Page` #1563
- [x] `viewmodels.ErrorPage` #1536
  - [x] `viewmodels.ErrorPage.user_error`  #1536
  - [x] `viewmodels.ErrorPage.server_error`  #1536
  - [x] `viewmodels.ErrorPage.not_found`  #1536
- [x] `viewmodels.PaymentProcessor` #1502
- [x] `viewmodels.TransitAgency` #1541

thekaveman:
We could move most of what view models are doing to using [Django template tags](https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/#writing-custom-template-tags). This seems like a good approach because:

* Removes need for view model classes, can pass needed data to each tag or get it from `context`
* Moves us closer to the concept of "components", will simplify and clarify templates and help reuse
* Likely simplifies view functions as well
* Tags should be straightforwardly testable

Django makes two types available OOTB:
[...]

PR #1534, opened 2023-07-14
Refactor: sign out button

### LINKED ISSUE #1519: Refactor sign out (Login.gov) copy/button in Enrollment Success
Follow up to #1508 and part of finishing that refactor per #1442, we need to address the final usage of `viewmodels.Button.logout()`.

## Acceptance Criteria

- [ ] We can entirely remove the `from benefits.core import viewmodels` from `enrollment/views.py`
- [ ] The sign out (Login.gov) button shows correctly for flows that support it (`senior`)
- [ ] The sign out (Login.gov) button does not show for flows that don't support it (`mst_courtesy_card`, `veteran`)

## Additional context

image

thekaveman:
Closes #1519

Needed for #1520 / #1531

Part of #1442

## Sign out link

image

## Sign out button

image


tests/pytest/conftest.py
```diff
@@ -44,12 +44,16 @@ def model_PemData():
...
-        sign_out_button_label="Sign out",
+        sign_out_button_template="core/includes/sign-out-button--senior.html",
+        sign_out_link_template="core/includes/sign-out-link--senior.html",
         client_name="Client",
         client_id="1234",
         authority="https://example.com",
     )
 
+    assert not auth_provider.supports_claims_verification
+    assert auth_provider.supports_sign_out
```
angela-tran, line 55: What is the reason for having these assertions here rather than in the test?

angela-tran (commented):
The refactor looks great! Just a minor question about test assertions

tests/pytest/conftest.py
```diff
@@ -59,16 +63,22 @@ def model_AuthProvider_with_verification(model_AuthProvider):
     model_AuthProvider.claim = "claim"
     model_AuthProvider.save()
 
+    assert model_AuthProvider.supports_claims_verification
```
angela-tran, line 66: Same question as above

tests/pytest/conftest.py
```diff
@@ -59,16 +63,22 @@ def model_AuthProvider_with_verification(model_AuthProvider):
...
     model_AuthProvider.scope = "scope"
     model_AuthProvider.claim = "claim"
-    model_AuthProvider.sign_out_button_label = None
+    model_AuthProvider.sign_out_button_template = None
+    model_AuthProvider.sign_out_link_template = None
     model_AuthProvider.save()
 
+    assert model_AuthProvider.supports_claims_verification
+    assert not model_AuthProvider.supports_sign_out
+
```
angela-tran, line 81: Same question as above

tests/pytest/conftest.py
```diff
@@ -78,16 +88,22 @@ def model_AuthProvider_without_verification(model_AuthProvider):
     model_AuthProvider.claim = None
     model_AuthProvider.save()
 
+    assert not model_AuthProvider.supports_claims_verification
```
angela-tran, line 91: Same question as above

tests/pytest/conftest.py
```diff
@@ -78,16 +88,22 @@ def model_AuthProvider_without_verification(model_AuthProvider):
...
 def model_AuthProvider_without_verification_no_sign_out(model_AuthProvider):
     model_AuthProvider.scope = None
     model_AuthProvider.claim = None
-    model_AuthProvider.sign_out_button_label = None
+    model_AuthProvider.sign_out_button_template = None
+    model_AuthProvider.sign_out_link_template = None
     model_AuthProvider.save()
 
+    assert not model_AuthProvider.supports_claims_verification
+    assert not model_AuthProvider.supports_sign_out
```
angela-tran, line 105: Same question as above

tests/pytest/conftest.py
thekaveman: Yeah there wasn't any explicit tests for these models, I could add them instead of asserting here.

angela-tran: Oh I see, `supports_sign_out` is not on `EligibilityVerifier` anymore and that's the model we have tests for

angela-tran: I'm ok with leaving them here

angela-tran (approved):

tests/pytest/conftest.py
thekaveman: I guess it just felt a little weird writing tests for the fixtures the tests are creating... 😆 

But we already do that, and the tests confirm the behavior in our standard way, so good call.

angela-tran (approved):
🚀

PR #1537, opened 2023-07-17
Fix: OAuth session checks and events

### LINKED ISSUE #1535: AttributeError: 'NoneType' object has no attribute 'auth_provider'
[Sentry Issue](https://sentry.calitp.org/organizations/sentry/issues/72335/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'auth_provider'
(1 additional frame(s) were not displayed)
...
  File "benefits/oauth/views.py", line 89, in cancel
    analytics.canceled_sign_in(request)
  File "benefits/oauth/analytics.py", line 59, in canceled_sign_in
    core.send_event(CanceledSignInEvent(request))
  File "benefits/oauth/analytics.py", line 27, in __init__
    super().__init__(request, "canceled sign in")
  File "benefits/oauth/analytics.py", line 13, in __init__
    self.update_event_properties(auth_provider=verifier.auth_provider.client_name)
```

thekaveman:
Closes #1535 

Makes all views in `benefits.oauth` require an `EligibilityVerifier` in the session with an `AuthProvider` that supports claims verification. 

This was an implicit assumption before, this PR makes it explicit via a middleware. Added tests to capture these assumptions as well.

### DIFF
--- a/benefits/enrollment/templates/enrollment/success.html
+++ b/benefits/enrollment/templates/enrollment/success.html
@@ -36,7 +36,7 @@
 {% endblock inner-content %}
 
 {% block call-to-action %}
-  {% if authentication and authentication.sign_out_button_template %}
+  {% if authentication and authentication.logged_in and authentication.sign_out_button_template %}
     <div class="row d-flex justify-content-start justify-content-lg-center">
       <div class="col-12 col-lg-8 pt-lg-5 mt-lg-0 pt-4 mt-2">
         <p>
--- a/benefits/oauth/analytics.py
+++ b/benefits/oauth/analytics.py
@@ -10,7 +10,8 @@ class OAuthEvent(core.Event):
     def __init__(self, request, event_type):
         super().__init__(request, event_type)
         verifier = session.verifier(request)
-        self.update_event_properties(auth_provider=verifier.auth_provider.client_name)
+        if verifier and verifier.uses_auth_verification:
+            self.update_event_properties(auth_provider=verifier.auth_provider.client_name)
 
 
 class StartedSignInEvent(OAuthEvent):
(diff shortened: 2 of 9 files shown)


angela-tran (approved):

thekaveman:
@angela-tran I found a small bug from #1534 while testing the oauth routes.

I'm also thinking of another use-case to fix -- why is this route callable by anyone? It should always correspond to a prior call to `login` or `logout` I think, right?

angela-tran:
> I'm also thinking of another use-case to fix -- why is this route callable by anyone? It should always correspond to a prior call to `login` or `logout` I think, right?

"This route" being `/cancel`, right? I believe it needs to be callable by anyone so that the IdG can configure it as a redirect URL.

What do you mean by it should always correspond to a prior call to `login` or `logout`?

thekaveman:
The user that triggered #1535 (seemingly) visited `/oauth/cancel` by directly entering the URL.

Even if the user has a correct `EligibilityVerifier` configured, they shouldn't be able to cancel their sign-in unless they first started it. But right now (even with the fix so far in this PR) you can get a verifier and then still manually GET the `/oauth/cancel` endpoint, triggering the view and the analytics event.

The endpoint should only be called by a legit OAuth flow, otherwise we should probably return a User Error?

angela-tran:
> The endpoint should only be called by a legit OAuth flow, otherwise we should probably return a User Error?

Oh I see. Yes, this sounds correct to me. Would this require that we maintain some session state to say that they're in an OAuth flow?

thekaveman:
Resolved merged conflicts

angela-tran (approved):

PR #1542, opened 2023-07-18
Refactor: remove icon/media item view models

thekaveman:
For #1442


angela-tran (approved):
This looks good to me. I tested out the app locally and didn't notice any regressions.

PR #1550, opened 2023-07-21
Fixes: CSS column alignment

### LINKED ISSUE #1548: Bug: Grid-width issue
Related to #1545 

Almost all instances of `col-lg-10` app should now be `col-lg-8`.

machikoyasuda:
closes #1548 

- Desktop: Container width for most* pages is now `col-lg-8`, not `col-lg-10`. Index page, Agency Index page and Enrollment Success is an outlier and is not part of this. No changes necessary for 404/400/500 pages. Also one page has a `col-lg-6` H1 wrapper container instead. Because some pages _are_ actually different from others, I recommend *not* codifying the `col-lg-8` width into the `base.html` itself.
- Desktop: Media Item grid gap is now lower to 24px. 
- Mobile: Should have no changes

### Screenshots

image
image
image
image

image
image

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -72,11 +72,11 @@
         <div class="container main-row">
           <div class="row justify-content-center">
             {% block headline %}
-              <div class="col-lg-10">
+              <div class="col-lg-8">
                 <h1>{{ page.headline }}</h1>
               </div>
             {% endblock headline %}
-            <div class="col-lg-10">
+            <div class="col-lg-8">
               {% block explanatory-text %}
               {% endblock explanatory-text %}
             </div>
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -6,7 +6,7 @@
   <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
     {% csrf_token %}
 
-    <div class="row form-container">
+    <div class="row form-container justify-content-center">
       <div class="{{ form.classes }}">
         {% for field in form %}
           <div class="row form-group mb-0">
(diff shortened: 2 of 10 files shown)


thekaveman (approved):
This looks good.

Just wondering about:

> I recommend not codifying the col-lg-8 width into the base.html itself.

Would that come in a later refactor?

machikoyasuda:
@thekaveman 

_Just wondering about:

    I recommend not codifying the col-lg-8 width into the base.html itself.

Would that come in a later refactor?_


No - because not all pages have `col-lg-8` width by default. So I'd rather not put it in.. because it's not on all pages. Help, Enrollment Success, Log Out, Agency Index, Index all are different. If "base" is truly to be "base", it should serve all pages.

thekaveman:
> @thekaveman
> 
> > _Just wondering about:
> > ```
> > I recommend not codifying the col-lg-8 width into the base.html itself.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would that come in a later refactor?_
> 
> No - because not all pages have `col-lg-8` width by default. So I'd rather not put it in.. because it's not on all pages. Help, Enrollment Success, Log Out, Agency Index, Index all are different. If "base" is truly to be "base", it should serve all pages.

@machikoyasuda that was my question, it's still in this PR -- will removing it be in a separate PR? (I guess so)

machikoyasuda:
@thekaveman Oh sorry I think I figured out your question.

https://github.com/cal-itp/benefits/pull/1550/files#diff-e2ae5f60bb472effafac24b079fcbb1ffd09dcc5d8df2d22cab07417d222f925R72-R87

I DO recommend `col-lg-8` be in `headline` and `explanatory-text` -- b/c all pages do use this, as this PR reflects. But I do *not* recommend `col-lg-8` to be coded into `base.html` for the `inner-content` portion, cos that's where it differs for the other pages.

Issue #1544, opened 2023-07-18
Update copy for Help page


angela-tran:
## Acceptance Criteria

- [x] Help page has copy that matches what is shown in Figma (Developer Benefits Full App)
- [x] For Spanish translations, just put TODOs for now

## Additional context
The Help page was updated to match copy shown in the help modal dialogs.

angela-tran:
Not sure if this is ready to be worked on. There are several comments in Figma on the Help Page screen in "Developer Benefits Full App" from 2 months ago that say "WIP" and "Fix".

Also, the punctuation on links is not consistent (some have a period, some don't) so that is an open question right now.

srhhnry:
Took care of the comment that said WIP (it was from way before we shifted to the new copy process), but don't see any other comments on those screens that are ambiguous (just tag me if you find any). 

Regarding the punctuation I've noted this in the Style Guide in the Text Layout section, but I can see that page not getting a lot of traffic. Essentially all sentences in body copy should receive punctuation, including sentences that end in icons. Unfortunately, the Google Sheets sync we use doesn't allow for adding extra spaces that would be consistent between desktop and mobile (with space added to account for the Figma component of the external link button). It's totally a pain, and I'm sorry it's not clear. I'll try to figure out a way to work around that limitation, but in the meantime, yes, punctuation on links.

angela-tran:
Thank you @srhhnry !

srhhnry:
lol, @angela-tran I just saw all of those comments on the mobile help page (they weren't loading or I was zoomed out or something crazy was happening). I think they primarily relate to that issue with Google sheets and punctuation, but if any of them don't I'll let you know.

angela-tran:
Question for @thekaveman and @machikoyasuda: since certain sections of the Help page are supposed to be the exact same copy as what's in the Help modals, should we use the same PO file entries? Or have separate, duplicate entries (one set for the modal, one set for the Help page)?

Here is the Login.gov help modal (in the app) and help page section (from Figma) for example:
![image](https://github.com/cal-itp/benefits/assets/25497886/3eb7cf9b-fd6a-491f-8031-02eb88e11dbb)

![image](https://github.com/cal-itp/benefits/assets/25497886/6a1c5f55-c4c4-4e17-a302-b6c34866f8f6)

machikoyasuda:
[...]

PR #1558, opened 2023-07-21
Chore: Help page - contactless cards section

angela-tran:
Part of #1544 

Depends on #1557 

This PR updates the section of the Help page explaining contactless cards. The copy should match what's on the Contactless Cards help modal.

image


benefits/core/templates/core/help.html
```diff
@@ -17,23 +17,24 @@ <h2 class="pt-8" id="what-is-cal-itp">{% translate "core.pages.help.about" %}</h
...
                alt="{% translate "core.icons.contactless" context "image alt text" %}" />
         </p>
-
-        <p class="pt-4">{% translate "core.pages.help.payment_options.p[1]" %}</p>
+        <p class="pt-4">{% translate "eligibility.pages.start.modal.p[2]" %}</p>
+        <p class="pt-4">{% translate "eligibility.pages.start.modal.p[3]" %}</p>
         <p class="pt-4">
-          {% blocktranslate with website="https://cash.app/help/us/en-us/14425-cal-transit" %}core.pages.help.payment_options.p[2][0]{{ website }}{% endblocktranslate %}
-          {% blocktranslate with website="https://venmo.com/about/debitcard/" %}core.pages.help.payment_options.p[2][1]{{ website }}{% endblocktranslate %}
+          {% if agency %}
```
angela-tran, line 32: Note that, unlike the case for the Contactless Cards modal which is on the Eligibility Start page, the user might not have chosen an `agency` yet when they go to the Help page. There wouldn't be specific link that we could show.

Therefore, here we're only showing this sentence if the user has chosen an agency. 

An alternative would be for product/design to provide us the copy to use when we're not in the context of a specific agency.

thekaveman: Should this `{% if %}` wrap the `<p>` tag then? Otherwise might this `<p>` be empty?

angela-tran: Yes, good point! Amended the commit as ed1935dcc995bdc88e2448a054d8e444121032d2

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -80,46 +80,47 @@ msgstr ""
...
+"<a href=\"%(agency_info_url)s\" target=\"_blank\" rel=\"noopener noreferrer"
+"\">"
 
-msgid "core.pages.help.payment_options.p[4]"
-msgstr ""
-"For updates on additional options, please check back on this website, or get "
-"in touch with your local transit provider."
+#, python-format
+msgid "eligibility.pages.start.modal.p[4.3]%(agency_short_name)s"
+msgstr "%(agency_short_name)s</a>."
```
thekaveman, line 123: I'm seeing a double period before and after the external link icon, I'm pretty sure this one here is not needed?

image

angela-tran: I think what you're seeing before the external link is part of the link's underline. The closing parenthesis makes it look more separate.

image

thekaveman: Yeah you are totally right @angela-tran! It was the parenthesis 🙄

thekaveman, line 123: Part of my question above, do we need this in the PO files?

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -80,46 +80,47 @@ msgstr ""
...
+"%(agency_short_name)s application process. For updates on additional "
+"options, please check back on this website, or contact "
 
-msgid "core.pages.help.payment_options.p[3]"
+#, python-format
+msgid "eligibility.pages.start.modal.p[4.2]%(agency_info_url)s"
 msgstr ""
-"You can still get your discount by going through your local transit "
-"provider’s application process."
+"<a href=\"%(agency_info_url)s\" target=\"_blank\" rel=\"noopener noreferrer"
```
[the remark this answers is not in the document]
thekaveman: Maybe the link does need to be here, because the word `contact` is supposed to be part of the link text, and that will be translated.

But it looks like the word `contact` is part of the previous message and not this one.

benefits/locale/en/LC_MESSAGES/django.po
thekaveman: See https://github.com/cal-itp/benefits/pull/1558#discussion_r1272757107.

I don't think we need this split into multiple pieces in the PO file, probably just one right?

angela-tran: Yeah, I had it as one in [the first commit](https://github.com/cal-itp/benefits/pull/1558/commits/c5b15bab6de198fb5d7d150d138d24fd113a932c#diff-8b77e09357e0ac2a23a88239a579cb8218a1c212a5d816af33125cf3690e7a16R108-R114) of this PR. In the second commit of this PR, I refactored the template to use the same msgids that the Contactless Card modal is using, as implemented in #1523.

I can update both the Help page and the modal to use one entry if that's the direction we'd like to go.

thekaveman: I see. Does the modal need it to be split up? I think we can assume one long sentence with the same `short_name` replacement used a few different times and the link within -- does that make sense?

angela-tran: No, I don't think the modal needs it split up. I made it one entry in https://github.com/cal-itp/benefits/pull/1558/commits/e9849f5da47092c681dce51ddeac1ca5d9a80da9 and also changed the `msgstr` to use placeholders like we're doing everywhere else until we have the real copy.

thekaveman: Ah, I think for this one I was hoping we could just finish it out -- we already have the `short_name` available in the context along with the `info_url`. The translation shouldn't change the agency's short name that we display.

angela-tran: Oh ok, got it. I've made that change in [9159af5](https://github.com/cal-itp/benefits/pull/1558/commits/9159af53560f46d6688df64a94eaf07ec931453a)

thekaveman (approved):
👍

PR #1559, opened 2023-07-21
Chore: Help page - Login.gov sections

angela-tran:
Part of #1544 

This PR updates the sections explaining Login.gov and identity verification via Login.gov. The copy should match the help modals with one exception being that the sentence about visiting the help center (shows up under the first section rather than the second).

As shown in Figma:
image


thekaveman (approved):
Ugh, this looks fine for now but this help page is kind of a bummer! Definitely going to break our "separate the strings into the apps that use them" idea unless we go full duplicate content 👎 Maybe some day we can get rid of it.

angela-tran:
> Ugh, this looks fine for now but this help page is kind of a bummer! Definitely going to break our "separate the strings into the apps that use them" idea unless we go full duplicate content 👎 Maybe some day we can get rid of it.

Yeah, I agree!

PR #1560, opened 2023-07-21
Chore: Help copy - agency card

angela-tran:
Part of #1544 

This PR does not change the copy for the "agency card" section of the Help page, but rather extracts that section out to be a template specified by the TransitAgency model.


angela-tran:
I'm not super sure about this PR... I was just following our pattern of using placeholders for variables that don't exist on models, but realistically, it'll be a while before we have another agency that this section applies to. So are we going to have another PR later that essentially reverts this one?

thekaveman:
> I'm not super sure about this PR... I was just following our pattern of using placeholders for variables that don't exist on models, but realistically, it'll be a while before we have another agency that this section applies to. So are we going to have another PR later that essentially reverts this one?

Yeah this doesn't feel all that useful to introduce the placeholders when all this text is known for MST/Courtesy Cards.

thekaveman:
Maybe a change is to refactor the [help page](https://github.com/cal-itp/benefits/blob/chore/help-copy-agency-card/benefits/core/templates/core/help.html#L51) to optionally show that section, with a `TransitAgency.help_template` field that points to an include?

angela-tran:
> Maybe a change is to refactor the [help page](https://github.com/cal-itp/benefits/blob/chore/help-copy-agency-card/benefits/core/templates/core/help.html#L51) to optionally show that section, with a `TransitAgency.help_template` (or `agency_card_info_template` or something) field that points to an include? (To get this ready for e.g. SacRT)

I took your suggestion to do this refactor instead of having placeholder-filled copy. I force-pushed a new set of commits to this branch, and this is ready for re-review

benefits/core/migrations/0001_initial.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 4.2.3 on 2023-07-14 04:51
+# Generated by Django 4.2 on 2023-07-26 20:23
```
thekaveman, line 1: Seems like you should reinstall packages in your devcontainer 😄 

```
pip install -e .
```

thekaveman:
This is looking really great otherwise! 👍

benefits/core/migrations/0001_initial.py
angela-tran: Oh, good catch! Re-installed and re-ran the script: [38f701d](https://github.com/cal-itp/benefits/pull/1560/commits/38f701d7c443e5fd6b5f6ee3f35393765cca0e4c)

thekaveman (approved):
Nice!

PR #1566, opened 2023-07-25
Refactor: Add alt tags to the landing pages

### LINKED ISSUE #1418: Home & Agency Index - Refactor homepage background image to use an img tag
Following on from #1392:

The primary image used on the homage of the app is brought in using the CSS `background` property:

image

This image could (should) be considered a key piece of informational content (vs. just decorative) and thus should be made accessible with corresponding `alt` text. 

This post on SO has a good discussion of various approaches: https://stackoverflow.com/q/48913759/453168

The consensus from that thread is that the [CSS property `object-fit`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) is widely supported enough (see https://caniuse.com/object-fit) that an `` tag can be dropped in as a replacement for `background` + `background-size: cover;` CSS properties.

## Acceptance Criteria

- [ ] The homepage splash image is defined in an `` tag with appropriate `alt` text applied for English and Spanish users.

## Additional context

The `alt` text needs Spanish translation, tracking in #1410.

machikoyasuda:
fix #1418 

- This PR moves the debug bar to the bottom, adds an `outline` of 5px (with an `outline-offset` of -5px to ensure this outline doesn't affect any widths). This was necessary to have the CSS work for the way the now absolutely-positioned div's logic from the top is calculated.
- Replaces `background-image` with 2 responsive `img` tags.
- Absolutely positions the div with text to be 135px from the top on Desktop, and set to bottom: 0 for Mobile
- Adds alt tag content to django.po file

## How to test this PR
- Test the Index page - Desktop (in all sizes), tablet, mobile. Test the modal works. Test that the footer links work correctly.
- Test the Agency Index page - Desktop (in all sizes), tablet, mobile

### Desktop
image
image
image
Observe that there are no height/width changes with the addition of the border

Short screens/weird width screens - Ensure the scaling/spacing/alignment is the same as Production
image
image

### Tablet
image

### Mobile
image
image

### DIFF
--- a/benefits/locale/en/LC_MESSAGES/django.po
+++ b/benefits/locale/en/LC_MESSAGES/django.po
@@ -2,11 +2,10 @@
 # Copyright (C) 2021 California Department of Transportation
 # This file is distributed under the same license as the benefits package.
 #
-#, fuzzy
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2023-07-25 01:52+0000\n"
+"POT-Creation-Date: 2023-07-25 18:34+0000\n"
 "Language: English\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -682,8 +681,6 @@ msgstr ""
 "The page you are looking for might be somewhere else or may not exist "
 "anymore."
 
-#, fuzzy
-#| msgid "core.icons.sadbus"
 msgctxt "image alt text"
 msgid "core.icons.sadbus"
 msgstr "Bus icon with flat tire"
--- a/benefits/locale/en/LC_MESSAGES/django.po
+++ b/benefits/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2023-07-21 16:47+0000\n"
+"POT-Creation-Date: 2023-07-25 01:52+0000\n"
 "Language: English\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -254,6 +254,10 @@ msgstr "Sign out of Login.gov"
 msgid "core.pages.index.button"
 msgstr "Choose your Provider"
 
+msgid "core.pages.index.alt"
+msgstr ""
+"A person holds a contactless debit card next to a card reader on a bus."
+
 msgid "core.pages.landing.h2"
 msgstr ""
 "Cal-ITP Benefits connects your transit benefit to your contactless card"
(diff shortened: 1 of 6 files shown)


benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -678,5 +682,8 @@ msgstr ""
 "The page you are looking for might be somewhere else or may not exist "
 "anymore."
 
+#, fuzzy
+#| msgid "core.icons.sadbus"
+msgctxt "image alt text"
 msgid "core.icons.sadbus"
```
machikoyasuda, line 688: I don't know why this fuzzy is here!

benefits/locale/es/LC_MESSAGES/django.po
```diff
@@ -691,5 +696,8 @@ msgstr ""
 "La página que estás buscando puede estar en otro lugar o puede que ya no "
 "exista."
 
+#, fuzzy
+#| msgid "core.icons.sadbus"
+msgctxt "image alt text"
 msgid "core.icons.sadbus"
```
machikoyasuda, line 702: I don't know why this fuzzy is here!

thekaveman:
Hmm, the whole point of that debug bar is it is loud and in our face to let us know debug mode is on. Moving it to the bottom could obscure it for certain pages. Can we find a way to do this without moving that to the bottom?

benefits/locale/en/LC_MESSAGES/django.po
thekaveman: It can happen when `gettext` thinks the string has changed. 

These should be manually reviewed and never committed -- when we run `python manage.py compilemessages` (in the init script) it skips all `fuzzy` entries so they never make it into the compiled language files.

https://docs.djangoproject.com/en/4.2/topics/i18n/translation/

> **Fuzzy entries**
>
> [makemessages](https://docs.djangoproject.com/en/4.2/ref/django-admin/#django-admin-makemessages) sometimes generates translation entries marked as fuzzy, e.g. when translations are inferred from previously translated strings. By default, fuzzy entries are not processed by [compilemessages](https://docs.djangoproject.com/en/4.2/ref/django-admin/#django-admin-compilemessages).

benefits/locale/es/LC_MESSAGES/django.po
thekaveman: https://github.com/cal-itp/benefits/pull/1566#discussion_r1273644826

thekaveman (dismissed):
Looks great!

**Desktop**

image

**Mobile**

image

thekaveman:
@machikoyasuda are there any `aria-` tags or other accessibility considerations we can make to help avoid a screenreader reading/seeing both desktop/mobile images back-to-back? Can be in a follow-up.

machikoyasuda:
@thekaveman I think a future refactor of using the `<picture>` element with `srcset` (where we can set numerous images by screen dimension/size) would take care of that issue.

machikoyasuda:
@thekaveman Putting this back into WIP for now b/c the site wasn't scaling on small / weird Desktop sizes.

machikoyasuda:
Closing this PR in favor of a different solution.

PR #1569, opened 2023-07-27
Fix: Add alt text to index, agency index background image

### LINKED ISSUE #1418: Home & Agency Index - Refactor homepage background image to use an img tag

machikoyasuda:
fixes #1418 

This PR takes a different approach from the previous PR - and avoids a lot of complications in new HTML, changing the debug bar, changing templates, etc.

Instead of changing the CSS structure, this PR simply adds an empty `span` tag with `role="img"` and `aria-label` with the alt tag, as suggested here: https://www.davidmacd.com/blog/alternate-text-for-css-background-images.html

image

### DIFF
--- a/benefits/core/templates/core/landing.html
+++ b/benefits/core/templates/core/landing.html
@@ -11,6 +11,7 @@
         <div class="box px-4 py-3">
           <h1 class="text-left p-0">{{ page.headline }}</h1>
           <h2 class="p-sm pb-lg-8 pt-1 pt-lg-4 pb-4">{% translate "core.pages.landing.h2" %}</h2>
+          <span role="img" aria-label="{% translate "core.pages.index.alt" %}"></span>
           {% block call-to-action %}
           {% endblock call-to-action %}
         </div>
--- a/benefits/locale/en/LC_MESSAGES/django.po
+++ b/benefits/locale/en/LC_MESSAGES/django.po
@@ -278,6 +278,10 @@ msgstr "Choose Provider"
 msgid "core.pages.index.headline"
 msgstr "Get reduced fare on public transportation when you tap to ride"
 
+msgid "core.pages.index.alt"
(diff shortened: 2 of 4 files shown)


machikoyasuda:
@thekaveman Decided to go with the span approach, because I couldn't figure out a more robust way to get the div to scale properly.

benefits/core/templates/core/landing.html
```diff
@@ -11,6 +11,7 @@
         <div class="box px-4 py-3">
           <h1 class="text-left p-0">{{ page.headline }}</h1>
           <h2 class="p-sm pb-lg-8 pt-1 pt-lg-4 pb-4">{% translate "core.pages.landing.h2" %}</h2>
+          <span role="img" aria-label="{% translate "core.pages.index.alt" %}"></span>
```
thekaveman, line 14: This also has the added benefit of inlining the alt text in the most appropriate place for the user, in context with the other (primary) content on the page.

thekaveman (approved):

Issue #1423, opened 2023-06-09
Color Workshop on Cal-ITP Blue


srhhnry:
@segacy1 and @srhhnry met to work on improving accessibility for Cal-ITP properties by increasing contrast on the primary blue used on these sites (known as Cal-ITP blue). We also adjusted the hover color to match the new blue. And for Benefits, the text link color was shifted to the new blue as well.

srhhnry:
Scheduled for June 21st

srhhnry:
@angela-tran @machikoyasuda 

The new Action Primary (Cal-ITP blue) color for Benefits, MobiMart, and Cal-ITP.org is: rgba(4, 91, 134, 1) or #045B86
The new Hover color for Benefits, MobiMart, and Cal-ITP.org is: rgba(4, 72, 105, 1) or #044869
The new Text Link color for Benefits is: rgba(4, 91, 134, 1) or #045B86
The new Hover link color for Benefits is:  rgba(4, 72, 105, 1) or #044869

Those colors have been updated in Benefits and MobiMart/Cal-ITP and can be seen in the Benefits Style Guide [here](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=4890%3A17182&t=w58kxu3smXZ6F1kD-1)

machikoyasuda:
@srhhnry Thanks for this update!

GitHub ProTip™️ - When typing in colors, if you wrap the color code in backticks (`), then a color chip will appear next to it. I went ahead and did that for your comment above, so we can see the colors!

machikoyasuda:
Can this ticket (https://github.com/cal-itp/benefits/issues/1389) be closed, in favor of this most current one?

srhhnry:
That protip is very cool. And yeah, I went ahead and closed #1389 after updating the typography page in figma

srhhnry:
@machikoyasuda checking in on whether I can close this ticket

machikoyasuda:
@srhhnry Yes! Once dev is ready to make this color change, we'll make a ticket to track it.

PR #1595, opened 2023-07-31
Add new primary color, hover color, footer hover color, footer visited color

### LINKED ISSUE #1577: Update blue color
From [#1423](https://github.com/cal-itp/benefits/issues/1423)

https://github.com/cal-itp/benefits/issues/1556#issuecomment-1652180169

machikoyasuda:
fix #1577 

- Design instructions are coming from here and reflected in Figma Style Guide: https://github.com/cal-itp/benefits/issues/1423#issuecomment-1601567969
- 4 updated colors: `primary`, `hover`, `footer-link-hover`, `footer-link-visited`

### Footer Link Hover

Used to be purple. Now it's a different shade of dark blue (`#0062ff`)

image

### Footer Link Visited

The color for :hover and :visited used to be the same. Now, only the visited is purple.

image

### Hover

image

Regular link hover is slightly darker.

### Primary

Primary color is now darker. It's most noticable in the nav:

image

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -1,6 +1,6 @@
 :root {
-  --primary-color: #046b99;
-  --primary-color-rgb: 4, 107, 153;
+  --primary-color: #045b86;
+  --primary-color-rgb: 4, 91, 134;
   --bs-blue: var(--primary-color);
   --bs-primary: var(--primary-color);
   --bs-primary-rgb: var(--primary-color-rgb);
@@ -17,7 +17,7 @@
   --bs-body-line-height: 1.5;
   --body-letter-spacing: 0.05em;
   --dark-color: #212121;
-  --hover-color: #2f4c65;
+  --hover-color: #044869;
   --error-color: #ea1010;
   --selected-color: #562b97;
   --error-color-rgb: 234, 16, 16;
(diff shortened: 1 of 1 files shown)


benefits/static/css/styles.css
```diff
@@ -1,6 +1,6 @@
 :root {
-  --primary-color: #046b99;
-  --primary-color-rgb: 4, 107, 153;
+  --primary-color: #045b86;
+  --primary-color-rgb: 4, 91, 134;
```
thekaveman, line 3: I thought we wanted to transition to using RGB everywhere since it is easier to diff and get the alpha correct?

It seems like this variable is only used to define (overwrite) the `--bs-primary-rgb` variable, whereas our `--primary-color` variable is used extensively.

machikoyasuda: `rgb` is used by Bootstrap for specific areas that need a rgbA. we could refactor later to get rid of the hex numbers entirely. it would require us to change the variable declaration a little bit though (something to make sure the variable is wrapped in`rgb()`).

machikoyasuda: I'm following Bootstrap CSS Variable conventions: https://getbootstrap.com/docs/5.2/customize/css-variables/#root-variables by declaring both.

thekaveman (approved):
LGTM 💙💜

Issue #980, opened 2022-09-29
Help Page Copy Update


srhhnry:
- [x] Add new agency-specific sections to Help and reorganize the Help page
- [x] Translate new Help page language to Spanish where need be

machikoyasuda:
English Help Page copy has been handed off to MST.

PR #1596, opened 2023-08-01
Help: Align desktop width to 8-col

### LINKED ISSUE #1576: Adjust Help page columns
Change col-10 to col-md-8 col-10

https://github.com/cal-itp/benefits/issues/1556#issuecomment-1648355118

machikoyasuda:
fix #1576 

image
image

### DIFF
--- a/benefits/core/templates/core/help.html
+++ b/benefits/core/templates/core/help.html
@@ -9,13 +9,13 @@
 {% block main-content %}
   <div class="container">
     <div class="row justify-content-center">
-      <div class="col-10">
+      <div class="col-md-8 col-10">
         <h1>{% translate "core.buttons.help" %}</h1>
       </div>
     </div>
 
     <div class="row justify-content-center">
-      <div class="col-10">
+      <div class="col-md-8 col-10">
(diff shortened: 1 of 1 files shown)


machikoyasuda:
Moving this outstanding question from #980 to here, for @srhhnry @angela-tran @thekaveman 

Should we remove the GitHub sentence after the `Go Back` button? Do we have to have it? If so, it should be in the Figma...and should it come before or after the `Go Back` button?

machikoyasuda:
@thekaveman Got it.

@srhhnry  Do you prefer the last line with the link to the source code to be below the Go back button or above it? (Or somewhere else entirely?) Eventually - it would be more appropriate on an About page.

image 

vs

image

I also wonder if the text needs some kind of subheader.

cc @indexing FYI

angela-tran:
> My suggestion is to make this a small section with a header (above the button, below the questions section). Definitely want to hear @indexing's thoughts on content, but could see a header title...

@thekaveman @machikoyasuda Can we create a separate GH issue for this idea? This GitHub sentence has been on the Help page all this time without any problem, and I don't think it blocks anything about updating the desktop width which is what this PR is about.

thekaveman (approved):

PR #1598, opened 2023-08-01
Fix: Desktop - Login.gov modal alignment

### LINKED ISSUE #1575: Fix Login.gov modal top padding
The distance from the top edge to the start of text should be 36px. This one's over 60px.

https://github.com/cal-itp/benefits/issues/1556#issuecomment-1646042932

- Do both Login.gov modals

machikoyasuda:
fix #1575 

## What this PR does

This PR ensures that the 2 Login.gov modals match the design in Figma and also match the design of the 2 other modals (Contactless pay and Littlepay) in padding, margin and spacing.

- Use `modal-info` includes for both Login.gov modals. (Side note: Why are 2 modal includes necessary? Because spacing and margins are different in seemingly small but noticeable ways between the Transit Provider Selection modal and the rest of the modals. The X icons aren't even the same. The header top padding height is totally different. The titles are differently aligned - to name a few.)
- Reduces the space between the Header text and the X to 36px.
- Reduces space between paragraphs

## Login.gov modal - Eligibility

image

## Login.gov modal - Eligibility Start

image

### DIFF
--- a/benefits/eligibility/templates/eligibility/includes/media-item--idcardcheck--start--senior.html
+++ b/benefits/eligibility/templates/eligibility/includes/media-item--idcardcheck--start--senior.html
@@ -23,5 +23,5 @@
         </div>
     </div>
 
-    {% include "eligibility/includes/modal--senior-start-help.html" with id="identity-verification-help" %}
+    {% include "eligibility/includes/modal--senior-start-help.html" with size="modal-lg" id="identity-verification-help" %}
 {% endblock body %}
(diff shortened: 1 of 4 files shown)


machikoyasuda:
@thekaveman I asked Sarah about this. She's going to be updating the Agency Provider modal for the SBMTD feature. And when she does that, she is going to align the Agency Provider modal to look more like the Info Modal. So I think we can combine the 2 modals into the same includes once she does that.

machikoyasuda:
@thekaveman Once we get the new Agency Provider modal design (for SBMTD) - which is part of Sprint 8, I suggest we:
- Consolidate modal includes to 1
- Fix the X icon. Export the X icon from Figma and use it on that 1 modal includes.

thekaveman (approved):

PR #1608, opened 2023-08-01
Fix: Login.gov button width on Tablet

### LINKED ISSUE #1574: Fix "Get started with Login.gov" button on Tablet widths
It should not go into 2 lines if we can avoid it cleanly.

https://github.com/cal-itp/benefits/issues/1556#issuecomment-1646036491

machikoyasuda:
fix #1574 

## What is the problem anyway?

At this specific range of widths (1279px - 992px), the Login.gov button breaks into 2 lines in a way that is not part of the design.

image
image

The button should only exist in 2 designs, as shown on Figma:

image

## What this PR does
- Make sure the Login.gov button only shows up in those 2 ways
- Ensure the 2 Login.gov buttons are the correct dimensions
- Modify the `start.html` so that I can adjust the columns for the `start--seniors` includes only, but keep the rest the same.
- Widened the container from `col-3` to `col-4` to allow for the button to remain at width: 289px as the window scales down from 1279px to the maximum of 992px. Add flex-box classes and a max-width declaration so that the button stays at the proper width and also is right-aligned. This widening is a one-off for _this includes with the Login.gov button only_. Not to be done for anything else.

## How to test
- Get to Elig Start for Login.gov/Seniors and check the button in all possible screen widths
- Ensure no regressions on the rest of the Elig Starts (Veterans, Agency Card)
- Make sure button looks good in focus, hover modes

### Desktop/Tablet
- Width: 289px
- Height: 41px
- Font size: 16px
- Space between top/bottom of the text and the edge: 12px (Add the 10px padding and 2 px border for total 12px)

image
image
image

### Mobile
- Width: 310px
- Height: 72px
- Font size: 20px
- Space between top/bottom of the text and the edge: 12px

image
image

### DIFF
--- a/benefits/eligibility/templates/eligibility/start--mst-courtesy-card.html
+++ b/benefits/eligibility/templates/eligibility/start--mst-courtesy-card.html
@@ -16,7 +16,7 @@
 {% endblock media-item %}
 
 {% block call-to-action %}
-  <div class="row d-flex flex-column-reverse justify-content-lg-end flex-lg-row">
+  <div class="row d-flex justify-content-lg-end">
     <div class="col-lg-3 offset-2 offset-sm-2 offset-lg-0 col-sm-8 col-8">
       {% url "eligibility:confirm" as button_url %}
       <a href="{{ button_url }}" class="btn btn-lg btn-primary" role="button">{% translate "eligibility.buttons.continue" %}</a>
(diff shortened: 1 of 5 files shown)


thekaveman (approved):
✅ Get to Elig Start for Login.gov/Seniors and check the button in all possible screen widths
✅ Ensure no regressions on the rest of the Elig Starts (Veterans, Agency Card)
✅ Make sure button looks good in focus, hover modes

Issue #1444, opened 2023-06-22
Design modal on the [home indirect entry] view to accommodate 4 agencies


srhhnry:
Ensure the indirect entry modal can accommodate up to 4 agencies (depending on evaluation of this approach). We need to accommodate at least three agencies by the end of the year, but four is a better target.

srhhnry:
Rough draft completed today. Can be [seen here](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=9799%3A52593&mode=design&t=tmnfTqohN6GLSQtu-1).

srhhnry:
@indexing these screens are now ready for review

indexing:
@srhhnry I reviewed the design. Overall it works! When we take the layout from rough draft to final design, I'd like to added the following as acceptance criteria. For some reason, I don't see and thus cannot edit the body for this issue.

# Acceptance criteria

- [ ] Create a component in Figma for the transit card to standardize the layout. (Use auto layout if possible.)
- [ ] Specify a uniform size for the agency image asset in the component using dimensions with a common denominator of 4, e.g. 64 x 128. (This will ensure we ask agencies for an asset will work with our design.)
- [ ] Specify a max character length for agency name. (Ensures we ask for a name that will work with the design.)
- [ ] Specify default sort order for the list of agencies.

srhhnry:
A note about the acceptance criteria:

1. Gov agencies are not the most most reliable sources of their own graphics. Files can be from years ago with no updates (e.g. .indd) have been copied over in weird ways (e.g. using a degraded .jpg), or the agency might not have a visual person on staff who can confidently hand over the right file. Our general approach has been to ask for the largest logo file size they have, and adjust from there. This approach works currently because we aren't onboarding an agency every month. 

I very much think asking for a uniform size or fidelity will be necessary at scale, but also anticipate we'll be shifting the design for agency selection (and benefit selection) before then. That ALL said, sizes for different types of icons listed in the next comment.

srhhnry:
Card components can be found [here in the components library](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=2-11&mode=design)

Working on adding autolayout and updating all designs to include these components

srhhnry:
Desktop autolayout update is almost complete--still working out a bug wherein I can swap logos seamlessly in the master component but not in an instance of the component.

[...]

PR #1610, opened 2023-08-02
Fix: Line-height across the app for H1, H2, H3, H4

### LINKED ISSUE #1573: Fix line-height for headlines
1.5 is now 1.3

> https://github.com/cal-itp/benefits/issues/1556#issue-1816156909

Double check that the line heights didn't change for other elements.

machikoyasuda:
fixes #1573 

Line-height should always be a unitless value. https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values Do not use rems, pixels for line-height.

## What this PR does
- Applies new line-height to h1, h2, h3, h4
- H1-H3 are 1.3
- H4 is 1.2
- These values are the same across Desktop and Mobile.

## Steps I took to run through this PR

1. Confirm on Figma that H1, H2, H3, H4 now have _different_ line-height values from the standard body line height.
2. Confirm that the `--bs-body-line-height` variable is correct, set at 1.5, and applied to all the places it should.
3. Create a new variable: `--heading-line-height` of value 1.3.
4. Create a new variable: `--h4-line-height` of value 1.2. H4 is only used in _one spot_ in the whole app, and that is for the Agency Selector modal, specifically, the name of the transit agency. The 1.2 value allows for the transit agency name to be tighter.
5. Apply the `--heading-line-height` to H1, H2, H3
6. Apply `--h4-line-height` outlier exception to H4.
7. Deleted the custom code setting `.card .card-title`'s line-height, which is an H4, to fall back and use the H4 line-height. 
8. Audit the CSS file for any other line-height declarations and make sure they're correct/appropriate.
9. Confirm local vs. dev site vs. Figma for desktop and mobile headlines 

### Modal titles on Mobile
Side-by-side comparison of Figma, this PR and the Dev site for modal titles. Modal titles on Dev have a line-height of 1.5, and now, with this PR, they are tightened to 1.3.
image

### Index/Agency Index on Desktop
Total height of the H1 and H2 types are tighter now
image

### 2-line Headlines on Desktop
image

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -14,7 +14,9 @@
   --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
     "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
   --bs-body-font-family: var(--bs-font-sans-serif);
-  --bs-body-line-height: 1.5;
+  --bs-body-line-height: 1.5; /* Displayed as 150% on Figma */
+  --heading-line-height: 1.3; /* Displayed as 130% on Figma */
+  --h4-line-height: 1.2; /* Displayed as 120% on Figma */
   --body-letter-spacing: 0.05em;
   --dark-color: #212121;
   --hover-color: #044869;
@@ -162,7 +164,6 @@ a[target="_blank"]::after {
 
 /* Headlines */
 /* All headlines */
-/* All headlines share font-weight, letter-spacing, line-height and margin */
 h1,
 h2,
 .h2,
@@ -171,7 +172,6 @@ h3,
 h4 {
   font-weight: var(--bold-font-weight);
   letter-spacing: var(--body-letter-spacing);
-  line-height: var(--bs-body-line-height);
   margin: 0;
 }
 
@@ -183,6 +183,7 @@ h1 {
   font-size: var(--h1-font-size);
   text-align: var(--h1-text-align);
   padding-top: calc(70rem / 16);
+  line-height: var(--heading-line-height);
 }
 
 /* H2 */
@@ -191,6 +192,7 @@ h1 {
 h2,
 .h2 {
   font-size: var(--h2-font-size);
+  line-height: var(--heading-line-height);
 }
 
 /* H3 */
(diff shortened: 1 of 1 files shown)


machikoyasuda:
@srhhnry How does this change look to you so far?

benefits/static/css/styles.css
```diff
@@ -199,6 +201,13 @@ h2,
...
 .h3 {
   font-size: var(--h3-font-size);
+  line-height: var(--heading-line-height);
+}
+
+/* H4 */
+/* Desktop only: Used for Agency Selector card, agency name */
+h4 {
+  line-height: var(--h4-line-height);
 }
```
machikoyasuda, line 211: One could argue we should be declaring the H4 font sizes here, but I opted not to for this PR. I did add that comment, thought, that it's only used on Agency Selector card title. Next time we do use the H4, it can be refactored.

angela-tran: I get that you're defining line-height on `h4` because the Agency Selector card title is the only place we actually use an `h4` element. And here you're saying that, by the same reasoning, we could also define all the styles for the Agency Selector card title on `h4`. 

I want to note that this is slightly different than what Figma says, since the Style Guide has `h4` with a certain font-size, line-height, etc. and the Components page has the Agency Selector Cards with a certain font-size, line-height, etc.

This is fine though since this code change gives us the net-effect we want and, as you noted,
> Next time we do use the H4, it can be refactored.

angela-tran (approved):

benefits/static/css/styles.css
machikoyasuda: Yea the next Letter-Spacing PR will apply the `h4` letter-spacing styles to this `h4`, and then the Agency Selector's H4 will get those styles. 

And we know we're gonna refactor the entire modal layout/includes after https://github.com/cal-itp/benefits/issues/1444 this ticket for SBMTD, we can revisit how the h4 on the modal is constructed and styled more thoroughly then.

thekaveman (approved):

srhhnry:
Late on this comment but line height looks fine to me from that screenshot.

machikoyasuda:
Awesome thank you @srhhnry. This PR is merged into `dev`, which means the changes are live and available to test out here dev-benefits.calitp.org/

Issue #1582, opened 2023-07-31
Design QA on dev-benefits.calitp.org


thekaveman:
Confirm design on dev-benefits.calitp.org matches Figma designs.

thekaveman:
@thekaveman to schedule a quick check-in with @srhhnry

srhhnry:
Created[ a new issue](https://github.com/cal-itp/benefits/issues/1609) per yesterday's [slack convo](https://compilerla.slack.com/archives/C02GQ7AJXQE/p1690909867570869) for this.

srhhnry:
QA complete with [checklist here](https://docs.google.com/document/d/1PlcoyMNSkwT2VJpPfmuS-cun4snYV5Bvse5zrYKWG08/edit?usp=sharing)

Let me know if anything needs additional explanation!

machikoyasuda:
Thanks @srhhnry 

Gonna start making tickets for these day and will ping you for more details on some of them!

PR #1639, opened 2023-08-10
Fix: Button uses primary-color for background

### LINKED ISSUE #1635: Bug: Buttons should be primary-color
Primary button appears to be #046b99 and not #045B86.

Change all buttons to #045B86 (override the CA State Web Template color), and make sure to do that for border-color too.

machikoyasuda:
fix #1635 

This fix came from @srhhnry's #1582 

## What this PR does
- Previously, button background color (and 2px border color of the same color) was coming from CA State Web Template. Now we're overriding the CA State Web Template and declaring it in the CSS itself.
- Add the hover color to button as well

image

## Buttons affected
- All CTAs with blue background
- Doesn't affect the button in `disabled` state

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -373,6 +373,8 @@ footer .footer-links li a.footer-link:visited {
 }
 
 .btn.btn-lg.btn-primary {
+  background-color: var(--primary-color);
+  border-color: var(--primary-color);
   border-width: 2px;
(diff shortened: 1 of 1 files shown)


angela-tran (approved):

machikoyasuda:
@angela-tran Rebased dev / fixed conflicts

angela-tran (approved):

PR #1642, opened 2023-08-10
Refactor: Footer - Re-do mobile footer alignment

### LINKED ISSUE #1636: Bug: Mobile footer spacing is incorrect
- Mobile footer needs to be wrapped in a row/container, so it can have the same spacing from the right as the Header
- Need to write a separate Mobile Footer.

image

machikoyasuda:
fixes #1636 

## What this PR does
- Creates an entirely separate mobile-only footer
- Toggle footer display at mobile/desktop width
- Ensure no Desktop regressions

## Mobile footer specs
- 3 different divs. Cannot be encapsulated into 1 parent div... doesn't work. The parent div has to be the `<footer class="navbar"` (which acts like a `row`, and is what the `<header>` uses). We want the footer link widths to be the same as that of the `<header>`. 
- 2 divs with the 2 different links are `container` width, to get that `container` scaling
- 1 div is `container-fluid` width with `p-0`, so that the horizontal white line can be 100% width of the desktop
- Removes all manual padding/margin that was previously positioning the mobile footer link, so the focus ring looks correct.

## Screenshots

I temporarily hid the `main` container from this view, so it's easy to see the Header/Footer alignment. 

Kapture 2023-08-10 at 14 50 50
Click [here](https://user-images.githubusercontent.com/3673236/259882532-d7777ec0-0a62-47e1-bc1b-6458eb475136.gif) to view gif again

image

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -127,7 +127,7 @@
       <div class="d-block d-lg-none container">
         <ul class="col-12 footer-links ps-0 mb-0">
           <li>
-            <a href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer" class="">Privacy Policy</a>
+            <a href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer" class="">{% translate "Privacy Policy" %}</a>
           </li>
         </ul>
       </div>
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -102,8 +102,8 @@
       {% endblock main-content %}
     </main>
 
-    <footer id="footer">
-      <div class="container">
+    <footer id="footer" class="navbar py-0">
+      <div class="container d-none d-lg-block">
         <ul class="footer-links m-0 p-0 list-unstyled d-lg-flex gap-lg-4">
           <li>
             <a class="footer-link m-0 ms-5 ps-lg-0" href="{% url "core:help" %}">{% translate "Help" %}</a>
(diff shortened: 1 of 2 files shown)


machikoyasuda:
There's a small footer regression on Dev right now: 
image


But this PR should fix it.

angela-tran:
> There's a small footer regression on Dev right now: 
> ...
> But this PR should fix it.

Do you know what was the cause? Just curious

machikoyasuda:
@angela-tran I think it was the Focus Ring PR

machikoyasuda:
Rebased dev, resolved conflicts + resolved bug:
- Ensure footer has no extra margin on Desktop
- Ensure footer focus rings are white on Mobile and Desktop
- Ensure footer padding alignment is still the same

image
image
image

benefits/core/templates/core/base.html
```diff
@@ -102,17 +102,36 @@
...
     </main>
 
-    <footer id="footer">
-      <div class="container">
+    <footer id="footer" class="navbar py-0">
+      <div class="container d-none d-lg-block">
         <ul class="footer-links m-0 p-0 list-unstyled d-lg-flex gap-lg-4">
           <li>
-            <a class="footer-link m-0 ms-5 ps-lg-0" href="{% url "core:help" %}">{% translate "Help" %}</a>
+            <a class="footer-link m-0 p-0" href="{% url "core:help" %}">{% translate "Help" %}</a>
```
machikoyasuda, line 109: @angela-tran Basically.. the bug came from changing `ps-5` to `ms-5` but _not_ also changing the `ps-lg-0` to `ms-lg-0`, so it added this extra margin-left you can see on Desktop https://dev-benefits.calitp.org/ right now

machikoyasuda: But it doesn't matter anymore cos this PR will not show this footer on Mobile anymore.

benefits/static/css/styles.css
```diff
@@ -257,38 +257,36 @@ h4 {
...
-/* Footer has same font styles on all screen widths */
-/* Mobile first: One link per row, each link is 50px height */
-/* Screen width above 992px - Footer is 50px height */
-/* Total footer height is 50px on Desktop */
-/* And 102 on Mobile */
-/* Footer breakpoint set to 992px */
-/* Note: Not all Footer styles are written in mobile-first */
-/* to style the full-width line on mobile */
-/* and the desktop container-width sizing */
-
```
angela-tran, line 276: 🙌

angela-tran (approved):
Nice!

PR #1654, opened 2023-08-14
Fix: Remove alt tag from decorative icons

### LINKED ISSUE #1651: Remove alt text from icons
Our icons are decorative elements and should not have alt text. The icons do not convey anything not stated in text and so adding alt text adds audio clutter. I added a clarifying flag to the components page: 
image

_Originally posted by @srhhnry in https://github.com/cal-itp/benefits/issues/1410#issuecomment-1678135953_

## Acceptance criteria

- [x] All of the icons in the image above have `alt=""` for their alt text attribute
- [x] All prior used alt text is removed from the PO files
- [x] Copy Master spreadsheet does not have alt text strings for icons.

machikoyasuda:
closes #1651 

## What this PR does
- Add `alt=""` to the `icon` includes. Remove alt validation
- Remove alt tag copy and call from happy bus, sad bus, contactless pay icons, id card icons

### DIFF
--- a/benefits/core/templates/core/includes/icon.html
+++ b/benefits/core/templates/core/includes/icon.html
@@ -1,9 +1,2 @@
-
-{% if not name %}
-  <p class="bg-grey-lightest color-standout">
-    Missing <strong><code>name</code></strong> argument
-  </p>
-{% else %}
-  {% load static %}
-  <img class="icon" width="150" height="150" src="{% get_static_prefix %}img/icon/{{ name }}.svg" alt="" />
-{% endif %}
+{% load static %}
+<img class="icon" width="150" height="150" src="{% get_static_prefix %}img/icon/{{ name }}.svg" alt="" />
--- a/benefits/core/templates/core/includes/icon.html
+++ b/benefits/core/templates/core/includes/icon.html
@@ -3,11 +3,7 @@
   <p class="bg-grey-lightest color-standout">
     Missing <strong><code>name</code></strong> argument
   </p>
-{% elif not alt %}
-  <p class="bg-grey-lightest color-standout">
-    Missing <strong><code>alt</code></strong> argument
-  </p>
 {% else %}
   {% load static %}
-  <img class="icon" width="150" height="150" src="{% get_static_prefix %}img/icon/{{ name }}.svg" alt="{{ alt }}" />
+  <img class="icon" width="150" height="150" src="{% get_static_prefix %}img/icon/{{ name }}.svg" alt="" />
 {% endif %}
--- a/benefits/eligibility/templates/eligibility/includes/modal--contactless.html
+++ b/benefits/eligibility/templates/eligibility/includes/modal--contactless.html
@@ -13,7 +13,7 @@
     </p>
     <p class="pt-4">{% translate "The contactless symbol is four curved lines, like this:" %}</p>
     <p class="contactless-symbol pt-3">
-      <img class="icon mx-auto d-block" width="30" height="37.5" src="{% static 'img/icon/contactless.svg' %}" alt="" />
+      <img class="icon mx-auto d-block" width="30" height="37.5" src="{% static 'img/icon/contactless.svg' %}" alt="{% translate "Four curved lines on contactless-enabled cards" context "image alt text" %}" />
     </p>
     <p class="pt-4">{% translate "Your card must include a Visa or Mastercard logo." %}</p>
     <p class="pt-4">
--- a/benefits/eligibility/templates/eligibility/includes/modal--contactless.html
+++ b/benefits/eligibility/templates/eligibility/includes/modal--contactless.html
@@ -13,7 +13,11 @@
     </p>
     <p class="pt-4">{% translate "The contactless symbol is four curved lines, like this:" %}</p>
     <p class="contactless-symbol pt-3">
-      <img class="icon mx-auto d-block" width="30" height="37.5" src="{% static 'img/icon/contactless.svg' %}" alt="{% translate "Four curved lines on contactless-enabled cards" context "image alt text" %}" />
+      <img class="icon mx-auto d-block"
+           width="30"
+           height="37.5"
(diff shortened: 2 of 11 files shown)


benefits/core/templates/core/includes/icon.html
```diff
@@ -3,11 +3,7 @@
   <p class="bg-grey-lightest color-standout">
     Missing <strong><code>name</code></strong> argument
```
thekaveman, line 4: Can we delete this too? This is a holdover I think we don't need it anymore.

benefits/eligibility/templates/eligibility/includes/modal--contactless.html
```diff
@@ -13,11 +13,7 @@ <h2 id="contactless-modal" class="me-4 me-md-0">{% translate "Learn more about c
     </p>
     <p class="pt-4">{% translate "The contactless symbol is four curved lines, like this:" %}</p>
     <p class="contactless-symbol pt-3">
-      <img class="icon mx-auto d-block"
-           width="30"
-           height="37.5"
-           src="{% static 'img/icon/contactless.svg' %}"
-           alt="{% translate "Four curved lines on contactless-enabled cards" context "image alt text" %}" />
```
thekaveman, line 20: @srhhnry @indexing this is the "contactless symbol" on the help page / modal -- is this an image/icon that deserves alt text? It is described but also referenced directly in the preceding sentence `The contactless symbol is four curved lines, like this:` so not sure where that falls.

thekaveman: @machikoyasuda depending on the answer to the question above, the alt text on the help page version of this icon would also need to be removed.

thekaveman (changes requested):

benefits/eligibility/templates/eligibility/includes/modal--contactless.html
machikoyasuda: I opted to keep it cos it's describing an _actual_ thing. Maybe we can improve the alt tag to be a bit more than the preceding sentence?

thekaveman: I hear you, this icon kind of feels different. I'm pointing out that this icon shows up in 2 places (both with alt text), and whether we remove the alt text or update the wording -- it should be done the same in each place.

srhhnry: So there's definitely a bit of an interpretative art to all of this alt text stuff (especially in whether or not something is decorative). I'm still new to understanding it, but one of my approaches is to read through the page with the alt text and try to assess how it feels as a person using a screenreader. In this case, we'd hear something like: "The contactless symbol is four curved lines, like this. Image. Four curved lines on contactless-enabled cards."

That feels redundant to me, and also in some ways assumes the user is encountering the symbol for the first time on our site (if we were a bank, I'd be more inclined to think that's true). 

My inclination is to go with an acknowledgement of the presence of the symbol, rather than explanation. Something like: "The contactless symbol is four curved lines, like this. Image. Contactless symbol." 

I'm not sure though. My thinking is to not overburden the user with redundant or complicated text and to get them through the flow quickly. I can see a case for describing the symbol though. I just don't know whether a low or no-sighted person is going to want to hear an in-depth description of a visual indicator meant to help a sighted person identify an attribute.

srhhnry: Like, I think this symbol counts as decorative BECAUSE it is described in the text, but the wording of the sentence would be awkward with a null value (The contactless symbol is four curved lines, like this. Your card....")

thekaveman: > Like, I think this symbol counts as decorative BECAUSE it is described in the text, but the wording of the sentence would be awkward with a null value

💯

srhhnry: Okay, took a look at what @machikoyasuda found on slack and I think the Capital One description is fine ("four curved lines that get bigger from left to right"). So the sentence would read: "The contactless symbol is four curved lines, like this. Image. Four curved lines that get bigger from left to right."

It's redundant. Maybe it's not necessary. But also, I am having trouble fully committing to that the idea that the symbol doesn't need to be described at all.

thekaveman: @indexing thoughts?

indexing: I agree with @srhhnry's approach. Include the alt text for the icon in this context using the CapitalOne description. I'll add it to the spreadsheet and we can get it translated for the Sept sprint.

machikoyasuda: So for now, we won't delete the alt text for contactless pay indicator. And we'll create a new alt text in a future ticket, in a future sprint.

Is the rest of this PR okay for review? @thekaveman

benefits/eligibility/templates/eligibility/includes/modal--contactless.html
```diff
@@ -13,11 +13,7 @@ <h2 id="contactless-modal" class="me-4 me-md-0">{% translate "Learn more about c
...
+      <img class="icon mx-auto d-block" width="30" height="37.5" src="{% static 'img/icon/contactless.svg' %}" alt="" />
```
thekaveman, line 16: Need to add the `alt` text back here per above conversation

thekaveman (changes requested):

benefits/eligibility/templates/eligibility/includes/modal--contactless.html
machikoyasuda: Added back: https://github.com/cal-itp/benefits/pull/1654/commits/bf860bc0f9fd03d52c51cc4357fd771eee6d9114

thekaveman (approved):
👍

PR #1658, opened 2023-08-15
Fix: Focus ring on Card on Google Chrome

### LINKED ISSUE #1659: Chrome: 1px extra outline offset on focus ring for card
## To Reproduce

1. Focus on Card on Chrome

## Expected behavior

Eliminate 1px offset on focus ring for card on Chrome

## Screenshots

image

- Chrome

machikoyasuda:
fixes #1659 

## What this PR does
- Explicitly set the `.card`'s focus ring's `outline-offset` to 0, to override Chrome's default browser CSS
- `.card`'s focus ring should be flush against the card, without any space in between:

imagevs. on Chrome dev right now:
image

## What accounts for the cross-browser differences?

Chrome's browser adds this CSS to all websites:

```css
a:-webkit-any-link:focus-visible {
    outline-offset: 1px;
}
```

image

Firefox and Safari do not set an `outline-offset` at all, so it defaults to `initial`, which is 0. I had only tested in Firefox and Safari, so I didn't notice that Chrome _does_ add an 1px offset.

This PR takes care of that browser difference.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -719,6 +719,7 @@ a.card:focus-visible {
 .card:focus,
 .card:focus-visible {
   outline: 3px solid var(--focus-color) !important; /* override CA State Web Template */
+  outline-offset: 0 !important;
   box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25) !important;
 }
 


thekaveman (approved):
Confirmed in Chrome:

image

angela-tran (approved):

PR #1663, opened 2023-08-15
Fix: Skip Nav - Remove skip nav page jank

### LINKED ISSUE #1661: Bug: Skip Nav causes page jank
## To Reproduce

broken
Note the page jank (the page scrolling back _up_ by itself), when the Previous Page button gets tabbed.

## Expected behavior

No page jank, like this:

fixed

Desktop, all browsers

Skip Nav not applicable for Mobile.

## Additional context

[StackOverflow](https://stackoverflow.com/questions/4086107/fixed-page-header-overlaps-in-page-anchors)

machikoyasuda:
closes #1661 

This was bothering me a lot 😓 and I finally looked up a solid, easy solution w/o any drawbacks. Currently, when using the Skip Nav to get to `#main-content` and then continuing to use the keyboard for navigation, the site will scroll you back _up_ the page when you go from `#main-content` to `Previous Page`. It creates a jarring user experience, where the page goes down, and then back up. This PR fixes it by adding a `scroll-padding-top` that's greater than the height of the nav (80px), so the screen doesn't move back up. This solution also doesn't affect anything else (like the footer position, main container height, etc).

| Before     | After     |
| --------  | -------- |
|  broken|  fixed |

- Uses `scroll-padding-top` from this: https://css-tricks.com/fixed-headers-on-page-links-and-overlapping-content-oh-my/
- Note: You have to test it with the DEBUG bar hidden. Doesn't work with the DEBUG bar on.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -52,6 +52,10 @@
   /* prettier-ignore */
 }
 
+html {
+  scroll-padding-top: 120px;
+}
+
 body {
   font-size: 100%;
   color: var(--text-primary-color);
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -53,7 +53,7 @@
 }
 
 html {
-  scroll-padding-top: 120px;
+  scroll-padding-top: 101px; /* Header Height (80px) + Skip Nav Height (21px) = 101px */
 }
 
 body {


benefits/static/css/styles.css
```diff
@@ -52,6 +52,10 @@
   /* prettier-ignore */
 }
 
+html {
+  scroll-padding-top: 120px;
```
thekaveman, line 56: Can you please add a little comment here on why this is necessary / the significance of the value?

machikoyasuda: I honestly just tried out a bunch of numbers until it worked.

machikoyasuda: Figured it out! 

https://github.com/cal-itp/benefits/pull/1663/commits/3711239b692f18f0ed4ad749ed55f0dfa2fbba47

Interesting observation: In Chrome, it works with a padding-top of 80px, but doesn't for Firefox. I had to test on both browsers, but ultimately Firefox needed a larger number. Ended up with:

```css
scroll-padding-top: 101px; /* Header Height (80px) + Skip Nav Height (21px) = 101px */
```

thekaveman (approved):

PR #1665, opened 2023-08-16
Feat: EligibilityVerifier active flag

### LINKED ISSUE #1662: Add active flag to EligibilityVerifier
Like the `TransitAgency.active` flag, which allows us to configure which agencies are active and not at runtime, it would be useful to have an `EligibilityVerifier.active` flag with the same behavior.

## Acceptance Criteria

- [x] Add a boolean field to the `EligibilityVerifier` model, default `False`
- [x] Read flag from environment in migration for each verifier model created
  - [x] environment set via Terraform from Key Vault secrets
- [x] `eligibility:index` only shows active verifiers

## Additional context

We'll want this to be able to deploy our code fixes to production and have more control over when the Veterans flow is turned on.

angela-tran:
Closes #1662

This PR adds `active` as a field to `EligibilityVerifier` and updates the `EligibilityVerifierSelectionForm` so that it only shows active verifiers.

The `BooleanField` on the model itself defaults to false, and then is overwritten by whatever the data migration specifies if no environment variable is found.

### To set verifiers as active locally
In .env:
```
MST_OAUTH_VERIFIER_ACTIVE=True
MST_VETERAN_VERIFIER_ACTIVE=True
COURTESY_CARD_VERIFIER_ACTIVE=True
SACRT_OAUTH_VERIFIER_ACTIVE=True
SBMTD_OAUTH_VERIFIER_ACTIVE=True
```
In terminal:
```
set -a
source .env
./bin/init.sh
```


angela-tran:
Working on fixing the failing Pytests. Will look into Cypress tests after that.

Side question: is anyone else not able to use VS Code's Testing integration? I'm getting this on any branch, not just this PR's:
```
  File "/home/calitp/.vscode-server/extensions/ms-python.python-2023.14.0/pythonFiles/vscode_pytest/__init__.py", line 613, in execution_post
    addr = ("localhost", int(testPort))
ValueError: invalid literal for int() with base 10: '0.0'
```
image

I'm on
```
Visual Studio Code
Version: 1.81.1
Commit: 6c3e3dba23e8fadc360aed75ce363ba185c49794
Date: 2023-08-09T22:22:42.175Z (6 days ago)
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Code/1.81.1 Chrome/108.0.5359.215 Electron/22.3.18 Safari/537.36
```

thekaveman:
@angela-tran RE: default values -- if it makes it easier, I think we can hardcode MST options as True, like we do for their Transit Agency. For now anyway...

angela-tran:
> @angela-tran RE: default values -- if it makes it easier, I think we can hardcode MST options as True, like we do for their Transit Agency. For now anyway...

Good idea. I think that'll fix the Cypress tests

thekaveman:
@angela-tran I'm able to run the tests on my VS Code. I am on 1.81.1

```
Version: 1.81.1 (user setup)
Commit: 6c3e3dba23e8fadc360aed75ce363ba185c49794
Date: 2023-08-09T22:22:42.175Z
Electron: 22.3.18
ElectronBuildId: 22689846
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Windows_NT x64 10.0.19045
```

machikoyasuda (approved):
image
with env set as
image

thekaveman (approved):
With the default for SacRT 🤣, works as expected:

image

PR #1705, opened 2023-09-06
Eligibility Start: Veterans - Use Older Adult copy

machikoyasuda:
closes #1693 

image

## What this PR does

- Copy: Use Older Adult copy for Veterans on Elig Start
- Button: Use Older Adult Login.gov button for Veterans on Elig Start
- Templates: Rename `senior` modal to `login-gov` modal.
- Make sure 2 templates affected on this PR are both set to `Spaces: 2` formatting.

Note: I noticed that the indentation on the Figma designs are a lot wider than it is on the app... but this ticket is just a copy ticket, so it's not going to address this discrepancy.

PR #1706, opened 2023-09-06
Eligibility Index: Veterans - Use Older Adult copy

machikoyasuda:
This PR depends on #1705 to be merged first.

## What this PR does
- Copy: On Elig Index, for the Veteran pathway, change the last sentence to read: `You will need to verify your identity with Login.gov.`, with a link to the Login.gov modal. *Both* links from the Veteran and the Older Adult radio input open the _same exact_ modal.
- Templates: Like in #1705, rename senior modal to login-gov modal.

## Includes, CSS changes
- CSS: Convert the `#login` id to a class. This is necessary b/c now, for the first time, a page has more than one Login.gov button on it. This requires changing _all_ calls to the `#login` id to the class. This affects the Eligibility Start page for Veterans, Eligibility Start page for Older Adults, and the Enrollment Success page.
- Modal trigger element: Previously the Modal trigger element needed an `id` field to pass in the `id` of `login`, to show the color Login.gov trigger image. Now, though, there is no `id` necessary to pass in. So I removed the field entirely from the includes template. It was actually causing an interesting bug on the radio button field page, where the id of the _radio button field_ was being used in the modal trigger instead - which was resulting in 2 links with the same id. And that was coming up in a11y testing.

## How to test this PR
- Test the regular app flow
- Pay attention to all buttons with the Login.gov icon: Elig Start, Log out on Enrollment Success, Login.gov modal triggers. Ensure no visual regressions.

image
image
image
image


benefits/locale/es/LC_MESSAGES/django.po
```diff
@@ -259,8 +259,8 @@ msgstr ""
 "califican para una serie de programas de tarifas reducidas. Este beneficio "
 "de tránsito debe renovarse en el futuro en función de la fecha de "
 "vencimiento de la tarjeta de cortesía. Obtenga más información en el <a "
-"href='https://mst.org/riders-guide/how-to-ride/courtesy-card/' "
-"target=\"_blank\" rel=\"noopener noreferrer\">MST Riders Guide</a>."
+"href='https://mst.org/riders-guide/how-to-ride/courtesy-card/' target="
+"\"_blank\" rel=\"noopener noreferrer\">MST Riders Guide</a>."
```
machikoyasuda, line 263: Django loves to move this every other pull request.

thekaveman: I tend to avoid committing this change if possible.

benefits/static/css/styles.css
```diff
@@ -375,6 +375,8 @@ footer .footer-links li a.footer-link:visited {
 
 .btn.btn-lg.btn-primary:focus,
 .btn.btn-lg.btn-primary:focus-visible,
+.btn.btn-lg:focus,
+.btn.btn-lg:focus-visible,
```
machikoyasuda, line 379: This is for the Login.gov Sign Out button on the Enrollment Success page for the Login.gov flows.

benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html
```diff
@@ -7,8 +7,9 @@
 
 {% block description %}
   {% blocktranslate trimmed %}
-    This option is for people who have served in the active military, naval, or air service, and who were discharged or
-    released therefrom under conditions other than dishonorable. You will need to
-    <a href="https://www.va.gov/resources/verifying-your-identity-on-vagov/" target="_blank" rel="noopener noreferrer">verify your identity through VA.gov</a>.
+    This option is for people who have served in the active military, naval, or air service, and who were discharged or released therefrom under conditions other than dishonorable. You will need to verify your identity with
   {% endblocktranslate %}
+
+  {% include "core/includes/modal-trigger.html" with classes="border-0 bg-transparent p-0 login" modal="modal--login-gov" login=True period=True %}
```
machikoyasuda, line 13: @thekaveman @angela-tran This modal trigger link opens the `modal--login-gov` modal, which is included into the page from the `selection-label--senior.html` file. That means that technically, under this design, if a transit agency *ONLY* offers Veterans benefits, but doesn't offer an Older Adult one, this template will have to be modified to also include the modal code. In other words, the Veteran selection label would require the Older Adult one to also be there.

Alternatively, I could also add this line here:

```
{% include "eligibility/includes/modal--login-gov-help.html" with id="modal--login-gov" size="modal-lg" header="p-md-2 p-3" body="pb-md-3 mb-md-3 mx-md-3 py-0 pt-0 absolute-top" %}
```

but then this HTML modal code would be duplicated on this page, and since there would be 2 modals with the same id (`modal--login-gov`), we'd have to have 2 different IDs pointing to duplicated HTML. Not sure which approach is better.

machikoyasuda:
Ready for review @thekaveman @angela-tran

benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html
angela-tran: Hmm, that is tricky. With your second idea - the one about including the login-gov-help modal in the veteran selection-label - maybe the selection-label include templates can have a variable that says whether to include the modal or not?

machikoyasuda: Hrmm - yeah, but that variable would have to live in `0002_data.py` because that's where the radio button selector templates get required for each agency: 
image

And that seems like a little too much for this situation?

Might just be better to go with adding the modal _all_ the time just to make each selection template independent.

machikoyasuda: Update: https://github.com/cal-itp/benefits/pull/1706/commits/ff4ee183ace750a19cbdab363386748be66758b1

The page now has 2 modals with 2 different IDs. The Veteran selector will have a modal with the ID `"modal--login-gov-veterans"` and the trigger will open that one.

angela-tran: > And that seems like a little too much for this situation?
>
> Might just be better to go with adding the modal all the time just to make each selection template independent.

Oh I see. Yeah, I didn't look too deeply at how the variable would get passed in, and what you're saying makes sense. Seems like having 2 modals is our best option here

benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html
```diff
@@ -7,8 +7,11 @@
 
 {% block description %}
   {% blocktranslate trimmed %}
-    This option is for people who have served in the active military, naval, or air service, and who were discharged or
-    released therefrom under conditions other than dishonorable. You will need to
-    <a href="https://www.va.gov/resources/verifying-your-identity-on-vagov/" target="_blank" rel="noopener noreferrer">verify your identity through VA.gov</a>.
+    This option is for people who have served in the active military, naval, or air service, and who were discharged or released therefrom under conditions other than dishonorable. You will need to verify your identity with
   {% endblocktranslate %}
+
+  {% include "core/includes/modal-trigger.html" with classes="border-0 bg-transparent p-0 login" modal="modal--login-gov-veterans" login=True period=True %}
```
thekaveman, line 13: Quibble: `veteran` should be singular in the modal ID here and below, to match the include naming format.

benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html
```diff
@@ -7,8 +7,11 @@
...
+
+  {% include "eligibility/includes/modal--login-gov-help.html" with id="modal--login-gov-veterans" size="modal-lg" header="p-md-2 p-3" body="pb-md-3 mb-md-3 mx-md-3 py-0 pt-0 absolute-top" %}
```
thekaveman, line 15: `veteran` singular in the modal id

thekaveman (changes requested):

benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html
machikoyasuda: Nice catch! Done here: https://github.com/cal-itp/benefits/pull/1706/commits/57553fd19fa0bb7a6925dbc82e7d4d8f7bed444e

Ready for rereview

thekaveman (approved):
👍

PR #1743, opened 2023-09-20
Docker: Revert platform change

### LINKED ISSUE #1730: Revert Fix: Docker platform issue
Revert #1354 as the `platform` designation is now causing a new issue running containers locally, and other related repositories like `eligibility-server` do not have that problem and do not have the `platform` designation.

## To Reproduce

1. On macOS M2, rebuild the benefits app container `DOCKER_BUILDKIT=0  docker compose build --no-cache client`
2. Try to run the container and attach to bash: `docker compose run --entrypoint bash client`
3. See error similar to `image with reference benefits_client:latest was found but does not match the specified platform: wanted linux/amd64, actual: linux/arm64/v8`

## Expected behavior

The container should run and the terminal should be attached to `bash` inside the container.

machikoyasuda:
closes #1730 

When I run `docker compose run --entrypoint bash client` on my machine now, with this PR, I'm able to get to bash properly.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -8,7 +8,6 @@ services:
       dockerfile: appcontainer/Dockerfile
     image: benefits_client:latest
     env_file: .env
-    platform: linux/amd64
     ports:
       - "${DJANGO_LOCAL_PORT:-8000}:8000"
 
@@ -22,7 +21,6 @@ services:
     entrypoint: sleep infinity
     depends_on:
       - server
-    platform: linux/amd64
     ports:
       - "${DJANGO_LOCAL_PORT:-8000}:8000"
     volumes:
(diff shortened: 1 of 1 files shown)


thekaveman:
And just to confirm @machikoyasuda all the _other_ Docker stuff works still, right?

Like `docker compose up -d client` and running the devcontainer and docs site locally?

machikoyasuda:
@thekaveman I did a full reinstall of everything just to be sure - and it all works: devcontainer, app with full login.gov flow, tests, docs

thekaveman (approved):
Ran the app and devcontainer locally 👍

angela-tran (approved):

PR #1769, opened 2023-11-01
Chore: clean up devcontainer

thekaveman:
Cleans up some deprecation notices (no longer just warnings) in the devcontainer.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,17 +12,10 @@
   "files.trimFinalNewlines": true,
   "files.trimTrailingWhitespace": true,
   "[python]": {
-    "editor.defaultFormatter": "ms-python.python"
+    "editor.defaultFormatter": "ms-python.black-formatter"
   },
-  "python.formatting.provider": "black",
   "python.languageServer": "Pylance",
-  "python.linting.enabled": true,
-  "python.linting.flake8Enabled": true,
-  "python.testing.pytestArgs": [
-    "tests/pytest",
-    "--import-mode=importlib",
-    "--no-migrations"
-  ],
+  "python.testing.pytestArgs": ["tests/pytest"],
   "python.testing.pytestEnabled": true,
   "python.testing.unittestEnabled": false,
   "[terraform]": {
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -12,17 +12,10 @@
...
-  "python.formatting.provider": "black",
   "python.languageServer": "Pylance",
-  "python.linting.enabled": true,
-  "python.linting.flake8Enabled": true,
-  "python.testing.pytestArgs": [
-    "tests/pytest",
-    "--import-mode=importlib",
-    "--no-migrations"
-  ],
+  "python.testing.pytestArgs": ["tests/pytest"],
```
angela-tran, line 18: 👀 I will try rebuilding my Benefits devcontainer and see if I can still run tests

angela-tran: I think we need the `--no-migrations` argument. Without it, there was one failing test:

image

With it, all tests pass:
image

I think the test failure was because our data migration created 3 agencies that the test doesn't expect.

thekaveman: Hmm... I'm running this locally using both the VS Code runner and the `tests/pytest/run.sh` script (matching the GitHub Actions) and all tests are passing for me.

They are also passing in GitHub? https://github.com/cal-itp/benefits/actions/runs/6725540014/job/18280144516?pr=1769

thekaveman: Ah, now I was able to trigger it... weird I'm not sure what changed.

angela-tran: I would expect the script and GitHub Actions to pass since `tests/pytest/run.sh` passes in `--no-migrations` and for the VS Code test runner to fail that one test 👍

thekaveman: OK I am tired of dealing with this, fixed the test :laughing:

angela-tran: Yeah, I think I've seen before that VS Code will blow away those args sometimes when updating `settings.json`, so this seems like a reasonable solution

angela-tran (approved):

PR #1770, opened 2023-11-01
Feat: parameterize Django database path/deletion

### LINKED ISSUE #1763: Parameterize database path and deletion
## Background

Currently, the `bin/init.sh` script deletes and then recreates (via Django's `manage.py` helpers) the Sqlite database file every time it is called. This happens both during local development and in the Azure app service -- `bin/init.sh` is called during the [app image's startup](https://github.com/cal-itp/benefits/blob/dev/bin/start.sh#L6).

The [Django app settings](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L153) put the Sqlite database file in the same directory as the source code within the running app (or dev) container. In the Azure app service, this is an ephemeral directory that is lost on app service restarts / redeploys, so it is "good" that the database file is recreated each app service restart.

However, our "data" is encoded in a large [data migration file](https://github.com/cal-itp/benefits/blob/dev/benefits/core/migrations/0002_data.py) that can read from the environment. We want to move away from configuring the app as code, and the first step is being able to maintain our database file between app service restarts.

## Acceptance Criteria

- [x] database is deleted if environment variable e.g. `DATABASE_RESET` if set and `true`
- [x] database stored in a location determined by an environment variable e.g. `DATABASE_DIR`; default to the current directory to maintain existing behavior if unset
- [x] local devcontainer workflow is unaffected / still works OOTB
- [x] `DATABASE_DIR` set to `/home/site/wwwroot` for Azure app service; https://github.com/cal-itp/benefits/issues/1673#issuecomment-1779933175
- [x] `DATABASE_RESET` set to `true` for Azure app service _for now to maintain existing behavior while we migrate to an Admin interface_

## Additional context

- #1312
- #1673

thekaveman:
Closes #1763 

## Testing / review

In your local `.env` file, try different values for the environment variables.

_Note the `DJANGO_DB_DIR` must already exist!_ 

E.g. to create the database file in a local (to your repo) directory called `db`:

```bash
# make sure the directory exists!
mkdir db
```

Then set the environment variable and reload your devcontainer:

```env
DJANGO_DB_DIR=./db
```

Verify `bin/init.sh` output:

```console
$ DJANGO_DB_DIR=./db bin/init.sh 
+ [[ true = true ]]
+ DB_DIR=./db
+ DB_FILE=./db/django.db
+ rm -f ./db/django.db
+ python manage.py migrate
Operations to perform:
  Apply all migrations: core, sessions
Running migrations:
  Applying core.0001_initial... OK
  Applying core.0002_data... OK
  Applying sessions.0001_initial... OK
```

Set `DJANGO_DB_RESET` to `false` to keep the database between restarts/reinits:

```env
DJANGO_DB_RESET=false
```

Verify `bin/init.sh` output:

```console
$ DJANGO_DB_RESET=false bin/init.sh 
+ [[ false = true ]]
+ python manage.py migrate
Operations to perform:
  Apply all migrations: core, sessions
Running migrations:
  No migrations to apply.
```

## Before merge

- [x] Create `django-db-dir` secret in Key Vault (x3), set to `/home/site/wwwroot`
- [x] Create `django-db-reset` secret in Key Vault (x3), set to `true`

## After merge

You shouldn't have to make any changes to your local setup, unless you want to per the above settings.

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -2,6 +2,7 @@
 Django settings for benefits project.
 """
 import os
+
 from benefits import sentry
 
 
@@ -147,10 +148,11 @@ TEMPLATES = [
 
 WSGI_APPLICATION = "benefits.wsgi.application"
 
+DATABASE_DIR = os.environ.get("DJANGO_DB_DIR", BASE_DIR)
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.sqlite3",
-        "NAME": "django.db",
+        "NAME": os.path.join(DATABASE_DIR, "django.db"),
     }
 }
 
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -1,15 +1,16 @@
 #!/usr/bin/env bash
 set -eux
 
-# construct the path to the database file from environment or default
-
-DB_DIR="${DJANGO_DB_DIR:-.}"
-DB_FILE="${DB_DIR}/django.db"
-
 # remove existing (old) database file
-# -f forces the delete (and avoids an error when the file doesn't exist)
 
-rm -f "${DB_FILE}"
+if [[ ${DJANGO_DB_RESET:-true} = true ]]; then
+    # construct the path to the database file from environment or default
+    DB_DIR="${DJANGO_DB_DIR:-.}"
+    DB_FILE="${DB_DIR}/django.db"
+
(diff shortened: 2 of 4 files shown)


angela-tran (approved):
These changes look good to me

PR #1775, opened 2023-11-02
Feat: create and mount Azure file share

### LINKED ISSUE #1773: Create and mount custom persistent Azure file share
This is a follow up to #1763

Now we can customize the location that Django creates and looks for its Sqlite database file 👍 But we can't write to a persistent storage location, because our app service container overwrites the Azure-provided persistent file system each restart. 

But we can create and mount a custom persistent file share in Azure, and maybe that works as a way to keep the Sqlite database.

## Acceptance Criteria

- [ ] Terraform creates and mounts an Azure File Share into app container directory
- [ ] App config `django-db-dir` points to mounted share

## Additional context

Azure docs: [Mount Azure Storage as a local share in App Service](https://learn.microsoft.com/en-us/azure/app-service/configure-connect-to-azure-storage?tabs=basic%2Cportal&pivots=container-linux)

thekaveman:
In the spirit of trying something; part of #1773

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -10,6 +10,10 @@ resource "azurerm_service_plan" "main" {
   }
 }
 
+locals {
+  data_mount = "/home/calitp/app/data"
+}
+
 resource "azurerm_linux_web_app" "main" {
   name                      = "AS-CDT-PUB-VIP-CALITP-${local.env_letter}-001"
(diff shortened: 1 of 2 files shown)


angela-tran (approved):

machikoyasuda (approved):
🤞

PR #1778, opened 2023-11-08
Feat: Unattended superuser creation

### LINKED ISSUE #1776: Allow for unattended superuser creation
Our [`bin/init.sh`](https://github.com/cal-itp/benefits/blob/dev/bin/init.sh) script will create a superuser if the app is configured to have the admin interface running.

However this requires a human to provide information like username and password on the command-line at the time of superuser creation.

[`django-admin` supports environment variables and passing a `--no-input` flag during superuser creation](https://docs.djangoproject.com/en/4.2/ref/django-admin/#createsuperuser), for scripts and other automated processes.

## Acceptance Criteria

- [x] `bin/init.sh` passes the `--no-input` flag to the call to `python manage.py createsuperuser`
- [x] Terraform references Key vault secrets to define the environment variables for the app service config
- [x] Required environment variables are documented under https://docs.calitp.org/benefits/configuration/environment-variables/#django (use a warning call out to indicate that each is required when `DJANGO_ADMIN` is true)
- [x] Key vault (x3) is configured with new secrets and value (randomly generate the password)

thekaveman:
Closes #1776 

## Reviewing 

In your `.env` file, add the new environment variables:

```env
DJANGO_SUPERUSER_EMAIL=you@example.com
DJANGO_SUPERUSER_PASSWORD=password
DJANGO_SUPERUSER_USERNAME=username

# also make sure DJANGO_ADMIN is true to enable the admin:
DJANGO_ADMIN=true
```

Rebuild and reopen the devcontainer.

Confirm in the init output that the admin/user migrations are run and superuser is created.

Launch the app with `F5`, and visit `localhost:PORT/admin` in your browser. Confirm you can login with the username/password above.

### DIFF
--- a/bin/init.sh
+++ b/bin/init.sh
@@ -20,7 +20,7 @@ python manage.py migrate
 # check DJANGO_ADMIN = true, default to false if empty or unset
 
 if [[ ${DJANGO_ADMIN:-false} = true ]]; then
-    python manage.py createsuperuser
+    python manage.py createsuperuser --no-input
 else
     echo "superuser: Django not configured for Admin access"
 fi
--- a/docs/configuration/environment-variables.md
+++ b/docs/configuration/environment-variables.md
@@ -120,6 +120,42 @@ By default the application sends logs to `stdout`.
 
 Django's primary secret, keep this safe!
 
+### `DJANGO_SUPERUSER_EMAIL`
+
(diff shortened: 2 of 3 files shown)


machikoyasuda (approved):
I'm in!

image

image

Issue #1762, opened 2023-10-26
Write small use-case for "enrollment retry" button


thekaveman:
Let's use #1755 as an example of writing up a small technical/feature use-case (different from a large "user flow" use-case like for Veterans).

thekaveman:
Introducing more technical / behavior-level use case documents was @angela-tran's suggestion. Giving Angela the first crack at this issue.

PR #1800, opened 2023-11-30
Docs: add extension to make nested lists work

angela-tran:
Useful for #1762 

This PR adds the `mdx_truly_sane_lists` extension to our Mkdocs configuration, which makes it so indenting a list item with 2 spaces in Markdown will make a nested list.

Without the extension, Mkdocs's default is to expect 4 spaces; this behavior is due to [Mkdocs's usage of Python-Markdown](https://github.com/mkdocs/mkdocs/issues/545). Prettier, our configured formatter, re-formats 4-space indents into 2-space indents.

So in our current docs site, we have lists like:

image

image

With this extension added, they are nested:

image

image

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,4 @@
+mdx_truly_sane_lists
 mkdocs
 mkdocs-awesome-pages-plugin
 mkdocs-macros-plugin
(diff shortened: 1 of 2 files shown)


machikoyasuda:
"Truly Sane Lists" 😆

thekaveman (approved):
Stop the insanity!!!!!

PR #1804, opened 2023-12-01
Fix: Re-enrollment

### LINKED ISSUE #1803: Adjust handling of enrollment when customer already exists in group
Our payment processor uses a `500` error response to indicate that a customer is already enrolled in a group.

This is not an actual problem for Benefits, so we have code that handles the response and returns a success to the user.

The error handling is brittle, relying to some degree on the returned error message. This message has now changed, thus our handling is broken.

## To Reproduce

1. Go to test
2. Verify as a senior with SacRT
3. Enroll a card
4. Do 2-3 again with the same card --> BOOM

## Expected behavior

Attempting to re-enroll with the same card should not cause an error for the end user.

## Additional context

New error message from the logs looks like:

```log
Customer xxxx already exists in group yyyy
```

The old one looked more like:

```
Duplicate customer xxxxx...
```

thekaveman:
Fixes #1803

### DIFF
--- a/benefits/enrollment/api.py
+++ b/benefits/enrollment/api.py
@@ -65,7 +65,7 @@ class CustomerResponse:
 class GroupResponse:
     """Benefits Enrollment Customer Group API response."""
 
-    def __init__(self, response, requested_id, payload=None):
+    def __init__(self, response, requested_id, group_id, payload=None):
         if payload is None:
             try:
                 payload = response.json()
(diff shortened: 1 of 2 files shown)


benefits/enrollment/api.py
```diff
@@ -74,18 +74,12 @@ def __init__(self, response, requested_id, payload=None):
         else:
             try:
                 # Group API uses an error response (500) to indicate that the customer already exists in the group (!!!)
-                # The error message should contain the customer ID we sent via payload and start with "Duplicate"
+                # The error message should contain the customer ID and group ID we sent via payload
```
angela-tran, line 77: 👍

angela-tran (approved):
This looks good to me

PR #1828, opened 2023-12-15
Docs: Various small fixes

machikoyasuda:
- Add Use Cases back to the sidebar: Renames the page to `Enrollment use cases` so it fits better in the hierarchy
image

- Fixes Alternate flows bullet bug:
image

- Updates redirect to go to the proper URL, removes College redirect (page is now deleted)

### DIFF
--- a/docs/product-and-design/.pages
+++ b/docs/product-and-design/.pages
@@ -1,5 +1,6 @@
 nav:
 
+- use-cases
 - analytics.md
 - copy-delivery.md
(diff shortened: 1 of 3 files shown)


angela-tran (approved):
These changes seem fine to me. I'm not sure if we'll want to segregate the use cases from each other, but I'm ok with leaving it as it is for now. I imagine there will be more rearranging of the docs site as we flesh out more of the documentation that has been introduced.

PR #1857, opened 2024-01-19
Skip to main content link

### LINKED ISSUE #1852: Fix color contrast in Skip Nav
Note: Skip Nav that is currently on https://test-benefits.calitp.org/ does not have a sufficient color contrast.

When designing links, keep in mind that links can be either their original color, or, it can be the `visited` color once the user has clicked it once already - which is purple `#562b97`. So the contrast has to be sufficient for both. In this case, the purple is sufficient, but the original blue is not.

Here's the full error message from an accessibility test tool. They recommend changing the blue from `#045b86` to a darker `#005681`.

>  • Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 4.2:1. Recommendation:  change text colour to #005681.
>    ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail
>    ├── #skip-to-content > div
>    └── <div class="container">Skip to Main Content</div>

cc @indexing @segacy1

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/issues/1652#issuecomment-1686887324_

machikoyasuda:
closes #1852 

## What this PR does
- Restyles Skip to main content link focus ring
- Refactors focus ring CSS to use variables, so both rings can use same variable

A11y testing this branch versus test:
image

image

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -44,7 +44,7 @@
     {% endif %}
     <header role="banner" id="header">
       <a id="skip-to-content"  href="#main-content" class="d-block w-100">
-        <div class="container">{% translate "Skip to Main Content" %}</div>
+        <div class="container">{% translate "Skip to main content" %}</div>
       </a>
       {% if messages %}
         {% for message in messages %}
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -44,7 +44,7 @@
     {% endif %}
     <header role="banner" id="header">
       <a id="skip-to-content"  href="#main-content" class="d-block w-100">
-        <div class="container">{% translate "Skip to main content" %}</div>
+        <div class="container"><span>{% translate "Skip to main content" %}</span></div>
       </a>
       {% if messages %}
         {% for message in messages %}
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -301,10 +301,6 @@ main {
   padding: 0 calc(4rem / 16);
 }
 
-.bg-danger {
-  display: none;
-}
-
 /* Footer */
 :root {
   --footer-background-color: var(--dark-color);
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -295,13 +295,13 @@ main {
 }
 
 #skip-to-content:focus span {
-  font-size: var(--font-size-18px);
-  line-height: 150%;
-  letter-spacing: 0.9px;
   border: 2px solid var(--focus-color);
-  border-radius: 4px;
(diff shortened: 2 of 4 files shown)


benefits/core/templates/core/base.html
```diff
@@ -44,7 +44,7 @@
     {% endif %}
     <header role="banner" id="header">
       <a id="skip-to-content"  href="#main-content" class="d-block w-100">
-        <div class="container">{% translate "Skip to Main Content" %}</div>
+        <div class="container"><span>{% translate "Skip to main content" %}</span></div>
```
machikoyasuda, line 47: Added this `span` here for semantic-ness.

benefits/static/css/styles.css
```diff
@@ -284,16 +284,21 @@ main {
...
+  height: auto;
   outline: none !important;
   box-shadow: none;
+  padding: 1rem 0;
+  background: var(--bs-white);
+}
+
+#skip-to-content:focus span {
+  border: 2px solid var(--focus-color);
+  border-radius: calc(4rem / 16);
```
machikoyasuda, line 299: Need to clear up one thing: https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3?node-id=438:1160&mode=dev#673932531

@indexing @segacy1 The standard yellow focus ring style has a border-radius of 2px and a border of 3px. This one is 1px off for both. Is that okay, or should I use the standard yellow focus ring style?

segacy1: Let's stick with what's standard! I updated Figma to match. ty!

benefits/core/templates/core/base.html
thekaveman: Can you elaborate a bit? Both `<div>` and `<span>` carry no semantic meaning on their own, the difference being denoting either block-level or inline content. Text content is allowed inside both of them... 

If there is an accessibility issue here with semantics, wouldn't a `<p>` be more appropriate?

machikoyasuda: Semantic was the wrong word here. Though I'm not sure what the right would be. Having a border (for the focus) around a `p` or any block element (like `p` or `div`) would make it look like this, with the yellow border around the entire div: 
image


But a `span` is an inline-display element, so the same CSS renders the focus like this instead:
image

So it's not semantic-ness but more like....... the native HTML element's inherit default block-or-inline-layout-ness.

benefits/static/css/styles.css
machikoyasuda: image
Now matches the other link focus ring style.

thekaveman (approved):
Looks great on my machine!

image

PR #1874, opened 2024-02-06
Refactor: model secret fields

### LINKED ISSUE #1847: Refactor handling of secret model fields
Implementation plan from #1764

## Implementation plan

- [x] Implement a simple helper script in the spirit of [this one in `data-infra`](https://github.com/cal-itp/data-infra/blob/main/packages/calitp-data-infra/calitp_data_infra/auth.py); use [Azure's python libraries](https://learn.microsoft.com/en-us/python/api/overview/azure/keyvault-secrets-readme?view=azure-python&source=recommendations) to read a value from a Key Vault by secret name. Ensure the Key Vault / environment (e.g. `dev`, `test`, `prod`) is not hardcoded; create a setting from an environment variable or similar. #1859
- [x] [Create a custom Django field type](https://docs.djangoproject.com/en/5.0/howto/custom-model-fields/) that inherits from `models.TextField`. The default field implementation stores the name of the secret. #1874
- [x] Using the [field mapping spreadsheet](https://docs.google.com/spreadsheets/d/1lPp_LJTfJoWCyOkRC4RaljABGG_jPWOgm4gaAHJigMU/edit#gid=176385027), update all secret fields to have this new field type. Update the field names to include a `_secret_name` postfix #1874
- [x] Add new properties to each model that use the helper script above and the `_secret_name` field value to get the actual value of the secret. These properties should be named the same as the original (pre-refactor) secret field (e.g. without the `_secret_name` postfix) #1874

thekaveman:
Defines a new field type `SecretNameField` that manages storage of the _name_ of the secret for models that need it.

Defines a custom [`RegexValidator`](https://docs.djangoproject.com/en/5.0/ref/validators/#regexvalidator) to ensure secret names are valid according to [Azure's rules](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftkeyvault).

Refactors secret fields on models according to the [field mapping spreadsheet](https://docs.google.com/spreadsheets/d/1lPp_LJTfJoWCyOkRC4RaljABGG_jPWOgm4gaAHJigMU/edit#gid=176385027) to store the name of their corresponding secret using the new field type. Secret names are defined in new fields on each model named with the postfix `_secret_name`.

Models then use the helper defined in #1859 to get the value on-demand via a `@property` using the original (sans-postfix) field name. This means calling code doesn't have to change.

Closes #1847

## Models to refactor

- [x] `AuthProvider`
- [x] `EligiblityVerifier`
- [x] `PemData`
- [x] ~`PaymentProcessor`~ Since the secret fields here are references to `PemData`, and that model has already been refactored, this model does not need to be further refactored.

## Reviewing this PR

It should be possible to set up your local environment to read secrets from `dev`:

1. Override `DJANGO_ALLOWED_HOSTS` to include `dev-benefits.calitp.org,localhost`
1. Override other environment variables for _non-secret_ fields, e.g. `AUTH_PROVIDER_AUTHORITY`
1. Run `bin/init.sh` to recreate the database
1. `F5` to launch the app locally
1. Verify the complete enrollment flow for the pathway(s) you set up


angela-tran (approved):
This looks good to me.

Since we'll be connecting to the Key Vault every time we need a secret, I looked up the Key Vault service limits: https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits#secrets-managed-storage-account-keys-and-vault-transactions

I think it should be fine for the scale at which we currently operate.

thekaveman:
> This looks good to me.
> 
> Since we'll be connecting to the Key Vault every time we need a secret, I looked up the Key Vault service limits: https://learn.microsoft.com/en-us/azure/key-vault/general/service-limits#secrets-managed-storage-account-keys-and-vault-transactions
> 
> I think it should be fine for the scale at which we currently operate.

Good point! I hadn't thought of that. 

> 4,000 transactions allowed in 10 seconds, per vault per region

Seems like WAY more than enough for now 😅 

However we could look at adding more of the [`@cached_property`](https://docs.python.org/3/library/functools.html#functools.cached_property) decorators like was done for `PemData` if we're worried about lag time.

machikoyasuda (approved):
- Ran locally sucessfully, going thru reading dev secrets

PR #1889, opened 2024-02-13
Refactor: use cal-itp/littlepay to check API access

### LINKED ISSUE #1814: Refactor Payment Processor API check workflow to use Back Office API
Given what happened and the fix for #1810, it is surprising that the [`check-api`](https://github.com/cal-itp/benefits/actions/workflows/check-api.yml) workflow continues to succeed for QA. From what we understand about the certificates, this should not be the case.

In addition, now that the Prod cert has been rotated, we would expect that check to be failing as the cert has not been rotated in GitHub secrets.

This task is to entirely refactor these checks to use the Back Office API via our [`littlepay`](https://github.com/cal-itp/littlepay) package.

## Acceptance criteria

- [x] The QA check uses the Back Office API via `littlepay`
- [x] The Prod check uses the Back Office API via `littlepay`
- [ ] All cert secrets are removed from this repository

## Additional context

Depends on cal-itp/littlepay#10

angela-tran:
Closes #1814 

This PR refactors the `check-api` workflow to use [cal-itp/littlepay](https://github.com/cal-itp/littlepay) instead. 

Like the previous implementation, this approach checks if it can get an access token for the QA and prod environments. Additionally it does those checks for each participant, as defined in the job matrix.

### Before merging
- [x] Copy contents of `config.yaml` for this API from LastPass, and add as a repository secret named `API_CHECK_CONFIG`

### Secrets cleanup to do
- [ ] Delete the old secrets used in the previous implementation

### DIFF
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -2,81 +2,39 @@ name: Check access to API
 
 on:
   workflow_dispatch:
-    inputs:
-      environment:
-        type: choice
-        description: Select the API environment
-        options: [all, prod, qa]
   schedule:
     - cron: "0 12 * * *"
 
 jobs:
   check-api:
     runs-on: ubuntu-latest
-    env:
-      SHOULD_RUN: |
-        ${{ github.event_name == 'schedule'
-        || github.event.inputs.environment == 'all'
-        || github.event.inputs.environment == matrix.name
-        }}
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - name: prod
-            cert: API_CHECK_PROD_CERT
-            key: API_CHECK_PROD_KEY
-            ca-cert: API_CHECK_PROD_CA_CERT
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
Wow, this is really clever to check out the Littlepay library inside the workflow and install from there!

Seeing that, I wonder if it might make more sense to have this check run in that repo instead?

I really like the new `matrix` configuration too :+1:

angela-tran:
> Seeing that, I wonder if it might make more sense to have this check run in that repo instead?

I could see arguments for either repo. The workflow does send a Slack notification to the Benefits channel if there's a failure, so that might be a reason to keep it here. On the other hand, `cal-itp/littlepay` pretty much exists to be used in Benefits, so you could say that it's just as valid for it to send notifications to the Benefits channel. 

I'll go ahead and merge it in to this repo. Let me know if you feel strongly about moving it.

PR #1890, opened 2024-02-13
Secrets: update guidance and parsing for local multiline env vars

thekaveman:
Docker, bash, etc. support multiline environment variables, by simply wrapping the value in single quotes with newlines, e.g. in an .env file:

```env
multi_line_value='first line
second line
third line'
```
    
Resulting in the expected:

```console    
$ echo "$multi_line_value"
first line
second line
third line
```
    
Due to a quirk in VS Code's Python extension, multiline values are not parsed, see https://code.visualstudio.com/docs/python/environments#_environment-variables
    
> ... Multiline values aren't supported ...
    
And more ongoing discussion at microsoft/vscode-python#18307
    
When running locally in e.g. Debug mode, and secrets are read dynamically from the environment, Python loses the multiline value and we end up with:

```console
>> value = os.environ.get("multi_line_value")
>> print(value)

first line
```
    
This changes the samples and docs so literal newlines are added to the value of the environment variable in an .env file:

```env
multi_line_value='first line\nsecond line\nthird line'
```
    
But the initial value read by Python contains _escaped_ newline characters:

```console
first line\\nsecond line\\nthird line
```
    
Hence unescaping in the secrets helper, so that local secrets contain the actual newline character:

```console
first line\nsecond line\nthird line
```

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -2,66 +2,14 @@ testsecret=Hello from the local environment!
 auth_provider_client_id=benefits-oauth-client-id
 courtesy_card_verifier_api_auth_key=server-auth-token
 mobility_pass_verifier_api_auth_key=server-auth-token
-client_private_key='-----BEGIN RSA PRIVATE KEY-----
-MIIEpAIBAAKCAQEA1pt0ZoOuPEVPJJS+5r884zcjZLkZZ2GcPwr79XOLDbOi46on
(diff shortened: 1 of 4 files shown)


thekaveman:
:thinking: Cypress tests are passing locally for me using the exact same steps as the workflow... ??

angela-tran (approved):
This looks good to me 👍

PR #1891, opened 2024-02-14
Update README.md removed directive from Mermaid chart

indexing:
I used custom styling for the roadmap timeline diagram. mkdocs is not displaying it correctly when published. I've removed the code that constitutes the directive to see if it mkdocs renders the diagram as expected.


thekaveman (approved):
Still looks good on the preview site:

image

Hopefully this works for the deployed version! :pray:

PR #1892, opened 2024-02-15
Update low-income.md changed ordered list markdown to all 1s

indexing:
The ordered list under basic flow isn't presenting sequence correctly.  Instead of adding an order to elements in the list, I've changed all numbering to one so the elements are ordered when they are presented.

Screenshot 2024-02-14 at 3 36 55 PM


thekaveman (approved):
:+1:

PR #1905, opened 2024-02-23
Refactor: enrollment with Backoffice API

angela-tran:
This PR refactors the enrollment phase to use the [cal-itp/littlepay](https://github.com/cal-itp/littlepay) library rather than having its own implementation against an older API.

From the user's perspective, everything should work exactly the same.

For testing this, you'll need to configure the new fields on `PaymentProcessor`: `client_id`, `audience`, and `client_secret_name`. The values for the first two can be found in LastPass, and for the secret, you'll need to set the name of the secret and then set the actual value in your `.env` file.

### Earlier notes

> Pushing this up for visibility, but please be aware that this branch will definitely be rebased since several things are WIP:
>  - [x] need to use `SecretNameField` for `client_secret` field
>  - [x] need to figure out how to properly assert that `oauth.ensure_active_token` is being called - having a hard time figuring out how to set up the mocked Client to work how I'd expect
>  - [x] need to delete old `enrollment.api` module and tests
>  - [x] need to update view tests for new implementation of enrollment
>  - [x] need to add back any valid test cases from deleted `enrollment.api` tests to be covered by view tests
>  - [x] need to generate migration for changes to `PaymentProcessor` model


angela-tran:
@thekaveman This is ready for review. I rebased and cleaned up the commits as much as I could. I think technically I could combine the two migration files into one, but that would require combining some changes that are currently in separate commits into one commit that might be a bit monstrous.

benefits/core/migrations/0002_paymentprocessor_audience_paymentprocessor_client_id_and_more.py
```diff
@@ -0,0 +1,35 @@
...
+    dependencies = [
+        ("core", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="paymentprocessor",
+            name="audience",
+            field=models.TextField(default="audience"),
+            preserve_default=False,
```
angela-tran, line 19: When generating this migration file, Django forced me to provide a default value for `audience` and `client_id` since the fields do not allow null.

We can use the admin interface to update the values.

thekaveman:
> I think technically I could combine the two migration files into one, but that would require combining some changes that are currently in separate commits into one commit that might be a bit monstrous.

I think how you have done it as a separate migration makes sense! We're now in the world of needing to migrate model changes, so I like it :+1:

thekaveman:
@angela-tran this is so exciting! Thank you for your hard work on this refactor!! :clap: 

I think we should wait to merge this so that it is _not_ going out to prod at the same time as all our secrets / Admin changes scheduled for Friday 3/1 (that seems like a lot to release all at once / harder to diagnose any issues that come up).

Let's plan for another release in the first week or two of March that includes this refactor.

angela-tran:
I rebased one more time so that I could rename `client_secret` to `client_secret_name` 😄

thekaveman:
Looking at this now

benefits/core/migrations/0003_remove_paymentprocessor_api_access_token_endpoint_and_more.py
```diff
@@ -0,0 +1,49 @@
+# Generated by Django 5.0.2 on 2024-02-28 23:18
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("core", "0002_paymentprocessor_audience_paymentprocessor_client_id_and_more"),
```
thekaveman, line 9: Since these are the first migrations we'll apply since moving to the Admin / stable DB, let's go ahead and combine these two (can be in a new commit to avoid rebase insanity). 

Just to make things clearer from a migrations perspective, since all of these changes are related to the Back Office API refactor, and really can all be made at once.

benefits/core/migrations/local_fixtures.json
```diff
@@ -324,18 +324,12 @@
...
-      "api_access_token_endpoint": "access-token",
-      "api_access_token_request_key": "request_access",
-      "api_access_token_request_val": "REQUEST_ACCESS",
+      "client_id": "",
+      "client_secret_name": "",
+      "audience": "",
       "card_tokenize_url": "http://server:8000/static/tokenize.js",
       "card_tokenize_func": "tokenize",
-      "card_tokenize_env": "test",
-      "client_cert": 5,
```
thekaveman, line 333: Let's go ahead and delete the sample fixtures for the (now) unused `PemData` references as well.

benefits/core/models.py
```diff
@@ -206,21 +206,12 @@ class PaymentProcessor(models.Model):
...
     card_tokenize_env = models.TextField()
-    # The certificate used for client certificate authentication to the API
-    client_cert = models.ForeignKey(PemData, related_name="+", on_delete=models.PROTECT)
-    # The private key, used to sign the certificate
-    client_cert_private_key = models.ForeignKey(PemData, related_name="+", on_delete=models.PROTECT)
-    # The root CA bundle, used to verify the server.
-    client_cert_root_ca = models.ForeignKey(PemData, related_name="+", on_delete=models.PROTECT)
-    customer_endpoint = models.TextField()
-    customers_endpoint = models.TextField()
-    group_endpoint = models.TextField()
```
thekaveman, line 215: Model refactor looks good, to store the name of the `client_secret` :+1: 

But we will still need the _value_ referenced by the `client_secret_name` for later passing into the `littlepay.api.Client`.

Suggest creating a `@property` like `client_secret` that uses our helper `get_secret_by_name()` -- see e.g. [`AuthProvider.client_id`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L102) for an example.

benefits/enrollment/views.py
```diff
@@ -37,8 +39,16 @@ def token(request):
     """View handler for the enrollment auth token."""
     if not session.enrollment_token_valid(request):
         agency = session.agency(request)
-        response = api.Client(agency).access_token()
-        session.update(request, enrollment_token=response.access_token, enrollment_token_exp=response.expiry)
+        payment_processor = agency.payment_processor
+        client = Client(
+            base_url=payment_processor.api_base_url,
+            client_id=payment_processor.client_id,
+            client_secret_name=payment_processor.client_secret_name,
```
thekaveman, line 46: Here we'll need to update the parameter passed to the `Client`: it wants the `client_secret` directly, not the _name_ of the secret.

See my review comment on the model for more.

benefits/enrollment/views.py
```diff
@@ -64,13 +75,34 @@ def index(request):
         logger.debug("Read tokenized card")
         card_token = form.cleaned_data.get("card_token")
 
-        response = api.Client(agency).enroll(card_token, eligibility.group_id)
-        if response.success:
+        client = Client(
+            base_url=payment_processor.api_base_url,
+            client_id=payment_processor.client_id,
+            client_secret_name=payment_processor.client_secret_name,
```
thekaveman, line 81: Same here, need to pass the _value_ of the referenced `client_secret`.

thekaveman (changes requested):
Overall this is looking really good :+1: 

I think the majority of the view changes and their tests are great.

Just a few more changes needed in the model / construction of the `littlepay.api.Client` and some minor cleanup of migrations and fixtures.

benefits/core/migrations/0003_remove_paymentprocessor_api_access_token_endpoint_and_more.py
thekaveman: I didn't realize this is what you were referencing with this comment: https://github.com/cal-itp/benefits/pull/1905#issuecomment-1970085455

I thought you were referencing the existing `0001` and a new `0002` migration, sorry. 

Let's combine as above, but a new commit is fine.

benefits/enrollment/views.py
angela-tran: Ah great catch, thank you! I must be getting tired because I just straight up replaced `client_secret` with `client_secret_name` 😅 

Edited the commit to use correct [parameter](https://github.com/cal-itp/benefits/pull/1905/commits/6e238c3bd24e2836770cb44b59c0ff5c8096f592#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R45) and [property](https://github.com/cal-itp/benefits/pull/1905/commits/6e238c3bd24e2836770cb44b59c0ff5c8096f592#diff-4d37ea08e0da7b3a7ea86f5fa29f0a86a74acb8089d2c9bd881173c430fad818R228-R230)

benefits/enrollment/views.py
angela-tran: Similar to above, I edited the commit for this implementation to use the correct [parameter](https://github.com/cal-itp/benefits/pull/1905/commits/fbc3a48ff532f457e9a67dd7cea5a6442e7599f4#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R81)

benefits/core/models.py
angela-tran: Thanks for catching that 🙏  

I added the [`@property`](https://github.com/cal-itp/benefits/pull/1905/commits/6e238c3bd24e2836770cb44b59c0ff5c8096f592#diff-4d37ea08e0da7b3a7ea86f5fa29f0a86a74acb8089d2c9bd881173c430fad818R228-R230) in [6e238c3](https://github.com/cal-itp/benefits/pull/1905/commits/6e238c3bd24e2836770cb44b59c0ff5c8096f592)

benefits/core/migrations/local_fixtures.json
angela-tran: Done in [ed26642](https://github.com/cal-itp/benefits/pull/1905/commits/ed266425587aee8d315eca23ca20d2961d7bbbc3)

benefits/enrollment/views.py
angela-tran: Just a remark, I find it a little concerning that our unit tests were not able to alert us that the `Client` instantiation here thrown a `TypeError: Client.__init__() got an unexpected keyword argument 'client_secret_name'`. I guess it's just the reality of needing to mock the `client` instance, so I'm not sure if there's something we could do better.

angela-tran:
I was able to make the requested changes and also tested the enrollment phase with those new code changes. Thanks for all the feedback @thekaveman. Let me know if you see anything else

image

benefits/enrollment/views.py
thekaveman: > Just a remark, I find it a little concerning that our unit tests were not able to alert us...

I agree... would changing how the mock is constructed, using the [`spec` parameter in `patch()`](https://docs.python.org/3/library/unittest.mock.html#patch) help here?

> The spec and spec_set keyword arguments are passed to the [MagicMock](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.MagicMock) if patch is creating one for you.
> 
> In addition you can pass spec=True or spec_set=True, which causes patch to pass in the object being mocked as the spec/spec_set object.

Read more about what the `spec` parameter does in the [docs for the `Mock` initializer](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock)

> spec: This can be either a list of strings or an existing object (a class or instance) that acts as the specification for the mock object. If you pass in an object then a list of strings is formed by calling dir on the object (excluding unsupported magic attributes and methods). Accessing any attribute not in this list will raise an [AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError).

E.g.

```python
mock_client_cls = mocker.patch("benefits.enrollment.views.Client", spec=True)
```

thekaveman (changes requested):
Sorry, one more requested change as I'm trying to get this running locally...

Can you please:

- [ ] remove existing sample payment processor secrets in `.env.sample` for the old API e.g. `mst_payment_processor_client_cert`
- [ ] add new sample payment processor secrets in `.env.sample` for the new API e.g. `mst_payment_processor_client_secret`
- [ ] update the sample fixture to reference these new sample secrets

angela-tran:
> Sorry, one more requested change as I'm trying to get this running locally...

For sure. I forgot about that `.env.sample` file. Done in [3bfb9e8](https://github.com/cal-itp/benefits/pull/1905/commits/3bfb9e86aee4ee8c0ab674083c2275aa82fc445c)

benefits/enrollment/views.py
angela-tran: This seems like an interesting path to explore. I tried out running the test with `mock_client_cls = mocker.patch("benefits.enrollment.views.Client", spec=True)`, and it gave me `AttributeError: Mock object has no attribute 'oauth'`, which I think has to do with the fact that it's patching the class.

I'll have to think / look more at this

thekaveman: Yeah definitely not something that has to be done for this PR

benefits/core/models.py
```diff
@@ -206,21 +206,16 @@ class PaymentProcessor(models.Model):
     id = models.AutoField(primary_key=True)
     name = models.TextField()
     api_base_url = models.TextField()
```
thekaveman, line 208: I think we are OK to proceed as-is, but I wanted to raise that the `littlepay.api.Client` also has a `version` parameter. 

This defaults to `v1` so it should work. However in the future we may need to add a field to this model if we need to change the version. Maybe worth doing now? Your call.

angela-tran: I hadn't thought much about that `version` parameter, so I'm glad you brought it up.

I'm thinking it would be better to add the field whenever we actually need it. If we added it now, it'd be just another field to set that we always set to the same thing so it'd be kind of annoying and another place for error.

thekaveman (approved):
Fantastic!! :tada:

I got this to work locally and tested the following flows:

* :ballot_box_with_check:  New enrollment in a group -- received the expected success response
* :ballot_box_with_check: Duplicate enrollment in a group -- received the expected `409` response, got the success page

We'll handle error flows in #1936

thekaveman:
## Post-merge

- [ ] Update the secrets in `dev`
  - [ ] Add the new `client_secret` secrets
  - [ ] Delete the old cert-based secrets
- [ ] Update the secrets in `test`
  - [ ] Add the new `client_secret` secrets
- [ ] Update the secrets in `prod`
  - [ ] Add the new `client_secret` secrets

We'll wait to delete the old secrets in `test` and `prod` until after we've deployed to those environments :sweat_smile:

thekaveman:
IMHO the best part of this PR :grin: :grin: :+1: 

image

angela-tran:
Awesome! I'm working on merging / post-merge actions now

angela-tran:
I finished the post-merge steps for `dev`. I would like to leave configuring the `test` and `prod` secrets for when we are doing the deployments for those.

PR #1910, opened 2024-02-23
Fix: check for existing preview comment before adding one

### LINKED ISSUE #1896: Fix docs preview comment workflow to only comment once
It is sort of annoying to get a new docs preview comment in a PR every single time a commit is pushed that changes the docs. Like, we get it, the preview is there!

image

## Acceptance Criteria

- [ ] When docs are updated in a PR, no matter how many commits are pushed until the PR is merged, we only get a single comment with the preview URL

thekaveman:
Closes #1896


thekaveman:
The second comment here was due to an error in the workflow file, I fixed it and force pushed, and we didn't get a 3rd! So it seems like it's working.

PR #1914, opened 2024-02-28
Fix: sample env and reset_db helper

thekaveman:
* wraps the sample secret value in quotes in `.env.sample` because bash can choke on the spaces
* fix an apparent syntax error in `bin/reset_db.sh` that was causing evaluation of the fixtures file to fail


angela-tran (approved):
This looks good to me

PR #1955, opened 2024-03-14
Feat: EligibilityType model fields for expiration

### LINKED ISSUE #1938: Enhance EligibilityType model with expiration-related field(s)
With the coming Low-income / CalFresh enrollment flow, we'll need a way to display certain information and apply logic in-code for discounts that expire, e.g.

* Does the discount / eligibility type expire at all?
* How long after enrollment does the discount expire?
* How long before expiration can a user re-enroll?

For CalFresh, this will be used in a number of places, e.g.

* In #1877, we'll calculate and send an expiration date with the group enrollment
* In #1919, we'll display the date of expiration
* In #1921, we'll display the date of expiration and the date at which re-enrollment can occur

We need a relatively simple way to make these decisions without (ideally) having to hardcode this logic into the templates / views, and to make this more modular for future, non-CalFresh flows that expire (e.g. students, enhancements to Agency cards).

## Acceptance criteria

The following fields are added to the `EligibilityType` model:

- [x] `supports_expiration`: a [`BooleanField`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#booleanfield) indicating if the `EligibiltyType` supports expiration, default to `False`
- [x] `expiration_days`: a [`PositiveSmallIntegerField`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#positivesmallintegerfield) indicating how many days after enrollment the associated discount expires, allow `null` but requiring a value `> 0` when `supports_expiration == True`; see Django docs on [Cleaning and validating fields that depend on each other](https://docs.djangoproject.com/en/5.0/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other) and [Validating model instances](https://docs.djangoproject.com/en/5.0/ref/models/instances/#validating-objects)
- [x] `expiration_reenrollment_days`: a [`PositiveSmallIntegerField`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#positivesmallintegerfield) indicating how many days _before_ expiration a user is allowed to re-enroll for the discount, allow `null` but requiring a value `> 0` when `supports_expiration == True`; see Django docs on [Cleaning and validating fields that depend on each other](https://docs.djangoproject.com/en/5.0/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other) and [Validating model instances](https://docs.djangoproject.com/en/5.0/ref/models/instances/#validating-objects)

angela-tran:
Closes #1938 

This PR adds fields to the `EligibilityType` model that will be needed to support expiration of benefits.

Since these fields can be edited in the admin interface, this PR also configures what [`form`](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.form) should be used so that we can perform [custom validation](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#admin-custom-validation) as specified on #1938.


benefits/core/admin.py
```diff
@@ -26,6 +27,33 @@
     admin.site.register(model)
 
 
+class EligibilityTypeForm(forms.ModelForm):
```
thekaveman, line 30: I was hoping this could be done without a custom admin form? 

I think implementing [`clean()`](https://docs.djangoproject.com/en/5.0/ref/models/instances/#django.db.models.Model.clean) in the model class should be enough?

angela-tran: Oh I see now that it could be done that way too and that we don't need the custom admin form. I'll put this back in draft and move the logic to be done on the model

angela-tran:
Superseded by #1957

Issue #1918, opened 2024-02-28
Create new CalFresh Cardholder informational modal


thekaveman:
![image](https://github.com/cal-itp/benefits/assets/1783439/b21ca1fc-269e-41d0-ac89-25ffd46b14ae)

## Figma links

* [CalFresh information modal (Desktop)](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=13761-6517&mode=design&t=7Gb5yK9MsYWUi5pt-4)
* [CalFresh information modal (Mobile )](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=13820-3376&mode=design&t=7Gb5yK9MsYWUi5pt-4)

PR #1958, opened 2024-03-15
Feat: add CalFresh option

### LINKED ISSUE #1915: `Eligibility index`: Add CalFresh Cardholder option
The order of items is important and based on the number of potential users:

1. Older Adult
2. CalFresh
3. US Veteran
4. Agency card

image

## Figma links

* [`Eligibility index` page with new CalFresh option (Desktop)](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=13339-13172&mode=design&t=7Gb5yK9MsYWUi5pt-4)
* [`Eligibility index` page with new CalFresh option (Mobile)](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=13362-16759&mode=design&t=7Gb5yK9MsYWUi5pt-4)

## Acceptance Criteria

- [ ] CalFresh option is between Older Adult and US Veteran
- [ ] New CalFresh Cardholder modal linked in radio button list, from #1918

angela-tran:
Closes #1915 and #1918 

This PR adds the radio button option for "CalFresh Cardholders" to the eligibility index. The description contains a link that opens a help modal about CalFresh Cardholder eligibility.

#### Steps for setting up to test
**_From existing database_**
```
./bin/init.sh
```
will run migrations, including a migration that sets the initial display order for your existing verifiers.

_**From scratch**_
```
./bin/reset_db.sh
./bin/init.sh
```
will recreate your database, run migrations, and load fixtures which define the values for `display_order` (since fixtures are loaded after migrations).

#### Steps for testing
Start your local app, and log into your local admin interface to edit the order of your `Eligibility verifiers` by drag-and-drop.

The radio buttons on the eligibility index should be in the order you specified.


angela-tran:
### Unexpected behavior with the ordering of benefit options

In dc62f355a, I added the CalFresh eligibility type and verifier as new fixtures and set their `id`s by incrementing the greatest `id` in their respective tables. This mimics how it would work if one were to use the admin interface to add CalFresh to the Benefits app database since the `id` is an [`AutoField`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#autofield).

When setting the references on a transit agency, I added the CalFresh eligibility type and verifier to be [second in the list](https://github.com/cal-itp/benefits/pull/1958/commits/dc62f355abe0bd3fe93089c6434c829af65b56a0#diff-cea7b783aff074158a2a778fb934f1b9d3e846d89346aaa9b9f7e5ab30169c37R331-R332) assuming / hoping that somehow Django would preserve the declaration order. We want CalFresh to be the second item when the app retrieves the verifiers for the eligibility selection form.

However, launching the app locally, I see that is not the case:
image

The eligibility selection form queries for the options using [`verifiers = agency.eligibility_verifiers.filter(active=True)`](https://github.com/cal-itp/benefits/blob/0ce39d2be27c212ca1d6659294faae84a402e39e/benefits/eligibility/forms.py#L28), and the result seems to be ordered by `id`. We can use [`order_by`](https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.order_by) to specify an order, but we are limited to ordering by the model's fields, i.e. the declaration order is no longer available or stored anywhere.

This makes sense because ultimately what gets put in the database are just rows in the junction tables `core_transit_agency_eligibility_types` and `core_transit_agency_eligibility_verifiers` that capture the relationship but not any ordering.
image

The admin interface makes it even more obvious that there's no "declaration order"; you can multi-select, but the order is clearly fixed.
image

I propose we add a `display_order` field to `EligibilityVerifier` (`PositiveSmallIntegerField` I guess?) but am also curious to hear any other ideas.

angela-tran:
@thekaveman and I discussed https://github.com/cal-itp/benefits/pull/1958#issuecomment-2000294887 and agreed that the solution must include a good user experience for editing the order.

This led to finding [jrief/django-admin-sortable2](https://github.com/jrief/django-admin-sortable2), which allows for maintaining ordering on whatever models you choose and for editing that order using drag-and-drop in the Django admin site.

We are following the approach described by these sections:
 - [Prepare the model classes](https://django-admin-sortable2.readthedocs.io/en/latest/usage.html#prepare-the-model-classes)
 - [Make the ListView sortable](https://django-admin-sortable2.readthedocs.io/en/latest/usage.html#in-django-s-admin-make-the-list-view-sortable)

### Alternatives considered

The approach described by [Make stacked and tabular inlines sortable](https://django-admin-sortable2.readthedocs.io/en/latest/usage.html#in-django-s-admin-detail-view-make-stacked-and-tabular-inlines-sortable) is for a one-to-many relationship and therefore not applicable to our TransitAgency and EligibilityVerifier models.

The approach described by [Sortable Many-to-Many Relations with Sortable Inlines](https://django-admin-sortable2.readthedocs.io/en/latest/usage.html#sortable-many-to-many-relations-with-sortable-inlines) is applicable, but we decided against it because it was getting complex: we'd need to use a custom `through` model to add `display_order` to the TransitAgency<>EligibilityVerifier association, write a data migration to [avoid losing any existing TransitAgency<>EligibilityVerifier associations](https://stackoverflow.com/a/51351406), and add a way to retrieve `display_order` from the `through` model when getting the `EligibilityVerifier`s since the [ordering doesn't propagate all the way through](https://stackoverflow.com/questions/1390556/default-ordering-for-m2m-items-by-intermediate-model-field-in-django?rq=4) to the original models.

angela-tran:
I don't think there are any unit tests to add here since all we're doing is adding a `PositiveSmallIntegerField` to a model and letting `django-admin-sortable2` do the rest of the logic.

benefits/core/migrations/local_fixtures.json
```diff
@@ -237,6 +252,27 @@
       "form_class": "benefits.eligibility.forms.SBMTDMobilityPass"
     }
   },
+  {
+    "model": "core.eligibilityverifier",
+    "pk": 7,
+    "fields": {
+      "name": "CalFresh oauth claims via Login.gov",
```
thekaveman, line 259: Nit, but can we stick with the pattern of prefixing the agency slug e.g. `(MST)`

benefits/eligibility/templates/eligibility/includes/selection-label--calfresh.html
```diff
@@ -0,0 +1,18 @@
+{% extends "eligibility/includes/selection-label.html" %}
+{% load i18n %}
+
+{% block label %}
+    {% translate "CalFresh Cardholder" %}
+{% endblock label %}
+
+{% block description %}
+    {% translate "You must have" %}
+    {% comment %} modal-trigger needs to be refactored to support translating the text {% endcomment %}
```
thekaveman, line 10: Interesting :thinking: 

Would you want to do that in this PR, since this is the first time it is being introduced? We don't have an issue for this otherwise.

thekaveman (changes requested):

benefits/eligibility/templates/eligibility/includes/selection-label--calfresh.html
angela-tran: I realized we don't actually need any refactor 😅 We can pass in the translated value, which is what I've done in [1a52320](https://github.com/cal-itp/benefits/pull/1958/commits/1a523202565c234688e1696dee87b37fdab34da6)

Django also probably wouldn't have been able to detect a variablized `{% translate text %}` in `modal-trigger`, so it's better to do it directly in the specific template rather than in the includes.


_From https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#standard-translation_
> (The caveat with using variables or computed values, as in the previous two examples, is that Django’s translation-string-detecting utility, [django-admin makemessages](https://docs.djangoproject.com/en/5.0/ref/django-admin/#django-admin-makemessages), won’t be able to find these strings. More on [makemessages](https://docs.djangoproject.com/en/5.0/ref/django-admin/#django-admin-makemessages) later.)

angela-tran:
Ready for re-review

benefits/core/migrations/local_fixtures.json
thekaveman: Just want to get this one in @angela-tran and then this PR looks G2G!

angela-tran: Forgot about this, thank you!

[5a58a8b](https://github.com/cal-itp/benefits/pull/1958/commits/5a58a8be12f3a3ec530d7d72d501bb686b927b04)

thekaveman (approved):
Looks great, I love the graphical re-ordering capability within the Admin!

PR #1973, opened 2024-03-21
Feat: include expiration date during enrollment

### LINKED ISSUE #1877: Include expiry_date with all CalFresh / Low income enrollments
Depends on https://github.com/cal-itp/littlepay/issues/17 and #1848 

## Acceptance Criteria

- [ ] Expiry date is 365 days from the day of enrollment
- [ ] Expiration should occur on midnight of the 366th day
- [ ] Expiry date is [ISO 8601 formatted datetime](https://en.wikipedia.org/wiki/ISO_8601) in UTC i.e. `YYYY-MM-DDTHH:mm:ssZ`

## Additional context

Example:

* User enrolls on March 12, 2024 at 14:30 Pacific
* Enrollment time in UTC corresponds to March 12, 2024 at 21:30
* Expiration is Wednesday, March 12, 2025 (+365 days)
* `expiry_date` should be set to Thursday, March 13, 2025 at 00:00 Pacific (midnight on the 366th day)
* `expiry_date` should be set to Thursday, March 13, 2025 at 07:00 UTC

angela-tran:
Closes #1877 

This is very much a WIP, but pushing up for visibility / feedback.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -102,9 +102,9 @@ def index(request):
                     return success(request)
                 # else if expiration date has passed, then re-enroll
                 elif expiration.is_expired(matching_group_funding_source):
-                    client.link_concession_group_funding_source(funding_source_id=funding_source.id, group_id=group_id)
-                    analytics.returned_success(request, group_id)
-                    return success(request)
+                    return _link_concession_group_funding_source(
+                        request, eligibility, client, matching_group_funding_source, group_id
+                    )
                 # else expiration date hasn't passed, so calculate re-enrollment date and show user
                 else:
                     reenrollment_date = expiration.calculate_reenrollment_date(  # noqa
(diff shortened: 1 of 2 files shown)


benefits/enrollment/views.py
```diff
@@ -84,25 +84,56 @@ def index(request):
...
+            else:
+                # if eligibility_type does not supports_expiration, then just link it
+                if not eligibility.supports_expiration:
+                    client.link_concession_group_funding_source(funding_source_id=funding_source.id, group_id=group_id)
+                # else eligibility_type supports_expiration, so calculate expiration date from today and include in request
+                else:
+                    expiry_date = expiration.calculate_expiry_date(matching_group_funding_source, eligibility.expiration_days)
+                    client.link_concession_group_funding_source(
+                        funding_source_id=funding_source.id, group_id=group_id, expiry_date=expiry_date
+                    )
```
angela-tran, line 126: @thekaveman I put all this business logic right here in the view function 1) so I could get it written down for starters, and 2) so that we could see what might make sense to extract out as helper functions.

I actually think this looks ok. It's kind of long, but in terms of level of abstraction, it seems appropriate to have the API calls here and delegate out the date calculation to a separate module.

Thoughts? Also are the comments helpful or just noisy?

thekaveman:
New version of `littlepay` published to PyPI with the concession expiry endpoints: https://github.com/cal-itp/littlepay/releases/tag/2024.03.2

benefits/enrollment/views.py
angela-tran: By "all this business logic" I mean lines [90-120](https://github.com/cal-itp/benefits/pull/1973/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R90-R120)

angela-tran:
@thekaveman and I discussed https://github.com/cal-itp/benefits/pull/1973#discussion_r1534780169 yesterday and came up with the full list of possible combinations to handle. We met with @indexing today to define the behavior for cases that are unlikely but still possible.

Closing this PR to start a new one that reflects what we decided.

thekaveman:
> @thekaveman and I discussed [#1973 (comment)](https://github.com/cal-itp/benefits/pull/1973#discussion_r1534780169) yesterday and came up with the full list of possible combinations to handle. We met with @indexing today to define the behavior for cases that are unlikely but still possible.
> 
> Closing this PR to start a new one that reflects what we decided.

If you think it makes sense, I think it would be OK to have the helper functions previously in the `expiration` module directly in the `views`, now that we won't need all of them.

PR #1974, opened 2024-03-22
Fix: path to coverage report

angela-tran:
Fixing a 404 I noticed on our docs site:

The `mkdocs.yml` GitHub workflow publishes the coverage report at [`docs/tests/coverage`](https://github.com/cal-itp/benefits/blob/2024.03.3/.github/workflows/mkdocs.yml#L100).

The link to the coverage report was on a page that got moved under `docs/tests/automated-tests` in https://github.com/cal-itp/benefits/commit/a5c63c54228209d94124f0725468ef79820da1f3, and we forgot to update the link.

### DIFF
--- a/docs/tests/automated-tests.md
+++ b/docs/tests/automated-tests.md
@@ -77,4 +77,4 @@ The report files include a local `.gitignore` file, so the entire directory is h
 
 ### Latest coverage report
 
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
Good catch. It's annoying we can't test this in the preview site but meh

PR #1976, opened 2024-03-22
Fix: allow custom media-item--heading tag

### LINKED ISSUE #1975: Enrollment index skips H2 heading level
While going through the app's typography, I noticed that there is 1 that does not have proper h1, h2, h3 ordering. Enrollment Index goes from H1 (the heading) to H3 (the Media Item title). On other page that uses Media Item, Eligibility Start, there's a H2 in between the heading and the Media Items, so that one's okay with Media Item titles being H3s.

This is somewhat related to this issue, because it's related to how we're sharing code through the Media Items include across pages, but it might lead to a situation like this where we have to change one small thing to get valid HTML.

image

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/issues/1528#issuecomment-1664710287_

thekaveman:
Fixes #1975 

Previously `h3` was hard-coded as the heading tag for media items.

This caused a bug in `enrollment:index`, where heading levels skip from `h1` (the page title) to `h3` (the first media item).

In `eligibility:start` (the other place where media items are used), there is an intervening `h2` that keeps the order correct.

This change allows a child media item to override the heading tag, while maintaining the `h3` styling for all `.media-item--heading`, regardless of tag used.

The default tag remains `h3` if a child media item does not choose to override.

Found this answer on SO which showed that, with an extra wrapping `{% block %}`, context from the child can be passed up to the parent using `{% with %}`: https://stackoverflow.com/a/46581444 TIL!

### Eligibility start -- correct heading levels

image

### Enrollment index -- `h2` is skipped

image

### Eligibility start -- correct heading levels

image

### Enrollment index -- `.media-item--heading` is overridden to `h2`, styling remains the same

image

### DIFF
--- a/benefits/core/templates/core/includes/media-item.html
+++ b/benefits/core/templates/core/includes/media-item.html
@@ -4,10 +4,12 @@
         {% endblock icon %}
     </div>
     <div class="media-body">
-        <h3 class="media-body--heading">
+        {% block heading_wrapper %}
+            <{{ heading_tag|default:"h3" }} class="media-body--heading h3">
             {% block heading %}
             {% endblock heading %}
-        </h3>
+            </{{ heading_tag|default:"h3" }}>
+        {% endblock heading_wrapper %}
         {% block body %}
         {% endblock body %}
     </div>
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
Nice!! 🔥

PR #1979, opened 2024-03-22
Fix(models): allow blank

### LINKED ISSUE #1977: Cannot save `EligibilityVerifier` without providing an `AuthProvider`
_(Prompted by post-merge steps for #1970)_

Intended behavior is that an `EligibilityVerifier` can exist without an `AuthProvider`. However, currently it is not possible to save one in that state from the admin interface.

The reason why we have existing `EligibilityVerifier`s with no `auth_provider` is because they were created externally rather than through the admin interface (see #1856 for notes on that process).

## To Reproduce

1. Go to `/admin`
2. Log in
3. Click on "Eligibility verifiers"
4. Choose an eligibility verifier that does not have an auth provider
5. Go into the form for editing it
6. Update some field (e.g. `Name`)
7. Click "Save"
8. Notice error saying the `Auth provider` field is required

## Expected behavior

The eligibility verifier is saved with the update you made.

## Screenshots
<details><summary>Expand to see</summary>

image
</details>

## Additional context

* Current field definition: [`auth_provider = models.ForeignKey(AuthProvider, on_delete=models.PROTECT, null=True)`](https://github.com/cal-itp/benefits/blob/2bcf0a37bb1da4105810975cb4a120649907e99d/benefits/core/models.py#L175)
* Django docs:
   * [`Field.blank`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#blank)
   * [`Field.null`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#null)
   * [`ForeignKey.on_delete`](https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.ForeignKey.on_delete)

Whether as a part of this issue or a separate one, we should go through and make sure all our model fields work as intended from the admin interface.

angela-tran:
Closes #1977 

This PR sets `blank` to `True` for all nullable fields on our models. This is so form validation on those fields will allow blank values.


thekaveman (changes requested):
I get that these are only changes that impact the Admin forms, but I still think a migration is required. 

1. Check out `dev` and run `bin/reset_db.sh` (now DB looks like it does on `dev`)
2. Check out this branch and run `bin/init.sh` (which is what the appcontainer will run in `dev`)
3. See output:

```log
Operations to perform:
  Apply all migrations: admin, auth, contenttypes, core, django_google_sso, sessions
Running migrations:
  No migrations to apply.
  Your models in app(s): 'core' have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
```

angela-tran:
> I get that these are only changes that impact the Admin forms, but I still think a migration is required.

Ah, thanks for catching this

thekaveman (changes requested):
This isn't totally related, but I feel like since you're updating models for editing in the Admin...

It seems like we forgot to list `AuthProvider` in the models that are registered for editing in the Admin: https://github.com/cal-itp/benefits/blob/dev/benefits/core/admin.py#L19

Do you think you can add that here so we can just be done with it?

thekaveman (approved):
:+1:

PR #1984, opened 2024-03-26
Added Mermaid code to replace image of product roadmap

indexing:
I kept the "directive" at the top of the chart to override default styling, but I'm not sure if mkdocs will display the chart properly with a directive. Please confirm display works locally if possible.

I also removed the link to the legacy product roadmap maintained in a Google doc as we are no longer using it.


thekaveman (approved):
This looks good, both in the preview site:

image


and locally:

image

thekaveman (changes requested):
Actually, sorry -- can you please also delete the image file `img/roadmap.png` so we don't get confused when it becomes out of date?

thekaveman (approved):

PR #1985, opened 2024-03-26
Feat: store/retrieve enrollment expiry in session

### LINKED ISSUE #1981: Session stores enrollment expiration, context processor makes available to templates
As part of #1877, we need to store the calculated expiration date in the user's session and make it available for use later on e.g. the Enrollment success page (#1919) and Enrollment error page (#1921).

## Acceptance Criteria

- [x] `benefits.core.session` defines a helper for getting the user's current expiration date as a `datetime` instance
- [x] `benefits.core.session.update` takes a `datetime` parameter for updating the user's current expiration date
- [x] A new context processor adds the `expiration_date` and `renenrollment_date` to the context for templates to use

thekaveman:
Closes #1981

New `session` helpers:

* `enrollment_expiry` gets the current user's expiration as a datetime, or None
* `enrollment_reenrollment` gets the current user's reenrollment as a datetime, or None
* `update` accepts `enrollment_expiry` as a datetime

New `context_processor`:

* Adds `expiration`, `reenrollment`, and `supports_expiration` under the `enrollment` key 

```
{ 
  "enrollment": {
    "expiration": datetime | None,
    "reenrollment": datetime | None, 
    "supports_expiration": bool
  }
}
```

Also removed most of the noisy logging from `session`.

### DIFF
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -69,6 +69,16 @@ def debug(request):
     return {"debug": session.context_dict(request)}
 
 
+def enrollment(request):
+    """Context processor adds enrollment information to request context."""
+    data = {
+        "expires": session.enrollment_expiry(request),
+        "supports_expiration": session.eligibility(request).supports_expiration if session.eligible(request) else None,
+    }
+
+    return {"enrollment": data}
+
+
 def origin(request):
     """Context processor adds session.origin to request context."""
     origin = session.origin(request)
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -71,9 +71,14 @@ def debug(request):
 
 def enrollment(request):
     """Context processor adds enrollment information to request context."""
+    eligibility = session.eligibility(request)
+    expiry = session.enrollment_expiry(request)
+    reenrollment = session.enrollment_reenrollment(request)
+
     data = {
-        "expires": session.enrollment_expiry(request),
-        "supports_expiration": session.eligibility(request).supports_expiration if session.eligible(request) else None,
+        "expires": expiry,
+        "reenrollment": reenrollment,
+        "supports_expiration": eligibility.supports_expiration if eligibility else False,
     }
 
     return {"enrollment": data}
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -2,6 +2,7 @@
 The core application: helpers to work with request sessions.
 """
 
+from datetime import datetime, timezone
 import hashlib
 import logging
 import time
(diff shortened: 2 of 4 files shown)


angela-tran (approved):
This looks great! I thought through the implications of having the session calculate the reenrollment date every time (as opposed to having the view set it on the session), and that seems fine to me.

angela-tran:
> I thought through the implications of having the session calculate the reenrollment date every time (as opposed to having the view set it on the session), and that seems fine to me.

I think it does mean that there's no need for any [`expiration.calculate_reenrollment_date`](https://github.com/cal-itp/benefits/pull/1973/files#diff-247184188cb343ecd9081acf1f50fd855df584114c353ac8aea227cefe3b6e1fR11) function

thekaveman:
> > I thought through the implications of having the session calculate the reenrollment date every time (as opposed to having the view set it on the session), and that seems fine to me.
> 
> I think it does mean that there's no need for any [`expiration.calculate_reenrollment_date`](https://github.com/cal-itp/benefits/pull/1973/files#diff-247184188cb343ecd9081acf1f50fd855df584114c353ac8aea227cefe3b6e1fR11) function

Hmm good point. 

I think for the first thought, it seems OK for the reenrollment date to be recalculated each time (regardless of where) since it is a fixed offset from the known/stored expiration date.

For the second thought (where to do this calculation) -- I'm open. My thinking was it made sense to put it here since the session will be the "source of truth" for the expiration date for the rest of the app, so it could also be the "source of truth" for the reenrollment date. It does sort of abuse the session and put business logic there, rather than just storing data... so I'm open to moving this out.

angela-tran:
> My thinking was it made sense to put it here since the session will be the "source of truth" for the expiration date for the rest of the app, so it could also be the "source of truth" for the reenrollment date.

Yeah, I think ultimately we need the reenrollment date in the context processor and therefore need it on the session. As for if the the calculation is done directly inside the `session` module or is in a separate module that is called by `session`, I don't think it makes a big difference in this PR. 

Maybe in my PR I can move the calculation into `benefits.enrollment.expiration`?

thekaveman:
> > My thinking was it made sense to put it here since the session will be the "source of truth" for the expiration date for the rest of the app, so it could also be the "source of truth" for the reenrollment date.
> 
> Yeah, I think ultimately we need the reenrollment date in the context processor and therefore need it on the session. As for if the the calculation is done directly inside the `session` module or is in a separate module that is called by `session`, I don't think it makes a big difference in this PR.
> 
> Maybe in my PR I can move the calculation into `benefits.enrollment.expiration`?

Hmmm but the problem is we can't/don't want `benefits.core.session` to depend on/import `benefits.enrollment.expiration`, the imports typically go the other way (and we might run into a circular import issue if we try to go this way).

angela-tran:
> Hmmm but the problem is we can't/don't want `benefits.core.session` to depend on/import `benefits.enrollment.expiration`, the imports typically go the other way (and we might run into a circular import issue if we try to go this way).

Oh I see. I'm leaning now towards just keeping it in `benefits.core.session`. The business logic is not complicated, and the value _is_ derived from something that was set on the session.

PR #1986, opened 2024-03-26
Delete roadmap image

indexing:
Removing this image, as we now have it rendered using Mermaid code in the docs.


thekaveman:
@indexing can you please make this change as part of your other PR #1984? They should all go at the same time.

PR #1987, opened 2024-03-26
Fix: enrollment expiry session

thekaveman:
Quick follow-up to #1985 

* Forgot to activate the new context processor in `settings.py`
* Forgot to clear the enrollment expiry in `session.reset()`


angela-tran (approved):
Ah, thanks!

PR #1988, opened 2024-03-26
Feat: CalFresh enrollment success with expiration

### LINKED ISSUE #1919: `Enrollment success`: Expiration date text; calculate and format date properly

thekaveman:
Closes #1919 

## How to test

1. Temporarily change the [`context_processors.enrollment()`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/context_processors.py#L75) to force a datetime for `expires`:

```diff
+ from datetime import datetime, timezone
...
- expiry = session.enrollment_expiry(request)
+ expiry = session.enrollment_expiry(request) or datetime.now(tz=timezone.utc)
```

2. In the Admin, update a flow that you can pass eligibility for to support expiration, e.g. MST Courtesy Cards
3. In the Admin, update the corresponding agency's PaymentProcessor to connect to the QA environment
4. Pass eligibility
5. Enroll a test credit card
6. See the new success page

## Screenshot - English

image

## Screenshot - Spanish

image

### DIFF
--- a/benefits/enrollment/templates/enrollment/success.html
+++ b/benefits/enrollment/templates/enrollment/success.html
@@ -21,10 +21,17 @@
   <div class="col-12 col-sm-12 col-lg-9">
     <div class="row flex-column-reverse flex-lg-row">
       <div class="col-12 col-lg-7">
-        <p class="pt-lg-4 mt-lg-3">
-          {% block success-message %}
-          {% endblock success-message %}
-        </p>
+        {# djlint:off #}
+        {% if enrollment.supports_expiration %}
+          <h2 class="h3 mb-1">{% translate "Your benefit will expire on" %} {{ enrollment.expires|date }}.</h2>
+          <p>
+        {% else %}
+          <p class="pt-lg-4 mt-lg-3">
+        {% endif %}
+            {% block success-message %}
+            {% endblock success-message %}
+          </p>
+        {# djlint:on #}
         <p class="pt-4">{% translate "Thank you for using Cal-ITP Benefits!" %}</p>
       </div>
       <div class="col-12 col-lg-5">
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -204,7 +204,11 @@ LOCALE_PATHS = [os.path.join(BASE_DIR, "benefits", "locale")]
 
 USE_I18N = True
 
-TIME_ZONE = "UTC"
+# See https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-TIME_ZONE
+# > Note that this isn’t necessarily the time zone of the server.
+# > When USE_TZ is True, this is the default time zone that Django will use to display datetimes in templates
+# > and to interpret datetimes entered in forms.
+TIME_ZONE = "America/Los_Angeles"
 USE_TZ = True
 
 # Static files (CSS, JavaScript, Images)


thekaveman:
I think we can actually merge this in, as it won't be activated for flows until we get back the expiration logic for enrollment (after #2052). And it doesn't have any conflicts with those changes, it just changed the `enrollment/success.html` template.

angela-tran (commented):
To test this locally, I: 
 - merged `dev` into my local `feat/calfresh-enrollment-success` branch so that I have the revert from #2052
 - edited my Courtesy Card eligibility type to support expiration
 - ran through the full Courtesy Card flow

I then see the success screen:
image

The expiration date doesn't show because currently there isn't any code that sets the `enrollment_expiry` session variable, and that's ok because this screen won't be accessible until after #2052 is reverted.

## Hard-coding in an `enrollment_expiry` to test date formatting
Just to test the date-formatting, I hard-coded a value in for `expiry` in the `session.enrollment_expiry` function:
```python
    # expiry = request.session.get(_ENROLLMENT_EXP)
    expiry = datetime(2024, 12, 1).timestamp()
```

English:

image

Spanish:
image

@thekaveman Do you know how the date formatting gets set? I'm not sure why I'm getting an abbreviated month in English

angela-tran:
With a March date, I see the full month name:

image

thekaveman:
> @thekaveman Do you know how the date formatting gets set? I'm not sure why I'm getting an abbreviated month in English

We're using the [`date` filter](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#date), which uses the [`DATE_FORMAT`](https://docs.djangoproject.com/en/5.0/ref/settings/#date-format) setting by default, which is:

```python
DATE_FORMAT = "N j, Y"
```

Broken down as ([from the table here](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#date)):

* `N`: _Month abbreviation in Associated Press style. Proprietary extension._
* `j`: _Day of the month without leading zeros._
* `Y`: _Year, 4 digits with leading zeros._

It looks like for `N`, some months are fully spelled out and others are abbreviated. I guess we should change the default and instead use one of:

* `E`: _Month, locale specific alternative representation usually used for long date representation._
* `F`: _Month, textual, long._

I think each language has its own default format string, so changing this one should only affect English.

thekaveman:
@angela-tran I rebased on `dev` so this branch now has the revert from #2052.

I introduced custom date formats per locale and now we get the expected full-month formatting for both English and Spanish:

image

image

angela-tran (approved):

angela-tran:
Nice!

angela-tran (commented):
The coverage report is flagging that the `formats.py` files aren't tested... should we do something about that?

thekaveman:
> The coverage report is flagging that the `formats.py` files aren't tested... should we do something about that?

Good call on this! I added tests for each in https://github.com/cal-itp/benefits/pull/1988/commits/9cdacdcf8e9768ec6e8e92895c57417195ad6be4

angela-tran (approved):
🚀

angela-tran:
Just noticed the infra pipeline failure... looking into it

angela-tran:
Re-ran it and it worked 😂

PR #1989, opened 2024-03-27
Feat: enrollments can expire

### LINKED ISSUE #1877: Include expiry_date with all CalFresh / Low income enrollments

angela-tran:
Closes #1877

This PR refactors the enrollment phase so that we no longer catch a 409 error to detect existing enrollments. Instead we see if the enrollment already exists for the group. This also allows us to inspect the expiration date of any existing enrollment.

So this PR also introduces support for enrollments that expire. See the table below for expected behavior.

# Combinations of states and their expected outcomes
_copied directly from [our dev meeting notes](https://docs.google.com/document/d/1vActZALSbRwO0HxDvm-YQK37p22SPbL4XuMdcRx-BI8/edit#heading=h.1qwtpkwx040s)_

<details><summary>Expand to view table</summary>

| Eligibility state           | Enrollment state | Expiration state                | Expected enrollment outcome              | Expected end page for user       |
| --------------------------- | ---------------- | ------------------------------- | ---------------------------------------- | -------------------------------- |
| Supports expiration         | Not enrolled     | No expiration                   | User can enroll, expiration date set     | Enrollment success               |
| Supports expiration         | Not enrolled     | Future expiration               | This state should not be possible        | N/A                              |
| Supports expiration         | Not enrolled     | Re-enrollment window expiration | This state should not be possible        | N/A                              |
| Supports expiration         | Not enrolled     | Past expiration                 | This state should not be possible        | N/A                              |
| Supports expiration         | Enrolled         | No expiration                   | Update expiration of existing enrollment | Enrollment success               |
| Supports expiration         | Enrolled         | Future expiration               | Re-enrollment error                      | Enrollment error with expiration |
| Supports expiration         | Enrolled         | Re-enrollment window expiration | Update expiration of existing enrollment | Enrollment success               |
| Supports expiration         | Enrolled         | Past expiration                 | Update expiration of existing enrollment | Enrollment success               |
| Does not support expiration | Not enrolled     | No expiration                   | User can enroll, no expiration date set  | Enrollment success               |
| Does not support expiration | Not enrolled     | Future expiration               | This state should not be possible        | N/A                              |
| Does not support expiration | Not enrolled     | Re-enrollment window expiration | This state should not be possible        | N/A                              |
| Does not support expiration | Not enrolled     | Past expiration                 | This state should not be possible        | N/A                              |
| Does not support expiration | Enrolled         | No expiration                   | No action                                | Enrollment success               |
| Does not support expiration | Enrolled         | Future expiration               | Remove expiration date                   | Enrollment success               |
| Does not support expiration | Enrolled         | Re-enrollment window expiration | Remove expiration date                   | Enrollment success               |
| Does not support expiration | Enrolled         | Past expiration                 | Remove expiration date                   | Enrollment success               |

</details>

# Some functions to be familiar with

In Python 3.9+, the standard library includes [`zoneinfo`](https://docs.python.org/3/library/zoneinfo.html) for working with time zones.

Django has a [`django.utils.timezone`](https://docs.djangoproject.com/en/5.0/ref/utils/#module-django.utils.timezone) module that exposes convenience methods for working with the built-in Python `datetime` module and incorporates its own [`USES_TZ`](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-USE_TZ) and [`TIMEZONE`](https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-TIME_ZONE) settings. Because of that, I did not need to use `zoneinfo` (or `pytz` which was a popular option before `zoneinfo` was a thing).

I thought it might be helpful to write out brief summaries of how the functions below are useful in this PR.

### [`django.utils.timezone.now()`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.now)
This function creates a standard Python `datetime.datetime` object representing the current point in time. 

More importantly, if `USE_TZ` is `True`, `timezone.now()` will return an aware `datetime` with UTC as its timezone. 

Benefits has `USE_TZ` set to `True`, so we can rely on this as an easy way to make an aware "now" `datetime` , for example in `_is_expired` and `_is_within_reenrollment_date`.

### [`django.utils.timezone.localtime(value=None, timezone=None)`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.localtime)
This function lets you give it an aware `datetime` and converts it to the `timezone` that you pass in.

If you don't give it a `value`, it will use `timezone.now()` for `value`.

This was useful for calculating the expiration date since we want to make sure the expiration time is midnight Pacific Time.

### [`django.utils.timezone.get_default_timezone()`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.get_default_timezone)
This function returns what you set the `TIME_ZONE` setting to, which Django refers to as the ["default time zone"](https://docs.djangoproject.com/en/5.0/topics/i18n/timezones/#default-time-zone-and-current-time-zone).

In this PR, we set Benefits's default timezone to `America/Los_Angeles`. This means we can call `timezone.get_default_timezone` and give it to `timezone.localtime` to create an aware `datetime` that is in Pacific Time.

_Aside_: "current time zone" is initially the same as the default time zone. If you call [`timezone.activate`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.activate) with some other time zone, that's when they'd be different. The way I think about this is that it's for cases like, for example, if we gave the user a drop-down of timezones and set their selection as the "current time zone" for their requests to the server. With that understanding, we can see that we don't want to use "current time zone" to calculate the expiration date.

### [`django.utils.timezone.make_aware(value, timezone=None)`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.make_aware)

This function lets you give it a naive `datetime` and a timezone to make an aware `datetime`. This is useful when you need a specific datetime, which I needed for testing.  If there is a more succinct way to do this, I'm open to it.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -89,14 +89,47 @@ def index(request):
             )
 
             already_enrolled = group_funding_source is not None
+            has_no_expiration_date = already_enrolled and group_funding_source.concession_expiry is None
+            has_expiration_date = already_enrolled and group_funding_source.concession_expiry is not None
 
-            if not already_enrolled:
-                # enroll user with no expiration date, return success
-                client.link_concession_group_funding_source(group_id=group_id, funding_source_id=funding_source.id)
-                return _success(request, group_id)
-            else:
-                # no action, return success
-                return _success(request, group_id)
+            if eligibility.supports_expiration:
+                # set expiry on session
+                if has_expiration_date:
+                    session.update(request, enrollment_expiry=group_funding_source.concession_expiry)
+                else:
+                    session.update(request, enrollment_expiry=_calculate_expiry(eligibility.expiration_days))
+
+                if not already_enrolled:
+                    # enroll user with an expiration date, return success
+                    pass
+                else:
+                    if has_no_expiration_date:
+                        # update expiration of existing enrollment, return success
+                        pass
+                    else:
+                        is_expired = _is_expired(group_funding_source.concession_expiry)
+                        is_within_reenrollment_window = _is_within_reenrollment_window(
+                            group_funding_source.concession_expiry, session.enrollment_reenrollment(request)
+                        )
+
+                        if is_expired or is_within_reenrollment_window:
+                            # update expiration of existing enrollment, return success
+                            pass
+                        else:
+                            # re-enrollment error, return enrollment error with expiration and reenrollment_date
+                            pass
+            else:  # eligibility does not support expiration
+                if not already_enrolled:
+                    # enroll user with no expiration date, return success
+                    client.link_concession_group_funding_source(group_id=group_id, funding_source_id=funding_source.id)
+                    return _success(request, group_id)
+                else:
+                    if has_no_expiration_date:
+                        # no action, return success
+                        return _success(request, group_id)
+                    else:
+                        # remove expiration date, return success
+                        pass
 
         except HTTPError as e:
             analytics.returned_error(request, str(e))
@@ -140,6 +173,21 @@ def _get_group_funding_source(client: Client, group_id, funding_source_id):
             break
 
     return matching_group_funding_source
+
+
+def _is_expired(concession_expiry):
+    """Returns whether the passed in datetime is expired or not."""
+    pass
+
+
+def _is_within_reenrollment_window(concession_expiry, enrollment_reenrollment_date):
+    """Returns whether the passed in datetime is within the reenrollment window."""
+    pass
+
+
+def _calculate_expiry(expiration_days):
+    """Returns the expiry datetime."""
+    pass
 
 
 @decorator_from_middleware(EligibleSessionRequired)
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -103,7 +103,10 @@ def index(request):
 
                 if not already_enrolled:
                     # enroll user with an expiration date, return success
-                    pass
+                    client.link_concession_group_funding_source(
+                        group_id=group_id, funding_source_id=funding_source.id, expiry_date=session.enrollment_expiry(request)
+                    )
+                    return _success(request, group_id)
                 else:
                     if has_no_expiration_date:
                         # update expiration of existing enrollment, return success
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -104,7 +104,7 @@ def index(request):
                     client.link_concession_group_funding_source(
                         group_id=group_id, funding_source_id=funding_source.id, expiry_date=session.enrollment_expiry(request)
                     )
-                    return _success(request, group_id)
+                    return success(request)
                 else:  # already_enrolled
                     if group_funding_source.concession_expiry is None:
                         # update expiration of existing enrollment, return success
@@ -113,7 +113,7 @@ def index(request):
                             funding_source_id=funding_source.id,
                             expiry_date=session.enrollment_expiry(request),
                         )
-                        return _success(request, group_id)
+                        return success(request)
                     else:
                         is_expired = _is_expired(group_funding_source.concession_expiry)
                         is_within_reenrollment_window = _is_within_reenrollment_window(
@@ -127,7 +127,7 @@ def index(request):
                                 funding_source_id=funding_source.id,
                                 expiry_date=session.enrollment_expiry(request),
                             )
-                            return _success(request, group_id)
+                            return success(request)
                         else:
                             # re-enrollment error, return enrollment error with expiration and reenrollment_date
                             return reenrollment_error(request)
@@ -135,11 +135,11 @@ def index(request):
                 if not already_enrolled:
                     # enroll user with no expiration date, return success
                     client.link_concession_group_funding_source(group_id=group_id, funding_source_id=funding_source.id)
-                    return _success(request, group_id)
+                    return success(request)
                 else:  # already_enrolled
                     if group_funding_source.concession_expiry is None:
                         # no action, return success
-                        return _success(request, group_id)
+                        return success(request)
                     else:
                         # remove expiration date, return success
                         raise NotImplementedError("Removing expiration date is currently not supported")
@@ -170,11 +170,6 @@ def index(request):
         logger.debug(f'card_tokenize_url: {context["card_tokenize_url"]}')
 
         return TemplateResponse(request, eligibility.enrollment_index_template, context)
-
-
-def _success(request, group_id):
-    analytics.returned_success(request, group_id)
-    return success(request)
 
 
 def _get_group_funding_source(client: Client, group_id, funding_source_id):
@@ -231,6 +226,7 @@ def retry(request):
 
 
 @pageview_decorator
+@decorator_from_middleware(EligibleSessionRequired)
 @decorator_from_middleware(VerifierSessionRequired)
 def success(request):
     """View handler for the final success page."""
@@ -239,10 +235,12 @@ def success(request):
 
     agency = session.agency(request)
     verifier = session.verifier(request)
+    eligibility = session.eligibility(request)
 
     if session.logged_in(request) and verifier.auth_provider.supports_sign_out:
         # overwrite origin for a logged in user
         # if they click the logout button, they are taken to the new route
         session.update(request, origin=reverse(ROUTE_LOGGED_OUT))
 
+    analytics.returned_success(request, eligibility.group_id)
     return TemplateResponse(request, agency.enrollment_success_template)
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -124,7 +124,12 @@ def index(request):
 
                         if is_expired or is_within_reenrollment_window:
                             # update expiration of existing enrollment, return success
-                            pass
+                            client.update_concession_group_funding_source_expiry(
+                                group_id=group_id,
+                                funding_source_id=funding_source.id,
+                                expiry_date=session.enrollment_expiry(request),
+                            )
+                            return _success(request, group_id)
                         else:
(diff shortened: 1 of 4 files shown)


benefits/settings.py
```diff
@@ -204,7 +204,7 @@ def RUNTIME_ENVIRONMENT():
 
 USE_I18N = True
 
-TIME_ZONE = "UTC"
+TIME_ZONE = "America/Los_Angeles"
```
thekaveman, line 207: I get why this setting is needed in this PR as well, and it makes sense to go here rather than #1988 

But can you please include the comments, both in code and in the commit message, that I included in 598c12f29f3a4ad882ec381d4a7347a13de85093? I think it's important context that isn't captured by the commit in this PR.

thekaveman (changes requested):

tests/pytest/enrollment/test_views.py
```diff
@@ -194,14 +250,358 @@ def test_index_eligible_post_valid_form_success(
     response = client.post(path, card_tokenize_form_data)
 
     mock_client.link_concession_group_funding_source.assert_called_once_with(
-        funding_source_id=mocked_funding_source.id, group_id=model_EligibilityType.group_id
+        funding_source_id=mocked_funding_source.id, group_id=model_EligibilityType.group_id, expiry_date=None
```
thekaveman, line 253: Maybe this test function should be renamed to `test_index_eligible_post_valid_form_success_no_expiry` or something?

benefits/enrollment/views.py
```diff
@@ -125,6 +184,54 @@ def index(request):
         return TemplateResponse(request, TEMPLATE_INDEX, context)
 
 
+def _enroll(client: Client, group_id, funding_source_id, expiry_date=None):
```
thekaveman, line 187: This helper function feels extraneous, it just calls a single method on the `client` and forwards all the args. I think we should unwind this one and go back to calls on the `client` directly, based on noting a few bugs above in its usage.

benefits/enrollment/views.py
```diff
@@ -84,25 +87,81 @@ def index(request):
...
+                        return _success(request, group_id)
+                    else:
+                        is_expired = _is_expired(group_funding_source.concession_expiry)
+                        is_within_reenrollment_window = _is_within_reenrollment_window(
+                            group_funding_source.concession_expiry, session.enrollment_reenrollment(request)
+                        )
+
+                        if is_expired or is_within_reenrollment_window:
+                            # update expiration of existing enrollment, return success
+                            _enroll(
```
thekaveman, line 136: Same as above

We don't want to call the underlying `client.link_concession_group_funding_source`, instead this should call `client.update_concession_group_funding_source_expiry`

benefits/enrollment/views.py
```diff
@@ -84,25 +87,81 @@ def index(request):
...
+                        client=client,
+                        group_id=group_id,
+                        funding_source_id=funding_source_id,
+                        expiry_date=session.enrollment_expiry(request),
+                    )
+                    return _success(request, group_id)
+                else:
+                    if has_no_expiration_date:
+                        # update expiration of existing enrollment, return success
+                        _enroll(
```
thekaveman, line 121: Related to my nit below on the `_enroll` helper:

This would not be the correct call here, right? We don't want to call the underlying `client.link_concession_group_funding_source`, instead this should call `client.update_concession_group_funding_source_expiry`

benefits/enrollment/views.py
```diff
@@ -84,25 +87,81 @@ def index(request):
...
 
         try:
-            client.link_concession_group_funding_source(funding_source_id=funding_source.id, group_id=eligibility.group_id)
+            group_funding_source = _get_group_funding_source(
+                client=client, group_id=group_id, funding_source_id=funding_source_id
+            )
+
+            already_enrolled = group_funding_source is not None
+            has_no_expiration_date = already_enrolled and group_funding_source.concession_expiry is None
+            has_expiration_date = already_enrolled and group_funding_source.concession_expiry is not None
```
thekaveman, line 100: I think these `has_*` bools are a bit confusing in that they combine the notion of being enrolled with having/not having an expiration date.

When looking at their usage below, it becomes less clear that we're actually branching based on 2 pieces of data. I suggest moving the `already_enrolled and` clause down into the usage below.

thekaveman (changes requested):
Good progress on this!

I have a few suggestions for cleaning up the code flow, and noticed a few bugs in the enrollment logic.

benefits/enrollment/views.py
angela-tran: Yes, you are absolutely right. I completely missed that, thank you!

benefits/enrollment/views.py
angela-tran: Yeah, I see what you mean. Will do

tests/pytest/enrollment/test_views.py
angela-tran: Yeah, I thought about that too and am ok with updating the name. I think then we should also rename things like `model_EligibilityType`, `mocked_funding_source`, etc. to say `no_expiry`. Any objection to that?

thekaveman: Sounds good!

benefits/settings.py
angela-tran: Cherry-picked your commit as [`2570489`](https://github.com/cal-itp/benefits/pull/1989/commits/2570489e43dc0a039ea3af974c8f89cda3366b68)

tests/pytest/enrollment/test_views.py
angela-tran: Renamed tests/fixtures in these two commits: https://github.com/cal-itp/benefits/pull/1989/files/15bf0c9bb93fbc0a8e80a905c8422f5efbaaeb98..9b157659ec748908ca70d2d2077c283587bfe7e8

benefits/enrollment/views.py
angela-tran: Done in rebase (dropped the commit)

benefits/enrollment/views.py
angela-tran: New commit that uses `client.update_concession_group_funding_source_expiry`: https://github.com/cal-itp/benefits/pull/1989/commits/58d99a10edf340de6a1207296c8c09be27939415

benefits/enrollment/views.py
angela-tran: Refactored the `if` conditions in https://github.com/cal-itp/benefits/pull/1989/commits/15bf0c9bb93fbc0a8e80a905c8422f5efbaaeb98

thekaveman (changes requested):
Just a minor change but this is looking great.

I can't add a line comment, but I think we can drop a few log statements that aren't very useful (trying to reduce noise and also reduce code in this file!):

[Line 76](https://github.com/cal-itp/benefits/pull/1989/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R76)

```python
logger.debug(f"Session contains an {models.EligibilityType.__name__}")
```

And [Line 78](https://github.com/cal-itp/benefits/pull/1989/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R78)

```python
logger.debug("Read tokenized card")
```

angela-tran:
> I can't add a line comment, but I think we can drop a few log statements that aren't very useful (trying to reduce noise and also reduce code in this file!):

Removed the log statements in [`e65d076`](https://github.com/cal-itp/benefits/pull/1989/commits/e65d0766e1516e1f277fcd725c96d1891d966c82)

benefits/enrollment/views.py
```diff
@@ -125,6 +176,48 @@ def index(request):
         return TemplateResponse(request, TEMPLATE_INDEX, context)
 
 
+def _success(request, group_id):
+    analytics.returned_success(request, group_id)
+    return success(request)
```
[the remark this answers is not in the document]
angela-tran: Great idea - if we're always calling them together, might as well move one inside the other.

I made changes in [`4a11cd4`](https://github.com/cal-itp/benefits/pull/1989/commits/4a11cd48c7488c3118704de960c23313f76e87a6) for this and only now just noticed the part in your comment about having `group_id` as a parameter. Is it ok that we use the `group_id` from the session eligibility, or do you still think a parameter would be better?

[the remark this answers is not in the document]
thekaveman: Oh I like your implementation better, getting it from the session! Nice :+1:

thekaveman (dismissed):
Awesome, nice work :rocket: 

I also really appreciate your PR description with notes and links of all the time zone calculations. Will be a very helpful bit of context to call up later if we need to.

angela-tran:
Rebased this to resolve merge conflicts 😄

thekaveman (approved):

PR #1993, opened 2024-03-29
CI: Optimize CodeQL analysis

### LINKED ISSUE #1740: Look into speeding up CodeQL checks
- [x] If possible, the checks should only run when files of the corresponding language are changed in `benefits/`
- [x] Ensure code in `tests/` is not being analyzed

thekaveman:
Closes #1740 

* Skips analysis for the `tests/` directory
* Conditionally analyzes languages (`javascript`, `python`) based on if corresponding files were changed


angela-tran (approved):
This looks good to me 👍

thekaveman:
Eh let's merge for now and see what happens / if we need to make adjustments, especially for the scheduled run (I'm thinking that it won't run for any languages since nothing will have changed, but we can figure that out later).

PR #2017, opened 2024-04-11
fix(docker): remove platform attribute from server service

lalver1:
Since `ghcr.io/cal-itp/eligibility-server:dev` is now a multi-platform image (built off the multi-platform `ghcr.io/cal-itp/docker-python-web:main` image), the `platform` attribute is no longer needed.


angela-tran (approved):

thekaveman (approved):
:+1:

Issue #2032, opened 2024-04-17
New `System enrollment error` page


angela-tran:
![image](https://github.com/cal-itp/benefits/assets/25497886/7a9ce18e-0440-4262-ba48-29b51eeec1e9)

## Acceptance Criteria

Front-end design implementation:
- [x] Updated headline
- [x] Updated body copy
- [x] Updated CTA button (Login.gov and non-login.gov variants - “Sign out of Login.gov” and “Return home”)
   - [x] Login.gov CTA variant signs them out and takes them back to agency index
   - [x] Non-login.gov CTA variant takes them back to the agency index

Back-end logic
- [x] System enrollment errors should still cause Sentry notifications and Amplitude events to be sent
- [x] 500-level HTTP errors during linking of funding-source and group should show system enrollment error - #2088 
- [x] Errors during the acquisition of card tokenization access token should send user to an error page - Javascript in `enrollment/index.html` will need to know where to send user
   - [x] 500-level HTTP errors show system enrollment error page - #2089
   - [x] 400-level HTTP errors and any other types of errors show server error page - #2117
- [x] `onError` callback in tokenization function needs to redirect
   - [x] to system enrollment error page if 500-level HTTP error - #2120
   - [x] to server error page if 400-level HTTP error - #2120

## Additional context

- Handles the case of our connection to Littlepay being broken e.g. because Littlepay’s system is down or returning errors.

angela-tran:
All subtasks have been completed

PR #2040, opened 2024-04-18
Chore: update enrollment retry page

### LINKED ISSUE #2031: New `User enrollment error` page
image

## Acceptance Criteria

- [x] Headline copy updated
- [x] Body copy updated
- [x] 24px spacing between icon, headline, body, and agency info
- [x] 64px spacing between CTA and the thing above it
- [x] Login.gov and non-login.gov variants (with and without the Sign out button)

## Additional context

 - Handles the case of the user entering the wrong card information (e.g. expiration date, security code)
 - Littlepay responds with an error in Hosted Card Tokenization flow (Javascript)

thekaveman:
Closes #2031 

The `/retry` page was functioning as both our "user" enrollment error page (i.e. #2031) and our "system" enrollment error page (i.e. #2032).

This PR updates the language so the page is more explicitly for "user" enrollment errors.

#2032 will have to make additional changes e.g. by creating a new route/template.

This PR also simplifies the view logic of `/retry`, to remove the unnecessary blocking of `GET` requests to that endpoint (which makes testing easier).

_Current on top of #2039_

## Reviewing

1. Launch the app locally
2. Verify eligibility for any flow
3. One verified, visit `/enrollment/retry`
4. See the updated copy and design

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -212,17 +212,8 @@ def reenrollment_error(request):
 @decorator_from_middleware(EligibleSessionRequired)
 def retry(request):
     """View handler for a recoverable failure condition."""
-    if request.method == "POST":
-        analytics.returned_retry(request)
-        form = forms.CardTokenizeFailForm(request.POST)
-        if form.is_valid():
-            return TemplateResponse(request, TEMPLATE_RETRY)
-        else:
-            analytics.returned_error(request, "Invalid retry submission.")
-            raise Exception("Invalid retry submission.")
-    else:
-        analytics.returned_error(request, "This view method only supports POST.")
-        raise Exception("This view method only supports POST.")
+    analytics.returned_retry(request)
+    return TemplateResponse(request, TEMPLATE_RETRY)
 
 
 @pageview_decorator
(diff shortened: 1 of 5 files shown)


benefits/enrollment/views.py
```diff
@@ -212,17 +212,8 @@ def reenrollment_error(request):
 @decorator_from_middleware(EligibleSessionRequired)
 def retry(request):
     """View handler for a recoverable failure condition."""
-    if request.method == "POST":
-        analytics.returned_retry(request)
-        form = forms.CardTokenizeFailForm(request.POST)
```
angela-tran, line 217: Does this mean we don't need the `forms.CardTokenizeFailForm` class anymore?

angela-tran: Oh wait, maybe we still need it so that the Javascript functions on Eligibility Index can POST back to the server

thekaveman: Yeah that is correct, we still use the form for the Javascript POST back -- it just makes that code a little cleaner/easier.

thekaveman: This change in the view was just to make this view a little more flexible. There isn't really a good reason to block GET requests. A normal app user should only ever make POST requests (via the Javascript / POST back) but this helps us test the view. And if a user does happen to GET this view... well really no harm there, it does require an eligible session and only sends them back to the index.

PR #2056, opened 2024-04-25
Feat: add properties for UTM codes

### LINKED ISSUE #2034: Add custom properties for UTM code information
We are proposing to use UTM codes in query strings for various Benefits outreach campaigns. Read more about UTM codes in this article from Hubspot: https://blog.hubspot.com/marketing/what-are-utm-tracking-codes-ht

We need to implement these properties into the Amplitude analytics tracking to be able to later gauge campaign effectiveness.

The proposal is to add them as `event_properties` and `user_properties` in the `viewed page` event. This is likely the first event that a user following a tracked URL will generate. And having the properties on both the event and the user makes later analysis in Amplitude more flexible.

## Acceptance Criteria

- [ ] The [`viewed page` event](https://github.com/cal-itp/benefits/blob/dev/benefits/core/analytics.py#L90) is modified to capture UTM code information
- [ ] Custom `event_properties` are added for each of the UTM codes:
  - [ ] `utm_campaign`
  - [ ] `utm_source`
  - [ ] `utm_medium`
  - [ ] `utm_content`
  - [ ] `utm_term`
- [ ] Custom `user_properties` are added for each of the UTM codes:
  - [ ] `utm_campaign`
  - [ ] `utm_source`
  - [ ] `utm_medium`
  - [ ] `utm_content`
  - [ ] `utm_term`

## Additional context

UTM code information comes from the query string, accessible via the Django [`request.GET`](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.GET) attribute, an instance of [`QueryDict`](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.QueryDict).

lalver1:
Closes #2034.

Uses [request.GET.get()](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.QueryDict.get) to get the UTM code. If a UTM code is not present in the query string, returns `null` for that UTM code.
## How to test

Using a local instance of `benefits`, send a request that looks something like `http://localhost:port/?utm_campaign=transit`. Verify that the property `utm_campaign` has been added to `event_properties` and `user_properties`.

| Browser | Debug Message |
|---|---|
| image | image |

### DIFF
--- a/tests/pytest/core/test_analytics.py
+++ b/tests/pytest/core/test_analytics.py
@@ -1,7 +1,41 @@
 import pytest
 
 import benefits.core.analytics
-from benefits.core.analytics import Event
+from benefits.core.analytics import Event, ViewedPageEvent
+from django.contrib.sessions.middleware import SessionMiddleware
+from django.middleware.locale import LocaleMiddleware
+
+
+@pytest.fixture
+def app_request_utm(rf):
+    """
+    Fixture creates and initializes a new Django request object similar to a real application request with UTM codes.
+    """
+    # create a request for the path with UTM codes, initialize
+    app_request = rf.get(
+        "/some/arbitrary/path?utm_campaign=campaign&utm_source=source&utm_medium=medium&utm_content=content&utm_id=id"
+    )
+
+    # https://stackoverflow.com/a/55530933/358804
+    middleware = [SessionMiddleware(lambda x: x), LocaleMiddleware(lambda x: x)]
+    for m in middleware:
+        m.process_request(app_request)
+
+    app_request.session.save()
+    benefits.core.analytics.session.reset(app_request)
+
+    return app_request
+
+
+@pytest.fixture
+def utm_code_data():
+    return {
+        "utm_campaign": "campaign",
+        "utm_source": "source",
+        "utm_medium": "medium",
+        "utm_content": "content",
+        "utm_id": "id",
+    }
 
 
 @pytest.mark.django_db
@@ -80,3 +114,21 @@ def test_Event_update_user_properties(app_request):
 
     assert key in event.user_properties
     assert event.user_properties[key] == value
+
+
+@pytest.mark.django_db
+def test_ViewedPageEvent_with_UTM(app_request_utm, utm_code_data):
+    event = ViewedPageEvent(app_request_utm)
+
+    for key, value in utm_code_data.items():
+        assert event.event_properties[key] == value
+        assert event.user_properties[key] == value
+
+
+@pytest.mark.django_db
+def test_ViewedPageEvent_without_UTM(app_request, utm_code_data):
+    event = ViewedPageEvent(app_request)
+
+    for key in utm_code_data:
+        assert event.event_properties[key] is None
+        assert event.user_properties[key] is None
--- a/benefits/core/analytics.py
+++ b/benefits/core/analytics.py
@@ -92,6 +92,18 @@ class ViewedPageEvent(Event):
 
     def __init__(self, request):
         super().__init__(request, "viewed page")
+        # Add UTM codes
+        utm_campaign = request.GET.get("utm_campaign")
+        utm_source = request.GET.get("utm_source")
+        utm_medium = request.GET.get("utm_medium")
+        utm_content = request.GET.get("utm_content")
+        utm_id = request.GET.get("utm_id")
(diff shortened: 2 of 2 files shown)


lalver1:
Looks like `test_Event_update_event_properties()` and `test_Event_update_user_properties()` in `test_analytics.py` already test the feature added in this PR, so new tests may not be required.

thekaveman (changes requested):
> Looks like `test_Event_update_event_properties()` and `test_Event_update_user_properties()` in `test_analytics.py` already test the feature added in this PR, so new tests may not be required.

True, those functions have been tested. But this PR is adding behavior to the `ViewedPageEvent`, and so tests could (should) be written for that.

lalver1:
Thanks, that makes sense. I added 4 new tests to test the new behavior of `ViewedPageEvent`. They are very similar to the `Event` tests but I think that this is ok since the new behavior of `ViewedPageEvent` is similar to the behavior of `Event`.

tests/pytest/core/test_analytics.py
```diff
@@ -80,3 +80,49 @@ def test_Event_update_user_properties(app_request):
...
+        assert utm_code in update_spy.call_args.kwargs
+        assert update_spy.call_args.kwargs.get(utm_code) is None
+
+
+@pytest.mark.django_db
+def test_ViewedPageEvent_update_event_properties(app_request):
+    key, value = "utm_campaign", "Transit"
+    event = ViewedPageEvent(app_request)
+
+    event.update_event_properties(**{key: value})
```
thekaveman, line 114: This is still a test of the `update_event_properties` method -- remember, what you are testing is simply the `ViewedPageEvent` constructor (with the new behavior you added to call `update_event_properties` etc.) We don't want additional calls to client code (e.g. a function on the `ViewedPageEvent`) because that introduces "another thing" that is under test.

The outline of the test could be something like:

```python
# prepare app_request so that is has UTM codes with values in the query parameters
utm_data = {"utm_campaign", "campaign", "utm_source": "source", etc.}
app_request.something = utm_data

# create the event, this is the only thing under test! No other methods from benefits should be called.
event = ViewedPageEvent(app_request)

# now assert that the event has the expected properties:
for key, value in utm_data:
    assert event.event_properties[key] == value
```

This time you have to set up the `app_request` somehow so that is has UTM information that the `ViewedPageEvent` can read (which is what you are later asserting happened).

tests/pytest/core/test_analytics.py
```diff
@@ -80,3 +80,49 @@ def test_Event_update_user_properties(app_request):
...
+@pytest.mark.django_db
+def test_ViewedPageEvent_sets_default_event_properties(app_request, mocker):
+    update_spy = mocker.spy(ViewedPageEvent, "update_event_properties")
+
+    ViewedPageEvent(app_request)
+
+    utm_codes = ["utm_campaign", "utm_source", "utm_medium", "utm_content", "utm_id"]
+    for utm_code in utm_codes:
+        assert utm_code in update_spy.call_args.kwargs
+        assert update_spy.call_args.kwargs.get(utm_code) is None
```
thekaveman, line 94: These asserts are good, in that they check that the `update_*_properties` function was called. I think of this like the "what" of the changes you made to the `ViewedPageEvent` class.

But we still want to check the result, or the "how" of your changes -- i.e. how the changes manifested when using the now updated `ViewedPageEvent` class.

We should see these new keys in the `*_properties` attributes on the event:

```python
event = ViewedPageEvent(app_request)

assert event.event_properties["utm_campaign"] is None
```

This asserts 2 things: 

1. `event.event_properties` has a key `"utm_campaign"` (otherwise a `KeyError` would be raised and the assert would fail)
2. the value of that key is `None`

These are the 2 expected outcomes of the changes you made when constructing a `ViewedPageEvent` with an `app_request` that has no UTM information on it.

thekaveman (changes requested):
Nice work to keep on this @lalver1! There's still some refinement needed for the tests, hopefully my comments make sense. LMK if you want to chat.

thekaveman (approved):
Fantastic work @lalver1! Thanks for sticking with the tests, looks great!

angela-tran (approved):
This looks great!

lalver1:
Thanks @thekaveman and @angela-tran!

PR #2067, opened 2024-05-01
Feat: enrollment success

### LINKED ISSUE #2015: `Enrollment success`: update copy and spacing
image

## Acceptance Criteria

- [x] Headline copy has changed 
- [x] Body copy has changed (mentions the transit agency more)
- [x] “You were not charged…” was moved down as its own separate sentence
- [x] Make sure the image is 40px below the heading
- [x] Make sure the spacing between heading and body copy should be 56px
- [x] Spacing is 16px between the body copy and the image

lalver1:
Closes #2015


benefits/enrollment/templates/enrollment/success--mst.html
```diff
@@ -3,7 +3,7 @@
 
 {% block success-message %}
   {% blocktranslate trimmed %}
-    You were not charged anything today. When boarding an MST fixed route bus, tap this card when you board and you will be
+    Your contactless card is now enrolled in an MST Courtesy Card transit benefit. When boarding an MST bus, tap this card and you will be
```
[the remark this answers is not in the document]
thekaveman: Yes I think you are both correct here @lalver1 @angela-tran -- it seems like we need to move `enrollment_success_template` from `TransitAgency` to `EligibilityType` (as it seems like most of our enrollment config is going there...)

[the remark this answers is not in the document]
angela-tran: And, as @lalver1 said in https://github.com/cal-itp/benefits/pull/2067#discussion_r1586880387, we'll need to know what copy to use for the non-agency-card flows. 

Namely we'll need to know what to put in place of `MST Courtesy Card transit benefit` in the sentence `Your contactless card is now enrolled in an MST Courtesy Card transit benefit`. There currently isn't anything in Figma showing the non-agency-card copy for this page (for example, the [Benefits Full App page](https://www.figma.com/file/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?type=design&node-id=9754-52799&mode=design&t=6RiCv7S7uC16tWAt-4) shows the enrollment success page variants but with copy for agency cards.)

For the `--sacrt` template, it just says `SacRT light rail transit benefit`, so maybe the copy could be something generic like that for non-agency-card flows for MST and SBMTD? But still we need to know what that copy should be.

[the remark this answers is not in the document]
angela-tran: Hmm, actually looking more closely at the Benefits Full App page, it looks like the Figma screens are saying we're keeping the `You were not charged anything today. ...` copy for non-agency-card flows. In other words, the copy is not changing at all for the non-agency-card enrollment success pages.

benefits/core/migrations/local_fixtures.json
```diff
@@ -69,7 +69,8 @@
     "fields": {
       "name": "senior",
       "label": "(MST) Senior Discount",
-      "group_id": "group123"
+      "group_id": "group123",
+      "enrollment_success_template": "enrollment/success--mst-non-agency.html"
```
thekaveman, line 73: These names with `-non-agency` are a little confusing. I think we should keep the default here as `success--mst` (since it covers every other case) and then for agency cards override the name to `success--mst-courtesy-cards` etc.

thekaveman (changes requested):

thekaveman:
Taking a look at this now :eyes:

tests/pytest/core/test_models.py
```diff
@@ -247,6 +247,16 @@ def test_EligibilityType_enrollment_index_template(model_EligibilityType):
     assert model_EligibilityType.enrollment_index_template == "test/enrollment.html"
 
 
+@pytest.mark.django_db
+def test_EligibilityType_enrollment_success_template(model_EligibilityType):
```
thekaveman, line 251: I think this test can be updated slightly to assert what we changed here -- the `EligibilityType.enrollment_success_template` should have a _default value_ of `enrollment/success.html`.

As written, this test is asserting that the fixture `model_EligibilityType` was configured with that value (in the conftest).

Instead, we should test that creating a brand new `EligibilityType` should get the default value.

thekaveman: In other words, this test should not _inject the fixture_ but instead create the object directly in the test (with no value given for the template) and assert the default.

benefits/enrollment/templates/enrollment/success--sacrt-light-rail.html
```diff
@@ -0,0 +1,21 @@
+{% extends "enrollment/success.html" %}
```
thekaveman, line 1: Is there a difference between `success--sacrt-light-rail.html` and `success--sacrt.html`?

We aren't yet live with SacRT and I think by the time we are, it will cover more than just light rail. The user has no way of choosing "light rail only" benefits, so I don't think this template is needed?

thekaveman (changes requested):
Looking great. Just some minor cleanups requested.

benefits/enrollment/templates/enrollment/success--sacrt-light-rail.html
lalver1: I created `success--sacrt-light-rail.html` for the eventual case when SacRT offers a benefit similar to an agency card but I think you're right, this template is not needed (and it's not even referenced in `local_fixtures.json` now so it's probably just confusing) so I'll remove it. By the time we go live with SacRT we'll revise them 👍

tests/pytest/core/test_models.py
lalver1: Thanks! This makes sense, I updated the test to create the object and then assert the default value. I had a choice to use 

- `new_eligibility_type = EligibilityType.objects.create()` or 
- `new_eligibility_type = EligibilityType()` 

to create the object, but it seemed that `new_eligibility_type = EligibilityType.objects.create()` was more appropriate for this test since it [creates and saves the object](https://docs.djangoproject.com/en/5.0/topics/db/queries/#creating-objects) even though either command seemed to work for the test.

benefits/enrollment/templates/enrollment/success--sacrt-light-rail.html
thekaveman: Yeah I think the other thing to keep in mind (a "philosophy" if you will) is that we try to avoid creating code for situations we think might come up in the future.

As it stands now, we haven't talked with SacRT about an agency card flow (not even sure if they have an agency card?) so better to come back to it later if we do end up needing it.

tests/pytest/core/test_models.py
lalver1: Oh and I guess this may be out of scope for this PR (not because it's a huge change but just better to leave it out of this PR), but I could also update `test_EligibilityType_enrollment_index_template()` since it seems that it also tests the fixture when it should test the creation of a new object.

thekaveman: I think that's a good idea for a follow-up to align these tests :+1: 

There is a _slight_ difference here (and it isn't obvious when looking at just the test) in that the fixture didn't set the `enrollment_index_template` so this test is in fact asserting the default. But I agree it would be better to do it in the same way -- namely, creating the model directly in the test and asserting the default.

lalver1: Great, that sounds good! But yeah, I see the difference, the fixture didn't set `enrollment_index_template` so it is testing the behavior we want to test.

benefits/enrollment/templates/enrollment/success--sacrt-light-rail.html
lalver1: I agree, it does keep the code more organized and it's easier to follow 👍

tests/pytest/core/test_models.py
thekaveman: Noted this in a follow-up #2108

thekaveman (approved):
Nice work @lalver1!!

lalver1:
Thanks @thekaveman! And thanks for the super helpful review!

PR #2078, opened 2024-05-03
Fix: model cleanups for the Admin

### LINKED ISSUE #2065: AuthProvider should have friendly / useful display name in the Admin
Currently the [`AuthProvider`]() model, given that is is never user-facing, does not have any fields that capture a label or name specifically for display to an end-user, e.g. within the Admin. So the default is a Django constructed named based on the class and ID:

image

We do have the field `client_name` that could be used, which distinguishes each `AuthProvider` in the configuration and for the IdG. This field is currently marked as `Read: Transit Agency` in the [admin config guide](https://docs.google.com/spreadsheets/d/1lPp_LJTfJoWCyOkRC4RaljABGG_jPWOgm4gaAHJigMU/edit#gid=176385027), which is the most open read permission (i.e. any user that can log in to the admin can read this field). 

## Expected behavior

- [ ] When viewing the list of `AuthProvider` in the Admin, the `client_name` field is used for the link text
- [ ] When viewing an individual `AuthProvider` in the Admin, the `client_name` field is used wherever the object name is displayed (e.g. in the page title, the admin breadcrumb, the main content pane)

## Additional context

The super simple way to do this is to implement the `__str__` method on the model class, to return the text of the label you want it to use (in this case, `client_name`). This is how other model classes implement their labels, e.g. [`PemData`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L59) and [`TransitAgency`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L276)

### LINKED ISSUE #2072: EligibilityVerifier.api_auth_key_secret_name required in the Admin
When trying to edit an existing or add a new `EligibilityVerifier` model in the Admin, the following error message is displayed:

image

This is despite the [field in the model being marked with `null=True, blank=True`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L170)

It looks like the bug is in this [override of the `blank` kwarg in `SecretNameField`](https://github.com/cal-itp/benefits/blob/dev/benefits/core/models.py#L41):

```python
# similar to max_length, enforce at the field (form) validation level to not allow blanks
kwargs["blank"] = False
```

We want `blank = True` to allow the form to pass validation in the Admin.

## To Reproduce

1. Open the Admin
2. Click on `Eligibility verifiers`
3. Click an existing one and click the `Save` button (no changes necessary)

## Expected behavior

This is an optional field. It should not be required to save an `EligibilityVerifier` in the Admin.

thekaveman:
Closes #2065
* implements the `__str__` method on `AuthProvider` so it gets a human-readable label in the Admin

Closes #2072 
* removes the override of `blank = False` from `SecretNameField` to prevent validation errors in the Admin

### DIFF
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -100,6 +100,9 @@ class AuthProvider(models.Model):
     def client_id(self):
         return get_secret_by_name(self.client_id_secret_name)
 
+    def __str__(self) -> str:
+        return self.client_name
+
 
 class EligibilityType(models.Model):
     """A single conditional eligibility type."""
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -37,8 +37,6 @@ class SecretNameField(models.SlugField):
         # although the validator also checks for a max length of 127
         # this setting enforces the length at the database column level as well
         kwargs["max_length"] = 127
-        # similar to max_length, enforce at the field (form) validation level to not allow blanks
-        kwargs["blank"] = False
         # the default is False, but this is more explicit
         kwargs["allow_unicode"] = False
         super().__init__(*args, **kwargs)
--- a/tests/pytest/core/test_models.py
+++ b/tests/pytest/core/test_models.py
@@ -96,6 +96,7 @@ def test_PemData_data_text_secret_name_and_remote__uses_remote(
 def test_model_AuthProvider(model_AuthProvider):
     assert not model_AuthProvider.supports_claims_verification
     assert model_AuthProvider.supports_sign_out
+    assert str(model_AuthProvider) == model_AuthProvider.client_name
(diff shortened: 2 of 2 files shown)


angela-tran (approved):
This looks good to me

thekaveman:
Confirming on `dev`:

`AuthProvider` has friendlier names now:

image

`EligibilityVerifier.api_auth_key_secret_name` can now be blank:

image

thekaveman:
I might open a follow up ticket to add a label field to the `AuthProvider` since the client names can be the same, and that doesn't really say as much. Anyway, better than `Auth provider (0)` for now.

PR #2086, opened 2024-05-13
Chore: update eligibilityverifier and pemdata  migrations

lalver1:
Follow up on #2078 
Adds a missing migration for the `EligibilityVerifier` and `PemData` models.


thekaveman (approved):
Thanks for catching this!

thekaveman (commented):
It's a small thing, but the name of this migration file is a little long / doesn't really describe what is changing. I might recommend something like `0010_alter_secret_name_field_blank`.

thekaveman (approved):

PR #2088, opened 2024-05-13
Feat: system enrollment error - 500 error when linking

angela-tran:
Part of #2032 

This PR implements the system enrollment error page template and adds error-handling to the `benefits.enrollment.views.index` function so that if a 500-level HTTP error occurs when we try to link a funding source to a group, the system enrollment error page will be shown. The CTA takes the user back to the agency index.

A Sentry notification will be sent - see [an example from my testing](https://sentry.calitp.org/organizations/sentry/issues/75377/?project=3&referrer=slack). **Note:** since I don't know how to cause an actual 500 error with the linking endpoint, I had to add some code locally to fake it (more details in steps for reviewing / testing).

### Screenshots
<details><summary>Expand</summary>

| Desktop | Mobile |
| --- | --- |
| image | image |
| image | image |

</details>

### Steps for reviewing / testing

To simulate a 500-level HTTP error, I added this to my `index` function:
```diff
       try:
+           class response:
+               def __init__(self):
+                   self.status_code = 500
+           raise HTTPError(response=response())
            client.link_concession_group_funding_source(funding_source_id=funding_source.id, group_id=eligibility.group_id)
        except HTTPError as e:
```

Then run through a Login.gov flow and non-Login.gov flow all the way through enrollment as normal to see the system enrollment error page.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -10,6 +10,7 @@ from django.urls import reverse
 from django.utils.decorators import decorator_from_middleware
 from littlepay.api.client import Client
 from requests.exceptions import HTTPError
+import sentry_sdk
 
 from benefits.core import session
 from benefits.core.middleware import EligibleSessionRequired, VerifierSessionRequired, pageview_decorator
@@ -90,6 +91,7 @@ def index(request):
                 return success(request)
             elif e.response.status_code >= 500:
                 analytics.returned_error(request, str(e))
+                sentry_sdk.capture_exception(e)
 
                 # overwrite origin so that CTA takes user to agency index
                 session.update(request, origin=agency.index_url)
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -90,6 +90,9 @@ def index(request):
                 return success(request)
             elif e.response.status_code >= 500:
                 analytics.returned_error(request, str(e))
+
+                # overwrite origin so that CTA takes user to agency index
+                session.update(request, origin=agency.index_url)
                 return system_error(request)
             else:
                 analytics.returned_error(request, str(e))
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -24,6 +24,7 @@ ROUTE_SUCCESS = "enrollment:success"
 ROUTE_TOKEN = "enrollment:token"
 
 TEMPLATE_RETRY = "enrollment/retry.html"
+TEMPLATE_SYSTEM_ERROR = "enrollment/system_error.html"
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


benefits/enrollment/views.py
```diff
@@ -88,6 +90,13 @@ def index(request):
             if e.response.status_code == 409:
                 analytics.returned_success(request, eligibility.group_id)
                 return success(request)
+            elif e.response.status_code >= 500:
+                analytics.returned_error(request, str(e))
+                sentry_sdk.capture_exception(e)
+
+                # overwrite origin so that CTA takes user to agency index
+                session.update(request, origin=agency.index_url)
+                return system_error(request)
```
thekaveman, line 99: I wonder if it makes sense to do the analytics, sentry capture, and the session update inside the `system_error` function? Otherwise I imagine this will need to be repeated every time this system error is called?

angela-tran: Yeah I think it would make sense to do that. I was hesitating on it though because I wasn't sure how we would pass arguments to the view function when in later PRs we'll need to call `reverse` on this view function and/or call it from Javascript in the front-end.

I'm seeing now that passing arguments in those situations [isn't too bad](https://docs.djangoproject.com/en/5.0/topics/http/urls/#reverse-resolution-of-urls) assuming that we're ok with having captured values in the URL, e.g. `/system_error/<str:error>/<agency:agency>`...? 

This approach feels weird to me, maybe because our code doesn't normally pass things in the URL just for analytics / Sentry. Any thoughts or alternatives I'm not seeing?

angela-tran:
I forgot to run `./bin/makemessages.sh`! 😱 Commit incoming

benefits/enrollment/views.py
thekaveman: Ah, no that's a good point -- I think we could get away with if it was just the agency, grabbing that out of the session. But I overlooked the exception, and yeah I agree sending that in the URL from Javascript is weird.

angela-tran: Good idea, thanks! I added a test assertion for the session update in https://github.com/cal-itp/benefits/pull/2088/commits/87e4dfca1554c12af0832a9b0064ff4fbbede3bd and moved the session update into the view function in https://github.com/cal-itp/benefits/pull/2088/commits/8d551832a95563129271720fba747e6907a16cdd

thekaveman:
I rebased on `dev` to clean up conflicts in the PO files

thekaveman:
Rebased on `dev` one more time to resolve conflicts with #2067, and added the Spanish translations for this page.

I also updated the page title in English and Spanish using the latest from the [copy spreadsheet](https://docs.google.com/spreadsheets/d/1_Gi_YbJr4ZuXCOsnOWaewvHqUO1nC1nKqiVDHvw0118/edit#gid=610152150).

thekaveman (approved):
This looks good!

I was able to trigger the error using the suggested code edit, and then confirmed that the Sentry and Amplitude events were sent, copy is correct, and the CTA takes the user back to the agency homepage.

PR #2089, opened 2024-05-13
Feat: system enrollment error - 500 error during /token

angela-tran:
Part of #2032 

 - [x] 500-level error that occurs during `/token` returns a JSON response that is used to send user to `/enrollment/error`
 - [x] `/enrollment/error` shows system enrollment error template
 - [x] Sentry notification is sent
 - [x] Analytic event is sent (`failed access token request` with `status_code` in `event_properties`)

### Reviewing / testing
Similar to review steps for #2088, you can add some code locally to simulate a 500 error.
To `/token`, add:
```diff
        try:
+          class response:
+             def __init__(self):
+                self.status_code = 500
+
+          raise HTTPError(response=response())
           response = client.request_card_tokenization_access()
        except Exception as e:
```

### DIFF
--- a/benefits/enrollment/templates/enrollment/index.html
+++ b/benefits/enrollment/templates/enrollment/index.html
@@ -41,6 +41,12 @@
       $.ajax({ dataType: "script", attrs: { nonce: "{{ request.csp_nonce }}"}, url: "{{ card_tokenize_url }}" })
           .done(function() {
           $.get("{{ access_token_url }}", function(data) {
+              if (data.redirect) {
+                // https://stackoverflow.com/a/42469170
+                // use 'assign' because 'replace' was giving strange Back button behavior
+                window.location.assign(data.redirect);
+              }
+
               $(".loading").remove();
               // remove invisible and add back visible, so we aren't left with
               // a div with an empty class attribute
--- a/benefits/enrollment/urls.py
+++ b/benefits/enrollment/urls.py
@@ -15,4 +15,5 @@ urlpatterns = [
     path("reenrollment-error", views.reenrollment_error, name="reenrollment-error"),
     path("retry", views.retry, name="retry"),
     path("success", views.success, name="success"),
+    path("error", views.system_error, name="system-error"),
 ]
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -22,6 +22,7 @@ ROUTE_INDEX = "enrollment:index"
 ROUTE_REENROLLMENT_ERROR = "enrollment:reenrollment-error"
 ROUTE_RETRY = "enrollment:retry"
 ROUTE_SUCCESS = "enrollment:success"
+ROUTE_SYSTEM_ERROR = "enrollment:system-error"
 ROUTE_TOKEN = "enrollment:token"
 
 TEMPLATE_RETRY = "enrollment/retry.html"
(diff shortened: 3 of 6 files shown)


benefits/enrollment/views.py
```diff
@@ -45,8 +46,20 @@ def token(request):
...
         )
         client.oauth.ensure_active_token(client.token)
-        response = client.request_card_tokenization_access()
-        session.update(request, enrollment_token=response.get("access_token"), enrollment_token_exp=response.get("expires_at"))
+
+        try:
+            response = client.request_card_tokenization_access()
+        except Exception as e:
+            if isinstance(e, HTTPError) and e.response.status_code >= 500:
+                sentry_sdk.capture_exception(e)
```
angela-tran, line 54: Send a new specific analytic event for us failing to get an access token for card tokenization

benefits/enrollment/urls.py
```diff
@@ -15,4 +15,5 @@
     path("reenrollment-error", views.reenrollment_error, name="reenrollment-error"),
     path("retry", views.retry, name="retry"),
     path("success", views.success, name="success"),
+    path("system_error", views.system_error, name="system_error"),
```
angela-tran, line 18:
```suggestion
    path("error", views.system_error, name="system_error"),
```

benefits/enrollment/views.py
```diff
@@ -45,8 +46,20 @@ def token(request):
...
+                data = {"redirect": reverse(ROUTE_SYSTEM_ERROR)}
+                return JsonResponse(data)
+            else:
+                raise e
```
angela-tran, line 58: Send the same analytic event class as above. It should include the error code.

thekaveman:
Rebased on `dev` to incorporate changes from #2088

benefits/enrollment/views.py
angela-tran: Will be done for 400-level error case of `/token` (see #2032 acceptance criteria)

benefits/enrollment/views.py
angela-tran: See [line 53](https://github.com/cal-itp/benefits/pull/2089/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R53)

thekaveman (approved):
This looks great. I made the suggested modification in the view and was sent to the expected error page when trying to acquire the access token.

benefits/enrollment/templates/enrollment/index.html
```diff
@@ -41,6 +41,12 @@ <h1 class="pb-lg-8 pb-4">
       $.ajax({ dataType: "script", attrs: { nonce: "{{ request.csp_nonce }}"}, url: "{{ card_tokenize_url }}" })
           .done(function() {
           $.get("{{ access_token_url }}", function(data) {
+              if (data.redirect) {
+                // https://stackoverflow.com/a/42469170
+                // use 'assign' because 'replace' was giving strange Back button behavior
+                window.location.assign(data.redirect);
```
thekaveman, line 47: I'm not worried about this as blocking this PR, but...

I wonder if `window.location.replace` is the right method here instead? Do we want to support the back button when the user is supposed to wait and not try again immediately?

angela-tran: I created #2119 to follow up on this

PR #2020, opened 2024-04-11
Q2 Benefits product roadmap updates

indexing:
- Commented out our milestone to enhance the Veterans pathway. The timeline for additional attributes to verify Veteran eligibility is unclear.
- Added milestone to capture effort to improve UX for the agency card flow.
- Updated the name of the SBMTD Mobility Pass to Reduced Fare Mobility ID.
- Moved diagram styling to the end of the code block, as I was seeing a syntax error in another editor when it was placed before the Mermaid diagram type.

PR #2091, opened 2024-05-14
Fix: allow mkdocs preview job to write comments on PRs

thekaveman:
This permission is needed to allow the step that adds a comment with the preview URL

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -29,6 +29,8 @@ jobs:
     runs-on: ubuntu-latest
     # only pull requests should generate a preview
     if: github.event.pull_request
+    permissions:
+      pull-requests: write
     steps:
       - name: Checkout
         uses: actions/checkout@v4


thekaveman:
> Preview url: https://benefits-2091--cal-itp-previews.netlify.app

It works!

angela-tran (approved):
Just to make sure I understand, it was broken specifically for dependabot PRs right?

thekaveman:
> Just to make sure I understand, it was broken specifically for dependabot PRs right?

This fix (I'm pretty sure) is unrelated to Dependabot PRs. There was a separate issue with Dependabot PRs not having access to the secrets (because there is a separate set of secrets specifically for Dependabot): https://github.com/cal-itp/benefits/pull/2087#pullrequestreview-2055535958

Once I fixed that, this issue came up where the workflow couldn't add the preview URL comment, and it was because of this missing permission: https://github.com/cal-itp/benefits/actions/runs/9070328822/job/24952931531#step:8:31

So I'm thinking all preview comments would have failed? I guess we haven't had a docs update in a while... :thinking:

angela-tran:
> I think the `GITHUB_TOKEN` for Dependabot PRs is more limited by default: [dependabot/dependabot-core#3253 (comment)](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544).
> 
> For example we had a docs update back in April and the preview URL comment was added successfully: [#2020 (comment)](https://github.com/cal-itp/benefits/pull/2020#issuecomment-2050549616)

Makes sense! I thought I had seen it work before, so this would explain the difference. Thanks 🙏

PR #2116, opened 2024-05-21
Refactor: use integer claims in OAuth

### LINKED ISSUE #2046: Refactor claims handling for integer claims
During the OAuth `authorize` flow, we look for [boolean claim values](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/views.py#L75) to determine if the user is eligible.

IdG is changing their claims implementation to cut down on the size of the token being sent to Benefits. Instead of booleans, they will use integers to indicate claim values:

* `0` will indicate `False` (i.e. the claim indicates eligibility failed)
* `1` will indicate `True` (i.e. the claim indicates eligibility succeeded)
* Any other integer `>= 10` will indicate an error code

**Note:** the claim values are transmitted in the token as `str`, and should be parsed to `int` before usage.

## Acceptance Criteria

- [ ] `authorize` processes integer claims as described above

## Additional context

While we work to implement this change, existing flows for Older Adults and Veterans will use both claim styles. New flows for CalFresh and the new Veterans API will ~only use the newer integer claim style, so this refactor is necessary for supporting those flows.~ also support both styles to allow us time to implement and cut over. There are an entirely new set of scopes created for the integer-based claims so as not to interfere with the existing implementation.

Once we have this change tested and deployed, IdG will cutover all flows to use the integer style only.

Mapping error codes to error messages and analytics will be handled in #2049.

See [this Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1714434750536319) from @johnatstate for more context.

lalver1:
Closes #2046 

## Steps for testing
Set Django's database `authprovider` fixtures to the new [scopes](https://cal-itp.slack.com/archives/C037Y3UE71P/p1715882088978089?thread_ts=1715881952.138259&cid=C037Y3UE71P)/[claims](https://cal-itp.slack.com/archives/C037Y3UE71P/p1715882216002199?thread_ts=1715881952.138259&cid=C037Y3UE71P). Run bin/reset_db.sh to reset database.

## Test results
  - [x] Successfully test older adult flow locally
  - [x] Successfully test veteran flow locally

### DIFF
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -69,11 +69,12 @@ def authorize(request):
 
         if userinfo:
             claim_value = userinfo.get(verifier_claim)
-            # the claim comes back in userinfo like { "claim": "True" | "False" }
+            # the claim comes back in userinfo like { "claim": "1" | "0" }
+            claim_value = int(claim_value) if claim_value else None
             if claim_value is None:
                 logger.warning(f"userinfo did not contain: {verifier_claim}")
-            elif claim_value.lower() == "true":
-                # if userinfo contains our claim and the flag is true, store the *claim*
+            elif claim_value == 1:
+                # if userinfo contains our claim and the flag is 1 (true), store the *claim*
                 stored_claim = verifier_claim
 
     session.update(request, oauth_token=id_token, oauth_claim=stored_claim)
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
This looks good to me. I also tested it out locally with the new scope / claim values

PR #2117, opened 2024-05-22
Feat: server error - 400 errors / other exceptions during /token

angela-tran:
Part of #2032 

 - [x] 400-level error or a non-`HTTPError` exception during `/token` returns a JSON response that is used to send user to `/error`
 - [x] `/error` shows server error template
 - [x] Sentry notification is sent
 - [x] Analytic event is sent (`failed access token request`)

### Reviewing / testing

**To test handling of 400-level HTTP errors**, you'll have to force it like in other PRs such as #2089.
To `/token`, add:
```diff
        try:
+          class response:
+             def __init__(self):
+                self.status_code = 400
+
+          raise HTTPError(response=response())
           response = client.request_card_tokenization_access()
        except Exception as e:
```

**To test other exceptions**:
- setting an incorrect `client_id` on `PaymentProcessor` causes `UnsupportedTokenTypeError` on `client.request_card_tokenization_access`
- setting an incorrect `api_base_url` on `PaymentProcessor` causes a `ConnectionError` on `client.ensure_active_token`
- setting an incorrect `client_secret_name` on `PaymentProcessor` could cause an error when creating `Client` (we read from `payment_processor.client_secret_name`)

### DIFF
--- a/benefits/enrollment/templates/enrollment/index.html
+++ b/benefits/enrollment/templates/enrollment/index.html
@@ -45,6 +45,7 @@
                 // https://stackoverflow.com/a/42469170
                 // use 'assign' because 'replace' was giving strange Back button behavior
                 window.location.assign(data.redirect);
+                return;
               }
 
               $(".loading").remove();
--- a/benefits/enrollment/templates/enrollment/index.html
+++ b/benefits/enrollment/templates/enrollment/index.html
@@ -45,7 +45,7 @@
                 // https://stackoverflow.com/a/42469170
                 // use 'assign' because 'replace' was giving strange Back button behavior
                 window.location.assign(data.redirect);
-                return;
+                return; // exit early so the rest of this function doesn't execute
               }
 
               $(".loading").remove();
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -50,6 +50,7 @@ def token(request):
         try:
             response = client.request_card_tokenization_access()
         except Exception as e:
+            logger.debug("Error occurred while requesting access token", exc_info=e)
             sentry_sdk.capture_exception(e)
 
             if isinstance(e, HTTPError):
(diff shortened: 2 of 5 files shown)


benefits/enrollment/views.py
```diff
@@ -21,6 +21,7 @@
 ROUTE_INDEX = "enrollment:index"
 ROUTE_REENROLLMENT_ERROR = "enrollment:reenrollment-error"
 ROUTE_RETRY = "enrollment:retry"
+ROUTE_SERVER_ERROR = "core:server-error"
```
thekaveman, line 24: Does it feel weird to have this definition here? Should it be imported from `core.urls` or something?

benefits/enrollment/templates/enrollment/index.html
```diff
@@ -45,6 +45,7 @@ <h1 class="pb-lg-8 pb-4">
                 // https://stackoverflow.com/a/42469170
                 // use 'assign' because 'replace' was giving strange Back button behavior
                 window.location.assign(data.redirect);
+                return;
```
thekaveman, line 48: Why is this needed? I thought it was working in #2089? Maybe add a small comment

thekaveman: Ah, I see the commit message now. I think a code comment would help thought!

benefits/enrollment/views.py
```diff
@@ -38,25 +38,35 @@ def token(request):
...
             sentry_sdk.capture_exception(e)
 
-            if isinstance(e, HTTPError) and e.response.status_code >= 500:
-                analytics.failed_access_token_request(request, e.response.status_code)
-                data = {"redirect": reverse(ROUTE_SYSTEM_ERROR)}
-                return JsonResponse(data)
+            if isinstance(e, HTTPError):
+                status_code = e.response.status_code
+
+                if e.response.status_code >= 500:
```
thekaveman, line 58: Small suggestion to use the `status_code` variable here instead

thekaveman (changes requested):
Very minor request, otherwise this looks good!

benefits/enrollment/views.py
angela-tran: Thanks for catching. Done in [fc291ed](https://github.com/cal-itp/benefits/pull/2117/commits/fc291ed866e52c07681a64bd81977554e0188f96)

thekaveman (approved):
:+1:

PR #2120, opened 2024-05-23
Feat: `onError` shows error pages rather than retry

angela-tran:
Part of #2032


lalver1 (approved):
This looks good to me 👍. I locally tested the 500-level HTTP error behavior using the method outlined in #2088 and saw the expected error page.

PR #2122, opened 2024-05-24
CI: check for up-to-date migrations

### LINKED ISSUE #2092: Add a PR check to ensure migrations are up to date
Sometimes we make model changes and forget to generate a new migration, e.g. with the case of #2078 and the follow-up #2086.

Let's create a PR check that runs the `bin/makemigrations.sh` script and errors if there are pending migrations.

When a migration is needed, the output looks like:

```console
$ bin/makemigrations.sh
Migrations for 'core':
  benefits/core/migrations/NNNN_changes.py
    - [description of changes]
```

When there are no pending migrations, the output looks like:

```console
$ bin/makemigrations.sh
No changes detected
```

## Acceptance Criteria

- [x] A new workflow runs for all PRs that update files in the `benefits/` directory
- [x] The workflow runs `bin/makemigrations.sh`
- [x] Unless the output is as above `No changes detected`, the workflow fails
- [x] The workflow is added as a required check to merge PRs

angela-tran:
Closes #2092 

I tested this locally with `act`:
image

I think this will need to be reviewed, approved, and merged into `dev` before we can see it in action on a PR.

### DIFF
--- a/.github/workflows/check-makemigrations.yml
+++ b/.github/workflows/check-makemigrations.yml
@@ -1,0 +1,8 @@
+name: Check for up-to-date Django migrations
+on:
+  pull_request:
+    paths:
+      - "benefits/**"
+  push:
+    paths:
+      - "benefits/**"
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
This looks good!

lalver1:
Nice! I hadn't had a chance to try `act` but I ran this workflow locally with it today and it looks good!

PR #2125, opened 2024-05-29
Feat: analytics for auth claims flow

### LINKED ISSUE #2049: Improve analytics for auth claims flow
With the move to integer claims in #2046, IdG will return error codes in place of True (`0`) or False (`1`) when an error occurs in processing the user's eligibility. Any integer that is greater than or equal to `10` is considered an error code.

We want to improve our analytics around the auth flow, in particular to capture any errors that occur as described by IdG error codes.

## Acceptance Criteria

- [x] The [`FinishedSignInEvent`](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/analytics.py#L32) is updated to allow an optional `error_code` event property.
- [x] The `error_code` property is the integer value of the claim error code (originally transmitted as a string)

## Additional context

In #2046, we will handle the regular True (`0`) / False (`1`) cases.

This issue has been scoped back to exclude lookup of the error message associated with a given code. This needs further discussion and design with CDT.

lalver1:
Closes #2049 

## How to test
To test the error condition, insert something like `claim_value = "20"` after line 73 in `benefits.oauth.views.authorize`. Then check that `"error_code": 20` appears as a property of `"event_properties"` in the log.

## Notes
The behavior originally described in these notes was fixed by #2127.

~~When running the test `test_authorize_success_without_verifier_claim`, the oauth middleware `VerifierUsesAuthVerificationSessionRequired` that decorates the `authorize` view seems to behave differently compared to debugging the application.~~

~~Specifically, [line 19](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/middleware.py#L19C12-L19C59) in `benefits/oauth/middleware.py` evaluates to a `<Mock>` object if running the test, but it evaluates to a boolean if debugging the application.~~

~~To see this behavior, debug the test and notice that [line 19](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/middleware.py#L19C12-L19C59) evaluates to a`<Mock>` object, so the `if` condition evaluates to `True` and returns `None` instead of returning the user error page.
However, when debugging the application and setting `claim` in `AuthProvider` to blank (to simulate a missing verifier claim in the database), [line 19](https://github.com/cal-itp/benefits/blob/dev/benefits/oauth/middleware.py#L19C12-L19C59) will evaluate to `False` and the conditional will return the user error page.~~

~~Also, seems like `verifier.auth_provider.claim = ""` in the test only sets `verifier_claim = verifier.auth_provider.claim` (line 64 in the view). For a test closer to the behavior being tested, we may need to set the `claim` to blank in a fixture similar to `model_AuthProvider_with_verification`.~~

### DIFF
--- a/benefits/oauth/views.py
+++ b/benefits/oauth/views.py
@@ -64,6 +64,8 @@ def authorize(request):
     verifier_claim = verifier.auth_provider.claim
     stored_claim = None
 
+    error_claim = None
+
     if verifier_claim:
         userinfo = token.get("userinfo")
 
@@ -76,10 +78,11 @@ def authorize(request):
             elif claim_value == 1:
                 # if userinfo contains our claim and the flag is 1 (true), store the *claim*
                 stored_claim = verifier_claim
+            elif claim_value >= 10:
+                error_claim = claim_value
 
     session.update(request, oauth_token=id_token, oauth_claim=stored_claim)
-
-    analytics.finished_sign_in(request)
+    analytics.finished_sign_in(request, error=error_claim)
 
     return redirect(ROUTE_CONFIRM)
 
--- a/tests/pytest/oauth/test_views.py
+++ b/tests/pytest/oauth/test_views.py
@@ -138,6 +138,27 @@ def test_authorize_success_with_claim_false(
 
 
 @pytest.mark.django_db
+def test_authorize_success_with_claim_error(
+    mocked_session_verifier_uses_auth_verification,
+    mocked_oauth_create_client,
+    mocked_analytics_module,
+    app_request,
+):
+    verifier = mocked_session_verifier_uses_auth_verification.return_value
+    verifier.auth_provider.claim = "claim"
+    mocked_oauth_client = mocked_oauth_create_client.return_value
+    mocked_oauth_client.authorize_access_token.return_value = {"id_token": "token", "userinfo": {"claim": "10"}}
+
+    result = authorize(app_request)
+
+    mocked_oauth_client.authorize_access_token.assert_called_with(app_request)
+    mocked_analytics_module.finished_sign_in.assert_called_with(app_request, error=10)
+    assert session.oauth_claim(app_request) is None
+    assert result.status_code == 302
+    assert result.url == reverse(ROUTE_CONFIRM)
+
+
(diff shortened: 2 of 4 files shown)


benefits/oauth/views.py
```diff
@@ -79,7 +79,10 @@ def authorize(request):
 
     session.update(request, oauth_token=id_token, oauth_claim=stored_claim)
 
-    analytics.finished_sign_in(request)
+    if claim_value is not None and claim_value >= 10:
```
angela-tran, line 82: I think we should move this check to be an `elif` after line 78 so that we are protected against the case in which the OAuth response (`token` on line 52) does not contain the claim for some reason.

This is the scenario being tested in `test_authorize_success_without_claim_in_response`. Once this PR has the fix from #2127, I believe `test_authorize_success_without_claim_in_response` will still fail because of this.

lalver1: Thanks! That makes sense, I'll move this check to be an `elif`.

lalver1:
I'm unsure what to do about missing tests for [lines 68-71](https://github.com/cal-itp/benefits/blob/4b253c2203bcba53ea48da6fa886506792701ac1/benefits/oauth/analytics.py#L68C5-L71C54) in `benefits.oauth.analytics.finished_sign_in` since `finished_sign_in` doesn't return a value; I likely need a mocker.

tests/pytest/oauth/test_views.py
```diff
@@ -137,6 +137,26 @@ def test_authorize_success_with_claim_false(
     assert result.url == reverse(ROUTE_CONFIRM)
 
 
+@pytest.mark.django_db
+@pytest.mark.usefixtures("mocked_analytics_module")
+def test_authorize_success_with_claim_error(
```
thekaveman, line 142: This is good, but sort of misses the actual meat of this PR's change, which was to send the code along with analytics.

thekaveman (changes requested):

tests/pytest/oauth/test_views.py
thekaveman: Suggest injecting `mocked_analytics_module` directly and asserting the call on that.

PR #2127, opened 2024-05-30
Fix: test assertions for missing claim on AuthProvider

angela-tran:
Prompted by @lalver1 's questions / notes for #2125

`test_authorize_success_without_verifier_claim` was giving us a false positive because the verifier fixture is also not behaving correctly when returning `uses_auth_verification`. Also see our [Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1717085288461349).

I'm not sure at what point this test became incorrect. It could be that when the test was written, we were using `VerifierSessionRequired` rather than `VerifierUsesAuthVerificationSessionRequired`, and so we expected to be able to reach `ROUTE_CONFIRM` which would then handle the fact that no claim was stored. See https://github.com/cal-itp/benefits/issues/1535 for why we changed the middleware behavior.

We can revisit if showing `user_error` is the correct UX here, but for now, the test should at least reflect the application behavior rather than giving us a false positive and blocking #2125.

### DIFF
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -274,7 +274,7 @@ def mocked_session_verifier_oauth(mocker, model_EligibilityVerifier_AuthProvider
 
 
 @pytest.fixture
-def mocked_session_verifier_auth_required(
+def mocked_session_verifier_uses_auth_verification(
     mocker, model_EligibilityVerifier_AuthProvider_with_verification, mocked_session_verifier_oauth
 ):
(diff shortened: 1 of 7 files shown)


angela-tran:
Ready to be reviewed

lalver1 (approved):
Looks good to me and all the tests passed 👍 . In particular, using `model_EligibilityVerifier_AuthProvider_with_verification` instead of the Mock verifier unblocks #2125.

PR #2137, opened 2024-06-05
CI: guard against empty language array for CodeQL runs

thekaveman:
Only run the CodeQL Analyze job if there are languages to analyze.

I'm not 100% this is the right syntax :crossed_fingers:


angela-tran (approved):
Let's try it 🧪

PR #2140, opened 2024-06-05
Fix: correct invalid syntax in CodeQL

thekaveman:
Follow-up to #2137, the syntax was in fact wrong :roll_eyes: 

Failed workflow run: https://github.com/cal-itp/benefits/actions/runs/9391520844


angela-tran (approved):
🤞

PR #2152, opened 2024-06-10
Refactor(oauth): resolve RuntimeWarning

### LINKED ISSUE #2003: Resolve `RuntimeWarning` about interacting with database during `AppConfig.ready()`
> Should we be worried about this / look into it further? What can go wrong if "the app registry has not been fully populated" / what does that even mean?

Looking into this a little further, I found some Django docs that correspond with the `RuntimeWarning` we're seeing. This section [warns against interacting with the database in `AppConfig.ready()`](https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.ready):

image

Our [`OAuthAppConfig`](https://github.com/cal-itp/benefits/blob/027d3aebcff0dc8c9f84215e0e1b0a855c85981c/benefits/oauth/apps.py#L17) interacts with the database by [querying for all `AuthProvider`s](https://github.com/cal-itp/benefits/blob/027d3aebcff0dc8c9f84215e0e1b0a855c85981c/benefits/oauth/client.py#L44).

The "Application intialization process" documentation describes how the app registry gets populated and why "premature database queries are discouraged": https://docs.djangoproject.com/en/5.0/ref/applications/#initialization-process

But I'm not sure if there's a better place for us to register our Authlib configurations.

Maybe we should register each `AuthProvider`'s configuration upon the first attempt to actually use it?

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/issues/1809#issuecomment-1874724416_

angela-tran:
Closes #2003 

The problem with calling `register_providers` in `ready()` is that it reads from the database (to get all the `AuthProvider`s), but `ready` is called during management commands, so it is not entirely safe to interact with the database at that time.

This PR refactors the function to take one provider model object and moves it from `ready()` to the view where the OAuth `client` object is needed. The [`register`](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/registry.py#L73-L87) method on `BaseOAuth` looks safe to call multiple times and returns a client object.

### When reviewing
You should see that the `RuntimeWarning` no longer occurs when starting up the app (or running any management command - e.g. `makemigrations`, `makemessages`).


benefits/oauth/views.py
```diff
@@ -24,7 +24,8 @@
 def login(request):
     """View implementing OIDC authorize_redirect."""
     verifier = session.verifier(request)
-    oauth_client = oauth.create_client(verifier.auth_provider.client_name)
+
+    oauth_client = register_provider(oauth, verifier.auth_provider)
```
thekaveman, line 28: This looks good but I think we also need this change in the `authorize` function down below.

thekaveman (changes requested):

benefits/oauth/client.py
```diff
@@ -42,23 +39,20 @@ def _authorize_params(scheme):
...
     Adapted from https://stackoverflow.com/a/64174413.
     """
-    logger.info("Registering OAuth clients")
-
-    providers = AuthProvider.objects.all()
+    logger.debug(f"Registering OAuth client: {provider.client_name}")
 
-    for provider in providers:
-        logger.debug(f"Registering OAuth client: {provider.client_name}")
+    client = oauth_registry.register(
```
lalver1, line 50: At first I was wondering about this line since the [Authlib documentation](https://docs.authlib.org/en/latest/client/api.html#module-authlib.integrations.django_client) does not mention a return type for `.register()`, but after running the app in debug mode and looking at the [Authlib code](https://github.com/lepture/authlib/blob/master/authlib/integrations/base_client/registry.py#L87C9-L87C40), it turns out that `.register()` does return a client (`<authlib.integrations.django_client.apps.DjangoOAuth2App object`), which is the same client shown in `oauth._clients` after calling `register()` on `oauth`.

lalver1 (commented):
This looks like it fixed the issue! 👍 I ran the app locally and saw that the `RuntimeWarning` no longer shows up.

I was wondering about the difference between
`oauth_client = register_provider(oauth, verifier.auth_provider)` and
`oauth_client = oauth.create_client(verifier.auth_provider.client_name)`

and when to use one over the other 🤔. It looks like both of them return the same object and calling `register_provider()` multiple times is safe, so they seem to be equivalent (as long as a provider has been registered). I guess this observation may relate to Kegan's [comment](https://github.com/cal-itp/benefits/pull/2152/files#r1633957729).

benefits/oauth/client.py
angela-tran: > I was wondering about the difference between `oauth_client = register_provider(oauth, verifier.auth_provider)` and `oauth_client = oauth.create_client(verifier.auth_provider.client_name)`
> 
> and when to use one over the other 🤔. It looks like both of them return the same object and calling `register_provider()` multiple times is safe, so they seem to be equivalent (as long as a provider has been registered). I guess this observation may relate to Kegan's [comment](https://github.com/cal-itp/benefits/pull/2152/files#r1633957729).

Yeah, so you'd want to use `register_provider(oauth, verifier.auth_provider)` instead of `oauth.create_client(verifier.auth_provider.client_name)` when it's possible that the AuthProvider's config hasn't been registered with Authlib yet. 

I used `register_provider` in the `login` view but not in `authorize` because we know `authorize` will always be preceded by `login` and therefore the client will already be registered. But as @thekaveman is suggesting, it wouldn't hurt to do that in `authorize` too, so I'm good with making that change.

Ideally I would've liked to somehow check if the `client_name` has already been registered, call `register` for it if it hasn't, and then call `create_client`. But Authlib doesn't provide a way for us to check the registry, so all we can do is call `register` every time.

benefits/oauth/views.py
angela-tran: I don't think it's needed since `authorize` always follows `login` currently, but it also doesn't hurt to have it in `authorize`. Made the change in [770f8e1](https://github.com/cal-itp/benefits/pull/2152/commits/770f8e185a192f802f1bec47b31c5b6e48a2b763)

benefits/oauth/client.py
thekaveman: > But Authlib doesn't provide a way for us to check the registry, so all we can do is call register every time.

What happens when we try to get a `client_name` that hasn't been registered yet? Exception, or does it just return `None`? Either way, we could handle that case and use it to register. Wrap both cases up in a helper function and call it a day :sunglasses:

angela-tran: I think an exception could mean a misconfiguration, so I'll leave that case out. But I went ahead and made the helper function that registers the client if we got `None` back - [4d2e1ec](https://github.com/cal-itp/benefits/pull/2152/commits/4d2e1ec854acb45f25b9b770492becdd636fc7c4). Hopefully this code reads as more intuitive.

thekaveman (approved):
Looks good :+1:

benefits/oauth/client.py
```diff
@@ -42,23 +39,32 @@ def _authorize_params(scheme):
...
-def register_providers(oauth_registry):
+def _register_provider(oauth_registry, provider):
     """
-    Register OAuth clients into the given registry, using configuration from AuthProvider models.
+    Register OAuth clients into the given registry, using configuration from AuthProvider model.
 
     Adapted from https://stackoverflow.com/a/64174413.
     """
-    logger.info("Registering OAuth clients")
+    logger.debug(f"Registering OAuth client: {provider.client_name}")
```
thekaveman, line 48: :+1:

PR #2155, opened 2024-06-10
Feat: check makemessages

angela-tran:
Similar to #2122, this PR adds a workflow to check that the Django message files are up-to-date.

Rather than making a new workflow, I added a step to the existing `check-makemigrations` workflow so that the two steps can use the same runner that has already checked out the repo, installed system packages, installed Python packages, etc. and to minimize the amount time it takes to check migrations and messages. (See https://github.com/cal-itp/benefits/issues/2092#issuecomment-2146300444 for more context on required checks.)


thekaveman (approved):
Nice, makes sense to combine into a single check.

angela-tran:
> Nice, makes sense to combine into a single check.
> 
> You'll need to update the branch protection in order to merge this PR, since the `check-makemigrations` check was renamed by this PR and won't ever run.

Thanks! I updated the branch protection for `dev`, `test`, and `prod`.

PR #2169, opened 2024-06-14
Fix: verifier help templates

### LINKED ISSUE #2168: Help page breaks when adding a new EligibilityVerifier through admin interface
## To Reproduce

1. Start with a database that has just loaded fixtures (`./bin/reset_db.sh`)
   - At this point, Help page works when starting from `/` or after choosing a transit provider
2. Go into admin interface, and add a new EligibilityVerifier, leaving the `help_template` blank, which is a valid thing to do. Click Save.
3. Add that Eligibility Verifier to one of the TransitAgencys
4. Go to `/` and select that TransitAgency
5. Click on the `Help` link at the bottom of the page
6. See error

## Expected behavior

The Help page should show the help sections for any of the TransitAgency's verifiers that specify a `help_template`.

## Screenshots

DEBUG off
image

DEBUG on
image

## Additional context

#1970 introduced `help_template`

Django docs on `null` parameter on `TextField`: https://docs.djangoproject.com/en/5.0/ref/models/fields/#null
> the Django convention is to use the empty string, not `NULL`

angela-tran:
This PR fixes #2168  by making sure the context variable `agency.help_templates` doesn't include empty strings.

The logic previously did not consider the case where the admin interface populates the `help_template` as an empty string rather than `None`. If you load an `EligibilityVerifier` fixture that doesn't specify `help_template`, the field gets set as `None`.

### DIFF
--- a/benefits/core/context_processors.py
+++ b/benefits/core/context_processors.py
@@ -15,7 +15,7 @@ def unique_values(original_list):
 def _agency_context(agency):
     return {
         "eligibility_index_url": agency.eligibility_index_url,
-        "help_templates": unique_values([v.help_template for v in agency.active_verifiers if v.help_template is not None]),
+        "help_templates": unique_values([v.help_template for v in agency.active_verifiers if v.help_template]),
         "info_url": agency.info_url,
         "long_name": agency.long_name,
         "phone": agency.phone,


angela-tran:
The Help page on `dev-benefits.calitp.org` was not working with MST as the selected transit provider because it had a CalFresh verifier with no `help_template`. I added `core/includes/help--calfresh.html` as the `help_template` and also made this fix so we can handle any new verifiers that don't need to add help sections.

lalver1 (approved):
This fixed it! 👍 I was able to locally recreate the issue in #2168 and confirmed that this PR filters out empty strings.

PR #2174, opened 2024-06-18
Refactor: funding source lookup

### LINKED ISSUE #2068: Refactor current enrollment check to check groups for funding source
When we first implemented enrollment expiration logic in #1989, the only way to check if the user (funding source) was already linked to a group was to look up all funding sources linked to the group, and check if the user's funding source was in the that list. Over time with increasing enrollments, this won't scale as every single funding source will have to be checked each time.

Now Littlepay offers an endpoint to look up the groups linked to a given funding source, i.e. the inverse relationship. See https://github.com/cal-itp/littlepay/issues/60.

This will be a much shorter list to check -- empty for new users (funding sources), and perhaps one or two and not many more for existing users.

## Acceptance Criteria

During enrollment:

- [ ] Instead of looking up funding sources for the group, look up groups for the funding source
- [ ] The user is considered already enrolled if the current eligibility's group is in the list

## Additional context

Depends on https://github.com/cal-itp/littlepay/issues/60.

This will be a refactor of the the work that comes back (but may be modified) with #2062.

angela-tran:
Closes #2068 

This PR uses the endpoint implemented in cal-itp/littlepay#60 to look up a funding source's groups. Methods/variables were renamed to reflect that.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -108,16 +108,16 @@ def index(request):
         group_id = eligibility.group_id
 
         try:
-            group_funding_source = _get_group_funding_source(
+            funding_source_group = _get_funding_source_group(
                 client=client, group_id=group_id, funding_source_id=funding_source.id
             )
 
-            already_enrolled = group_funding_source is not None
+            already_enrolled = funding_source_group is not None
 
             if eligibility.supports_expiration:
                 # set expiry on session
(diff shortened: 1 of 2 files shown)


thekaveman:
Let's go ahead and close this now that all the enrollment logic is refactored.

We can revisit this change later against the refactored code.

PR #2179, opened 2024-06-20
Fix: changing languages on the enrollment success page

### LINKED ISSUE #2058: Changing languages on the success page takes user back to Enrollment Index
Opening this ticket to document what I noticed. Not sure if this is a "it's a feature not a bug" situation

## To Reproduce

1. Go through a full flow so that you're on the success screen
2. Click on "Español" button
3. Notice you are now on the Enrollment Index

## Expected behavior
I would've thought the success screen should just be shown again in Spanish.

lalver1:
Closes #2058 

On the enrollment success page, the Django [`set_language`](https://github.com/django/django/blob/main/django/views/i18n.py#L30) redirect view was redirecting the user to the enrollment index page when the language was changed. Looking at the [Django documentation](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#the-set-language-redirect-view), adding a `next` parameter in the `POST` data and explicitly including the enrollment success page url in the context seems to be what was missing.

For all other views, when changing languages, I think we don't see this behavior because the `Referer` request header contains the address of the same view, but for the enrollment success view, `Referer` is `/enrollment/` (I think because it is called from `/enrollment/` initially), so using the `next` parameter is required.

### DIFF
--- a/benefits/core/templates/core/includes/lang-selector.html
+++ b/benefits/core/templates/core/includes/lang-selector.html
@@ -14,6 +14,7 @@ If more language support is added, the hidden input can be replaced by a select
     <form method="post" action="{% url "set_language" %}">
       {% csrf_token %}
       {% get_language_info for code as langinfo %}
+      <input name="next" type="hidden" value="{{ redirect_to }}" />
       <input name="language" type="hidden" value="{{ code }}" />
       <button class="btn btn-lg btn-outline-light" type="submit">{{ langinfo.name_local | capfirst }}</button>
     </form>
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -289,4 +289,5 @@ def success(request):
         session.update(request, origin=reverse(ROUTE_LOGGED_OUT))
 
     analytics.returned_success(request, eligibility.group_id)
-    return TemplateResponse(request, eligibility.enrollment_success_template)
+    context = {"redirect_to": request.path}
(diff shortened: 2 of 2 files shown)


benefits/core/templates/core/includes/lang-selector.html
```diff
@@ -14,8 +14,9 @@
     <form method="post" action="{% url "set_language" %}">
       {% csrf_token %}
       {% get_language_info for code as langinfo %}
+      <input name="next" type="hidden" value="{{ redirect_to }}" />
       <input name="language" type="hidden" value="{{ code }}" />
-      <button class="btn btn-lg btn-outline-light" type="submit">{{ langinfo.name_local | capfirst }}</button>
+      <input class="btn btn-lg btn-outline-light" type="submit" value="{{ langinfo.name_local | capfirst }}">
```
angela-tran, line 19: Just wanted to confirm: this change wasn't actually needed to fix the bug, right? That is, there's no behavioral difference, and it's just a different way of defining the same submit button.

angela-tran (dismissed):
Nice fix and explanation! Thanks for linking the Django docs

benefits/core/templates/core/includes/lang-selector.html
lalver1: Thanks @angela-tran , that's a good point! Changing `button` to `input` is actually not needed to fix the bug so I'll revert that change by pushing one more time. I think doing this keeps things cleaner.

angela-tran (approved):
🚀

Issue #1527, opened 2023-07-13
Consider a "Deploy on Release" model


thekaveman:
Today, our [release process](http://docs.calitp.org/benefits/deployment/release/) requires a number of manual steps, and at least 1 Pull Request requiring interaction from at least 2 people (the author and an approver). This works well but is cumbersome, and could be more efficient especially for getting urgent or timely changes out.

Further, we have 3 separate "versions" of the code across the `dev`, `test`, and `prod` branches. The explicit branch naming makes it easy to see where each environment is, but it also means we're constantly merging and needing to remember the state of each branch (e.g. is `test` ready to release to `prod`? Did we merge that fix into `test` yet?)

The final step in our release process today is to tag and create a [GitHub release](https://github.com/cal-itp/benefits/releases). 

I propose we invert this model and instead make the tagging and GitHub release the initial step that kicks off the rest of the deployment.

### Branches

* We remove the 3 environment branches: `dev`, `test`, and `prod`
* All new code is merged to a single default branch `main`
* `dev` would essentially be renamed to `main` but otherwise function the same

### Deployments

* Merging to `main` deploys to the `dev` environment (just like merging to `dev` today)
* Creating a GitHub pre-release deploys to the `test` environment
* Create a GitHub release deploys to the `prod` environment

## Discussion

Creating a GitHub Release requires an associated `git tag`. 

Today we tag the `prod` branch at the end of the process, just before creating the GitHub Release.

Under this new proposed model:

* All the code (generally) is in a single branch / timeline `main`
* We tag at the beginning of the process to signal a release
* We have far more tags on the `main` branch, as both `test` and `prod` releases are tagged
* Releases are just specific points in the timeline ("forward" or "backward")
* We could think about using tag names or partial names to signify specific features (e.g. `sacrt` or `veterans`)
* `release == tag == commit`

### Pros

* Simpler release process: choose a commit, tag it, push the tag, make a (pre) release in GitHub
* Reduced friction to release: no more PRs for approval, drastically reduced [release checklist](https://github.com/cal-itp/benefits/issues/new?assignees=&labels=release&projects=&template=release.yml&title=Make+a+Release)
* Easier to get features and fixes out sooner
* Features and fixes can be released to `test` / `prod` independent of where or whether the commit is in `main`

### Cons

* There will be a lot of tags
* It may be _too easy_ to make a deployment? E.g. broken code, tests not passing
[...]

PR #2202, opened 2024-07-05
Chore: Rename `dev` branch to `main`

lalver1:
This PR implements the first step of code changes needed to transition the `benefits` repository to tag-based deployments as described in #1527. The code changes are associated with:

- GitHub workflows
- GitHub Issues template
- Terraform
- References to `dev` in the documentation or code comments

To merge these code changes:

- [x] [Rename](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch#renaming-a-branch) the `dev` branch to `main` in the repository's settings. Renaming instead of creating a new branch avoids having to set up branch protection rules again for `main`.
- [x] Create the `main` environment in GitHub’s Environments settings page following the same configuration as the `dev` environment
  - [ ] Set `AZURE_WEBAPP_PUBLISH_PROFILE`
  - [ ] Set `AZURE_WEBAPP_NAME`
- [ ] Merge this PR since GitHub will automatically [update the base branch of this PR to `main`](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch#about-renaming-branches).

## Notes

These may be helpful when reviewing [workspace.py](https://github.com/cal-itp/benefits/pull/2202/files#diff-33a8fda4d235c86352a2629618f4d0829b626df6084edef709ad22d97e8f80fc):

- `REASON` = `Build.Reason`: The event that caused the build to run.
- `OTHER_SOURCE` = `System.PullRequest.SourceBranch`: The branch that is being reviewed in a pull request.
- `INDIVIDUAL_SOURCE` = `Build.SourceBranchName`: The name of the branch in the triggering repo the build was queued for.
- `TARGET` = `System.PullRequest.TargetBranch`: The branch that is the target of a pull request.

### DIFF
--- a/terraform/pipeline/workspace.py
+++ b/terraform/pipeline/workspace.py
@@ -5,29 +5,30 @@ import os
 import sys
 
 REASON = os.environ["REASON"]
+
 # the name of the variable that Azure Pipelines uses for the source branch
 # depends on the type of run, so need to check both
 SOURCE = os.environ.get("OTHER_SOURCE") or os.environ["INDIVIDUAL_SOURCE"]
+
 TARGET = os.environ["TARGET"]
 
-# the branches that correspond to environments
-ENV_BRANCHES = ["dev", "test", "prod"]
+# branch to workspace mapping
+WORKSPACES = {"main": "dev", "test": "test", "prod": "default"}
+# workspace to service connection mapping
+SERVICE_CONNECTIONS = {"dev": "Development", "test": "Development", "default": "Production"}
 
-if REASON == "PullRequest" and TARGET in ENV_BRANCHES:
+if REASON == "PullRequest" and TARGET in WORKSPACES:
     # it's a pull request against one of the environment branches, so use the
     # target branch
-    environment = TARGET
-elif REASON in ["IndividualCI", "Manual"] and SOURCE in ENV_BRANCHES:
+    workspace = WORKSPACES[TARGET]
+elif REASON in ["IndividualCI", "Manual"] and SOURCE in WORKSPACES:
     # it's being run on one of the environment branches, so use that
-    environment = SOURCE
+    workspace = WORKSPACES[SOURCE]
 else:
     # default to running against dev
-    environment = "dev"
+    workspace = "dev"
 
-# matching logic in ../init.sh
-workspace = "default" if environment == "prod" else environment
-
-service_connection = "Production" if environment == "prod" else "Development"
+service_connection = SERVICE_CONNECTIONS[workspace]
 
 # just for troubleshooting
 if TARGET is not None:
--- a/.github/ISSUE_TEMPLATE/release.yml
+++ b/.github/ISSUE_TEMPLATE/release.yml
@@ -55,8 +55,8 @@ body:
 
         Reference the diagram and discussion on [the release process docs](https://docs.calitp.org/benefits/deployment/release/).
 
-        * `Regular` release: propagates from `dev` to `test`, and then `test` to `prod`. Only possible if `dev` is ready to deploy.
-        * `Hotfix` release: fix is merged into `prod`. After release, fix is propagated to `dev` and `test`.
+        * `Regular` release: propagates from `main` to `test`, and then `test` to `prod`. Only possible if `main` is ready to deploy.
+        * `Hotfix` release: fix is merged into `prod`. After release, fix is propagated to `main` and `test`.
   - type: dropdown
     id: type
     attributes:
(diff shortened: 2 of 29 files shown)


.github/workflows/labeler-deploy-main.yml
```diff
@@ -12,10 +12,10 @@ on:
...
 jobs:
-  label-deployment-dev:
+  label-deployment-main:
     runs-on: ubuntu-latest
     steps:
       - name: add-label
         uses: andymckay/labeler@master
         with:
-          add-labels: "deployment-dev"
+          add-labels: "deployment-main"
```
thekaveman, line 21: Let's actually keep this label as `deployment-dev` and the workflow name / job name the same. It is a marking a deployment to the `dev` environment.

But we do need to change the branch as you've done here.

terraform/pipeline/workspace.py
```diff
@@ -5,18 +5,20 @@
...
+
 # the name of the variable that Azure Pipelines uses for the source branch
 # depends on the type of run, so need to check both
 SOURCE = os.environ.get("OTHER_SOURCE") or os.environ["INDIVIDUAL_SOURCE"]
+
 TARGET = os.environ["TARGET"]
 
 # the branches that correspond to environments
-ENV_BRANCHES = ["dev", "test", "prod"]
+ENV_BRANCHES = {"main": "dev", "test": "test", "prod": "prod"}
```
thekaveman, line 16: This is great to use a dictionary here instead of a list, couple of comments:

* The `key` of the items is the branch name; the `value` of the items is the workspace/environment name
* This allows us to have a different workspace name from the branch name
* It also means in the code, there's no need to have both an `environment` and `workspace` variable, since those were both used to help make the distinction between branch name and workspace name, that is now encoded in the dict

So a couple changes are still needed:

* Rename `environment` to `workspace`
* In both line 21 and line 24, use the dict to get the value for the `TARGET` or `SOURCE` branch
* Delete line 30, since `workspace` is now set appropriately
* Update line 32 `service_connection` to check `workspace` instead of `environment`
  * Further idea, create a new dict mapping workspace name to service connection name and use a lookup

I think this will really help simplify the code here.

thekaveman (changes requested):
Looking really good :+1: I have just a couple of comments on a few different areas, we can talk through everything at our pairing session.

terraform/pipeline/workspace.py
lalver1: Thanks for the comments! I was still not completely clear about how the workspace/environment was being used but this explanation is really helpful, I think I got it now. Only `workspace` and `service_connection` are actually passed back to `azure-pipelines.yml` in `env_select` so these changes do make the code simpler 👍

angela-tran: The `prod` branch should map to the `default` workspace

terraform/pipeline/workspace.py
```diff
@@ -5,29 +5,32 @@
...
+    workspace = ENV_BRANCHES[SOURCE]
 else:
     # default to running against dev
-    environment = "dev"
+    workspace = "dev"
 
 # matching logic in ../init.sh
-workspace = "default" if environment == "prod" else environment
+# workspace to service connection mapping
+service_connections = {"dev": "Development", "test": "Development", "prod": "Production"}
```
angela-tran, line 31: I suggest replacing `prod` with `default` here (assuming the `ENV_BRANCHES` dict is updated to map to the `prod` branch to the `default` workspace)

terraform/pipeline/workspace.py
thekaveman: Here's a reminder of the mapping: https://docs.calitp.org/benefits/deployment/infrastructure/#environments

angela-tran: Maybe we rename `ENV_BRANCHES` to `WORKSPACES`?

terraform/pipeline/workspace.py
lalver1: Ah that's true, thank you!

terraform/pipeline/workspace.py
lalver1: Thanks for catching that the `prod` branch should map to the `default` workspace!
I renamed `ENV_BRANCHES` to `WORKSPACES` to match the naming pattern of `service_connections` in line 31 but we can change the naming if we find something more descriptive.

terraform/pipeline/workspace.py
```diff
@@ -5,29 +5,32 @@
...
+    workspace = WORKSPACES[SOURCE]
 else:
     # default to running against dev
-    environment = "dev"
+    workspace = "dev"
 
 # matching logic in ../init.sh
-workspace = "default" if environment == "prod" else environment
+# workspace to service connection mapping
+service_connections = {"dev": "Development", "test": "Development", "default": "Production"}
```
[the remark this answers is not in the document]
lalver1: Yep, that makes sense, I'll make that change 👍 
I'll also remove the `# matching logic in ../init.sh` comment since the `environment` to `workspace` mapping is in `../init.sh` and it looks like we don't need to mention it here anymore.

thekaveman (approved):

Issue #2189, opened 2024-07-01
Configure sample California State Transit agency in `dev` and `test`


thekaveman:
We've reached a point of stability in the Admin interface, and also a point where we need to be able to demo and test in-progress enrollment flows that don't necessarily have a representative agency (yet!).

All along, our designs have included a fictitious agency **CST** (or the California State Transit agency) used for mocking, spacing, etc. 

This task is about configuring the app in both `dev` and `test` with a `TransitAgency` instance and the related data for CST.

## Acceptance Criteria

Code:
- [x] #2197
- [x] #2205 
- [x] #2209 

`dev` and `test` configuration:

- [x] New groups are created in QA
- [x] New `EligibilityType` instances are created
  - [x] Older Adult: use CST success template
  - [x] CalFresh Cardholder: use CST success template
  - [x] US Veteran: use CST success template
  - [x] Agency Card: use CST Agency Card success template
- [x] New `EligibilityVerifier` instances are created
  - [x] Age claims via Login.gov
  - [x] CalFresh claims via CA Department of Social Services
  - [x] Veteran claims via VA.gov
  - [x] Eligibility server verifier (copy from MST)
- [x] A new `PaymentProcessor` instance is created
  - [x] Using `cst` sample agency Back Office credentials
- [x] A new `TransitAgency` instance is created
  - [x] `slug`: cst
  - [x] `short name`: CST
  - [x] `long name`: California State Transit
  - [x] `agency id`: cst
  - [x] `info_url`: https://www.agency-website.com
  - [x] `phone`: 1-800-555-5555
  - [x] `active`: true

thekaveman:
This is done!

https://cal-itp.slack.com/archives/C037Y3UE71P/p1720652348690329

PR #2205, opened 2024-07-09
Fix: CST templates

thekaveman:
Quick follow-up to #2197, part of #2189

I thought we could get away with reusing the `MSTCourtesyCard` class for CST's agency card, but I forgot that some of the copy comes from that form class. So this PR adds a new class for CST.

_Note that in `dev` and `test`, CST Agency Card verification will use the corresponding MST Courtesy Card verification server._

Also fixed a small translation issue on the Agency Card enrollment success for CST.


lalver1 (approved):
I tested locally and it looked good 👍

thekaveman:
Sorry, I added some quick unit tests for the new form.

lalver1 (approved):
Ah I had forgot about `test_forms.py`. Looks good!

PR #2207, opened 2024-07-10
Fix: Eligibility Index - Clicking on modal backdrop should not click radio button

### LINKED ISSUE #1526: Modal dialog link in radio button has inconsistent select behavior
There's some inconsistent behavior with closing the modal and what happens to the radio button.

See this GIF where I close the modal by:
 - clicking the X
 - pressing the Esc key
 - clicking off the modal

In the last case, the radio button gets selected.
benefits-login-gov-help-modal

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/issues/1524#issuecomment-1633250362_
            
### Additional context
An attempt to fix this was made in #1524 by consuming the click event, but because the modal content has clickable elements, we can't use that solution.

machikoyasuda:
closes #1526 

## What does this PR do

Clicking on a modal's backdrop should close the modal, and should not trigger any other changes in the DOM. Currently, this works fine on all the pages with a modal, like the home page, with the exception of the Eligibility Index page. On this page, there are radio buttons *underneath* the modal backdrop. A backdrop click event will bubble down into a radio button click event as well. I wrote code that looks for a click on `.modal`, and stops the bubbling of events - if the click is *not* clicking on the Login.gov help link (or any other future link).

I added this code to `eligibility/index.html` instead of the modal includes because I want to isolate this code to *just* this form for now. I don't see this behavior affecting the home page, for example, so I don't want to add the code to all modals.

GIF of the bug by Angela:

benefits-login-gov-help-modal

## How to test:

Test in isolation: https://codepen.io/machikoyasuda/pen/vYqENKx

Test in app: Go to the Eligibility Index page, and try opening a modal with your keyboard. Use all 4 methods to close a modal and the behavior should be the same: click on the X, use <kbd>enter</kbd> on the X, type <kbd>Escape</kbd>, and finally, try clicking on the backdrop. No radio button should be selected. (Unless of course, you have already selected a radio button, it should remain selected. The selected button should not change.) Then, try clicking on the Login.gov Help Page link and make sure that works.


thekaveman (approved):
Tested locally, looks great! Nice simple fix.

angela-tran (approved):
🙌

PR #2213, opened 2024-07-11
Refactor: implement tag-based deployment

### LINKED ISSUE #1527: Consider a "Deploy on Release" model

lalver1:
This PR implements the second step of code changes needed to transition the benefits repository to tag-based deployments as described in https://github.com/cal-itp/benefits/issues/1527. 

The code changes are associated with:

- GitHub workflows
- Azure DevOps/Terraform pipeline
- Python package versioning using `git` metadata

The last step needed to close #1527, will be to remove the `test` and `prod` branches and remove references to them in all the GitHub workflows. 

## Acceptance Criteria

- [ ] Merging to `main` deploys to the `dev` environment
- [ ] Pushing a pre-release tag deploys to the `test` environment
- [ ] Pushing a release tag deploys to the `prod` environment and triggers the release job
- [ ] The conditional `environment`, `dev` and `prod` Azure jobs (`terraform` stage) run correctly
- [ ] Python package versioning uses `git` metadata
- [ ] There is a guarantee that the required checks have passed on the commit to be deployed

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -12,10 +12,3 @@ RUN pip install -e .[dev,test]
 # docs requirements are in a separate file for the GitHub Action
 COPY docs/requirements.txt docs/requirements.txt
 RUN pip install -r docs/requirements.txt
-
-# install pre-commit environments in throwaway Git repository
-# https://stackoverflow.com/a/68758943
-COPY .pre-commit-config.yaml .
-RUN git init . && \
-    pre-commit install-hooks && \
-    rm -rf .git
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,18 +5,44 @@ on:
   push:
     branches:
       - main
-      - test
-      - prod
+    tags:
+      # pre-release tag
+      - "202[3-9].[0-9][0-9].[0-9]+-rc[0-9]+"
+      # release tags
+      - "202[3-9].[0-9][0-9].[0-9]+"
 
 defaults:
   run:
     shell: bash
 
+concurrency:
+  # this ternary operator like expression gives us the name of the deployment environment (see https://docs.github.com/en/actions/learn-github-actions/expressions#example)
+  group: ${{ github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}
+  cancel-in-progress: true
+
 jobs:
+  tests-ui:
+    uses: ./.github/workflows/tests-ui.yml
+    if: github.ref_type == 'tag'
+
+  tests-pytest:
+    uses: ./.github/workflows/tests-pytest.yml
+    if: github.ref_type == 'tag'
+
+  tests-cypress:
+    uses: ./.github/workflows/tests-cypress.yml
+    if: github.ref_type == 'tag'
+
+  check-migrations-and-messages:
+    uses: ./.github/workflows/check-migrations-and-messages.yml
+    if: github.ref_type == 'tag'
+
   deploy:
     runs-on: ubuntu-latest
-    environment: ${{ github.ref_name }}
-    concurrency: ${{ github.ref_name }}
+    needs:
+      [tests-ui, tests-pytest, tests-cypress, check-migrations-and-messages]
+    if: (!cancelled())
+    environment: ${{ github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}
 
     steps:
       - name: Checkout
@@ -69,3 +95,23 @@ jobs:
           app-name: ${{ vars.AZURE_WEBAPP_NAME }}
           images: ghcr.io/${{ github.repository }}:${{ github.sha }}
           publish-profile: ${{ secrets.AZURE_WEBAPP_PUBLISH_PROFILE }}
+
+  release:
+    needs: deploy
+    if: ${{ github.ref_type == 'tag' && !contains(github.ref, '-rc') }}
+    runs-on: ubuntu-latest
+    permissions:
+      # https://github.com/softprops/action-gh-release#permissions
+      contents: write
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Release
+        uses: softprops/action-gh-release@v2
+        with:
+          prerelease: false
+          generate_release_notes: true
--- a/terraform/azure-pipelines.yml
+++ b/terraform/azure-pipelines.yml
@@ -4,8 +4,8 @@ trigger:
   branches:
     include:
       - main
-      - test
-      - prod
+      - refs/tags/20??.??.?*-rc?*
+      - refs/tags/20??.??.?*
   # only run for changes to Terraform files
   paths:
     include:
@@ -26,8 +26,14 @@ stages:
             value: $[variables['Build.SourceBranchName']]
           - name: TARGET
             value: $[variables['System.PullRequest.TargetBranch']]
+          - name: IS_TAG
+            value: $[startsWith(variables['Build.SourceBranch'], 'refs/tags/')]
         steps:
-          - bash: python terraform/pipeline/workspace.py
+          - bash: |
+              python terraform/pipeline/workspace.py
+
+              TAG_TYPE=$(python terraform/pipeline/tag.py)
+              echo "##vso[task.setvariable variable=tag_type;isOutput=true]$TAG_TYPE"
             displayName: Set environment-related variables
             # save the values
             # https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#use-outputs-in-a-different-job
@@ -39,18 +45,22 @@ stages:
         condition: eq(dependencies.environment.outputs['env_select.service_connection'], 'Development')
         variables:
           workspace: $[ dependencies.environment.outputs['env_select.workspace'] ]
+          tag_type: $[ dependencies.environment.outputs['env_select.tag_type'] ]
         steps:
           - template: pipeline/deploy.yml
             parameters:
               service_connection: Development
               workspace: $(workspace)
+              tag_type: $(tag_type)
       - job: prod
         dependsOn: environment
         condition: eq(dependencies.environment.outputs['env_select.service_connection'], 'Production')
         variables:
           workspace: $[ dependencies.environment.outputs['env_select.workspace'] ]
+          tag_type: $[ dependencies.environment.outputs['env_select.tag_type'] ]
         steps:
           - template: pipeline/deploy.yml
             parameters:
               service_connection: Production
               workspace: $(workspace)
+              tag_type: $(tag_type)
--- a/terraform/pipeline/deploy.yml
+++ b/terraform/pipeline/deploy.yml
@@ -3,6 +3,8 @@ parameters:
     type: string
   - name: workspace
     type: string
+  - name: tag_type
+    type: string
 
 steps:
   # https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformInstaller#readme
@@ -48,7 +50,13 @@ steps:
       # service connection
       environmentServiceNameAzureRM: "${{ parameters.service_connection }}"
     # the plan is done as part of the apply (below), so don't bother doing it twice
-    condition: notIn(variables['Build.SourceBranchName'], 'main', 'test', 'prod')
+    condition: |
+      ${{ and(
+            ne(variables['Build.SourceBranchName'], 'main'),
+            ne(parameters.tag_type, 'test'),
+            ne(parameters.tag_type, 'prod')
+          )
+      }}
   - task: TerraformTaskV3@3
     displayName: Terraform apply
     inputs:
@@ -59,5 +67,11 @@ steps:
       workingDirectory: "$(System.DefaultWorkingDirectory)/terraform"
       # service connection
       environmentServiceNameAzureRM: "${{ parameters.service_connection }}"
-    # only run on certain branches
-    condition: in(variables['Build.SourceBranchName'], 'main', 'test', 'prod')
+    # only run on main branch OR if it's a tag for test or prod
+    condition: |
+      ${{ or(
+            eq(variables['Build.SourceBranchName'], 'main'),
+            eq(parameters.tag_type, 'test'),
+            eq(parameters.tag_type, 'prod')
+          )
+      }}
--- a/.devcontainer/postAttach.sh
+++ b/.devcontainer/postAttach.sh
@@ -4,4 +4,9 @@ set -eu
 # initialize pre-commit
 
 git config --global --add safe.directory /calitp/app
-pre-commit install --overwrite
+
+# initialize hook environments
+pre-commit install --install-hooks --overwrite
+
+# manage commit-msg hooks
+pre-commit install --hook-type commit-msg
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,3 @@
-.git/
 .github/
 .vscode/
 **/__pycache__/
--- a/.github/workflows/check-migrations-and-messages.yml
+++ b/.github/workflows/check-migrations-and-messages.yml
@@ -1,5 +1,5 @@
 name: Check for up-to-date Django migrations and messages
-on: [push, pull_request]
+on: [push, pull_request, workflow_call]
 
 jobs:
   check-migrations-and-messages:
--- a/.github/workflows/tests-cypress.yml
+++ b/.github/workflows/tests-cypress.yml
@@ -1,6 +1,6 @@
 name: Cypress tests
 
-on: push
+on: [push, workflow_call]
 
 defaults:
   run:
--- a/.github/workflows/tests-pytest.yml
+++ b/.github/workflows/tests-pytest.yml
@@ -1,6 +1,6 @@
 name: Pytest
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_call]
 
 jobs:
   pytest:
--- a/.github/workflows/tests-ui.yml
+++ b/.github/workflows/tests-ui.yml
@@ -1,6 +1,7 @@
 name: UI & a11y tests
 
 on:
+  workflow_call:
   workflow_dispatch:
   pull_request:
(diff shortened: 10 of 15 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Create GitHub release
```
thekaveman, line 1: I'd like this workflow to be steps in the `deploy` workflow, or else called specifically by the `deploy` workflow (rather than `on.push` etc.). 

Otherwise, we get the GitHub release published well before the code is actually deployed, which feels a little out of order.

thekaveman (changes requested):

.github/workflows/release.yml
lalver1: Thanks! I agree, the `release` job should run after the `deploy` job is finished, so I'll add some steps in `deploy.yml` to achieve this. Oh and `release` should only run on an actual release, not a release candidate so I'll be careful about keeping this behavior unchanged.

lalver1:
Now `deploy` and `release` are two sequential jobs in the `Deploy` workflow. I grouped the two `release` steps into the `release` job to set the `if` condition and the `permissions` just once at the job level instead of twice if they had been at the step level. 

Unfortunately I couldn't set `${{ github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}` as an environment variable using `env` in the workflow and avoid duplicating writing it out since the `env` [context is not available](https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/contexts#context-availability) at the `job` level, only at the `step` level.

With this implementation we'll have the following behavior in `main`:

- push a commit:
  - set `environment = main`
  - push image to GHCR
  - deploy to https://dev-benefits.calitp.org/
- push an rc tag:
  - set `environment = test`
  - push image to GHCR
  - deploy to https://test-benefits.calitp.org/
- push a tag:
  - set `environment = prod`
  - push image to GHCR
  - deploy to https://benefits.calitp.org/
  - if `deploy` succeeded and `deployment environment == prod` , make a GitHub release

thekaveman (changes requested):
This is looking really good so far! I think the workflow updates all look correct.

But we still need the updates to the Azure DevOps/Terraform pipeline, which is looking at the old `test` and `prod` branches etc. 

Take a look at Server for how those pipelines work with tag-based.

.github/workflows/deploy.yml
```diff
@@ -69,3 +76,23 @@ jobs:
           app-name: ${{ vars.AZURE_WEBAPP_NAME }}
           images: ghcr.io/${{ github.repository }}:${{ github.sha }}
           publish-profile: ${{ secrets.AZURE_WEBAPP_PUBLISH_PROFILE }}
+
+  release:
+    needs: deploy
+    if: ${{ (github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod') == "prod" }}
```
thekaveman, line 82: I get that you're just re-using the same expression here, which is fine... but really all this check cares about is if it's a production tag, we don't need the env name or anything else here.

Maybe this could be simplified to something like `${{ github.ref_type == "tag" && !contains(github.ref, "-rc") }}`?

lalver1: That's true, this expression works better! 👍

thekaveman (dismissed):
Also, don't forget that tag-based deployment means we get the version number from the tag instead of hardcoding it in the source.

That means changes at least to [`pyproject.toml`](https://github.com/cal-itp/benefits/blob/main/pyproject.toml#L3)

See this PR for how it was done in Server:

- https://github.com/cal-itp/eligibility-server/pull/382

thekaveman:
@lalver1 you should be able to rebase on `main` now to get the Cypress tests passing.

Fixed by #2214

lalver1:
Thanks @thekaveman, trying it now!

.github/workflows/deploy.yml
```diff
@@ -69,3 +76,23 @@ jobs:
           app-name: ${{ vars.AZURE_WEBAPP_NAME }}
           images: ghcr.io/${{ github.repository }}:${{ github.sha }}
           publish-profile: ${{ secrets.AZURE_WEBAPP_PUBLISH_PROFILE }}
+
+  release:
+    needs: deploy
+    if: ${{ github.ref_type == "tag" && !contains(github.ref, "-rc") }}
```
thekaveman, line 82: I think these have to be single quotes :roll_eyes:

lalver1:
Thanks for the review, I had overlooked the Azure DevOps/Terraform pipeline and the dynamic versioning part. I think we should now have all the parts ready for tag-based deployment.

The only thing I don't feel very confident about is if we need `COPY .git .git`in the [dev container's Dockerfile](https://github.com/cal-itp/benefits/pull/2213/commits/48046721c0eb9beaff628e8e1eec2a5af2b69ee6#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6R10) since we already have this line in the app container's Dockerfile. If indeed we don't need it, I guess including it has no impact in how it works, it would just make the code cleaner.

terraform/pipeline/deploy.yml
```diff
@@ -48,7 +50,7 @@ steps:
       # service connection
       environmentServiceNameAzureRM: "${{ parameters.service_connection }}"
     # the plan is done as part of the apply (below), so don't bother doing it twice
-    condition: notIn(variables['Build.SourceBranchName'], 'main', 'test', 'prod')
+    condition: notIn(variables['Build.SourceBranchName'], 'main')
```
thekaveman, line 53: This is correct to remove the old branch names here, but we still want to skip the `plan` under the same conditions as before. Namely, when we are doing a `test` or `prod` deployment. 

I think you can use a similar conditional as below in the `apply`.

thekaveman (changes requested):
Getting really close @lalver1, this is looking great.

Just a few more small notes and I think it's ready.

thekaveman:
> The only thing I don't feel very confident about is if we need COPY .git .gitin the [dev container's Dockerfile](https://github.com/cal-itp/benefits/pull/2213/commits/48046721c0eb9beaff628e8e1eec2a5af2b69ee6#diff-13bd9d7a30bf46656bc81f1ad5b908a627f9247be3f7d76df862b0578b534fc6R10) since we already have this line in the app container's Dockerfile. If indeed we don't need it, I guess including it has no impact in how it works, it would just make the code cleaner.

I was wondering the same thing. I see we do it this way in Server, so there must have been a reason. No worries, it is copied over immediately once we launch the devcontainer, so I think you can just leave it as-is.

.devcontainer/Dockerfile
```diff
@@ -6,6 +6,9 @@ USER root
 RUN curl -sL https://aka.ms/InstallAzureCLIDeb | bash
 USER $USER
 
+# copy files needed for version metadata
+COPY .git .git
```
thekaveman, line 10: On second thought (and after some testing), let's remove this -- I removed it locally and tried rebuilding the devcontainer after the appcontainer, and got an error:

```log
> [dev 8/8] RUN git init . &&     pre-commit install-hooks &&     rm -rf .git:
0.101 error: could not lock config file /calitp/app/.git/config: Permission denied
0.101 fatal: could not set 'core.repositoryformatversion' to '0'
```

This is a problem with the last few lines of this file, which should also be removed -- since the `.git` directory now exists, we can't call `git init` and do the temporary pre-commit install.

```
# install pre-commit environments in throwaway Git repository
# https://stackoverflow.com/a/68758943
COPY .pre-commit-config.yaml .
RUN git init . && \
    pre-commit install-hooks && \
    rm -rf .git
```

Even trying to install the hooks without `git init` fails due to some Docker layers weirdness... in any case, this matches how we have it in Server.

thekaveman (changes requested):

.devcontainer/Dockerfile
thekaveman: In fact, take a look at what calitp.org does in `postAttach`, I think we should apply this here and in Server: https://github.com/cal-itp/calitp.org/blob/main/.devcontainer/postAttach.sh

.github/workflows/deploy.yml
lalver1: Ah that's true, thanks for catching that!

.devcontainer/Dockerfile
lalver1: Yep, I verified the behavior you described and made the suggested changes. The dev container loaded without a problem so we can make this change to eligibility-server as you mentioned.

terraform/pipeline/deploy.yml
lalver1: Yep, that makes sense. I used `parameters.tag_type` to determine if it is a `test` or `prod` deployment similar to the `apply` step and updated the condition. The condition may look a bit odd (maybe someone would expect keeping the `or` and just changing the `eq`s to `ne`s) but as far as I can tell the `and` is required for this step to run only for the cases when `apply` will not run. So this condition ensures that `# the plan is done as part of the apply (below), so don't bother doing it twice` following the comment for this `task`.

thekaveman (approved):
Nice work :+1: I'd like @angela-tran to take a look at this before you merge.

lalver1:
Thanks @thekaveman! Yep, I won't merge until @angela-tran  has had a chance to look at it.

angela-tran (dismissed):
This looks good to me! 

The only thing (which I don't think needs to block this PR) is in the GitHub deploy workflow, we aren't running the tests before deploying. Is this being done some other way?

thekaveman (changes requested):
@angela-tran 

> The only thing (which I don't think needs to block this PR) is in the GitHub deploy workflow, we aren't running the tests before deploying. Is this being done some other way?

Good call, let's get this in as part of this PR.

Can we run the migrations and messages check as well?

@lalver1 see https://github.com/cal-itp/eligibility-server/blob/main/.github/workflows/docker-publish.yml#L24

terraform/azure-pipelines.yml
```diff
@@ -26,8 +26,14 @@ stages:
             value: $[variables['Build.SourceBranchName']]
           - name: TARGET
             value: $[variables['System.PullRequest.TargetBranch']]
+          - name: IS_TAG
+            value: $[startsWith(variables['Build.SourceBranch'], 'refs/tags/')]
         steps:
-          - bash: python terraform/pipeline/workspace.py
+          - bash: |
+              python terraform/pipeline/workspace.py
```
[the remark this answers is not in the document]
thekaveman: I think it's probably because this line didn't change, in that the `workspace` is still set as an output variable by `workspace.py`: https://github.com/cal-itp/benefits/pull/2213/files/f4304d09dd1bf08a2797f165ba7d603cd0743a48#diff-33a8fda4d235c86352a2629618f4d0829b626df6084edef709ad22d97e8f80fcR52

terraform/azure-pipelines.yml
```diff
@@ -26,8 +26,14 @@ stages:
...
+
+              TAG_TYPE=$(python terraform/pipeline/tag.py)
+              echo "##vso[task.setvariable variable=tag_type;isOutput=true]$TAG_TYPE"
```
thekaveman, line 36: These lines could be normalized to move the `TAG_TYPE` and `echo` in `tag.py` similar to how it works in `workspace.py`, but not a priority for this PR.

lalver1: I agree. Great, I'll create an issue to come back to this later.

lalver1:
Thanks @angela-tran, that's a good point about the tests! I'm thinking about reusing these workflows:

- `tests-ui`
- `tests-pytest`
- `tests-cypress`
- `check-migrations-and-messages`

I don't think we need to run `codeql` and `check-api` 🤔. If this makes sense, I'll add `on: workflow_call` to those workflows that don't already have it, and will call them from `deploy.yml` using `uses:` before the `deploy` job.

Oh and one more question, I'm thinking about setting `if: github.ref_type == 'tag'` when calling these tests, since we already run them as part of the branch protection status checks when `ref_type` is `branch` and we probably don't want to run them again?

angela-tran:
> Thanks @angela-tran, that's a good point about the tests! I'm thinking about reusing these workflows:
> 
>   * `tests-ui`
>   * `tests-pytest`
>   * `tests-cypress`
>   * `check-migrations-and-messages`
> 
> 
> I don't think we need to run `codeql` and `check-api` 🤔. If this makes sense, I'll add `on: workflow_call` to those workflows that don't already have it, and will call them from `deploy.yml` using `uses:` before the `deploy` job.

That list looks good to me. And we'll want `deploy` to depend on the tests succeeding.

 
> Oh and one more question, I'm thinking about setting `if: github.ref_type == 'tag'` when calling these tests, since we already run them as part of the branch protection status checks when `ref_type` is `branch` and we probably don't want to run them again?

Yep, that's definitely needed. [`eligibility-server` does it too](https://github.com/cal-itp/eligibility-server/blob/main/.github/workflows/docker-publish.yml#L26) in its GitHub workflow.

lalver1:
@angela-tran @thekaveman I added the tests to the `deploy` workflow and [required them to successfully complete](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) before the `deploy` job starts.

I did notice that `eligibility-server` has these [lines](https://github.com/cal-itp/eligibility-server/blob/main/.github/workflows/docker-publish.yml#L30-L32) in its `docker-publish` workflow
```
deploy:
    needs: test
    # !cancelled() is needed because if the whole workflow was cancelled, we don't want this job to run.
    # (github.ref_type != 'tag' || needs.test.result == 'success') is needed because if `test` did run, we only want this to run if `test` succeeded.
    if: (!cancelled() && (github.ref_type != 'tag' || needs.test.result == 'success'))
```

and I'm wondering if I should add `if: (!cancelled()` to the benefits `deploy` job since I guess we could manually want to cancel a deploy before it finishes running. I think the second part `&& (github.ref_type != 'tag' || needs.test.result == 'success')` may not be needed because 

```
needs:
      [tests-ui, tests-pytest, tests-cypress, check-migrations-and-messages]
```
covers this case?

angela-tran:
> I'm wondering if I should add `if: (!cancelled()` to the benefits `deploy` job since I guess we could manually want to cancel a deploy before it finishes running. I think the second part `&& (github.ref_type != 'tag' || needs.test.result == 'success')` may not be needed because
> 
> ```
> needs:
>       [tests-ui, tests-pytest, tests-cypress, check-migrations-and-messages]
> ```
> 
> covers this case?

Great catch @lalver1 -- yes, I think we should add `if: (!cancelled())`. 

I took a closer look at that second part from `eligibility-server`, and I think you're right that it's not needed. GitHub docs on [`needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) confirms that the jobs listed must complete successfully before the dependent job will run, so it would be redundant to include this condition. I think it's a remnant from an [older version](https://github.com/cal-itp/eligibility-server/pull/329#discussion_r1341766075) of the expression.

Long story short 😄, we don't need the second part here, and it could also be removed in `eligibility-server`.

lalver1:
Thanks @angela-tran! I added `if: (!cancelled())` to the `deploy` job and will make a ticket to remove the second part of the expression in `eligibility-server`; I think this one is ready for a review 🙂

thekaveman (approved):
I think this is ready to go :tada:

PR #2224, opened 2024-07-16
Feat: Cal-ITP staff can view configuration values

### LINKED ISSUE #2199: Admin interface: Cal-ITP Benefits staff users can view existing configuration
Implementation:
 - [x] Add a data migration that gives the `Cal-ITP` group permission to `view` the models listed under Acceptance Criteria
 - [x] Define `ModelAdmin`s for each model and implement [`get_exclude`](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_exclude) to hide fields from users in the `Cal-ITP` group according to the spreadsheet

## Acceptance Criteria

- [x] Cal-ITP Benefits staff users can view existing `EligibilityType`s, `EligibilityVerifier`s, `AuthProvider`s, `PaymentProcessor`s, and `TransitAgency`s
- [x] Cal-ITP Benefits staff users can view fields on those existing objects if the field is listed as readable or writable for Cal-ITP in the [Benefits admin configuration spreadsheet](https://github.com/cal-itp/benefits/issues/2001#issue-2226668714)

## Additional context

Depends on #2198 
Part of #2001

angela-tran:
Closes #2199 

### Screenshots

This is what the user sees upon logging in:

image

When viewing model objects, the user sees a read-only view of fields that are not in the `exclude` list for each model.

Here is an example using the `TransitAgency` model:

image

### DIFF
--- a/benefits/core/admin.py
+++ b/benefits/core/admin.py
@@ -24,7 +24,7 @@ admin.site.register(models.PemData)
 @admin.register(models.AuthProvider)
 class AuthProviderAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
-        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+        if not request.user.is_superuser:
             return ["client_id_secret_name"]
         else:
             return super().get_exclude(request, obj)
@@ -33,7 +33,7 @@ class AuthProviderAdmin(admin.ModelAdmin):
 @admin.register(models.EligibilityType)
 class EligibilityTypeAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
-        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+        if not request.user.is_superuser:
             return []
         else:
             return super().get_exclude(request, obj)
@@ -42,7 +42,7 @@ class EligibilityTypeAdmin(admin.ModelAdmin):
 @admin.register(models.EligibilityVerifier)
 class SortableEligibilityVerifierAdmin(SortableAdminMixin, admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
-        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+        if not request.user.is_superuser:
             return [
                 "api_auth_header",
                 "api_auth_key_secret_name",
@@ -59,7 +59,7 @@ class SortableEligibilityVerifierAdmin(SortableAdminMixin, admin.ModelAdmin):
 @admin.register(models.PaymentProcessor)
 class PaymentProcessorAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
-        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+        if not request.user.is_superuser:
             return [
                 "client_id",
                 "client_secret_name",
@@ -72,7 +72,7 @@ class PaymentProcessorAdmin(admin.ModelAdmin):
 @admin.register(models.TransitAgency)
 class TransitAgencyAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
-        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+        if not request.user.is_superuser:
             return [
                 "private_key",
                 "public_key",
--- a/benefits/core/admin.py
+++ b/benefits/core/admin.py
@@ -41,7 +41,19 @@ class EligibilityTypeAdmin(admin.ModelAdmin):
 
 @admin.register(models.EligibilityVerifier)
 class SortableEligibilityVerifierAdmin(SortableAdminMixin, admin.ModelAdmin):
-    pass
+    def get_exclude(self, request, obj=None):
+        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
+            return [
+                "api_auth_header",
+                "api_auth_key_secret_name",
+                "public_key",
+                "jwe_cek_enc",
+                "jwe_encryption_alg",
+                "jws_signing_alg",
+                "form_class",
+            ]
+        else:
+            return super().get_exclude(request, obj)
 
 
 @admin.register(models.PaymentProcessor)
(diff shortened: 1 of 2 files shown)


angela-tran:
> ## Coverage report
> 
>  
> Click to see where and how coverage changed

The code changes in `admin.py` are exclusively using Django features, i.e. subclassing `ModelAdmin`, implementing `get_exclude` and `get_readonly_fields`, so I don't think we need to write tests to assert that Django works.

benefits/core/admin.py
```diff
@@ -17,21 +17,69 @@
...
-    models.PemData,
-    models.TransitAgency,
-]:
-    logger.debug(f"Register {model.__name__}")
-    admin.site.register(model)
+
+@admin.register(models.AuthProvider)
+class AuthProviderAdmin(admin.ModelAdmin):
+    def get_exclude(self, request, obj=None):
+        if request.user.groups.contains(Group.objects.get(name=STAFF_GROUP_NAME)):
```
thekaveman, line 27: This check seems correct, but I wonder if it could be future-proofed just slightly by instead checking if the user is _not_  a superuser, and if not returning these exclude lists.

This will cover the Staff use-case and also the eventual Transit Agency use-case for these fields that should only be visible to superusers.

angela-tran: That will work, and I'm ok with changing the check to that. It's easy enough to change later if needed.

thekaveman: Yes we can always add more conditions as needed. I think checking the superuser status is a little more explicit for what we are doing here: these are fields that should not be visible to _most_ users.

angela-tran:
> ## Coverage report
> 
>  
> Click to see where and how coverage changed
> 
> File	Statements	Missing	Coverage	Coverage
> (new stmts)	Lines missing
>   benefits/core
>   [admin.py](https://github.com/cal-itp/benefits/pull/2224/files#diff-04cabe21a2e4c17dd214c2bc58eacfc5bb09a462137b9a5d4d81239ef2110934) 	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-04cabe21a2e4c17dd214c2bc58eacfc5bb09a462137b9a5d4d81239ef2110934)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-04cabe21a2e4c17dd214c2bc58eacfc5bb09a462137b9a5d4d81239ef2110934)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-04cabe21a2e4c17dd214c2bc58eacfc5bb09a462137b9a5d4d81239ef2110934)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-04cabe21a2e4c17dd214c2bc58eacfc5bb09a462137b9a5d4d81239ef2110934)	
> Project Total 	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-4b0bf2efa3367c0072ac2bf1e234e703dc46b47aaa4fe9d3b01737b1a15752b1)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-4b0bf2efa3367c0072ac2bf1e234e703dc46b47aaa4fe9d3b01737b1a15752b1)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-4b0bf2efa3367c0072ac2bf1e234e703dc46b47aaa4fe9d3b01737b1a15752b1)	[](https://github.com/cal-itp/benefits/pull/2224/files#diff-4b0bf2efa3367c0072ac2bf1e234e703dc46b47aaa4fe9d3b01737b1a15752b1)	 
> 
> This report was generated by [python-coverage-comment-action](https://github.com/py-cov-action/python-coverage-comment-action)

Was able to exclude the `ModelAdmin`s from test coverage report: https://github.com/cal-itp/benefits/pull/2224/commits/449c8e29153f08e616a81d5eff6edbcc29f54980

benefits/core/admin.py
angela-tran: Changed to check if the user is not a superuser: https://github.com/cal-itp/benefits/pull/2224/commits/b8df7b9fe417a7a53a0307e971e1950686b6d6f0

benefits/core/admin.py
```diff
@@ -17,21 +17,69 @@
...
+    def get_exclude(self, request, obj=None):
+        if not request.user.is_superuser:
+            return [
+                "api_auth_header",
+                "api_auth_key_secret_name",
+                "public_key",
+                "jwe_cek_enc",
+                "jwe_encryption_alg",
+                "jws_signing_alg",
+                "form_class",
```
thekaveman, line 53: NB: this field is marked as `read` for Cal-ITP in the spreadsheet, but I think it makes sense to hide such a code-specific config from them.

benefits/core/admin.py
```diff
@@ -17,21 +17,69 @@
 GOOGLE_USER_INFO_URL = "https://www.googleapis.com/oauth2/v3/userinfo"
 STAFF_GROUP_NAME = "Cal-ITP"
 
+logger.debug("Register models with admin site")
+admin.site.register(models.PemData)
```
thekaveman, line 21: The spreadsheet lists the `text_secret_name` field in `PemData` as readable only by superusers.

thekaveman (changes requested):
1 change, 1 comment.

benefits/core/admin.py
angela-tran: I can update the spreadsheet to reflect this

benefits/core/admin.py
angela-tran: Oh, somehow I had the impression non-superusers wouldn't see `PemData`s at all, but you're totally right - the spreadsheet definitely says they can read everything but `text_secret_name`.

Fixed this in https://github.com/cal-itp/benefits/pull/2224/commits/3f3b37d7bf93ada92c9fc1bc628627e12bbf273b

benefits/core/admin.py
angela-tran: Updated it

image

benefits/core/admin.py
thekaveman: Oh, I see -- you had it so they couldn't see the model at all. I think that's better :sweat_smile: sorry! It doesn't make a lot of sense for non-superusers to see that model.

angela-tran: No worries! I think it might've just been the way I filtered the spreadsheet that made me think that, but I also like hiding the `PemData`s altogether better. I'll drop the commit and update the spreadsheet

angela-tran: Done and done

thekaveman (approved):
:tada:

PR #2227, opened 2024-07-17
Feat: handle OAuth errors

### LINKED ISSUE #2019: Implement error handling during authentication with the IDG for good UX
## Overview

This is a follow-up to the minor incident documented in #2009.

We want to ensure that if errors are generated during the Auth interaction with IdG they are handled appropriately by the Benefits app, and users are given a meaningful error message.

This is related to the new error page design and scoping happening for #1939

## Eligibility verification error

This error state addresses errors result after a rider has authenticated with Login.gov and the Benefits requests the CDT Identity Gateway verify eligibility for the selected transit Benefit. The fix is outside the user's control; resolution will require support from the CDT team and may take hours to resolve.

The genesis for this error message is the incident documented in this [report](https://docs.google.com/document/d/15QLfupcL6KVfT54ZtBODZk8Mtahdvy9BDcqJ40VQKvE/edit#heading=h.78n6lxqhffn). 

“The technical manifestation of this issue was that when the Benefits application sent its request to the IdG, and the user completed their sign in with [Login.gov](https://github.com/orgs/cal-itp/projects/8/views/Login.gov), the IdG attempted to sign the token response back to Benefits with an expired certificate. This led to a failure in the IdG, with the resulting 500 response back to the Benefits app, causing a failure to parse the token response, which ultimately generated the Sentry notification.”

### Acceptance criteria:

**Design criteria**

- [x] View content communicates the next step for the user.
- [x] Copy explains to the user what happened using plain language that can be comprehended at an 8th grade reading level.
- [x] Copy is translated into Spanish

**Engineering criteria**

- [x] Implement this version
image

- [x] No surprises on spacing, colors, fonts, etc -- all the same as other error pages
- [x] In `oauth.views`, any error condition related to creating or using the `oauth_client` shows this error page
- [x] Whenever this error page is shown, a Sentry event captures the exception (this will be an explicit Sentry call)
- [x] We aren't handling the `error_code` case returned from IdG in this ticket

thekaveman:
Closes #2019 

* Implements a new error screen in the `oauth` app
* Handles errors on creation/registration of the OAuth client
* Handles errors for all OAuth client method calls
* Sends a new analytics event `oauth error` in all cases
  * `operation` property indicates the specific client call: `init`, `authorize_redirect`, `authorize_access_token`, and `load_server_metadata`
  * `message` property describes the exception
* Captures an exception in Sentry in all cases

## How to test

1. Checkout the branch locally
2. Run `bin/init.sh` to get the PO file updates
3. Launch the app with `F5`
4. Select an agency or enter via their slug URL
5. Manually enter the URL in the browser: `/oauth/error`: see the error page
6. Attempt a Login.gov flow for the agency (without configuring the `AuthProvider`): see the error page
7. Misconfigure an `EligibilityVerifier` to not specify an `AuthProvider` _and_ without an Eligibility API base URL or form class and attempt to sign in with this verifier selected: see the error page

### DIFF
--- a/benefits/oauth/middleware.py
+++ b/benefits/oauth/middleware.py
@@ -1,7 +1,13 @@
 import logging
 
+from django.shortcuts import redirect
+import sentry_sdk
+
 from benefits.core import session
 from benefits.core.middleware import VerifierSessionRequired, user_error
+
+from . import analytics
+from .redirects import ROUTE_SYSTEM_ERROR
 
 
 logger = logging.getLogger(__name__)
@@ -16,8 +22,19 @@ class VerifierUsesAuthVerificationSessionRequired(VerifierSessionRequired):
             # from the base middleware class, the session didn't have a verifier
             return result
 
-        if session.verifier(request).uses_auth_verification:
+        verifier = session.verifier(request)
+
+        if verifier.uses_auth_verification:
+            # all good, the chosen verifier is configured correctly
             return None
+        elif not (verifier.api_url or verifier.form_class):
+            # the chosen verifier doesn't have Eligibility API config OR auth provider config
+            # this is likely a misconfiguration on the backend, not a user error
+            message = f"Verifier with no API or IDP config: {verifier.name} (id={verifier.id})"
+            analytics.error(request, message=message, operation=request.path)
+            sentry_sdk.capture_exception(Exception(message))
+            return redirect(ROUTE_SYSTEM_ERROR)
         else:
+            # the chosen verifier was for Eligibility API
             logger.debug("Session not configured with eligibility verifier that uses auth verification")
             return user_error(request)
--- a/benefits/locale/en/LC_MESSAGES/django.po
+++ b/benefits/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2024-07-09 10:45-0700\n"
+"POT-Creation-Date: 2024-07-17 12:05-0700\n"
 "Language: English\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
@@ -779,6 +779,17 @@ msgstr ""
 msgid "Sign out of"
 msgstr ""
 
+msgid "Service is down"
+msgstr ""
+
+msgid "There was a problem with the service that checks your identity."
+msgstr ""
+
+msgid ""
+"We’re working to solve the problem. Please wait 24 hours and try to enroll "
+"again, or contact your transit agency for help."
+msgstr ""
+
 msgid "Start over"
 msgstr ""
 
@@ -791,9 +802,6 @@ msgid ""
 msgstr ""
 
 msgid "Start Here"
-msgstr ""
-
-msgid "Service is down"
 msgstr ""
 
 msgid "Sorry! Service for this site is down."
--- a/benefits/locale/es/LC_MESSAGES/django.po
+++ b/benefits/locale/es/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2024-07-09 10:45-0700\n"
+"POT-Creation-Date: 2024-07-17 12:05-0700\n"
 "Language: Español\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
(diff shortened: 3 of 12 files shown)


angela-tran (commented):
These code changes look good to me!

I'm not sure if this is just me, but when I try to log out, the IdG doesn't redirect me back to Benefits. I end up at this screen:

image

When I try log out on https://dev-benefits.calitp.org, it redirects me back to Benefits.

Does this happen for anyone else?

thekaveman:
Hmm I'm seeing an error page now (both locally and for `dev`) just trying to sign in with IdG, so I wonder if they are working on some changes in `dev` on their side?

image

lalver1:
The code looks good and I did see the new error screen when going through a flow that uses a misconfigured `AuthProvider` (for example using an invalid `Authority`) 👍 
As I was testing I did notice these 2 use cases that maybe can show the new error screen too:

- Not selecting an AuthProvider for a flow that requires one. The middleware seems to catch this case, `DEBUG benefits.oauth.middleware:22 Session not configured with eligibility verifier that uses auth verification`, so the user sees [`200-user-error.html`](https://github.com/cal-itp/benefits/blob/main/benefits/templates/200-user-error.html). Perhaps we can instead show the new error screen [`system_error.html`](https://github.com/cal-itp/benefits/blob/feat/oauth-error/benefits/oauth/templates/oauth/system_error.html) in this case.

- Wrong scope/claim configured. If a wrong scope/claim has been configured, attempting a Login.gov flow will return an error screen generated by the `Authority`, for example, `Sorry, there was an error : Invalid scope`. Perhaps we can instead show the new error screen [`system_error.html`](https://github.com/cal-itp/benefits/blob/feat/oauth-error/benefits/oauth/templates/oauth/system_error.html) in this case.

thekaveman:
@lalver1: 
 
> * Not selecting an AuthProvider for a flow that requires one. The middleware seems to catch this case, `DEBUG benefits.oauth.middleware:22 Session not configured with eligibility verifier that uses auth verification`, so the user sees [`200-user-error.html`](https://github.com/cal-itp/benefits/blob/main/benefits/templates/200-user-error.html). Perhaps we can instead show the new error screen [`system_error.html`](https://github.com/cal-itp/benefits/blob/feat/oauth-error/benefits/oauth/templates/oauth/system_error.html) in this case.

Good call, I added a check to that middleware for a misconfigured verifier, i.e. one that has neither Eligibility API nor IDP configuration, and in that case show the new system error page: [`f5c7820`](https://github.com/cal-itp/benefits/pull/2227/commits/f5c78205e11b4eb13f21823964fecba0902d97d2)

> * Wrong scope/claim configured. If a wrong scope/claim has been configured, attempting a Login.gov flow will return an error screen generated by the `Authority`, for example, `Sorry, there was an error : Invalid scope`. Perhaps we can instead show the new error screen [`system_error.html`](https://github.com/cal-itp/benefits/blob/feat/oauth-error/benefits/oauth/templates/oauth/system_error.html) in this case.

I assume you mean this screen from IdG? 

image

While I agree it would be nice to show the user our error page here, unfortunately we have no control over this page per se. This is the IdG error page, as we have already redirected the user away from Benefits and this is the response IdG gave to the user.

That said, we could talk to CDT about the possibility of redirecting the user _back_ to our error page if this condition occurs on their side, which would take a little bit of scoping and work to get right I imagine.

thekaveman:
> > These code changes look good to me!
> > I'm not sure if this is just me, but when I try to log out, the IdG doesn't redirect me back to Benefits. I end up at this screen:
> > [image]
> > When I try log out on https://dev-benefits.calitp.org, it redirects me back to Benefits.
> > Does this happen for anyone else?
> 
> @angela-tran I'm seeing this behavior now locally, but not in `dev` -- I can't see how it would be due to any changes in this PR. I'll follow up in Slack.

I found the issue, it was a change from this PR :disappointed: I'm going to revert it and rebase... stand by.

angela-tran:
> Fixed now, I dropped out [the commit](https://github.com/cal-itp/benefits/commit/c3cfe716270c3d0c336669783944c0b3fd748bca) that changed the URL of our `post_logout` endpoint, which is registered with IdG 

Ah that makes sense

angela-tran (dismissed):
Looks good!

benefits/oauth/middleware.py
```diff
@@ -16,8 +22,19 @@ def process_request(self, request):
             # from the base middleware class, the session didn't have a verifier
             return result
 
-        if session.verifier(request).uses_auth_verification:
+        verifier = session.verifier(request)
+
+        if verifier.uses_auth_verification:
+            # all good, the chosen verifier is configured correctly
             return None
+        elif verifier.api_url is None and verifier.form_class is None:
```
lalver1, line 30: Not sure if it's only in my configuration, but if I don't set an Auth provider and leave both Eligibility API base URL and form_class blank, it looks like `verifier.api_url` and `verifier.form_class` evaluate to `""` and `""` doesn't evaluate to `None` in the condition so the new error page is not shown. If I change the condition to
```
elif verifier.api_url == "" and verifier.form_class == "":
```
then the condition does evaluate to true in this case and I see the new error page.

lalver1 (commented):
The logic looks good @thekaveman for the misconfigured verifier case, but there may be a small typo related to checking for `is None` and `== ""` that prevents the new error page from displaying.

benefits/oauth/middleware.py
thekaveman: Thanks @lalver1 -- I updated the conditional and squashed it into the latest commit [`5bce79d`](https://github.com/cal-itp/benefits/pull/2227/commits/5bce79d175967e3e382187fb959e961390552213)

lalver1 (approved):
Looks good! 👍

PR #2232, opened 2024-07-19
Fix: terraform plan and apply conditions

lalver1:
This PR is a follow-up to PR #2213. The terraform `plan` and `apply` steps were not running when expected because the conditions on the steps were not evaluating correctly. Contrary to the [documentation](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops#parameters-in-conditions), we noticed that [`parameters` don't get evaluated in `conditions`](https://github.com/cal-itp/benefits/blob/2f628637fefe66fde24cebecdcd4cf3a7d6ddb1b/terraform/pipeline/deploy.yml#L74-L75). Because of this, even though `tag_type` was set in [`Terraform apply`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/deploy.yml#L60), this step wouldn't run when a tag was pushed.

This PR fixes this behavior by changing `parameters.tag_type` to `variables['tag_type']` for [`Terraform plan`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/deploy.yml#L41) and [`Terraform apply`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/deploy.yml#L60) which requires:

- moving the step that generates the `tag_type` from `azure-pipelines.yml` into `deploy.yml` so that `tag_type` doesn't get passed as a parameter anymore
- using `tag.py` to determine the value of `IS_TAG` instead of an Azure pipeline expression since `env:` [does not support evaluating expressions](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-bash?view=azure-pipelines) and `variables:` is [out of scope](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#variable-scopes) in `deploy.yaml`.

### DIFF
--- a/terraform/azure-pipelines.yml
+++ b/terraform/azure-pipelines.yml
@@ -29,11 +29,7 @@ stages:
           - name: IS_TAG
             value: $[startsWith(variables['Build.SourceBranch'], 'refs/tags/')]
         steps:
-          - bash: |
-              python terraform/pipeline/workspace.py
-
-              TAG_TYPE=$(python terraform/pipeline/tag.py)
-              echo "##vso[task.setvariable variable=tag_type;isOutput=true]$TAG_TYPE"
+          - bash: python terraform/pipeline/workspace.py
             displayName: Set environment-related variables
             # save the values
             # https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#use-outputs-in-a-different-job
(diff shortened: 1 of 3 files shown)


angela-tran (approved):
This reads very nicely and cleanly to me! Nice work @lalver1

thekaveman (approved):
Nice work sticking with this one!

PR #2250, opened 2024-07-24
Feat: set language code on Littlepay overlay

### LINKED ISSUE #2195: Send user's current language code to Littlepay overlay
The Littlepay Hosted Card Tokenization overlay supports multiple languages. We can grab the user's current language and send it with the template context to use in the configuration of that overlay.

## Acceptance Criteria

- [ ] The Littlepay Hosted Card Tokenization overlay displays in English if the user is currently viewing Benefits in English
- [ ] The Littlepay Hosted Card Tokenization overlay displays in Spanish if the user is currently viewing Benefits in Spanish

## Additional context

[Littlepay docs on localization of the overlay](https://docs.littlepay.io/back-office/littlepay-hosted-card-tokenisation/#localising-language)

Note that Littlepay supports 2 dialects of Spanish:

| Code | Description |
| ------|-------------|
| `es` | Spanish |
| `es_419` | Spanish (Latin America) |

We should use the Latin America dialect as that is likely closer to our target population in California.

lalver1:
Closes #2195 

Given that the Littlepay Hosted Card Tokenization overlay supports multiple languages, this PR uses this feature to set the overlay's language to the user's language in Benefits.

Note that the Littlepay docs list `es_419` as the code for `Spanish (Latin America)`, but this seems to be a typo since the overlay did not work with this code, instead it does work with `es-419`.

## Steps for testing

Run through a complete flow until you get to the enrollment index page. Then, change languages in Benefits and close/open the overlay to see the overlay in the language that was selected in Benefits.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -186,6 +186,9 @@ def index(request):
         tokenize_system_error_form = forms.CardTokenizeFailForm(ROUTE_SYSTEM_ERROR, "form-card-tokenize-fail-system-error")
         tokenize_success_form = forms.CardTokenizeSuccessForm(auto_id=True, label_suffix="")
 
+        # mapping from Django's I18N LANGUAGE_CODE to Littlepay's overlay language code
+        overlay_language = {"en": "en", "es": "es-419"}.get(request.LANGUAGE_CODE, "en")
+
         context = {
             "forms": [tokenize_retry_form, tokenize_server_error_form, tokenize_system_error_form, tokenize_success_form],
             "cta_button": "tokenize_card",
@@ -197,6 +200,7 @@ def index(request):
             "form_server_error": tokenize_server_error_form.id,
             "form_success": tokenize_success_form.id,
             "form_system_error": tokenize_system_error_form.id,
+            "overlay_language": overlay_language,
         }
 
         logger.debug(f'card_tokenize_url: {context["card_tokenize_url"]}')
--- a/tests/pytest/enrollment/test_views.py
+++ b/tests/pytest/enrollment/test_views.py
@@ -249,6 +249,18 @@ def test_index_eligible_get(client, model_EligibilityType):
     assert "token_field" in response.context_data
     assert "form_retry" in response.context_data
     assert "form_success" in response.context_data
+    assert "overlay_language" in response.context_data
+
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures("mocked_session_agency", "mocked_session_verifier", "mocked_session_eligibility")
+@pytest.mark.parametrize("LANGUAGE_CODE, overlay_language", [("en", "en"), ("es", "es-419"), ("unsupported", "en")])
+def test_index_eligible_get_changed_language(client, LANGUAGE_CODE, overlay_language):
+    path = reverse(ROUTE_INDEX)
+    client.post(reverse("set_language"), data={"language": LANGUAGE_CODE})
+    response = client.get(path)
+
+    assert response.context_data["overlay_language"] == overlay_language
 
 
 @pytest.mark.django_db
--- a/benefits/enrollment/templates/enrollment/index.html
+++ b/benefits/enrollment/templates/enrollment/index.html
@@ -67,7 +67,8 @@
                   element: '#{{ cta_button }}',
                   envUrl: '{{ card_tokenize_env }}',
                   options: {
-                      color: '#046b99'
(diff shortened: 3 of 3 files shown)


tests/pytest/enrollment/test_views.py
```diff
@@ -249,6 +249,17 @@ def test_index_eligible_get(client, model_EligibilityType):
...
     assert "form_success" in response.context_data
+    assert "overlay_language" in response.context_data
+
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures("mocked_session_agency", "mocked_session_verifier", "mocked_session_eligibility")
+@pytest.mark.parametrize("LANGUAGE_CODE, overlay_language", [("en", "en"), ("es", "es-419")])
+def test_index_eligible_get_changed_language(client, LANGUAGE_CODE, overlay_language):
+    path = reverse(ROUTE_INDEX)
+    response = client.get(path, headers={"accept-language": LANGUAGE_CODE})
```
angela-tran, line 260: Nice, this works well!

This line made me curious about how Django handles setting and reading the current language, and I thought I'd share what I learned. 

The docs about [How Django discovers language preference](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#how-django-discovers-language-preference) lists out the various places that the current language could come from, in order of precedence:
 - the URL (doesn't apply to us because we're not using [`i18n_patterns`](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#how-django-discovers-language-preference) in our [root URLconf](https://github.com/cal-itp/benefits/blob/main/benefits/urls.py#L22))
 - the Django language cookie
 - the `Accept-Language` header sent by the browser
 - the global LANGUAGE_CODE setting

Our `lang-selector.html` template makes use of the built-in [`set_language` view](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#the-set-language-redirect-view) when the form is [submitted](https://github.com/cal-itp/benefits/blob/main/benefits/core/templates/core/includes/lang-selector.html#L14), so to set the language in a way similar to how a user would, I think we could replace this line with 
```
client.post(reverse("set_language"), data={"language": LANGUAGE_CODE})
response = client.get(path)
```

Either one is perfectly sufficient for the purpose of this test I think, so no change necessary

tests/pytest/enrollment/test_views.py
```diff
@@ -249,6 +249,17 @@ def test_index_eligible_get(client, model_EligibilityType):
     assert "token_field" in response.context_data
     assert "form_retry" in response.context_data
     assert "form_success" in response.context_data
+    assert "overlay_language" in response.context_data
+
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures("mocked_session_agency", "mocked_session_verifier", "mocked_session_eligibility")
+@pytest.mark.parametrize("LANGUAGE_CODE, overlay_language", [("en", "en"), ("es", "es-419")])
```
thekaveman, line 257: We should add a case here for an unknown/unsupported language code.

thekaveman (changes requested):
Small request to tighten up the code a bit.

angela-tran (approved):
This looks great!

image

benefits/enrollment/views.py
```diff
@@ -186,6 +186,11 @@ def index(request):
         tokenize_system_error_form = forms.CardTokenizeFailForm(ROUTE_SYSTEM_ERROR, "form-card-tokenize-fail-system-error")
         tokenize_success_form = forms.CardTokenizeSuccessForm(auto_id=True, label_suffix="")
 
+        # mapping from Django's I18N LANGUAGE_CODE to Littlepay's overlay language code
+        overlay_language_codes = {"en": "en", "es": "es-419"}
+
+        overlay_language = overlay_language_codes[request.LANGUAGE_CODE]
```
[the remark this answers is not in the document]
lalver1: Ah good idea 👍

tests/pytest/enrollment/test_views.py
lalver1: Thanks @angela-tran ! This is good to know, and I like the idea that the test should follow as close as possible how the application is used so I think it's worth it that I make this change 👍

lalver1:
Thanks for the suggestions @thekaveman and @angela-tran , I think this one is ready for your re-review.

thekaveman (approved):
Looks good! Tested locally and it works great :+1:

angela-tran (approved):
🎉

PR #2253, opened 2024-07-29
Adding Medicare cardholders

indexing:
- added Medicare cardholders to supported enrollment pathways table

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,13 +27,13 @@ The following California transit agencies have launched Cal-ITP Benefits for the
 
 The Cal-ITP Benefits app supports the following enrollment pathways that use the corresponding eligibility verification methods:
 
-| Enrollment pathway                                             | Eligibility verification                                                              | Status | Launch                                                                |
-| -------------------------------------------------------------- | ------------------------------------------------------------------------------------- | ------ | --------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


indexing:
For the new Medicare cardholders row, I followed the path structure of the links to other enrollment pathways from the table.
`[**Medicare cardholders**](/benefits/enrollment-pathways/medicare-cardholders)`

However, it seems the links for the other enrollment pathways are broken on the live page. Click low-income as an example. I get a 404. https://github.com/cal-itp/benefits/tree/main/docs

Screenshot 2024-08-16 at 1 09 44 PM

The link is `[**Low-income**](/benefits/enrollment-pathways/low-income)` in the code.
The link in the published documentation is 
https://github.com/cal-itp/benefits/blob/main/benefits/enrollment-pathways/low-income

The relative path looks correct, but do we need to add the full URL—including the subdomain—for the link? 
https://docs.calitp.org/benefits/enrollment-pathways/low-income/

@thekaveman Something is not resolving correctly in the docs file structure.

thekaveman:
@indexing that's just how the preview site works sometimes, unfortunately. The live site is working fine: https://docs.calitp.org/benefits/#supported-enrollment-pathways

It's because the live site is all under a sub-directory of the domain, e.g. `docs.calitp.org/benefits/#supported-enrollment-pathways`

But the preview site is at the top-level of the domain, e.g. `benefits-2253--cal-itp-previews.netlify.app/#supported-enrollment-pathways`

I don't know if its worth digging into more at this point. That's a shared preview site used by many repos in Cal-ITP. We may be able to change the way links are entered in our docs to resolve correctly no matter what, that could be something to look into... but would be a big refactor.

thekaveman (approved):

PR #2254, opened 2024-07-30
Chore: normalize lines for tag_type

### LINKED ISSUE #2240: Normalize lines in `tag.py`
Modify [`tag.py`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/tag.py) to be more consistent with [`workspace.py`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/workspace.py):

[These lines](https://github.com/cal-itp/benefits/blob/main/terraform/azure-pipelines.yml#L35-L36) could be normalized to move the `TAG_TYPE` and `echo` in `tag.py` similar to how it works in `workspace.py`, but not a priority for this PR.

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/2213#discussion_r1680022276_

lalver1:
Closes #2240 

This PR normalizes setting `tag_type`. It moves the setting of `tag_type` from the shell into `tag.py`, consistent with how setting the environment-related variables is done in [`workspace.py`](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/workspace.py#L52-L53).

### DIFF
--- a/terraform/pipeline/deploy.yml
+++ b/terraform/pipeline/deploy.yml
@@ -36,18 +36,12 @@ steps:
       workingDirectory: "$(System.DefaultWorkingDirectory)/terraform"
       # service connection
       environmentServiceNameAzureRM: "${{ parameters.service_connection }}"
-  - bash: |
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Looks good :+1: 

I see you pushed an RC tag @lalver1; pipeline ran and the `tag_type` was set correctly:

image

PR #2277, opened 2024-08-06
Fix: Eligibility Confirm - Unnest explanatory-text-wrapper div

### LINKED ISSUE #2271: Visual regression: Agency card form padding on Mobile
The vertical alignment on the Agency card form is funky on both Mobile and Desktop

## To Reproduce

1. Go to CST
2. Click on Agency Card

## Expected behavior

image

## Screenshots

image
image
image

- Mac OS
- Tested on Chrome and Firefox

- iPhone
- Safari

machikoyasuda:
closes #2271 

The problem: Vertical alignment of the `We use the ...` paragraph is off for both Desktop and Tablet.
image

The issue:
image

These two divs need to be siblings, not a parent/child relationship. Having a `row justify-content-center` _inside_ another `row justify-content-center` is giving the paragraph in the child div a more padding left and right. The heading text and the paragraph text need to be siblings, not a parent/child.

This parent/child relationship was created in the `base.html` file that the `confirm.html` is based off of.

Since `explanatory-text-wrapper` is _only_ used on this `confirm.html` page, I was able to safely change the `base.html` to fix the issue, and know that this is will only change the confirm page and not any other pages.

Test is by looking at all pages for other regressions, and testing the Confirm page in desktop and mobile

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -79,13 +79,13 @@
           <div class="row justify-content-center">
             {% block headline %}
             {% endblock headline %}
-            {% block explanatory-text-wrapper %}
+          </div>
+          {% block explanatory-text-wrapper %}
             <div class="col-lg-8">
               {% block explanatory-text %}
               {% endblock explanatory-text %}
             </div>
-            {% endblock explanatory-text-wrapper %}
-          </div>
+          {% endblock explanatory-text-wrapper %}
           <div class="row justify-content-center">
             {% block inner-content %}
             {% endblock inner-content %}


thekaveman:
@machikoyasuda just as a reminder if you're reviewing/smoke testing in `prod`:

image

Make sure your user agent string is overridden to `cal-itp/benefits-smoke-test`.

> The problem: Vertical alignment of the We use the ... paragraph is off for both Desktop and Tablet.

Am I understanding that it is the left and right padding/margin around this paragraph (pushing it inwards) that is the issue? I think that is the _horizontal_ alignment right?

thekaveman (approved):

machikoyasuda:
@thekaveman Yeah you're right - the wording is off. The horizontal spacing is off, which is taking the divs out of vertical alignment. All the divs should all start from the same point along the blue line I added in the screenshots.

PR #2291, opened 2024-08-07
Refactor: move IdG config to EligibilityVerifier

### LINKED ISSUE #2237: Model refactor: move IdG config to `EligibilityVerifier`
Depends on #2233 being done first.

## Acceptance criteria

- [ ] `ClaimsProvider.scope` moved to `EligibilityVerifier.claims_scope`
- [ ] `ClaimsProvider.claim` moved to `EligibilityVerifier.claims_claim`
- [ ] ~`ClaimsProvider.supports_claims_verification` moved to `EligibilityVerifier.supports_claims_verification`~ Conditions in `ClaimsProvider.supports_claims_verification` moved to `EligibilityVerifier.uses_claims_verification` inline
- [ ] `EligibilityVerifier.claims_scheme` added as optional (leave the field as required on `ClaimsProvider`); default to using the value on `ClaimsProvider` if no value on `EligibilityVerifier`
- [ ] Existing Django config data is migrated
- [ ] Docs are updated

## Impact on the code

- [ ] [Code search for `claim` in the repo](https://github.com/search?q=repo%3Acal-itp%2Fbenefits+claim&type=code)
- [ ] [Code search for `scheme` in the repo](https://github.com/search?q=repo%3Acal-itp%2Fbenefits+scheme&type=code)
- [ ] [Code search for `scope` in the repo](https://github.com/search?q=repo%3Acal-itp%2Fbenefits+scope&type=code)

## Impact on existing data

- [ ] Data migrations needed to move field data from `ClaimsProvider` instances to `EligibilityVerifier` instances

## Impact on analytics

- [ ] These fields are not logged in analytics

## Impact on Admin permissions

- [ ] [Read only fields](https://github.com/cal-itp/benefits/blob/main/benefits/core/admin.py#L37) need to be updated for the moved fields

lalver1:
Closes #2237

## Testing locally

1. Checkout the `main` branch
1. Make sure that your `local_fixtures.json` (or whatever fixture you are using) has a `ClaimsProvider` and an `EligiblityVerifier` that has a foreign key reference to that claims provider.
1. Run `bin/reset_db.sh` (now your local DB mirrors the structure of the `dev` environment)
1. Checkout this PR's branch
1. Run `bin/init.sh` (running this PR's migration on top of existing DB structure)

- [ ] For all `EligiblityVerifier`s that have a `ClaimsProvider`, confirm that the data in `EligiblityVerifier.claims_provider.scope` and `EligiblityVerifier.claims_provider.claim` was migrated to `EligiblityVerifier.claims_scope` and `EligiblityVerifier.claims_claim`, respectively.
- [ ] Confirm that when the code calls an`EligiblityVerifier.claims_scheme` instance, the value of `claims_scheme` equals the value of the associated `ClaimsProvider.scheme` if `claims_scheme` is blank in the database. Otherwise, `claims_scheme` should equal the value set for `EligiblityVerifier` in the database.
- [ ] Flows using scopes/claims from IdG should still work
- [ ] In the admin panel, when not logged in as a super user, check that `Claims scheme` under a selected `Eligibility verifiers` is read-only.

## Notes/Questions

- Is the [setter method](https://github.com/cal-itp/benefits/blob/refactor/move-IdG-config/benefits/core/models.py#L252) for `claims_scheme`
```python
@claims_scheme.setter
def _set_claims_scheme(self, value):
    self._claims_scheme = value
```
actually needed? I included it because I thought that it would be required for setting the value using the admin interface, but I did some testing (with it removed) and it doesn't seem to be needed. We also don't set `claims_scheme` in the code, so maybe I should remove it?


benefits/core/models.py
```diff
@@ -213,10 +209,14 @@ def eligibility_api_public_key_data(self):
         """This Verifier's public key as a string."""
         return self.eligibility_api_public_key.data
 
+    @property
+    def supports_claims_verification(self):
+        return bool(self.claims_scope) and bool(self.claims_claim)
```
angela-tran, line 214: I think we can put this logic inline in `uses_claims_verification` and remove this method.

benefits/core/models.py
```diff
@@ -235,6 +235,24 @@ def by_id(id):
         logger.debug(f"Get {EligibilityVerifier.__name__} by id: {id}")
         return EligibilityVerifier.objects.get(id=id)
 
+    _claims_scheme = models.TextField(
+        help_text="The authentication scheme to use (Optional). If blank, defaults to the value in Claims providers",
+        default=None,
+        null=True,
+        blank=True,
+        verbose_name="Claims scheme",
+    )
```
thekaveman, line 244: I think this implementation could be improved, the naming the field (with the underscore) and setter are a little strange. Is a setter even needed?

One suggestion is to just rename the field to be more explicit like `claims_scheme_override` or something.

Then the `@property` works the same, and the setter isn't needed.

thekaveman (changes requested):
This is largely looking good :+1: 

I think the `claims_scheme` field needs a little more work though.

benefits/core/models.py
thekaveman: > Notes/Questions
> Is the [setter method](https://github.com/cal-itp/benefits/blob/refactor/move-IdG-config/benefits/core/models.py#L252) for claims_scheme actually needed? 

Haha sorry, I didn't see this note -- I don't think it is.

benefits/core/models.py
lalver1: I agree, removing this method made the model a bit simpler 👍

benefits/core/models.py
lalver1: 😄 thanks for confirming the setter isn't needed. And I agree about the naming, I like `claims_scheme_override` so I made the change.

thekaveman: Great! But let's group all the fields together at the top of the class, which is kind of the standard practice.

benefits/core/models.py
```diff
@@ -235,6 +231,20 @@ def by_id(id):
...
+
+    @property
+    def claims_scheme(self):
+        if self.claims_scheme_override == "":
```
thekaveman, line 244: the default value above is `None`, so probably better to check

```python
if not self.claims_scheme_override:
```

benefits/core/migrations/local_fixtures.json
```diff
@@ -63,8 +59,6 @@
       "client_name": "calfresh-benefits-oauth-client-name",
       "client_id_secret_name": "claims-provider-client-id",
       "authority": "https://example.com",
-      "scope": "verify:calfresh",
-      "claim": "calfresh",
       "scheme": "dev-cal-itp_benefits"
     }
   },
```
thekaveman, line 64: Now these `claimsprovider` fixtures are all the same -- we only need one of them. Make sure to update the references in the `eligibilityverifier` fixtures below.

benefits/core/models.py
thekaveman: Great! But the new field should be added up above with the other fields. Not in the middle of some methods and properties where it is harder to see and potentially introduces more complicated merges.

thekaveman (changes requested):
Getting closer

benefits/core/migrations/local_fixtures.json
lalver1: Ah, that's true. Updated in d6c6d5e0f330cad52127ebe745b029847ad2907d

benefits/core/models.py
lalver1: For sure, moved it so it's next to the other fields 👍

benefits/core/models.py
lalver1: Yep, this works better than `== ""`

thekaveman (approved):
This LGTM! And I tested it locally with dev IdG and the older adults flow, worked :heavy_check_mark:

PR #2300, opened 2024-08-13
Feat: introduce separate app for in-person eligibility / enrollment

### LINKED ISSUE #2284: Set up URLs for in-person enrollment pages
### Implementation
- ~~[ ] Add view functions to `TransitAgencyAdmin` that each return just a blank template~~
   - ~~[ ] `eligibility` (with URL `/eligibility`)~~
   - ~~[ ] `enrollment` (with URL `/enrollment`)~~
- ~~[ ] Implement [`get_urls`](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_urls) on `TransitAgencyAdmin` to map URL paths to the view functions described above, wrapping them in Django's built-in `admin_view` function to ensure basic admin site permissions are enforced~~
   - ~~The expected URLs resulting from this are: `<base URL>/admin/core/transit_agency/eligibility` and `<base URL>/admin/core/transit_agency/enrollment`~~

In #2295, we decided it makes sense to introduce a new Django app that handles the in-person logic and templates. The admin interface simply provides the entrypoint to the in-person logic via the home dashboard (#2246).

This means the URLs to the two views are
- `/in_person/eligibility`
- `/in_person/enrollment`

## Acceptance Criteria

- [x] These URLs exist and go to a blank page
- [x] These views behave as admin views (i.e. require admin-interface login)

## Additional context

This ticket by itself is not that meaningful to a user, but it is foundation that's needed to then work on #2243 (permission to access specific transit agencies) and #2246, #2269, #2244 (implementing the actual pages)

### LINKED ISSUE #2295: Set up file structure for in-person enrollment code (views, templates, forms)
Needed by #2284, #2269, #2246 at least

Need to figure out how to organize the code implementing in-person enrollment (it can't just all be in `admin.py`)

### Decision

The approach we've decided to take is to introduce a new Django app called `in_person`. This centralizes all the view, form, template code for the in-person eligibility verification feature.

The reason for not naming this app `admin` is because app [`label`](https://docs.djangoproject.com/en/5.1/ref/applications/#django.apps.AppConfig.label)s in Django need to be unique, and `admin` is already taken by `django.contrib.admin`. 

Naming conflict or not though, this approach also just feels more aligned with Django's philosophy to separate specific and composable features into [separate apps](https://docs.djangoproject.com/en/5.0/ref/applications/#projects-and-applications). This is as opposed to having an `admin` app that contains all features we would want to expose in the admin interface. 

Another approach we considered was to implement the in-person views, forms, templates, etc. within the existing `eligibility` and `enrollment` apps and then wire up the URLs in a `ModelAdmin` in `core.admin`.

angela-tran:
Closes #2295 and closes #2284 

## Test locally

**Not logged in yet**
- Start the app
- Use browser's URL bar to go to `/in_person/eligibility` or `/in_person/enrollment`
   - You should be prompted with the admin interface login screen
- Log in
- You should see an empty page  

**Logged in**
- Start the app
- Go to `/admin`
- Log in
- Use URL bar to go to `/in_person/eligibility` or `/in_person/enrollment`
- You should see an empty page

### DIFF
--- a/benefits/urls.py
+++ b/benefits/urls.py
@@ -25,6 +25,7 @@ urlpatterns = [
     path("enrollment/", include("benefits.enrollment.urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("oauth/", include("benefits.oauth.urls")),
+    path("in_person/", include("benefits.in_person.urls")),
 ]
 
 if settings.DEBUG:
--- a/benefits/in_person/apps.py
+++ b/benefits/in_person/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class InPersonConfig(AppConfig):
+    name = "benefits.in_person"
+    label = "in_person"
--- a/benefits/in_person/urls.py
+++ b/benefits/in_person/urls.py
@@ -1,10 +1,11 @@
 from django.urls import path
 
+from benefits.core.admin import admin
 from . import views
 
 
 app_name = "in_person"
 urlpatterns = [
-    path("eligibility/", views.eligibility, name="eligibility"),
-    path("enrollment/", views.enrollment, name="enrollment"),
+    path("eligibility/", admin.site.admin_view(views.eligibility), name="eligibility"),
+    path("enrollment/", admin.site.admin_view(views.enrollment), name="enrollment"),
 ]
--- a/benefits/in_person/urls.py
+++ b/benefits/in_person/urls.py
@@ -1,0 +1,10 @@
+from django.urls import path
+
+from . import views
+
+
+app_name = "in_person"
+urlpatterns = [
+    path("eligibility/", views.eligibility, name="eligibility"),
+    path("enrollment/", views.enrollment, name="enrollment"),
+]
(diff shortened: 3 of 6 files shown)


benefits/urls.py
```diff
@@ -25,6 +25,7 @@
     path("enrollment/", include("benefits.enrollment.urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("oauth/", include("benefits.oauth.urls")),
+    path("in_person/", include("benefits.in_person.urls")),
```
thekaveman, line 28: Can we make this top level URL `in-person/` instead? I think it looks a little more natural in a URL than underscores.

thekaveman (changes requested):
This is great, just a small cosmetic change request.

Thanks for the tests too!

benefits/urls.py
thekaveman: Ah, disregard. Looks like we mostly use underscores everywhere else so I would rather be consistent.

thekaveman (approved):
Nice!

angela-tran:
Thanks @thekaveman !

Issue #2297, opened 2024-08-08
Add separate view for Django login


angela-tran:
Depends on #2295 

### Acceptance criteria
- [ ] `/admin/login/django` takes you to login page with fields for native login

thekaveman:
Let's call this "superuser login" to be really explicit

PR #2304, opened 2024-08-14
Hide native Django login fields

### LINKED ISSUE #2264: Hide native login fields
To reduce confusion, we should hide the native login fields on the view for `/admin`.

## Acceptance Criteria

- [ ] The login page only has buttons for logging in with SSO

## Additional context
[Hiding the login form](https://megalus.github.io/django-google-sso/customize/#hiding-the-login-form)

machikoyasuda:
closes #2264 

image

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -72,6 +72,7 @@ GOOGLE_SSO_SCOPES = [
     "https://www.googleapis.com/auth/userinfo.email",
     "https://www.googleapis.com/auth/userinfo.profile",
 ]
+SSO_SHOW_FORM_ON_ADMIN_PAGE = False
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",


benefits/settings.py
```diff
@@ -72,6 +72,7 @@ def RUNTIME_ENVIRONMENT():
     "https://www.googleapis.com/auth/userinfo.email",
     "https://www.googleapis.com/auth/userinfo.profile",
 ]
+SSO_SHOW_FORM_ON_ADMIN_PAGE = False
```
machikoyasuda, line 75: @angela-tran Should this an environment variable or can we have this False all the time?

angela-tran: No need for an environment variable, especially since we'll have the separate view in #2297. What you have looks good!

angela-tran (approved):

Issue #2243, opened 2024-07-23
Ensure user is associated to a `TransitAgency` before they can access in-person enrollment views


angela-tran:
Depends on #2295 and #2284 

The in-person enrollment pages assume they are in the context of some transit agency.
Therefore, we wouldn't want the user to **not** be associated to a transit agency and attempt to view the in-person enrollment pages.

Previously, the idea was that an admin from Cal-ITP would manually associate the user to their transit agency by adding the user to a group that is referenced by the `TransitAgency`.

However, we can minimize the time in which a user is transit-agency-less by automatically associating the user to a transit agency based on the domain from their email. 

We don't necessarily want every user associated with the transit agency to be able to do in-person enrollment, so we need a separate group to represent having permission to do that. That is the group to which they need to be manually added by a Cal-ITP admin. If the user is not in that group yet, they will not see the part of the UI for "In-person enrollment."

## Technical details

Each `TransitAgency` will have a reference to a `Group` from the Django admin site via a `staff_group` field. This represents being associated with the `TransitAgency` (i.e. "belonging to" the `TransitAgency`.)

They will each also have a reference to a `Group` via a `customer_service_group` field. This represents having permission to do in-person enrollment. 

### Model changes
- [x] Add a `staff_group` field to `TransitAgency` which is a `ForeignKey` to the `Group` model from `django.contrib.auth`
   - The field can be null and can be set to blank
   - `on_delete` should be `models.PROTECT` which means if someone tries to delete the `Group`, they need to first go remove the reference from the `TransitAgency`
- [x] Add another field similar to `staff_group` but for `customer_service_group` instead
- [x] Add a `sso_domain` field to `TransitAgency` which is used to map to the user's email domain

### View changes
No view changes needed. It will be very unlikely for the user to not have a transit agency, due to the automatic association using `sso_domain`. We're ok with the app showing some generic error screen if somehow that very unlikely situation happens.

## Acceptance Criteria
- [x] Users can be associated with transit agencies
- [x] Users are automatically associated with a transit agency if their email domain matches the SSO domain for the transit agency
- [x] User sees a message when trying to access a view they are not allowed to access

angela-tran:
Split out the permissions/access of model configuration views into a separate ticket (#2278)

This ticket is now solely focused on the permissions/access for in-person enrollment views

angela-tran:
This ticket is really saying "make sure the user is associated to a TransitAgency upon login".

angela-tran:
[...]

PR #2299, opened 2024-08-13
Refactor: move EligibilityType fields to EnrollmentFlow

lalver1:
Closes #2283

- [x] Fields from `EligibilityType` are moved to `EnrollmentFlow`
- [x] Migrate the data from `EligibilityType.name` to `EnrollmentFlow.name` in the migration file
- [x] The `label` field is used as the Admin display field for `EnrollmentFlow` instances
- [x] `TransitAgency.eligibility_types` field (and all related calculated fields) are removed
- [x] `EligibilityType` model is removed
- [x] `session.eligibility_types` is removed
- [x] `session.eligible` (get `bool`) used to indicate if the user was found eligible for their selected `flow`, updated through `session.update(eligible=bool)` like other session data
- [x] [`EligibilityType` read-only fields](https://github.com/cal-itp/benefits/blob/main/benefits/core/admin.py#L55) are moved to the new model
- [x] Analytics events that currently include `eligibility_types` properties (event/user) now use `enrollment_flows.eligibility_name` for the `eligibility_types` property values
- [x] Update tests

## Testing locally

1. Checkout the `main` branch
1. Make sure that your `local_fixtures.json` (or whatever fixture you are using) has an `EnrollmentFlow` that has a foreign key reference to an `EligibilityType`.
1. Run `bin/reset_db.sh` (now your local DB mirrors the structure of the `dev` environment)
1. Checkout this PR's branch
1. Run `bin/init.sh` (running this PR's migration on top of existing DB structure)

PR #2307, opened 2024-08-16
Feat: transit agency user group

angela-tran:
Part of #2243 

This PR associates a user to a `TransitAgency` through their group membership.


benefits/core/models.py
```diff
@@ -352,6 +353,14 @@ class TransitAgency(models.Model):
         help_text="The name of the secret containing this agency's client_secret value used to access the TransitProcessor's API.",  # noqa: E501
         default="",
     )
+    group = models.OneToOneField(
```
thekaveman, line 356: Let's name this something a little more explicit - like `staff_group` or `user_group` or something. 

E.g. we use `EnrollmentFlow.group_id` so this may be confusing.

benefits/core/migrations/0022_transitagency_group.py
```diff
@@ -0,0 +1,27 @@
...
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("auth", "0012_alter_user_first_name_max_length"),
+        ("core", "0021_rename_eligibilityverifier_enrollmentflow"),
```
thekaveman, line 11: Interesting... it did this automatically, to list 2 different `dependencies` there?

thekaveman (changes requested):
Looks like a good start, and I see that we have the reverse access so that's nice :+1: 

Small request to rename the field for clarity. Also I think you'll want to rebase on #2299 -- that PR has a migration as well, and it is nearly done.

benefits/core/migrations/0022_transitagency_group.py
angela-tran: Yes, it did that automatically

benefits/core/models.py
angela-tran: Done in [9b36c7c](https://github.com/cal-itp/benefits/pull/2307/commits/9b36c7c3af566d144cac61c9fca62599b0cdeb13)

thekaveman (approved):
:+1: :+1:

Issue #2252, opened 2024-07-29
Create a route to support sending users directly to `eligibility:confirm`


thekaveman:
As we work on the SBMTD flyer/mailer to message out the Benefits app to their riders with a _Reduced Fare Mobility ID_, we'd like to include a URL that will shortcut users as much as possible through the enrollment flow.

To that end, we should support a direct route into `eligibility:confirm` for the agency card flows.

## Acceptance Criteria

- [x] Create a URL that deep-links in the app so users can immediately enter agency card details without having to click though: Get Started > Select agency > Choose Benefit > Prepare (> Enter agency card details) 
- [x] A route exists to pre-configure a users' `TransitAgency` and ~`EligibilityType`/`EligibilityVerifier`~ `EnrollmentFlow`
- [x] The route takes the user to `eligibility:confirm` e.g. the eligibility confirmation form page
- [x] This route only works for flows utilizing the Eligibility API / forms -- show the user error page for other flows

## Additional context

See #1204 for how agency selection was implemented for `eligibility:index`. The work for this ticket can be done in a similar fashion.

thekaveman:
Suggest waiting for #2236 to be complete before working on this.

thekaveman:
For the URL:

* We don't expect users to actually type this in, rather scanning from a QR code or entering a bitly / shortname
* URL pattern will look something like `/<agency>/agency-card` e.g. `/sbmtd/agency-card`

thekaveman:
Could start on this on top of #2293

indexing:
@thekaveman I am encoding both URLs for the mailer today. To confirm, this is what I'm an using for each of the full URLs.

**QRCODE**
https://benefits.calitp.org/sbmtd/agency-card?utm_source=printed_mailer&utm_medium=QRCODE&utm_campaign=SBMTD_mobility-ID&utm_id=sep2024&utm_content=mobilityID_EN

**SHORT URL (https://bit.ly/sbmtd-enroll)**
https://benefits.calitp.org/sbmtd/agency-card?utm_source=printed_mailer&utm_medium=short_URL&utm_campaign=SBMTD_mobility-ID&utm_id=sep2024&utm_content=mobilityID_EN

thekaveman:
@indexing in the current implementation they look like `/{slug}/agencycard` (i.e. no dash in agencycard).

I can update it if its too late.

indexing:
@thekaveman Yes. Please add the dash so we use `/{slug}/agency-card`. 🙏

PR #2308, opened 2024-08-16
Refactor: centralize management of Django routes

### LINKED ISSUE #2303: Chore: update route for re-enrollment error
The current route of [`/enrollment/reenrollment-error`](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/urls.py#L15) is the only instance where we use a hyphen to separate a multi-word URL fragment. Let's make this one consistent with the others.

## Acceptance Criteria

- [ ] New route uses an underscore instead: `/enrollment/reenrollment_error`; OR
- [ ] New route builds on the existing error route like: `/enrollment/error/reenrollment`

## Additional context

Came up in PR review of the new `in_person` app setup: https://github.com/cal-itp/benefits/pull/2300#discussion_r1716050482

thekaveman:
Closes #2303 

Part of #2252 

## Context setting

Right now we have Django route strings, e.g. `app_name:view_name` all over the place: in views, in middleware, in models, in templates, in URL config, in tests... it is truly a mess and difficult to get a handle on.

This PR refactors every route string in the app into a top-level `benefits/routes.py` module. Everywhere else mentioned above that needs a route string, uses this central module to get it.

## Implementation details

* Create a class `benefits.routes.Routes` with an `@property` for every route in the app
* Property names are exactly like the current `ROUTE_NAME` variables, cleaned up and organized by app e.g.
  * `INDEX` for the homepage
  * `HELP` for the help page
  * `ELIGIBILITY_INDEX` for the entry to eligibility phase
  * `ENROLLMENT_INDEX` for the entry to enrollment phase
  * etc.
* _Note: I decided against `CORE_` routes, it felt a little extra... but happy to add that for consistency._
* Property docstrings `"""Doc strings"""` are used to describe what the route is for, handy for devs using these props
* Define an instance of the class in the module: `routes = Routes()`
* This instance is imported and used elsewhere e.g. in views: 
  ```python
  
  redirect(reverse(routes.INDEX))
  ```
* Helper function on the class `Routes.name(route: str) -> str` gets the name portion, e.g. `name` from `app:name`
* `urls.py` uses this `Routes.name()` helper when registering URLs, avoiding duplication
* New `benefits.core.context_processors.routes()` puts every route in the template context, so all hardcoded routes could be removed from templates:
  ```html
  <a href="{% url routes.INDEX %}">Go home</a>
  ```


benefits/eligibility/views.py
```diff
@@ -37,7 +29,7 @@ def index(request, agency=None):
         else:
             session.update(request, eligible=False, origin=agency.index_url)
     else:
-        session.update(request, agency=agency, eligible=False, origin=reverse(ROUTE_CORE_INDEX))
+        session.update(request, agency=agency, eligible=False, origin=agency.index_url)
```
lalver1, line 32: I think this is correct, since if the session has an agency, the origin is `agency.index_url` --> `routes.AGENCY_INDEX` --> `"core:agency_index"` (url `"/cst"`). But originally `origin` was set to `"core:index"` via `ROUTE_CORE_INDEX` (url `"/"`) so just wanted to make sure this change was intended.

lalver1 (dismissed):
The code looks good to me and I didn't see any dangling routes (searched for the regex `"[A-Za-z0-9_]*:[A-Za-z0-9_]*"` and the results are all in `routes.py`). I only saw one place where the route changed but I think the change makes sense 👍

benefits/eligibility/views.py
thekaveman: Yes, thanks for noticing this and asking. 

I did intentionally make this change because it seemed wrong to direct the user to `core:index` if they had an agency selected already.

thekaveman:
I think it would be good if @angela-tran takes a look before this is merged.

benefits/core/context_processors.py
```diff
@@ -92,3 +94,17 @@ def origin(request):
...
+
+def routes(request):
+    """Context processor adds information about each application route to the context."""
+
+    # get a dict[str,str] mapping property name --> value
+    # for each @property in the Routes collection
+    routes_dict = dict(
+        (prop, str(getattr(app_routes, prop)))
+        for prop in dir(app_routes)
+        if prop not in dir(Routes) or isinstance(getattr(Routes, prop), property)
```
angela-tran, line 107: Just for my understanding, what is `if prop not in dir(Routes)` for?

I see that `app_routes` is just an instance of `Routes`, so I'm not getting why it's needed.

Is there somewhere that adds to `app_routes`?

angela-tran (approved):
This refactor looks good to me. I left a question about how the `routes` context processor works, which I think is more just for my understanding rather than pointing out a needed change.

benefits/core/context_processors.py
thekaveman: Ah, yeah sorry I think this may have been a leftover from a refactor or something, and it doesn't make a lot of sense even to me looking at it now...

I think all we need is the second clause `isinstance(getattr(Routes, prop), property)` right?

thekaveman: And probably would would make it most clear is testing `app_routes`:

```python
if isinstance(getattr(app_routes, prop), property)
```

thekaveman: OK sorry about that!

I updated the code in [c37ef28](https://github.com/cal-itp/benefits/pull/2308/commits/c37ef28020585c60e3742b773089600d1da30812) and hopefully it is more clear now.

`app_routes` is just an alias:

```python
from benfits.routes import routes as app_routes
...
```

Because the function under test `benefits.context_processor.routes` has a conflicting name.

angela-tran: This is clearer, thanks!

angela-tran (approved):

PR #2309, opened 2024-08-19
Fix: getting a group's transit agency

angela-tran:
Follow-up to #2307 

If we use `group.transitagency` to get a `Group`'s `TransitAgency`, Django will raise a `RelatedObjectDoesNotExist` error if the group does not have a `TransitAgency`.

The Django docs recommend either catching the exception or checking for the attribute. This PR checks if the group has the attribute before trying to get the object.

https://docs.djangoproject.com/en/5.1/topics/db/examples/one_to_one/

image

### DIFF
--- a/benefits/core/models.py
+++ b/benefits/core/models.py
@@ -403,7 +403,7 @@ class TransitAgency(models.Model):
     def for_user(user: User):
         group = user.groups.first()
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
Nice catch!

PR #2312, opened 2024-08-20
Feat: Agency card direct route

### LINKED ISSUE #2252: Create a route to support sending users directly to `eligibility:confirm`

thekaveman:
Closes #2252

_Edited the URL to be `/{slug}/agency-card` per https://github.com/cal-itp/benefits/issues/2252#issuecomment-2305900492_

In addition to the tasks of #2252, this also refactors the one other agency-specific shortcut route to normalize it into the same pattern as the now (2) other agency-specific shortcut routes.

Instead of:

* `/cst/agency-card`
* `/cst/publickey`
* `/eligibility/cst`

We now have:

* `/cst/agency-card`
* `/cst/publickey`
* `/cst/eligibility`

This route is used on the agency selector modal: it is the route the user goes through to get to `eligibility:index` when they start at the `core:index` (which stores the transit agency selection in the user's session).

This differs slightly from the agency index page `core:agency_index`; in that case, although the transit agency is also stored in the user's session, the user's CTA on the page is a single button that does not use this shortcut, but goes directly to the `/eligibility` page (since the transit agency is already configured in the session).

I also refactored the pytest model fixtures a bit to get them into a shape that I could more easily use to test this new functionality. 

## Testing

1. Run this branch locally with `F5`
1. Visit `/cst/agency-card` --> you should be redirected to the `eligibility:confirm` page
1. Create a brand new Eligibility API flow instance; add _both_ to CST
1. Visit `/cst/agency-card` --> you should be redirected to the `eligibility:confirm` page with the _newer_ flow selected
1. In the admin, configure CST to not have any Eligibility API flows
1. Visit `/cst/agency-card` --> you should be redirected to the user error page


lalver1:
The code looks good to me and it was helpful to see how to normalize the structure of all agency redirects 👍
I also followed the testing steps and got the expected results ✅

angela-tran (approved):
This looks good to me. Tested it locally too

Issue #2246, opened 2024-07-23
Transit agency staff should see a home dashboard upon logging in


angela-tran:
## Acceptance criteria
- [ ] `/admin` for **a transit agency user** shows the home dashboard UI (see mockup below)
- [ ] `/admin` for a superuser or Cal-ITP user shows the default Django admin index
- [ ] Buttons link to URLs that were defined in #2284 

#2268 will implement the "Transit processor" link, so don't worry about that part of the UI for this ticket.

![image](https://github.com/user-attachments/assets/deb3d808-e3a3-403c-9418-f54d11b19ff2)

angela-tran:
Thanks @lalver1 for updating the issue description 🙏 

Here's a section from Django docs on CSS variables that drive the admin site's theming, in case it's helpful: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#theming-support

lalver1:
No problem @angela-tran 🙂
Ah, thanks for that! I was actually looking at how to override the template that `admin` uses, so this is very helpful!

angela-tran:
@lalver1 and I met earlier to talk through how to conditionally show the home dashboard and the default index page. I had originally been thinking we would just override `admin/index.html` and use conditionals on the user to determine what to show, but after going through some of it with Luis, we came to the consensus that it'll be much cleaner to have a separate template.

The default index page is returned by the [`AdminSite.index`](https://github.com/django/django/blob/main/django/contrib/admin/sites.py#L552-L571) function, so in order to return a different template for transit agency staff users, we need to override this function.

Luis and I worked through putting the [necessary configuration](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#overriding-the-default-admin-site) in place correctly so that there are no regressions and so that he can override `index`.

angela-tran:
This ticket is pretty big. The plan is to do it in 3 PRs:
- [ ] Create a base template which defines the header navbar and includes the styles that every page will need at a minimum
- [ ] Implement and return the dashboard template for transit agency staff users in the `index` view function, except for the top-right `{% usertools %}` block
- [ ] Implement the top-right `{% usertools %}` block

cc @lalver1

PR #2313, opened 2024-08-20
Feat: transit agency home dashboard

lalver1:
Part of #2246

This PR implements overriding the default admin site so that transit agency staff see a home dashboard upon logging in:
image

 A superuser or Cal-ITP user still sees the default Django admin site. 

Another PR will implement the top right corner "user tools" panel for transit agency staff, and another PR will implement a base `html` template that will be extended by all templates in the `in_person` app.

## Steps to test:
- Reset database to use local_fixtures (`bin/reset_db.sh`)
- Set GOOGLE_SSO_ALLOWABLE_DOMAINS, GOOGLE_SSO_STAFF_LIST, GOOGLE_SSO_SUPERUSER_LIST to appropriate values (your email and domain) in `.env`
- Go to `localhost/admin` and sign in with your email address
- Add a `CST` group
- Add the `CST` group to CST transit agency’s “staff_group” (it will ask for `audience` and `client_id`, use dummy values)
- Change your user to not superuser
- Remove `Cal-ITP` from your user’s chosen groups
- Add the `CST` group to your user’s chosen groups
- Save (will log you out)

> **_NOTE:_** If you logout and need to reach the agency home dashboard page again, you can skip the steps shown above by commenting out `add_staff_user_to_group(user, request)` in `benefits/core/admin.py`.  Then, when you log in again, your user will not be automatically added to the `Cal-ITP` group in the Google SSO prelogin hook (`pre_login_user`). If you don’t do this, you’ll need to reset the database again since you’ll never see the agency home dashboard page. The reason is that you will be part of `Cal-ITP`, but you won't be able to edit the database since you are not a superuser.

- Go to `localhost/admin`, verify that the agency home dashboard appears


thekaveman:
> This kind of makes me feel like we should have a real `benefits/admin` app? Aren't we sort of faking it here with everything at the top-level? Would that pose a problem? Is it better to do it this way for some reason?

@angela-tran walked me through the thinking (and docs!) on this setup, thank you :clap: :raised_hands: Makes total sense and please disregard my comment above.

benefits/admin.py
```diff
@@ -0,0 +1,31 @@
+from django.contrib import admin
+from django.contrib.auth.models import Group
+from django.template.response import TemplateResponse
+
+
+STAFF_GROUP_NAME = "Cal-ITP"
```
angela-tran, line 6: Is it possible to reuse the existing `STAFF_GROUP_NAME` variable from `benefits.core.admin`?

(I'm not sure if maybe this is just here temporarily while you focus on the bigger task of implementing the templates. But in any case, let's avoid duplicating this string.)

lalver1: Yep, it's here temporarily to make the local debugging setup easier (changing from `Cal-ITP` to `CST`, for example) but for sure I'll reuse the existing `STAFF_GROUP_NAME` 👍

lalver1:
Thanks for the comments @thekaveman  and @angela-tran! Quick update on this PR:

- The BenefitsAdminSite's `index` view gets the user's transit agency (assuming a database that has been set up) and updates the session so that the template will display it
- An `index` template that still needs work has been added
- Need to use `routes.py` instead of the hardcoded url in the template
- Update tests

benefits/admin.py
thekaveman: Would it be possible to make this a setting and use it that way everywhere? I think that would be cleaner than a Python variable import.

benefits/admin.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class BenefitsAdminSite(admin.AdminSite):
+
+    def index(self, request, extra_context=None):
+        """
+        Display the main admin index page, which lists all of the installed
+        apps that have been registered in this site.
+        """
```
machikoyasuda, line 17: I appreciate these comments!

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,68 @@
...
+                <img class="sm d-lg-none" src="{% static "img/logo-sm.svg" %}" width="90" height="51.3" alt="{% translate "California Integrated Travel Project: Benefits logo (small)" context "image alt text" %}" />
+                <img class="lg d-none d-lg-block" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="{% translate "California Integrated Travel Project: Benefits logo (large)" context "image alt text" %}" />
+            </span>
+        </div>
+        <h1 class="p-0">Administrator</h1>
+    </div>
+{% endblock branding %}
+
+{% comment %} {% block coltype %}
+    colMS
```
machikoyasuda, line 37:
```suggestion
    w-100
```

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,68 @@
...
+{% block content_title %}
+{% endblock content_title %}
+
+{% block content %}
+    <div class="container">
+        <div class="row justify-content-start">
+            <div class="col-lg-6">
+                <h1>{{ agency.long_name }}</h1>
+                <h1>In-person enrollment</h1>
+                <div>Verify transit benefit eligibility using agency criteria and complete a rider's open-loop card enrollment.</div>
```
machikoyasuda, line 59: This is extremely nit-picky of me:

`rider’s`

benefits/admin.py
lalver1: They came from Django but I agree that they are helpful! 🙂

benefits/admin.py
lalver1: Yep, I made it a setting in ee80de36ed8c2ef456c7fba44a01fa6f2cb1a4a6. Then, I made a small update to [`core/admin.py`](https://github.com/cal-itp/benefits/blob/main/benefits/core/admin.py) and its test in 62a2e8ff0bceb409cffbe1513dca74f01ce81a07 as part of this PR.

lalver1:
I think this one is ready for a review, but I still need to fix these items for sure:

-  Make the `Administrator` header white. I could've put an inline style to make it white, but it seems there should be a better solution.
image
- Not sure if the value for `STAFF_GROUP_NAME` should be set in `settings` (as of this PR) or read as an environment variable. Actually, I'm now leaning to an environment variable since this value will depend on where the app is deployed.

machikoyasuda:
I see Administrator as yellow: 
image

I'm not in Dark mode

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,78 @@
...
+{% block content_title %}
+{% endblock content_title %}
+
+{% block content %}
+    <div class="row">
+        <div class="col-lg-12">
+            <h1 class="pt-0 pb-3 text-left">{{ agency.long_name }}</h1>
+            <div class="border border-3 p-3">
+                <h2 class="pt-0 pb-2 text-left">In-person enrollment</h2>
+                <p>Verify transit benefit eligibility using agency criteria and complete a rider's open-loop card enrollment.</p>
```
machikoyasuda, line 64:
```suggestion
                <p>Verify transit benefit eligibility using agency criteria and complete a rider’s open-loop card enrollment.</p>
```

benefits/static/css/styles.css
```diff
@@ -395,6 +395,7 @@ footer .footer-links li a.footer-link:visited {
   outline: 3px solid var(--focus-color) !important;
   outline-offset: 2px !important;
   box-shadow: none !important; /* override CA State Web Template */
+  text-decoration: none;
```
machikoyasuda, line 398: Just wanted to leave a comment here for everyone else: This is necessary to override a Django Admin CSS style for buttons. Without this, there would be an underline on the button.

benefits/templates/admin/agency-dashboard-index.html
machikoyasuda: Just this small thing, otherwise this PR LGTM.

There are some questions about how we approach this Header section for example across all the new pages, but I think that kind of refactoring can happen later when we need it. And that for now, this is a good start.

machikoyasuda:
Just adding something I noticed here for people to see, but it doesn't need fixing or anything:

image


This PR made the log out area type small, and there's something funky going on with the underlines (the two links have different line heights, and there's a border-radius applied to one link). It's b/c of the way our current app CSS styles links. This is for documentation purposes only.

lalver1:
@angela-tran and @machikoyasuda, I was looking at the other in-person eligibility tickets and I'm wondering what you think if I add one minor thing to this PR. I would make a new template called something like `base-in-person.html` in the `in_person/templates/in_person` directory that will extend `admin/base.html` and will have all the common elements (like `{% block extrastyle %}` and `{% block branding %}`) from [`agency-dashboard-index`](https://github.com/cal-itp/benefits/blob/988b5b8d6006a16f69c92ba4eb0f5fd34e754559/benefits/templates/admin/agency-dashboard-index.html) that `eligibility.html`, `enrollment.html` and the current `agency-dashboard-index.html` need. Then, `eligibility.html`, `enrollment.html` and `agency-dashboard-index.html` will extend from `base-in-person.html` and that'll ensure that they have the same base styling.

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,78 @@
...
+        html[data-theme="dark"], :root {
+            --primary: #045b86;
+            --secondary: #045b86;
+            --link-fg: #ffffff;
+            --body-quiet-color: #000000;
+          }
+    </style>
+    <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet" type="text/css">
+    <link href="https://california.azureedge.net/cdt/statetemplate/6.0.7/css/cagov.core.min.css" rel="stylesheet">
+    <link href="{% static "css/styles.css" %}" rel="stylesheet">
```
angela-tran, line 21: I think it might not be necessary to pull in all the styles from this file. I know we said we want to reuse the Cal-ITP color palette, but I don't think that necessarily means we have to reuse this file.

And moreover, pulling in all the styles then causes small oddities like what @machikoyasuda mentioned in https://github.com/cal-itp/benefits/pull/2313#issuecomment-2305355509

@lalver1 Maybe a pairing session would be good to talk over this and about the mockups?

angela-tran (changes requested):

benefits/templates/admin/agency-dashboard-index.html
lalver1: Yep, that makes sense. A pairing session would be good 👍 I can send you an invite for early this afternoon if that works.

lalver1:
Thanks for the feedback last week @angela-tran. The tests were failing earlier but now they are all passing so I think this one is ready for a re-review. The key changes are:
 - The original description for this PR has been updated to reflect the change in scope
 - A `styles.css` was added for `admin` in 51ce1fd5a92f462444a204eaa854dd3947534994 to avoid the situation mentioned in [#2313 (comment)](https://github.com/cal-itp/benefits/pull/2313#issuecomment-2305355509) and to clean up the styling of the home dashboard

benefits/settings.py
```diff
@@ -338,6 +338,7 @@ def RUNTIME_ENVIRONMENT():
     "'unsafe-inline'",
     "https://california.azureedge.net/",
     "https://fonts.googleapis.com/css",
+    "https://cdn.jsdelivr.net",
```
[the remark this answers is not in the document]
thekaveman: I'm wondering if it is worth it to just download the Bootstrap source file and stick it in our repo?

benefits/static/css/admin/styles.css
```diff
@@ -0,0 +1,124 @@
...
+  --font-size-15px: calc(15rem / 16);
+  --font-size-16px: calc(16rem / 16);
+  --font-size-14px: calc(14rem / 16);
+  --font-size-12px: calc(12rem / 16);
+  --border-width: calc(2rem / 16);
+  --border-radius: calc(3rem / 16);
+  --focus-style: 3px solid var(--focus-color);
+  --focus-border-radius: calc(2rem / 16);
+  --focus-outline-offset: calc(2rem / 16);
+  --focus-border-radius: calc(2rem / 16);
```
thekaveman, line 48: RE: https://github.com/cal-itp/benefits/pull/2313/files#r1731506620

It feels like a lot of these variables could probably be moved to a separate file for re-use.

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,70 @@
+{% extends "admin/base.html" %}
+{% load i18n static %}
+
+{% block extrastyle %}
+    <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet" type="text/css">
```
thekaveman, line 5: Same here, maybe we just download this font file and serve it locally?

benefits/admin.py
```diff
@@ -0,0 +1,34 @@
...
+            "title": self.index_title,
+            "subtitle": None,
+            "app_list": app_list,
+            **(extra_context or {}),
+        }
+
+        request.current_app = self.name
+
+        staff_group = Group.objects.get(name=settings.STAFF_GROUP_NAME)
+        agency = TransitAgency.for_user(request.user)
```
thekaveman, line 29: Might this fail for a superuser or Cal-ITP staff that doesn't have an associated transit agency? Seems like it should be wrapped under the `if` below.

benefits/templates/admin/agency-dashboard-index.html
machikoyasuda: I don't think we should do that unless we are actually committed to Roboto, which I don't think we are.

benefits/static/css/admin/styles.css
machikoyasuda: Yeah we could in the future, but I'd only be comfortable doing that if Design is 10000% committed to these design tokens.

benefits/templates/admin/agency-dashboard-index.html
angela-tran: > There are some questions about how we approach this Header section for example across all the new pages, but I think that kind of refactoring can happen later when we need it. 

Thanks for noting this; I think this is an important question to answer sooner than later. 

@lalver1 and I talked more about his idea from https://github.com/cal-itp/benefits/pull/2313#issuecomment-2307464485 when we paired on Friday, and the plan is for him to open a separate PR that extracts out a base template that all the new pages can use.

benefits/settings.py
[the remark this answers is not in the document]
lalver1: Yep, that makes sense. I just tested it @angela-tran but I don't think it worked because the styles did not get applied:
image

benefits/static/css/admin/styles.css
angela-tran: > I'd only be comfortable doing that if Design is 10000% committed to these design tokens.

If Design changes their minds about some of these variables, wouldn't it be easier to handle that update if we had a single `variables.css` file?

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -0,0 +1,70 @@
...
+
+{% block branding %}
+    <div id="header-container" class="navbar navbar-expand-sm navbar-dark justify-content-between">
+        <div class="container">
+            <span class="navbar-brand p-0">
+                <img class="sm d-lg-none" src="{% static "img/logo-sm.svg" %}" width="90" height="51.3" alt="{% translate "California Integrated Travel Project: Benefits logo (small)" context "image alt text" %}" />
+                <img class="lg d-none d-lg-block" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="{% translate "California Integrated Travel Project: Benefits logo (large)" context "image alt text" %}" />
+            </span>
+        </div>
+        <h1 class="p-0 branding-title">Administrator</h1>
```
machikoyasuda, line 26:
```suggestion
        <h1 class="p-0 m-0 branding-title">Administrator</h1>
```

Adding `m-0` here will vertically center the Administrator text and the logo

benefits/static/css/admin/styles.css
thekaveman: > Yeah we could in the future, but I'd only be comfortable doing that if Design is 10000% committed to these design tokens.

Does this matter? Are we really coordinated with design on these tokens right now?

machikoyasuda: Yeah but I guess what I mean by "committing" is that if they change one thing in one place, they have to commit to changing it everywhere. Like, let's say, we bump up the H1 size on one page, or focus ring style on one page, we have to commit to those tokens being changed on all pages. Or else we have to add new variables.

benefits/static/css/admin/styles.css
```diff
@@ -0,0 +1,124 @@
...
+}
+
+@font-face {
+  font-family: "Public Sans";
+  font-weight: var(--bold-font-weight);
+  font-style: normal;
+  /* prettier-ignore */
+  src: local("PublicSans"), url("../../fonts/PublicSans-Bold.woff") format("woff");
+  /* prettier-ignore */
+}
```
machikoyasuda, line 58: I think this can be deleted.

benefits/static/css/admin/styles.css
```diff
@@ -0,0 +1,124 @@
...
+  letter-spacing: calc(var(--font-size-20px) * var(--letter-spacing-2));
+}
+
+:root {
+  --branding-title-color: #ffffff;
+}
+.branding-title {
+  color: var(--branding-title-color);
+}
+
```
machikoyasuda, line 117: You can also just use `text-white` utility class from Bootstrap so we don't have to have this one-off style. Documented here: https://getbootstrap.com/docs/5.3/utilities/colors/#colors

benefits/settings.py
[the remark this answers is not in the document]
lalver1: Putting `https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/` in `CSP_STYLE_SRC` worked! Just for documentation, if I leave out the trailing `/` it won't work.

benefits/templates/admin/agency-dashboard-index.html
angela-tran: Here is the full plan: https://github.com/cal-itp/benefits/issues/2246#issuecomment-2310896594

It is the same as what @lalver1 wrote in this PR's description.
> Another PR will implement the top right corner "user tools" panel for transit agency staff, and another PR will implement a base html template that will be extended by all templates in the in_person app.

benefits/templates/admin/agency-dashboard-index.html
angela-tran: @machikoyasuda pointed out that we can actually just remove this line since we don't actually need or use the font in the admin

machikoyasuda: Oh, sorry I meant we're not commited to Robot for the Analog site.

benefits/admin.py
lalver1: Yep, updated the docstring in 772f369554c6e8eded3cb4eb65913ed93ba9e8a5

benefits/static/css/admin/styles.css
lalver1: Created a `variables.css` file in `static/css` and imported it into `css/admin/styles.css` in c7621550bd559933dfa129da1a924a63bc9589b0. Made a minor refactor to import `variables.css` into the benefits app's `css/styles.css` in a88cc400a2e4be1b0ec9e68723d0ea4c10a986db

benefits/admin.py
lalver1: Yep, that's a good point. I moved it under the `if` in 772f369554c6e8eded3cb4eb65913ed93ba9e8a5

benefits/static/css/admin/styles.css
lalver1: 👍 removed it in c7621550bd559933dfa129da1a924a63bc9589b0

machikoyasuda (dismissed):
Front-end changes LGTM 👍 👍  This is a great foundation for the rest of the In Person work.

benefits/templates/admin/agency-dashboard-index.html
lalver1: 👍 removed it in 57b852291a8df3241570d194c2bf13bc1d7547d7

angela-tran (approved):
I left some comments about naming in the tests that you can commit if you agree with them.

And there is one comment that no action is needed on -- I'll create a separate ticket for it.

Nice work 👍 👍

tests/pytest/test_admin.py
```diff
@@ -0,0 +1,58 @@
...
+    response = BenefitsAdminSite().index(app_request)
+
+    assert response.status_code == 200
+    assert response.template_name == "admin/index.html"
+
+
+@pytest.mark.django_db
+def test_admin_override_super_user(app_request, model_SuperUser):
+    app_request.user = model_SuperUser
+    response = BenefitsAdminSite().index(app_request)
```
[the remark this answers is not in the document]
angela-tran: Created #2321

tests/pytest/test_admin.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+from django.conf import settings
+from django.contrib.auth.models import Group, User
+from benefits.admin import BenefitsAdminSite
+
+
+@pytest.fixture
+def model_User():
+    return User.objects.create(is_active="1", is_staff="1")
```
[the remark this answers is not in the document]
lalver1: Not at all, this makes sense to use `True` 👍 I was looking at the database and got confused because I saw `True` gets stored as `1`, but we are working with the Django model, not the database value and we should use `True` for setting the boolean.

angela-tran (approved):
🚀

machikoyasuda:
Woohoo! 🔥 🔥 🔥
Great job @lalver1 on a big PR!

PR #2315, opened 2024-08-21
Admin: Test that transit user does not get added to calitp group

### LINKED ISSUE #2242: Set `is_staff` to true for transit agency staff users
In order to access the admin site, a user must have [`is_staff`](https://docs.djangoproject.com/en/5.0/topics/auth/customizing/#django.contrib.auth.is_staff) set to true. 

Any user who can authenticate with an agency domain (via `*_ALLOWABLE_DOMAIN` env var) can log in (has the experience of a user that is not in any group, can do nothing) - i.e. is set to `is_staff=True`

Let's make sure to have unit tests covering this logic.

## Acceptance Criteria

- [ ] A user in the list is able to login with the currently supported SSO

machikoyasuda:
closes #2242 

This PR does not add any code, but it adds 2 tests. 

This PR's tests test the `add_staff_user_to_group` method, which runs in the `pre_login_user` callback. In these tests, we are testing a new scenario that will come up when the Admin app starts allowing Transit Agency admin users:

- Engineers add a new transit agency's domain to the `GOOGLE_SSO_ALLOWABLE_DOMAINS` list, like `cst.org`
- Users with a `cst.org` email address can create Admin accounts via the site's OAuth button.
- But, these users should not be added to the Cal-ITP group, because their email addresses are not specifically in the `GOOGLE_SSO_STAFF_LIST` list.

These 2 specs test out both scenarios of ensuring a transit agency user does not get added to the Cal-ITP group, and to ensure that people with email addresses in the `GOOGLE_SSO_STAFF_LIST` _do_ get added to this group properly.

### DIFF
--- a/tests/pytest/core/test_admin.py
+++ b/tests/pytest/core/test_admin.py
@@ -78,12 +78,12 @@ def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
 
 
 @pytest.mark.django_db
-def test_pre_login_user_does_not_add_transit_staff_to_group(mocker, model_AdminUser):
+def test_pre_login_user_does_not_add_transit_staff_to_group(mocker, model_AdminUser, settings):
     mocked_request = mocker.Mock()
     mocked_request.session.get.return_value = None
 
-    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_STAFF_LIST", [model_AdminUser.email])
-    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_ALLOWABLE_DOMAINS", ["cst.org"])
+    settings.GOOGLE_SSO_STAFF_LIST = [model_AdminUser.email]
+    settings.GOOGLE_SSO_ALLOWABLE_DOMAINS = ["cst.org"]
     agency_user = User.objects.create_user(username="agency_user", email="manager@cst.org", is_staff=True)
 
     pre_login_user(agency_user, mocked_request)
--- a/tests/pytest/core/test_admin.py
+++ b/tests/pytest/core/test_admin.py
@@ -92,21 +92,3 @@ def test_pre_login_user_does_not_add_transit_staff_to_group(mocker, model_AdminU
     staff_group = Group.objects.get(name=STAFF_GROUP_NAME)
     assert staff_group.user_set.count() == 0
     assert agency_user.groups.count() == 0
-
-
-@pytest.mark.django_db
-def test_pre_login_user_adds_calitp_staff_to_group(mocker, model_AdminUser):
-    mocked_request = mocker.Mock()
-    mocked_request.session.get.return_value = None
-
-    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_STAFF_LIST", [model_AdminUser.email])
-    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_ALLOWABLE_DOMAINS", ["cst.org"])
-    calitp_user = User.objects.create_user(username="calitp_user", email=model_AdminUser.email, is_staff=True)
-
-    pre_login_user(calitp_user, mocked_request)
-
-    # assert that a user whos email is in the STAFF_LIST does get added to group
-    staff_group = Group.objects.get(name=STAFF_GROUP_NAME)
-    assert staff_group.user_set.count() == 1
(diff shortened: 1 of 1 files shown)


tests/pytest/core/test_admin.py
```diff
@@ -75,3 +75,38 @@ def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
...
+    assert agency_user.groups.count() == 0
+
+
+@pytest.mark.django_db
+def test_pre_login_user_adds_calitp_staff_to_group(mocker, model_AdminUser):
+    mocked_request = mocker.Mock()
+    mocked_request.session.get.return_value = None
+
+    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_STAFF_LIST", [model_AdminUser.email])
+    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_ALLOWABLE_DOMAINS", ["cst.org"])
```
machikoyasuda, line 103: Technically, this line is not necessary here.

tests/pytest/core/test_admin.py
```diff
@@ -75,3 +75,38 @@ def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
...
     staff_group = Group.objects.get(name=STAFF_GROUP_NAME)
     assert model_AdminUser.groups.contains(staff_group)
+
+
+@pytest.mark.django_db
+def test_pre_login_user_does_not_add_transit_staff_to_group(mocker, model_AdminUser):
+    mocked_request = mocker.Mock()
+    mocked_request.session.get.return_value = None
+
+    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_STAFF_LIST", [model_AdminUser.email])
```
thekaveman, line 85: To override settings for a test, we want to use the `settings` fixture provided by `pytest-django` (already installed): https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

See e.g. https://github.com/cal-itp/benefits/blob/main/tests/pytest/test_secrets.py#L80

tests/pytest/core/test_admin.py
thekaveman: Yes let's go ahead and remove it if it is not relevant to the test.

tests/pytest/core/test_admin.py
```diff
@@ -75,3 +75,38 @@ def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
...
+    pre_login_user(agency_user, mocked_request)
+
+    # assert that a transit agency user does not get added to the Cal-ITP user group
+    staff_group = Group.objects.get(name=STAFF_GROUP_NAME)
+    assert staff_group.user_set.count() == 0
+    assert agency_user.groups.count() == 0
+
+
+@pytest.mark.django_db
+def test_pre_login_user_adds_calitp_staff_to_group(mocker, model_AdminUser):
```
thekaveman, line 98: This test feels really similar to the existing test? [`test_pre_login_user_add_staff_to_group`](https://github.com/cal-itp/benefits/blob/main/tests/pytest/core/test_admin.py#L68).

I.e. it's creating a new user but the email address is the same, and then the assertions are all essentially the same?

thekaveman (commented):
Some minor syntax comments.

tests/pytest/core/test_admin.py
machikoyasuda: Oh yeah it is!

angela-tran (changes requested):
I think `test_pre_login_user_does_not_add_transit_staff_to_group` is definitely the test coverage we're looking for for this ticket. I noticed some of the mocking of the settings could be cleaned up a little. Suggestions below

tests/pytest/core/test_admin.py
angela-tran: Agree with both comments

tests/pytest/core/test_admin.py
angela-tran: Ah good catch. I think this means `test_pre_login_user_add_staff_to_group`, which existed before this PR, should do the same

angela-tran: Created https://github.com/cal-itp/benefits/issues/2316 for the test unrelated to this PR

angela-tran (approved):
This looks great @machikoyasuda , thank you!

PR #2317, opened 2024-08-22
Admin: Eligibility View

### LINKED ISSUE #2269: Transit agency staff should be able to select the eligibility they are verifying and indicate their verification
## Acceptance Criteria

- [ ]

machikoyasuda:
closes #2269 

http://localhost:11369/in_person/eligibility/

Light mode
image

Dark mode
image

## How to test
- Follow https://github.com/cal-itp/benefits/pull/2325#issue-2489892807 to get into the dashboard
- Click New Enrollment button
- Test clicking submit without choosing anything
- Test clicking submit with only checkbox
- Test clicking submit with only radio button
- Use <kbd>enter</kbd> or <kbd>spacebar</kbd> and <kbd>tab</kbd> for form completion and submission
- Use <kbd>tab</kbd> to skip to main content correctly
- Accessibility: Passes WCAG AA (except for 1 known ARIA form issue coming from `form.html` - which isn't going to be changed in this PR)
- Check that the Console warning area is empty

### Known issues/questions
- The log out user tools area does not get rendered.

### Form
- Has jQuery library loaded
- Does not have recaptcha
- Uses Django base form inputs (`RadioSelect` and `BooleanField`) on top of `form.html` from `core`
- Uses HTML5 validation and Django default back-end validation
- Has some hacky CSS written for styles, instead of adding utility classes through Django or adding logic to the Django template. 
- This PR aims to not make any changes to `forms.html` and just go with defaults. The required `*` is black because the `forms.html` utility classes make it so, and overriding a style that already has an `!important` style declaration is kinda messy. So we're keeping it black for now.

### Copy
- Is not internationalized.

### Tests
- [x] Write tests for form submission

### CSS
- Adds an Outline Button style, use with `"btn btn-lg btn-outline-primary"` (On view and hover)
image
image

### DIFF
--- a/benefits/static/css/admin/styles.css
+++ b/benefits/static/css/admin/styles.css
@@ -2,6 +2,7 @@
 
 /* Buttons */
 /* Primary Button: Use all three classes: btn btn-lg btn-primary */
+/* Outline Primary Button: Use all three classes: btn btn-lg btn-outline-primary */
 /* Set button width in parent with Bootstrap column */
 /* Height: 60px on Desktop; 72 on mobile*/
 
@@ -15,8 +16,16 @@
   }
 }
 
+.btn.btn-lg.btn-outline-primary:not(:hover) {
+  color: var(--primary-color);
+}
+
 .btn.btn-lg.btn-primary {
   background-color: var(--primary-color);
+}
+
+.btn.btn-lg.btn-primary,
+.btn.btn-lg.btn-outline-primary {
   border-color: var(--primary-color);
   border-width: 2px;
   font-weight: var(--medium-font-weight);
@@ -26,7 +35,8 @@
   padding: var(--primary-button-padding);
 }
 
-.btn.btn-lg.btn-primary:hover {
+.btn.btn-lg.btn-primary:hover,
+.btn.btn-lg.btn-outline-primary:hover {
   background-color: var(--hover-color);
   border-color: var(--hover-color);
 }
@@ -62,4 +72,20 @@ html[data-theme="light"],
 #user-tools,
 #logout-form button {
   text-transform: unset;
+}
+
+.checkbox-parent .form-group:last-of-type .col-12 {
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: start;
+  column-gap: 0.5rem;
+  margin-top: 2rem;
+}
+
+.checkbox-parent,
+.checkbox-parent .form-group .col-12,
+.checkbox-parent .form-group .col-12 #id_flow {
+  display: flex;
+  flex-direction: column;
+  row-gap: 1rem;
 }
--- a/tests/pytest/in_person/test_views.py
+++ b/tests/pytest/in_person/test_views.py
@@ -5,6 +5,15 @@ import pytest
 from benefits.routes import routes
 
 
+@pytest.fixture
+def mocked_eligibility_auth_request(mocked_eligibility_request_session, mocked_session_agency):
+    """
+    Stub fixture combines mocked_eligibility_request_session and mocked_session_oauth_token
+    so that session behaves like in an authenticated request to the eligibility app
+    """
+    pass
+
+
 @pytest.mark.django_db
 @pytest.mark.parametrize("viewname", [routes.IN_PERSON_ELIGIBILITY, routes.IN_PERSON_ENROLLMENT])
 def test_view_not_logged_in(client, viewname):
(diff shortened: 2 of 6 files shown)


benefits/static/css/admin/styles.css
```diff
@@ -63,3 +73,19 @@ html[data-theme="light"],
...
+  margin-top: 2rem;
+}
+
+.checkbox-parent,
+.checkbox-parent .form-group .col-12,
+.checkbox-parent .form-group .col-12 #id_flow {
+  display: flex;
+  flex-direction: column;
+  row-gap: 1rem;
+}
```
machikoyasuda, line 91: As utility classes: `d-flex flex-row-reverse justify-content-start gap-0 row-gap-1` 

`d-flex flex-column gap-0 row-gap-3` 

But just decided to do it like this for now. Named `checkbox-parent` because the main thing this does is flip the checkbox rendering to show the checkbox input first, then label, which is the opposite of the rest of the pages.

benefits/eligibility/forms.py
```diff
@@ -42,6 +42,33 @@ def clean(self):
...
+
+    cancel_url = routes.ADMIN_INDEX
+
+    def __init__(self, agency: models.TransitAgency, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        flows = agency.enrollment_flows.all()
+
+        self.classes = "checkbox-parent"
+        flow_field = self.fields["flow"]
+        flow_field.choices = [(f.id, f.label) for f in flows]
```
machikoyasuda, line 63: This sets the form field's label as the `label`, which are currently named like this:

```
(CST) Senior Discount
CalFresh
(CST) Veteran Discount
(CST) Agency Card Discount
```

It's slightly different from what we want (Older adult, CalFresh card holder, Courtesy Card holder, etc.) - which are specified in these files for the Benefits app: https://github.com/cal-itp/benefits/blob/1e2ff89cfcdbcddc6895701399aaf1702b3b2fc1/benefits/eligibility/templates/eligibility/includes/selection-label--senior.html

 But didn't want to bog down this PR with implementing all of that.

benefits/eligibility/forms.py
```diff
@@ -42,6 +42,29 @@ def clean(self):
...
 
+class InPersonEligibilityForm(forms.Form):
+    """Form to capture enrollment flow selection."""
+
+    action_url = routes.IN_PERSON_ENROLLMENT
+    id = "form-flow-selection"
+    method = "POST"
+
+    flow = forms.ChoiceField(label="Choose an eligibility type to qualify this rider.", widget=forms.widgets.RadioSelect)
+    verified = forms.BooleanField(label="I have verified this person’s eligibility for a transit benefit.", required=True)
```
machikoyasuda, line 53: Adding `required=True` will add a validation check 

Documentation: https://docs.djangoproject.com/en/5.1/ref/forms/fields/#booleanfield

image


Code: https://github.com/django/django/blob/stable/5.1.x/django/forms/fields.py#L838

machikoyasuda:
Status update: Everything in this PR is finished, commented/documented and tested _except_ I have not yet written specs for the form. The one other "unknown" I haven't yet gotten around to figuring out is why the admin user's name and the sign out link doesn't appear, as noted above.

angela-tran:
@machikoyasuda You can do something similar to https://github.com/cal-itp/benefits/pull/2327/commits/e9614fe12edd467d5721db68c6cfb648340a297c to get the usertools to show up.

angela-tran:
I think another unknown is why the custom validity messages are not working

benefits/eligibility/forms.py
```diff
@@ -42,6 +42,29 @@ def clean(self):
             raise forms.ValidationError("reCAPTCHA failed")
 
 
+class InPersonEligibilityForm(forms.Form):
```
angela-tran, line 45: Small note for code organization: let's have this in a file called `benefits/in_person/forms.py`

benefits/eligibility/forms.py
angela-tran: Created separate issue https://github.com/cal-itp/benefits/issues/2336 to cover this

angela-tran:
> I think another unknown is why the custom validity messages are not working

Created separate issue https://github.com/cal-itp/benefits/issues/2337 to cover this

benefits/eligibility/forms.py
machikoyasuda: Done here: https://github.com/cal-itp/benefits/pull/2317/commits/563b134f85bfbd34c56470aefe2abfa9e07bcc70

angela-tran (commented):
This looks great @machikoyasuda! I think we can remove an unused fixture, so after that, this is ready to merge

tests/pytest/in_person/test_views.py
```diff
@@ -5,6 +5,15 @@
 from benefits.routes import routes
 
 
+@pytest.fixture
+def mocked_eligibility_auth_request(mocked_eligibility_request_session, mocked_session_agency):
```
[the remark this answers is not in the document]
machikoyasuda: oops no! i forgot about this. thank you for catching.

angela-tran (approved):
🎉

PR #2323, opened 2024-08-26
Feat: create base template for in-person eligibility

lalver1:
Part of #2246

This PR creates a base template which defines the header navbar and includes the styles that every in-person eligibility/enrollment page will need at a minimum. It extends Django's [admin base template](https://github.com/django/django/blob/main/django/contrib/admin/templates/admin/base.html). 

The `agency-dashboard-index.html` template is also refactored to use this new base template.

### DIFF
--- a/benefits/templates/admin/agency-dashboard-index.html
+++ b/benefits/templates/admin/agency-dashboard-index.html
@@ -1,50 +1,9 @@
-{% extends "admin/base.html" %}
+{% extends "admin/agency-base.html" %}
 {% load i18n static %}
-
-{% block extrastyle %}
-    <link href="{% static "css/admin/styles.css" %}" rel="stylesheet">
-    <link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
-    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
-          rel="stylesheet"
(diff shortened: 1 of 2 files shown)


benefits/in_person/templates/in_person/base.html
angela-tran, line 1: The contents of this file look good. I think it makes more sense as a template under `benefits/templates/admin` though.

benefits/templates/admin/agency-dashboard-index.html
```diff
@@ -1,51 +1,10 @@
-{% extends "admin/base.html" %}
+{% extends "in_person/base.html" %}
```
angela-tran, line 1: Going along with the comment above, this would become

```suggestion
{% extends "admin/base.html" %}
```

angela-tran (changes requested):

benefits/in_person/templates/in_person/base.html
angela-tran: Hmm... I guess that could cause the templates from the Django admin site to end up using this file too. So maybe we still move this under `benefits/templates/admin` but have the filename be `agency-base.html`.

lalver1: Yes! Sorry, I forgot we had looked at the location so I'll go ahead and move it there 👍

angela-tran (approved):

PR #2324, opened 2024-08-27
Feat: implement the top-right usertools block

### LINKED ISSUE #2246: Transit agency staff should see a home dashboard upon logging in

lalver1:
Closes #2246

This PR is the last part of #2246 and it implements the top right usertools block for in-person eligibility/enrollment pages. It overrides the `text-transform` property it inherits from Django's base admin site so that the welcome message is capitalized (WELCOME, USER) but the sign out message isn't (Sign out). ~It uses the same idea to remove the underline from `Sign out`.~ In addition, it adds a [Bootstrap icon](https://icons.getbootstrap.com/icons/box-arrow-right/) and uses it as the `Sign out` icon.

### DIFF
--- a/benefits/static/css/admin/styles.css
+++ b/benefits/static/css/admin/styles.css
@@ -58,3 +58,8 @@ html[data-theme="light"],
   --link-fg: #ffffff;
   --body-quiet-color: var(--dark-color);
 }
+
+#user-tools,
+#logout-form button {
+  text-transform: unset;
+}
--- a/benefits/templates/admin/agency-base.html
+++ b/benefits/templates/admin/agency-base.html
@@ -17,8 +17,8 @@
     <div id="header-container" class="navbar navbar-expand-sm navbar-dark justify-content-between">
         <div class="container">
             <span class="navbar-brand p-0">
-                <img class="sm d-lg-none" src="{% static "img/logo-sm.svg" %}" width="90" height="51.3" alt="{% translate "California Integrated Travel Project: Benefits logo (small)" context "image alt text" %}" />
-                <img class="lg d-none d-lg-block" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="{% translate "California Integrated Travel Project: Benefits logo (large)" context "image alt text" %}" />
+                <img class="sm d-lg-none" src="{% static "img/logo-sm.svg" %}" width="90" height="51.3" alt="California Integrated Travel Project: Benefits logo (small)" />
+                <img class="lg d-none d-lg-block" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="California Integrated Travel Project: Benefits logo (large)" />
             </span>
         </div>
         <h1 class="p-0 m-0 text-white">Administrator</h1>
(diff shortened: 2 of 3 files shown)


thekaveman:
I edited the title, please don't put code like that in PR titles.

benefits/static/css/admin/styles.css
```diff
@@ -58,3 +58,16 @@ html[data-theme="light"],
   --link-fg: #ffffff;
   --body-quiet-color: var(--dark-color);
 }
+
+#user-tools,
+#logout-form button {
+  padding: 0;
+  font-weight: 300;
+  font-size: 0.6875rem;
+  letter-spacing: 0.5px;
```
machikoyasuda, line 67: I believe this CSS Is unneccessary. These styles are already being set by `base.css`. The only one necessary is `text-transform: unset`.

benefits/templates/admin/agency-base.html
```diff
@@ -26,6 +26,31 @@ <h1 class="p-0 m-0 text-white">Administrator</h1>
...
+                        <a href="{{ docsroot }}">{% translate "Documentation" %}</a> /
+                    {% endif %}
+                {% endif %}
+                {% if user.has_usable_password %}
+                    <a href="{% url 'admin:password_change' %}">{% translate "Change password" %}</a> /
+                {% endif %}
+                <img class="icon" width="15" height="15" src="{% static "img/icon/box-arrow-right.svg" %}" alt="" />
+                <form id="logout-form" method="post" action="{% url 'admin:logout' %}">
+                    {% csrf_token %}
+                    <button type="submit">{% translate "Sign out" %}</button>
```
machikoyasuda, line 49: Adding a `class="border-0"` will remove the border. Then you can delete: https://github.com/cal-itp/benefits/pull/2324/files#diff-77a3765ab5e3cdb49fb5b92ca214d69e2821102f48112dc9eba1a9429ec902fcR71-R73

benefits/templates/admin/agency-base.html
```diff
@@ -26,6 +26,31 @@ <h1 class="p-0 m-0 text-white">Administrator</h1>
 {% endblock branding %}
 
 {% block usertools %}
+    {% if has_permission %}
+        <div id="user-tools">
+            {% block welcome-msg %}
+                <span class="text-uppercase text-white">
+                    {% translate "Welcome," %}
+                    <strong>{% firstof user.get_short_name user.get_username %}</strong>. </span>
```
machikoyasuda, line 34: `<span class="fw-bold">` instead of `<strong>` - the <strong> doesn't make the font bold for me.

benefits/static/css/admin/styles.css
machikoyasuda: If we're okay with `Sign Out` instead of `Sign out`, we can use Bootstrap's `text-capitalize` class instead. Unfortunately there is no `text-unset` class for `text-transform: unset`...... Then we could get all the styles in via utility classes (the preferred method).

machikoyasuda:
Just some small changes @lalver1  - otherwise looks good to me. Ping me if you have any Qs!

benefits/templates/admin/agency-base.html
```diff
@@ -26,6 +26,31 @@ <h1 class="p-0 m-0 text-white">Administrator</h1>
 {% endblock branding %}
 
 {% block usertools %}
+    {% if has_permission %}
+        <div id="user-tools">
+            {% block welcome-msg %}
+                <span class="text-uppercase text-white">
+                    {% translate "Welcome," %}
```
angela-tran, line 33: To be consistent with how copy is maintained in our admin / in-person templates, I suggest not using `translate` at all in this template.

I realize you probably just grabbed this from the original Django template and might be thinking of our previous discussion about preserving original functionality if it doesn't hurt anything. But in this case with copy, we've said we don't need to worry about translation for now.

angela-tran (changes requested):

benefits/static/css/admin/styles.css
lalver1: That's true, thanks for pointing that out. I only left the necessary CSS so that we match the capitalization in the mock-up (`Sign out`).

benefits/templates/admin/agency-base.html
lalver1: Ah, that makes sense. Cleaned it up in aa73dca7d3981c5bd06504cb9aa5d23011dfc39e and 51f2462ab1d09b4a0ca1bc12e45b9f3612c45e25

benefits/templates/admin/agency-base.html
lalver1: Thanks, made the change in aa73dca7d3981c5bd06504cb9aa5d23011dfc39e

angela-tran (approved):
Thanks @lalver1! Let's make sure @machikoyasuda approves too

benefits/templates/admin/agency-base.html
lalver1: Ah that's right, thanks for reminding me about that 👍 I made the change in aa73dca7d3981c5bd06504cb9aa5d23011dfc39e.

machikoyasuda (approved):
LGTM 💯 💯 💯

PR #2325, opened 2024-08-27
Feat: automatically associate user to agency

### LINKED ISSUE #2243: Ensure user is associated to a `TransitAgency` before they can access in-person enrollment views

angela-tran:
Closes #2243 

This PR makes it so that users are added to their transit agency's `staff_group` upon login. This is done with the help of a field called `sso_domain` on the transit agency.

The user needs to be manually added to their transit agency's `customer_service_group` to see the part of the dashboard for "In-person enrollment".

## Testing locally

- Set your `GOOGLE_SSO_STAFF_LIST` environment variable to `*`
- Make sure your `GOOGLE_SSO_ALLOWABLE_DOMAINS` environment variable is `compiler.la`

Those two steps make it so your Google account will be seen as a "transit agency staff user" - i.e. not put into the "Cal-ITP" group but still marked as `is_staff`.

- Set `SSO_SHOW_FORM_ON_ADMIN_PAGE` to `True` temporarily in `benefits/settings.py`
- Reset your database to use local fixtures - `./bin/reset_db.sh`
- Set up the SSO domain mapping
   - Launch the app and go to `/admin` and login as your superuser (based on your [`DJANGO_SUPERUSER_USERNAME`](https://docs.calitp.org/benefits/configuration/environment-variables/#django_superuser_username) and `DJANGO_SUPERUSER_PASSWORD` environment variables)
   - Edit the "CST" `TransitAgency` so that it has an `sso_domain` of `compiler.la`
   - Create a new group and set it as the `staff_group` for CST
   - Create another new group and set it as the `customer_service_group` for CST
- Log out
- Log back in using your compiler.la Google account

**Expected result:** your user has been added to the `staff_group` for CST upon login. When you go to the agency dashboard, you see the transit agency name, but no button or anything for in-person enrollment.

To elevate the user to having permission to do in-person enrollment:
- Log in as the superuser
- Add your Google user to the CST customer service group
- Log back in as your Google user

**Expected result:** When you go to the agency dashboard, you see the the transit agency name as well as the in-person enrollment section.

## Screenshots
User in CST's `staff_group` but not the `customer_service_group`

User in CST's `staff_group` and `customer_service_group`


machikoyasuda:
Starting to review now

machikoyasuda:
image

machikoyasuda (dismissed):
LGTM! Thanks for the instructions.

(When I first tested this with your instructions, I duh forgot to run `source .env` so it didn't work as expected. After I ran `source` and confirmed the env vars changed, it worked.)

benefits/core/admin.py
```diff
@@ -151,3 +152,11 @@ def add_staff_user_to_group(user, request):
...
         staff_group = Group.objects.get(name=settings.STAFF_GROUP_NAME)
         user.groups.add(staff_group)
+
+
+def add_transit_agency_staff_user_to_group(user, request):
+    user_sso_domain = user.email.split("@")[1]
+    if user_sso_domain:
+        agency = models.TransitAgency.objects.filter(sso_domain=user_sso_domain).first()
+        if agency is not None and agency.staff_group:
+            user.groups.add(agency.staff_group)
```
thekaveman, line 162: This feels backwards? Why don't we add the user to the staff group? I guess it results in the same thing, but just looks a little strange.

benefits/core/admin.py
```diff
@@ -151,3 +152,11 @@ def add_staff_user_to_group(user, request):
     if user.email in settings.GOOGLE_SSO_STAFF_LIST:
         staff_group = Group.objects.get(name=settings.STAFF_GROUP_NAME)
         user.groups.add(staff_group)
```
thekaveman, line 154: I guess it's the same thing here...

benefits/core/admin.py
angela-tran: Yeah, it can be done either way. I can change it

machikoyasuda: It is a little weird code-wise. But it also kinda makes sense visually in the Admin tool too, because in order to add a User into a Group in the Admin, I go find the user (http://localhost:11369/admin/auth/user/3/change/) and then add or remove the User's group from this list:
image

machikoyasuda: Groups view - Pretty much only has Group name and permissions. You can't even see who is in the group from this view by default
image

thekaveman: Thanks for those screenshots @machikoyasuda. Either way, whatever makes the most sense to everyone.

angela-tran: I went ahead and changed it in https://github.com/cal-itp/benefits/pull/2325/commits/04d3c194d80ae4be4afffd92fefeef986410e93a which I think does match the function name better.

And thanks for those screenshots @machikoyasuda to note how the admin interface presents it.

benefits/core/admin.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/2325/commits/04d3c194d80ae4be4afffd92fefeef986410e93a

thekaveman (approved):
Nice 👍

angela-tran:
Thank you both!

PR #2329, opened 2024-08-29
Chore: Admin - Add jQuery to agency base template

### LINKED ISSUE #2328: Add jQuery to Agency Base
Both #2327 and #2317 need jQuery, so we should add it to Agency Base.

machikoyasuda:
closes #2328 

This PR adds jQuery to the `agency-base` template, so all Admin pages that use this template will have access to jQuery. jQuery is used in `form.html` - the main form base template that the app uses for all forms.

### DIFF
--- a/benefits/templates/admin/agency-base.html
+++ b/benefits/templates/admin/agency-base.html
@@ -8,6 +8,10 @@
           rel="stylesheet"
           integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
           crossorigin="anonymous">
+    <script nonce="{{ request.csp_nonce }}"
+            src="https://cdn.jsdelivr.net/npm/jquery@3.6.1/dist/jquery.min.js"
+            integrity="sha256-o88AwQnZB+VDvE9tvIXrMQaPlFFSUTR+nldQm1LuPXQ="
+            crossorigin="anonymous"></script>
 {% endblock extrastyle %}
 
 {% block title %}


benefits/templates/admin/agency-base.html
```diff
@@ -8,6 +8,10 @@
           rel="stylesheet"
           integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
           crossorigin="anonymous">
+    <script nonce="{{ request.csp_nonce }}"
+            src="https://cdn.jsdelivr.net/npm/jquery@3.6.1/dist/jquery.min.js"
+            integrity="sha256-o88AwQnZB+VDvE9tvIXrMQaPlFFSUTR+nldQm1LuPXQ="
+            crossorigin="anonymous"></script>
 {% endblock extrastyle %}
```
machikoyasuda, line 16: @angela-tran Should I also add this line from `/calitp/app/benefits/core/templates/core/base.html` to pull in the recaptcha script, in this PR?

```liquid
    {% if request.recaptcha %}
      <script nonce="{{ request.csp_nonce }}" src="{{ request.recaptcha.script_api }}"></script>
    {% endif %}
```

machikoyasuda: I would have preferred to be able to add this to the very bottom of the `<body>` tag, but there isn't an easy place to override and add this.

thekaveman: FWIW, I think we can ignore reCAPTCHA for this effort -- the user must have access to the Admin, and I think we can assume they are legit if they can get in and initiate an in-person flow.

angela-tran (approved):
👍

PR #2327, opened 2024-08-28
Admin: Enrollment View

angela-tran:
Closes #2244 

WIP

image

PR #2338, opened 2024-09-05
Refactor: introduce enrollment module

angela-tran:
Part of #2327 

Refactors out logic from `benefits.enrollment.views` for calling the transit processor API out into a separate module. This will make it easier to reuse the logic for the in-person enrollment view.

Uses an [`Enum`](https://docs.python.org/3/library/enum.html) as the result of the enrollment attempt.

## Manual testing
There shouldn't be any regressions to enrollment in the Benefits app.


lalver1 (dismissed):
Looks good! 👍 I tested locally (also simulated a 500-level HTTP error) and enrollment worked as expected.

thekaveman (approved):
This is such a clean refactor, well done :100: :1st_place_medal: 

I only have 2 non-blocking comments:

1. It would be helpful to have some docstrings for the new `enroll` helper, since that is used by callers external to the `enrollment` module.
2. It seems like the `token` view function could also be refactored in a similar way, and re-use the same `Status` enum concept. 

One or both of these can definitely be follow-ups.

angela-tran:
> 1. It would be helpful to have some docstrings for the new `enroll` helper, since that is used by callers external to the `enrollment` module.

Good idea. I added that in [e4b4d7a](https://github.com/cal-itp/benefits/pull/2338/commits/e4b4d7a72ce8814d0f23be98f0970750f21ca23a) - open to any suggestions or tweaks

> 2. It seems like the `token` view function could also be refactored in a similar way, and re-use the same `Status` enum concept. 

Yep, definitely. I think it makes sense as a follow-up PR. It is definitely needed for completeness in handling system errors and server errors, which I will create separate tickets to cover.

lalver1 (approved):
👍

PR #2347, opened 2024-09-10
Admin test: Use settings instead of mocker

### LINKED ISSUE #2316: Chore: use `settings` fixture in unit test `test_pre_login_user_add_staff_to_group`
> Ah good catch. I think this means `test_pre_login_user_add_staff_to_group`, which existed before this PR, should do the same

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/pull/2315#discussion_r1725950151_

machikoyasuda:
fixes #2316

### DIFF
--- a/tests/pytest/core/test_admin.py
+++ b/tests/pytest/core/test_admin.py
@@ -1,5 +1,4 @@
 import pytest
-from django.conf import settings
 from django.contrib.auth.models import User, Group
 
 import benefits.core.admin
@@ -67,7 +66,7 @@ def test_pre_login_user_no_session_token(mocker, model_AdminUser):
 
 
 @pytest.mark.django_db
-def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
+def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser, settings):
     mocked_request = mocker.Mock()
     mocked_request.session.get.return_value = None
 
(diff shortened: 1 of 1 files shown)


angela-tran (dismissed):
👍

tests/pytest/core/test_admin.py
```diff
@@ -71,8 +71,7 @@ def test_pre_login_user_add_staff_to_group(mocker, model_AdminUser):
     mocked_request = mocker.Mock()
     mocked_request.session.get.return_value = None
 
-    mocker.patch.object(benefits.core.admin.settings, "GOOGLE_SSO_STAFF_LIST", [model_AdminUser.email])
-
+    settings.GOOGLE_SSO_STAFF_LIST = [model_AdminUser.email]
```
thekaveman, line 74: I don't think this is quite right. This is the `settings` object imported from `django.conf`

It should be the `settings` fixture from `pytest-django` (e.g. injected to the test method).

See https://github.com/cal-itp/benefits/blob/main/tests/pytest/test_secrets.py#L80

thekaveman (changes requested):

tests/pytest/core/test_admin.py
machikoyasuda: https://github.com/cal-itp/benefits/pull/2347/commits/d35edf76ad79dfa2d958f265da998e1cd7368954

thekaveman (approved):
Nice :+1:

Issue #2244, opened 2024-07-23
Implement `enrollment` view and template


angela-tran:
Depends on #2284

This view allows the user to enroll their payment card to receive their transit benefit using the Littlepay hosted verification flow.

## Acceptance Criteria

- [x] User is able to enroll their payment card using the Littlepay hosted verification flow

angela-tran:
I was thinking of the success template as a part of this ticket and decided it should be a separate PR so that #2340 wasn't too big.

PR #2348, opened 2024-09-10
Refactor: enrollment token logic

angela-tran:
Part of #2244 

Extracts logic from `benefits.enrollment.token` view into `benefits.enrollment.enrollment` module. This is needed for the in-person enrollment view.

Similar to #2338 

## Testing locally

There shouldn't be any regressions with enrollment in the Benefits app.

### DIFF
--- a/benefits/enrollment/enrollment.py
+++ b/benefits/enrollment/enrollment.py
@@ -33,7 +33,7 @@ class CardTokenizationAccessResponse:
     status_code: int = None
 
 
-def request_card_tokenization_access(request):
+def request_card_tokenization_access(request) -> CardTokenizationAccessResponse:
     """
     Requests an access token to be used for card tokenization.
     """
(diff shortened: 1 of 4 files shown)


benefits/enrollment/enrollment.py
```diff
@@ -23,6 +24,53 @@ class Status(Enum):
...
+@dataclass
+class CardTokenizationAccessResponse:
+    status: Status
+    access_token: str
+    expires_at: str
+    exception: Exception = None
+    status_code: int = None
+
+
+def request_card_tokenization_access(request):
```
thekaveman, line 36: Very minor but it would be great to have a type hint here for the return type.

benefits/enrollment/enrollment.py
```diff
@@ -23,6 +24,53 @@ class Status(Enum):
...
+        else:
+            status_code = None
+            status = Status.EXCEPTION
+
+    return CardTokenizationAccessResponse(
+        status=status, access_token=None, expires_at=None, exception=exception, status_code=status_code
+    )
+
+
 def enroll(request, card_token):
```
thekaveman, line 74: Also here for the type hint, since it is tuple of values.

thekaveman (changes requested):
One hopefully small request

thekaveman (approved):
Nice :+1:

PR #2360, opened 2024-09-11
Enrollment success: All flows now have same success headline copy

### LINKED ISSUE #2353: `enrollment:success` - update headline for all success templates
We want to use the copy from the Agency card success templates:

> You can now use your contactless card to tap to ride with a reduced fare!

This is already translated into Spanish as it is the live copy in production.

Remove any unused strings from the PO files.

machikoyasuda:
close #2353


lalver1 (approved):
Looks good! At first I was wondering about the use of `blocktranslate trimmed` in `success.html` but I see that it makes for a clean `.po` file because of the indentation 👍

angela-tran (approved):
This LGTM!

PR #2361, opened 2024-09-11
Feat: customize the login form using an environment variable

lalver1:
This PR sets the login form configuration using an environment variable.
Before merging this PR, ensure that:
- [ ] `dev` has a new secret in the key vault and a new environment variable that refers to this secret in the app settings
- [ ] `test` has a new secret in the key vault and a new environment variable that refers to this secret in the app settings
- [ ] `prod` has a new secret in the key vault and a new environment variable that refers to this secret in the app settings

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -72,7 +72,7 @@ GOOGLE_SSO_SCOPES = [
     "https://www.googleapis.com/auth/userinfo.email",
     "https://www.googleapis.com/auth/userinfo.profile",
 ]
-SSO_SHOW_FORM_ON_ADMIN_PAGE = False
+SSO_SHOW_FORM_ON_ADMIN_PAGE = os.environ.get("SSO_SHOW_FORM_ON_ADMIN_PAGE", "False").lower() == "true"
 STAFF_GROUP_NAME = "Cal-ITP"
 
 MIDDLEWARE = [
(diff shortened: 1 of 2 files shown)


benefits/settings.py
```diff
@@ -72,7 +72,7 @@ def RUNTIME_ENVIRONMENT():
     "https://www.googleapis.com/auth/userinfo.email",
     "https://www.googleapis.com/auth/userinfo.profile",
 ]
-SSO_SHOW_FORM_ON_ADMIN_PAGE = False
+SSO_SHOW_FORM_ON_ADMIN_PAGE = os.environ.get("SSO_SHOW_FORM_ON_ADMIN_PAGE", "False").lower() == "true"
```
machikoyasuda, line 75: https://megalus.github.io/django-google-sso/customize/#hiding-the-login-form

machikoyasuda (approved):
Tested:
image

angela-tran (approved):
This LGTM!

angela-tran (changes requested):
Luis noticed that there is a missing comma in the line above the one he added 😅 so I'm dismissing my review

angela-tran (approved):

lalver1:
It's interesting @angela-tran , looks like commas are not required according to the [terraform docs](https://developer.hashicorp.com/terraform/language/expressions/types#maps-objects) since [app_settings](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/app_service#app_settings) are a key-value pair, but it's prob best to include them (it also makes the VS Code syntax highlighting look uniform).

PR #2362, opened 2024-09-12
Refactor: use client to test the BenefitsAdminSite view

### LINKED ISSUE #2321: Chore: refactor unit tests of AdminSite view functions
I think to be more consistent with the way we usually test our view functions, we would use a `client` to `.get` the path for the view function. 

So I think it would be something like:
```
path = routes.ADMIN_INDEX
client.force_login(model_SuperUser)
client.get(path)
```

instead of calling the view function directly.

What you have here works though and provides the right test coverage, so I think I'll make a separate chore ticket for this.

(pytest-django docs on [`force_login`](https://pytest-django.readthedocs.io/en/latest/helpers.html#client-django-test-client))

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/pull/2313#discussion_r1731833307_

lalver1:
Closes #2321

This PR refactors the test for `BenefitsAdminSite` to use a `client` to `.get` the path for the view function instead of calling the view function directly.

### DIFF
--- a/tests/pytest/test_admin.py
+++ b/tests/pytest/test_admin.py
@@ -1,7 +1,8 @@
 import pytest
 from django.conf import settings
 from django.contrib.auth.models import Group, User
-from benefits.admin import BenefitsAdminSite
+from django.urls import reverse
+from benefits.routes import routes
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


machikoyasuda (approved):
This is much cleaner/readable!

lalver1:
I agree, using `client` is much nicer! 🙂

PR #2371, opened 2024-09-16
Admin: Log in page

### LINKED ISSUE #2265: Add styling to login page
We should add some styling to the login page which aligns with the general Cal-ITP Benefits brand to avoid creating doubt with transit agency staff users.

Depends on #2264 

Cal-ITP Benefits_ Admin Interface - Login

machikoyasuda:
closes #2265 

image
image
image

## What this PR does
- Sets `site_header`, `site_title` for BenefitsAdminSite
- CSS: Add default Benefit focus states to login inputs, buttons
- CSS: Removes darkmode from Agency Base and Login
- Templates: Login and Agency Base - Moves Django's Login.css code to Styles.css and changes some of the style declarations
- Login template: Creates login.html to override parts of Django's login.html's branding section to add the Cal-ITP image.

## How to test
- Test with `SSO_SHOW_FORM_ON_ADMIN_PAGE` set to False 
- Test with `SSO_SHOW_FORM_ON_ADMIN_PAGE` set to True
- Test with an error/info/warning message
- Test on dark (browser set to dark mode) and light mode - Should only show light mode
- Test focus states
- Test with a wrong password, error message should show up

This PR also resolves all outstanding WCAG 2.0 issues: 
image

### DIFF
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -46,7 +46,3 @@ class BenefitsAdminSite(admin.AdminSite):
                 )
 
             return TemplateResponse(request, "admin/agency-dashboard-index.html", context)
-
-
-BenefitsAdminSite.site_header = "Administrator"
-BenefitsAdminSite.site_title = "Cal-ITP Benefits - Administrator"
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -46,3 +46,7 @@ class BenefitsAdminSite(admin.AdminSite):
                 )
 
             return TemplateResponse(request, "admin/agency-dashboard-index.html", context)
+
+
+BenefitsAdminSite.site_header = "Administrator"
+BenefitsAdminSite.site_title = "Cal-ITP Benefits - Administrator"
--- a/benefits/static/css/admin/styles.css
+++ b/benefits/static/css/admin/styles.css
@@ -1,5 +1,16 @@
 @import "../variables.css";
 
+/* CSS Variable overrides */
+
+html[data-theme="light"],
+:root {
+  --primary: #ffffff;
+  --secondary: var(--dark-color);
+  --link-fg: #ffffff;
+  --body-quiet-color: var(--dark-color);
+  --button-bg: #ffffff;
+}
+
 /* Buttons */
 /* Primary Button: Use all three classes: btn btn-lg btn-primary */
 /* Outline Primary Button: Use all three classes: btn btn-lg btn-outline-primary */
@@ -20,11 +31,13 @@
   color: var(--primary-color);
 }
 
+.login input[type="submit"],
 .btn.btn-lg.btn-primary {
   background-color: var(--primary-color);
 }
 
 .btn.btn-lg.btn-primary,
+.login input[type="submit"],
 .btn.btn-lg.btn-outline-primary {
   border-color: var(--primary-color);
   border-width: 2px;
@@ -36,6 +49,7 @@
 }
 
 .btn.btn-lg.btn-primary:hover,
+.login input[type="submit"]:hover,
 .btn.btn-lg.btn-outline-primary:hover {
   background-color: var(--hover-color);
   border-color: var(--hover-color);
@@ -48,7 +62,11 @@
 .btn-outline-dark:focus,
 .btn-outline-dark:focus-visible,
 .btn-outline-light:focus,
-.btn-ouline-dark:focus-visible {
+.btn-ouline-dark:focus-visible,
+input[type="text"]:focus,
+input[type="password"]:focus,
+.google-login-btn a:focus,
+.login input[type="submit"]:focus {
   outline: 3px solid var(--focus-color) !important;
   outline-offset: 2px !important;
   box-shadow: none !important; /* override CA State Web Template */
@@ -61,19 +79,13 @@
   letter-spacing: calc(var(--font-size-20px) * var(--letter-spacing-2));
 }
 
-html[data-theme="light"],
-:root {
-  --primary: #ffffff;
-  --secondary: var(--dark-color);
-  --link-fg: #ffffff;
-  --body-quiet-color: var(--dark-color);
-}
-
 #user-tools,
-#logout-form button {
+#logout-form button,
+.login .submit-row input[type="submit"] {
   text-transform: unset;
 }
 
+/* Eligibility Page */
 .checkbox-parent .form-group:last-of-type .col-12 {
   display: flex;
   flex-direction: row-reverse;
@@ -90,9 +102,94 @@ html[data-theme="light"],
   row-gap: 1rem;
 }
 
-/* Enrollment */
+/* Enrollment Page*/
 iframe.card-collection {
   border: 0px;
   width: 100%;
   height: 60vh;
+}
+
+/* Login Page */
+.login #header {
+  padding: 0 !important;
+}
+
+.login #branding {
+  flex-direction: column;
+  width: 100%;
+}
+
+.login #container {
+  background: var(--body-bg);
+  border: 1px solid var(--hairline-color);
+  overflow: hidden;
+  width: 28em;
+  min-width: 300px;
+  margin: 100px auto;
+  height: auto;
+}
+
+.login #login-form {
+  display: grid;
+  gap: 1em;
+}
+
+.login .form-row label {
+  display: block;
+  line-height: 2em;
+}
+
+.login .form-row #id_username,
+.login .form-row #id_password {
+  padding: 8px;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.login .password-reset-link {
+  text-align: center;
+}
+
+.login .clear {
+  display: none;
+}
+
+/* Login Page: Google SSO Button */
+.login .login-btn-area {
+  width: 100%;
+}
+
+.login .google-login-btn {
+  border: 1px solid var(--bs-primary);
+  background-color: var(--white);
+  height: 100%;
+  border-radius: 4px;
+  padding: 0;
+}
+
+.login .google-login-btn .btn-content {
+  padding: 13.5px 0;
+  border-radius: 4px;
+}
+
+.login .google-btn-logo {
+  padding-right: 20px;
+  background-color: transparent;
+}
+
+.login .google-login-btn:hover {
+  color: var(--white);
+  background-color: var(--bs-primary);
+}
+
+.login .google-login-btn:hover .google-btn-label,
+.login .google-login-btn:hover a {
+  color: var(--bs-white);
+}
+
+.login .google-btn-label {
+  color: var(--bs-primary);
+  width: auto;
+  font-weight: var(--medium-font-weight);
+  font-size: var(--font-size-20px);
 }
--- a/benefits/templates/admin/login.html
+++ b/benefits/templates/admin/login.html
@@ -1,0 +1,28 @@
+{% extends "admin/login.html" %}
+{% load i18n static %}
+
+{% block dark-mode-vars %}
+{% endblock dark-mode-vars %}
+
+{% block extrastyle %}
+    <link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
+    <script nonce="{{ request.csp_nonce }}"
+            src="https://cdn.jsdelivr.net/npm/jquery@3.6.1/dist/jquery.min.js"
+            integrity="sha256-o88AwQnZB+VDvE9tvIXrMQaPlFFSUTR+nldQm1LuPXQ="
+            crossorigin="anonymous"></script>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
+          rel="stylesheet"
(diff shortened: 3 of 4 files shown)


machikoyasuda:
Starting to notice a slew of weirdness having Bootstrap loaded alongside Django Admin's default variables:

- The ordering of CSS variable files matter. I had to move our own styles.css _last_ here https://github.com/cal-itp/benefits/pull/2371/files#diff-cf7d89ad332a4e833961cb19b0e583671aa226405c701eea97bade4f51fc1790R14 for the variables to properly override Bootstrap's.
- This is what happens in Dark mode: Bootstrap sets the base `body` font color to `var(--bs-body-color)`, so that overrides the Django `base.css`'s `var(--body-fg)`. 
image
- Django's login template seems to be inserting it's own jQuery into the HTML 
image

benefits/templates/admin/login.html
```diff
@@ -0,0 +1,28 @@
...
+    <script nonce="{{ request.csp_nonce }}"
+            src="https://cdn.jsdelivr.net/npm/jquery@3.6.1/dist/jquery.min.js"
+            integrity="sha256-o88AwQnZB+VDvE9tvIXrMQaPlFFSUTR+nldQm1LuPXQ="
+            crossorigin="anonymous"></script>
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css"
+          rel="stylesheet"
+          integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3"
+          crossorigin="anonymous">
+    <link href="{% static "css/admin/styles.css" %}" rel="stylesheet">
+{% endblock extrastyle %}
```
machikoyasuda, line 18: Question! @lalver1 @angela-tran - I put this here b/c the login.html template is not already inheriting the extrastyle block from agency-base. Why is that? Is a way I can get login.html to inherit from agency-base?

Also note here: I had to move the call to `styles.css` to line 17 so that the variable overrides set in there actually do override the Bootstrap variables. That's why I also made this same change in agency-base.

benefits/templates/admin/login.html
```diff
@@ -0,0 +1,28 @@
+{% extends "admin/login.html" %}
+{% load i18n static %}
+
+{% block dark-mode-vars %}
+{% endblock dark-mode-vars %}
```
machikoyasuda, line 5: This _removes_ the dark mode CSS styles, effectively disabling dark mode on Admin. cc @thekaveman @angela-tran @lalver1 

Right now, the Bootstrap body class declarations are overriding Django's, and we can't really change that. The extrastyle block comes _after_ the style files from Django, and CSS reads the last file's overrides. So the dark mode looks messed up - the text is all dark (as set by Bootstrap). If we could easily set Django's CSS files to be declared _last_ and Bootstrap's first, then we could implement dark mode. But for now it's kinda complex so I think we should just soft launch this site w/o dark mode.

benefits/static/css/admin/styles.css
```diff
@@ -90,9 +102,89 @@ html[data-theme="light"],
...
+.login .form-row #id_username,
+.login .form-row #id_password {
+  padding: 8px;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.login .password-reset-link {
+  text-align: center;
+}
```
machikoyasuda, line 146: This is directly from https://github.com/django/django/blob/9ca1f6eff6f19d1ae074d289c6c4209073351805/django/contrib/admin/static/admin/css/login.css#L59 but I don't know how to trigger it or if it shows up in our app

benefits/static/css/admin/styles.css
```diff
@@ -90,9 +102,89 @@ html[data-theme="light"],
...
 
-/* Enrollment */
+/* Enrollment Page*/
 iframe.card-collection {
   border: 0px;
   width: 100%;
   height: 60vh;
 }
+
+/* Login Page */
```
machikoyasuda, line 112: This is copied + pasted from https://github.com/django/django/blob/9ca1f6eff6f19d1ae074d289c6c4209073351805/django/contrib/admin/static/admin/css/login.css#L1 except with a few modifications necessary for the design of our page.

benefits/admin.py
```diff
@@ -39,3 +39,7 @@ def index(self, request, extra_context=None):
                 context.update({"has_permission_for_in_person": has_permission_for_in_person})
 
             return TemplateResponse(request, "admin/agency-dashboard-index.html", context)
+
+
+BenefitsAdminSite.site_header = "Administrator"
+BenefitsAdminSite.site_title = "Cal-ITP Benefits - Administrator"
```
machikoyasuda, line 45: @indexing What should the default title of the Admin app be?

This will set it so that each page is something like, `Log in | Cal-ITP Benefits - Administrator`

benefits/static/css/admin/styles.css
```diff
@@ -48,7 +62,11 @@
 .btn-outline-dark:focus,
 .btn-outline-dark:focus-visible,
 .btn-outline-light:focus,
-.btn-ouline-dark:focus-visible {
+.btn-ouline-dark:focus-visible,
+input[type="text"]:focus,
+input[type="password"]:focus,
+.google-login-btn a:focus,
+.login input[type="submit"]:focus {
```
machikoyasuda, line 69: Adds Benefits-style focus styles to all buttons and inputs

machikoyasuda:
I'm marking this PR as complete. 

Some deviations from the mocks:
- I did not add the `Choose an authentication method...` b/c we only have one SSO method so far
- I removed the dark/light mode toggle b/c it would require more work to get the styling right

benefits/templates/admin/login.html
lalver1: @machikoyasuda I think the reason is that `benefits/templates/admin/login.html` extends `contrib/admin/login.html` and `agency-base` is not in that 'chain' of inheritance (`agency-base` extends `contrib/admin/base.html`) so `login.html` is not inheriting the extrastyle block from `agency-base`.

What you have works, but I guess you could also have `benefits/templates/admin/login.html` extend `agency-base` and duplicate some code from `contrib/admin/login.html` into `benefits/templates/admin/login.html`. Maybe that duplication would be ok since we are basically overriding the whole `login` template.

lalver1: I tested this alternate approach about extending `agency-base.html` instead and `login.html` would look like this (we no longer need the `extrastyle` block):
```
{% extends "admin/agency-base.html" %}
{% load i18n static %}

{% block bodyclass %}
    {{ block.super }} login
{% endblock bodyclass %}

{% block usertools %}
{% endblock usertools %}

{% block nav-global %}
{% endblock nav-global %}

{% block content %}
    {% if form.errors and not form.non_field_errors %}
        <p class="errornote">
            {% blocktranslate count counter=form.errors.items|length %}Please correct the error below.{% plural %}Please correct the errors below.{% endblocktranslate %}
        </p>
    {% endif %}

    {% if form.non_field_errors %}
        {% for error in form.non_field_errors %}<p class="errornote">{{ error }}</p>{% endfor %}
    {% endif %}

    <div id="content-main">

        {% if user.is_authenticated %}
            <p class="errornote">
                {% blocktranslate trimmed %}
                    You are authenticated as {{ username }}, but are not authorized to
                    access this page. Would you like to login to a different account?
                {% endblocktranslate %}
            </p>
        {% endif %}

        <form action="{{ app_path }}" method="post" id="login-form">
            {% csrf_token %}
            <div class="form-row">
                {{ form.username.errors }}
                {{ form.username.label_tag }} {{ form.username }}
            </div>
            <div class="form-row">
                {{ form.password.errors }}
                {{ form.password.label_tag }} {{ form.password }}
                <input type="hidden" name="next" value="{{ next }}">
            </div>
            {% url 'admin_password_reset' as password_reset_url %}
            {% if password_reset_url %}
                <div class="password-reset-link">
                    <a href="{{ password_reset_url }}">{% translate "Forgotten your password or username?" %}</a>
                </div>
            {% endif %}
            <div class="submit-row">
                <input type="submit" value="{% translate "Log in" %}">
            </div>
        </form>

    </div>
{% endblock content %}

{% block branding %}
    <div class="d-flex justify-content-center w-100 bg-primary">
        <img class="my-3" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="California Integrated Travel Project: Benefits logo (large)" />
    </div>

    <div id="site-name">
        <h1 class="text-center text-white fs-3 py-3 m-0">{{ site_header }}</h1>
    </div>
{% endblock branding %}
```

benefits/admin.py
lalver1: We may want to set these 2 values in the templates themselves @machikoyasuda since `site_title` can be different across the pages of in-person enrollment and `site_header` is only used in the login page. So setting `site_title` using the `title` block and setting `site_header` directly in the `branding` block.

benefits/templates/admin/login.html
machikoyasuda: Thank you for checking this! It makes sense to me now, how the inheriting happens. I think I want to err on the side of caution of not wanting to accidentally import jQuery twice, so I decided I actually don't want to inherit `extrastyles` from `agency-base`. So I am good with how the `login.html` template looks like now: minimal block overriding, just overriding `branding` basically, and an `extrastyles` that removes the jQuery script.

benefits/templates/admin/login.html
```diff
@@ -0,0 +1,28 @@
...
+
+{% block title %}
+    Log in | Cal-ITP Benefits Administrator
+{% endblock title %}
```
machikoyasuda, line 9: I hardcoded the title here, instead of setting variables for `site_header` and `site_title`. I discussed with Andy and he's going to be adding official titles into Miro so that we can update all of the titles throughout the Admin app in a clean way, and I made a ticket to document that these titles are still TBD, but need to fixed in the future #2381.

machikoyasuda:
@thekaveman @angela-tran @lalver1 This PR is ready for review again.

I did a design review of this PR with @indexing earlier today, and we identified some design/copy requirements that were missing from the Miro/the ticket. I made tickets for them - #2380 and #2381 - and Andy is working on the design/copy requirements and updating Miro for it, so we can work on it in the future. This current PR will _not_ address these and it can be addressed in the future. Both are small changes, but I didn't want these things to block this PR.

benefits/static/css/admin/styles.css
angela-tran: If your database has the user from the sample fixtures (CST User) and you log in as that user, you'll see a "Change password" link in the usertools section.

image   


... speaking of which, that link takes us to a page that we have not overridden / that doesn't use our `extrastyles`, so it uses all the default Django styles. 

image

I'm not sure if what we should have this link or if we should override it to use our extra styles. I'll create a separate ticket for this.

angela-tran: Looks like the link shows up based on if the user "has a changeable password":

https://github.com/django/django/blob/1857b6663b664847a426e8c79b8aec42994660ba/django/contrib/admin/templates/admin/base.html#L56-L58

angela-tran: Created #2384

angela-tran: Oh wait, `password-reset-link` is actually a different thing from what I was talking about.

It's a link on the login page. I guess we wouldn't show it because we don't have a `'admin_password_reset'` route defined.

https://github.com/search?q=repo%3Adjango%2Fdjango%20password-reset-link&type=code

https://github.com/django/django/blob/1857b6663b664847a426e8c79b8aec42994660ba/django/contrib/admin/templates/admin/login.html#L59-L61

angela-tran: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#adding-a-password-reset-feature

I don't think we should worry about adding this now since we ideally want to use SSO long-term

benefits/templates/admin/login.html
```diff
@@ -0,0 +1,28 @@
...
+
+{% block extrastyle %}
```
angela-tran, line 11: Maybe add a comment here to say why we're not bringing in jQuery here / why we're using our own `extrastyle` block here instead of extending from `agency-base.html`?

angela-tran (approved):
This looks good to me! I saw a place where it might be helpful to have a comment, but I won't block the PR for it. I can re-approve if you decide to add it.

benefits/templates/admin/login.html
machikoyasuda: Good idea! Added here: https://github.com/cal-itp/benefits/pull/2371/commits/27bbf27b7c66b926f7a09c964ed8794b3c7628c6

benefits/static/css/admin/styles.css
thekaveman: On the other hand, any kind of self-service we can offer to users _while_ we have to manage them locally would be handy. 

Let's create another ticket to figure out what to do about the Reset link on the login page.

angela-tran: Created https://github.com/cal-itp/benefits/issues/2386 for the reset password link feature

angela-tran (approved):

Issue #2245, opened 2024-07-23
Send analytic events for in-person eligibility verification / enrollment


angela-tran:
The implementation details still need to be fleshed out, but the acceptance criteria below seem correct based on dev workshop discussion on 2024-09-11.

## Acceptance Criteria

- [ ] As part of in-person eligibility verification / enrollment, analytic events are sent for:
   - [ ] `started eligibility`
   - [ ] `returned eligibility`
   - [ ] ~`started payment connection`~ `started card tokenization`
   - [ ] ~`closed payment connection`~ `ended card tokenization`
   - [ ] `returned enrollment`
   - [ ] `failed access token request`
- [ ] Analytic events capture if the eligibility verification / enrollment is from an in-person interaction or not

angela-tran:
@indexing @thekaveman and I discussed the above and decided we don't need to worry about capturing "Verified by" for this ticket or the September release.

thekaveman:
Let's align the property naming here with what we choose for the field name in #2287

@indexing suggests `method` for the name of the field/property.

Devs suggest possible values of `digital` or `in_person`.

angela-tran:
This ticket needs #2244 and #2345 to be done first, so moving this to Blocked

indexing:
In this interest of more intuitive naming for parameters in our analytics tool, I propose changing `method` to `enrollment_method`. The values for this parameter remain as `digital` and `in_person`.

angela-tran:
@lalver1 https://github.com/cal-itp/benefits/pull/2382 has the code in a state that's ready to have analytics added to it. You could either build off that branch or wait for it to be merged into `main`.

PR #2383, opened 2024-09-19
Add `supported_methods` field to `EnrollmentFlow`

### LINKED ISSUE #2287: Enhance `EnrollmentFlow` model with field to indicate it is a pathway that can only be done in-person
## Acceptance Criteria

- [x] We need a way to represent that a flow is digital-only, in-person only, or supports both.

machikoyasuda:
closes #2287 

## What this PR does
- **Note**: This PR currently does _not_ add any checks throughout the app to check if a flow is supported in the in-person or digital apps.
- Adds https://pypi.org/project/django-multiselectfield/ package
- Adds `supported_enrollment_methods` field to `EnrollmentFlow`, as radio buttons that can be both selected
- The default for a new flow is having both selected: `http://localhost:11369/admin/core/enrollmentflow/add/` 
- Updates local fixtures: All existing enrollment flows are supported both digitally and in-person, adds a sample In-Person Only and Digital Only --- but does not add those flows to CST.
- Add spec

image


machikoyasuda:
Not sure why both Pytest and Cypress tests are failing here. Both suites pass for me locally. 🤨

machikoyasuda:
@angela-tran and I learnt something new today:

A GitHub repo for a PyPI package might not reflect the latest release! The release page for https://github.com/goinnn/django-multiselectfield shows that the latest version is [v0.1.12](https://github.com/goinnn/django-multiselectfield/releases/tag/v0.1.12). But we noticed that when I had installed `django-multiselectfield` with the command, `pip install django-multiselectfield`, version 0.1.13 was installed. And we went to the PyPI page for the package directly and saw a 0.1.13 release: https://pypi.org/project/django-multiselectfield/0.1.13/, which fixed the bug that was breaking my tests.

So I updated the version to 0.1.13, and all the tests were fixed 🎉 No hacky fixes required.

Lesson learnt: Check both the GitHub and PyPI page for the package you are using.

thekaveman (changes requested):
This looks great!

> This PR currently does not update existing enrollment flows in the fixture data, nor does it add any checks throughout the app to check if a flow is supported in the in-person or digital apps.

I think for existing fixtures e.g. in `dev`, `test`, `prod` -- this is fine. Since the field defines a default value, we'll be good there and the flows will be updated in the database next time they are edited.

I do think it would be worth it to update our local fixtures in this PR though, just for completeness of the data when starting from scratch. And to show a full example of how the config works together.

The other minor change, can we change the label shown to admin users for the new field to: `Supported enrollment methods`? This is to align more closely with @indexing's comment on the analytics side: https://github.com/cal-itp/benefits/issues/2245#issuecomment-2353987835

machikoyasuda:
@thekaveman That makes sense. Should I set all Enrollment Flows to accept both methods?

thekaveman:
> @thekaveman That makes sense. Should I set all Enrollment Flows to accept both methods?

Let's set all our existing fixtures to both :+1: 

Maybe add one basic sample each of a `digital` only and `in_person` only, if that feels reasonable?

machikoyasuda:
@thekaveman Updated all:
- Existing flows are supported by both enrollment methods
- Adds a sample Digital only and In-Person only method -- but does not add them to the CST Agency. https://github.com/cal-itp/benefits/pull/2383/commits/9b2e3f6b360da8cfc4ca9ddcdfddacef767316ab
- Renames the field to `supported_enrollment_method` -- this is the cleanest way to change the display name, plus it makes for more readable code, I think.

thekaveman (approved):
Sorry, I just realized one more thing, but it can be left for a follow-up.

PR #2397, opened 2024-09-23
Admin: Log out page styled

### LINKED ISSUE #2359: Add styling to logout page
Similar to #2265 

The default log out page that Django shows is this:

image

- [x] Create a Miro mock-up for what this page should look like
- [x] Implement

Miro: https://miro.com/app/board/uXjVKwXyUy0=/?moveToWidget=3458764599849444989&cot=14

machikoyasuda:
closes #2359 

image

This PR creates a template, `logged_out.html`, which extends `agency-base` to get Bootstrap and Button styles, and combines code from:

- the default Django log out page for the HTML https://github.com/django/django/blob/39de2e97a06d0317973b280bc159ca6f89fc64e3/django/contrib/admin/templates/registration/logged_out.html
- the default Django log in page css style (adding the `login`class to the body element) brings in default Django login page styles
- the customizations created in app's Log in page https://github.com/cal-itp/benefits/blob/main/benefits/templates/admin/login.html for the repeated styles of the branding header and the main div.

0 WCAG AA issues.

### DIFF
--- a/benefits/templates/registration/logged_out.html
+++ b/benefits/templates/registration/logged_out.html
@@ -1,0 +1,35 @@
+{% extends "admin/agency-base.html" %}
+{% load static %}
+
+{% block title %}
+    Logged out | Cal-ITP Benefits Administrator
+{% endblock title %}
+
+{% block extrastyle %}
+    {{ block.super }}
+    <link rel="stylesheet" href="{% static "admin/css/login.css" %}">
+{% endblock extrastyle %}
+
+{% block bodyclass %}
+    {{ block.super }} login
+{% endblock bodyclass %}
+
+{% block branding %}
+    <div class="d-flex justify-content-center w-100 bg-primary">
+        <img class="my-3" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="California Integrated Travel Project: Benefits logo (large)" />
+    </div>
+
+    <div id="site-name">
+        <h1 class="text-center text-white fs-3 py-3 m-0">Administrator</h1>
+    </div>
+{% endblock branding %}
+
+{% block content %}
+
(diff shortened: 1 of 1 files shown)


benefits/templates/registration/logged_out.html
```diff
@@ -0,0 +1,35 @@
...
+{% endblock bodyclass %}
+
+{% block branding %}
+    <div class="d-flex justify-content-center w-100 bg-primary">
+        <img class="my-3" src="{% static "img/logo-lg.svg" %}" width="220" height="50" alt="California Integrated Travel Project: Benefits logo (large)" />
+    </div>
+
+    <div id="site-name">
+        <h1 class="text-center text-white fs-3 py-3 m-0">Administrator</h1>
+    </div>
```
machikoyasuda, line 24: This is the same exact code as https://github.com/cal-itp/benefits/blob/main/benefits/templates/admin/login.html#L21-L29 so I _could_ turn it into an `includes`... but I didn't think we'd re-use this a third time, so I opted not to.

lalver1: Yep, I can't see where we would re-use it since it's specific to the log in and log out pages.

benefits/templates/registration/logged_out.html
```diff
@@ -0,0 +1,35 @@
...
+{% endblock branding %}
+
+{% block content %}
+
+    <div id="content-main">
+        <p class="text-center mt-4">You have logged out successfully.</p>
```
lalver1, line 30: Just noticed that we have `You've logged out successfully.` in Miro but since we don't have a strict copy I think it's ok.

lalver1 (approved):
It looks really nice!

PR #2399, opened 2024-09-24
Only show supported flows for In-person and Digital flows

### LINKED ISSUE #2396: Use `supported_enrollment_methods` field to filter flows in digital/in-person views

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/2383#pullrequestreview-2323672886_
            
## Acceptance criteria

- [ ] A flow that is digital-only shows up only on the digital side
- [ ] A flow that is in-person only shows up only on the in-person side
- [ ] A flow that supports both shows up on both sides

machikoyasuda:
closes #2396 

## What this PR does
- Adds filter() calls to both Eligibility form methods
- Updates test mock to return an array of flows
- Adds tests to the Digital and In-Person flows, testing a scenario in which an agency has 3 flows (digital, in-person and both( and the page should only show 2.

## How to test
- Log into Admin as a superuser and set different enrollment flows to be either in-person-only or digital-only.
- Log into Admin as a CS user to see that digital-only is filtered out
- Go to Benefits digital to see that in-person-only is filtered out

Note: Either digital-only or in-person-only HAVE to be checked. You can't set a flow to be neither.


tests/pytest/eligibility/test_views.py
```diff
@@ -103,6 +104,7 @@ def test_index_get_agency_single_flow(mocker, model_TransitAgency, model_Enrollm
     mock_manager = mocker.Mock()
     mock_manager.all.return_value = [model_EnrollmentFlow]
     type(mock_agency).enrollment_flows = mocker.PropertyMock(return_value=mock_manager)
```
angela-tran, line 106: I'm curious if these lines are still needed or if they can be removed.

thekaveman: Hmm probably not, since we aren't calling the `.all` helper anymore, right? I guess we just need to mock the `.filter` helper.

machikoyasuda: Yeah! It _shouldn't_ need it but for some reason right now it does for tests to pass.

Marking this PR as In Progress again b/c I'm trying to write a different spec to test the filtering function.

tests/pytest/eligibility/test_views.py
```diff
@@ -71,15 +71,55 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
...
+def test_index_show_digital_flow(mocked_session_agency, model_EnrollmentFlow_only_digital, client):
+    path = reverse(routes.ELIGIBILITY_INDEX)
+    response = client.get(path)
+
+    assert mocked_session_agency.return_value.enrollment_flows.count() == 1
+    assert mocked_session_agency.return_value.enrollment_flows.first().supported_enrollment_methods == ["digital"]
+    assert response.status_code == 200
+    assert response.status_code == 200
+    assert "form" in response.context_data
+    len(response.context_data["form"].fields["flow"].choices) == 1
```
machikoyasuda, line 110: These specs are testing:

- When you have an agency with only 1 flow, and that flow is only supported for `in-person`, the form should have 0 fields. (The flow is filtered out from the digital flow form view).
- When you have an agency with only 1 flow, and that flow is `digital`, the form should have 1 field. (The flow is NOT filtered out from the digital flow form view).

I would have liked to have been able to have an assertion on the last line of these specs that are more like "The response's form should have a field that is supported by digital", but the `context_data["form"].fields["flow"].choices` only returns an id of the flow, not name of the flow or anything more. So I decided to test it like this instead. I'm open to other ideas on how to test this!

PR #2402, opened 2024-09-25
Feat: analytics for in-person eligibility/enrollment

lalver1:
- **feat(analytics): add enrollment_method to analytics base Event**
- **feat(analytics): add started and finished eligibility event**


benefits/in_person/views.py
```diff
@@ -30,6 +32,7 @@ def eligibility(request):
             flow_id = form.cleaned_data.get("flow")
             flow = models.EnrollmentFlow.objects.get(id=flow_id)
             session.update(request, flow=flow)
+            eligibility_analytics.started_eligibility(request, flow, enrollment_method=models.EnrollmentMethods.IN_PERSON)
```
thekaveman, line 35: I would think we also want to send the `selected enrollment flow` event here, right? For consistency with the `digital` side.

benefits/in_person/views.py
```diff
@@ -91,20 +97,31 @@ def enrollment(request):
...
+                )
                 sentry_sdk.capture_exception(exception)
                 return redirect(routes.IN_PERSON_SERVER_ERROR)
 
             case Status.REENROLLMENT_ERROR:
                 return redirect(routes.IN_PERSON_ENROLLMENT_REENROLLMENT_ERROR)
     # GET enrollment index
     else:
+        flow = session.flow(request)
+        eligibility_analytics.returned_success(request, flow, enrollment_method=models.EnrollmentMethods.IN_PERSON)
```
thekaveman, line 124: I think in general, we should move away from sending events in response to `GET` requests where possible.

It becomes easier to spam the events if all you need to do is visit the URL. Especially for these types of events where we do already have a `POST`-back, I think ideally the event should be sent there.

benefits/in_person/views.py
```diff
@@ -137,6 +155,7 @@ def enrollment(request):
 
 def reenrollment_error(request):
     """View handler for a re-enrollment attempt that is not yet within the re-enrollment window."""
+    enrollment_analytics.returned_error(request, "Re-enrollment error.", enrollment_method=models.EnrollmentMethods.IN_PERSON)
```
thekaveman, line 158: Could be moved up into `POST`-back

benefits/in_person/views.py
```diff
@@ -147,6 +166,7 @@ def reenrollment_error(request):
 
 def retry(request):
     """View handler for card verification failure."""
+    enrollment_analytics.returned_retry(request, enrollment_method=models.EnrollmentMethods.IN_PERSON)
```
thekaveman, line 169: This one can't be moved since the user is sent here from JavaScript.

Would _could_ enforce that this route is `POST`-only and make the from `POST` back to here... that doesn't have to happen in this PR though.

thekaveman (changes requested):
This is really good so far.

benefits/in_person/views.py
lalver1: Yep, good idea. Added it in 159594176a75282614d75a815a6409af8501f07a

benefits/in_person/views.py
lalver1: Ah that makes sense! Moved it to the `POST` section in e9f5ead8ea7e813c6f360d63bcd4ca23218cf5e6

benefits/in_person/views.py
lalver1: Yep, moved it up in f5a376abb9e483dcd8b7f50ab34dedd24ff6ffb6

benefits/in_person/views.py
lalver1: Right, I'll make a ticket for it 👍

PR #2403, opened 2024-09-25
Refactor Enrollment Flow for consistent labeling across apps

### LINKED ISSUE #2336: Refactor `EnrollmentFlow` model so in-person eligibility view can use copy from the selection label template
Original context is https://github.com/cal-itp/benefits/pull/2317#discussion_r1737187630.

Basically we want the radio buttons in the in-person eligibility form to use the same copy that the Benefits app uses in its enrollment flow selection form, but that copy is currently embedded within the `EnrollmentFlow`s' `selection_label_template`s, e.g. ["Older Adult"](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/templates/eligibility/includes/selection-label--senior.html#L5), ["US Veteran"](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/templates/eligibility/includes/selection-label--veteran.html#L5), etc.

Let's discuss how to refactor the model to achieve this, and then do it. It would be nice to not have to add another field to the model.

## Plan we came up with after discussing
The reason why we prepend the transit agency's name to the `EnrollmentFlow`'s label is because when we look at a list of `EnrollmentFlow`s, it is hard to distinguish between flows without it. For example, this happens when we're looking at either the `EnrollmentFlow` model list or when looking at the `enrollment_flows` field on a `TransitAgency`.

<table>
<tr>
<td>
</td><td></td>
</tr>
</table>

However, we can solve this another way by
- [x] adding a column to the `EnrollmentFlow` table to show the associated transit agency (do this by configuring `list_display` on the `EnrollmentFlowAdmin` - see `EnrollmentEventAdmin` as an example)
- [x] refactoring the relationship between `EnrollmentFlow` and `TransitAgency` so that the field is on `EnrollmentFlow` instead, and while we're at it, go ahead and make it a many-to-one relationship instead of many-to-many since that is more accurate - #2403 

Then, `EnrollmentFlow.label` is free to say whatever we need it to. 

Some notes from a maintenance perspective: 
- the copy we put in `EnrollmentFlow.label` is independent of the copy inside the `selection_label_template`s being used on the digital side, so that is to say the copy will be duplicated.
- `EnrollmentFlow.label` is just an English string, and we haven't thought about how we'd handle translation of it if we ever need to. 

We're ok with both of these things for this feature.

## Acceptance Criteria

- [x] In-person eligibility form uses the desired copy
- [x] No regressions to the Benefits enrollment flow selection form

lalver1:
closes #2336

## What this PR does
- [x] This PR changes the implementation of the relationship between `EnrollmentFlow` and `TransitAgency`. `EnrollmentFlow` now has a `TransitAgency` foreign key field (many flows to one transit agency). This will allow us to have more flexibility with the names we can use in `EnrollmentFlow.label` (for example, having the name of the transit agency in this field won't be necessary anymore).
- [x] Update EnrollmentFlow labels in fixture data, to match the names used in form fields
- [x] Updates fixture data: Removed the `digital-only` and `in-person only` enrollment flows for now. It was breaking the Cypress tests, because these flows did not have selection labels and all the necessary HTML templates. Can be re-added when the enrollment flow filtering goes into effect.
- [x] Updates test failures, test fixtures. 1 failing test.
- [x] Update Admin view for Enrollment Flow to show Agency | Flow label | Supported methods
- [x] Add migration for existing data
- [x] Updates helptext on a flow's agency
- [x] EnrollmentFlow's agency is now not required and can be blank.

## How to test
- Run `./bin/reset_db.sh`
- Test both in-person and digital flows
- Test re-ordering enrollment flows in Admin
- Run all the tests

## Screenshots
- In Person Eligibility index: Same code, new database data
image

- Admin: Enrollment Flows displayed with label, agency and supported methods
image


machikoyasuda:
I'll take the rest of the work of this PR - fixing tests and anything else necessary.

benefits/core/migrations/0028_remove_transitagency_enrollment_flows_and_more.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.RemoveField(
+            model_name="transitagency",
+            name="enrollment_flows",
+        ),
+        migrations.AddField(
+            model_name="enrollmentflow",
+            name="transit_agency",
+            field=models.ForeignKey(default=1, on_delete=django.db.models.deletion.PROTECT, to="core.transitagency"),
+            preserve_default=False,
+        ),
```
angela-tran, line 23: Just wanted to note here that we should make sure to update this generated migration so that instead of
- remove the field
- add field

it's

- add the field
- migrate the existing data
- remove the field

so that we don't lose the existing relationships.

angela-tran: To migrate the data, you'll want to use a `migrations.RunPython` operation and give it a function that iterates over all the `TransitAgency`s, iterates over each agency's `EnrollmentFlow`s, and sets the new `transit_agency` field for each `EnrollmentFlow`.

We have some examples of `RunPython` in previous migrations.

angela-tran:
@machikoyasuda Can you update the title of this PR to reflect that it covers more than just the relationship refactor? I think it is covering all of #2336

angela-tran:
> @angela-tran I couldn't think of a short way to summarize all the changes into one good PR headline 😅

😆 It is a bit long of a title, but it does cover all the technical changes, so I think it's ok.

I guess from a more abstract level, this PR is making it so we don't need to have the transit agency name in the label, so I think another title that could work is `Refactor: move transit agency name from EnrollmentFlow.label to admin UI` or something like that.

Up to you, either one is fine with me!

thekaveman:
Shorter is better I think... remember this will be the commit message when the PR is merged!

benefits/core/migrations/local_fixtures.json
```diff
@@ -62,15 +86,16 @@
       "eligibility_start_template": "eligibility/start--senior.html",
       "claims_scope": "verify:senior",
       "claims_claim": "senior",
-      "supported_enrollment_methods": ["digital", "in_person"]
+      "supported_enrollment_methods": ["digital", "in_person"],
+      "transit_agency_id": 1
```
angela-tran, line 90: This does work, but to be more consistent with how other ForeignKeys are defined in these fixtures, let's have this as
```
"transit_agency": 1
```
instead.

Same for the rest of the instances of `transit_agency_id` in this file.

tests/pytest/conftest.py
```diff
@@ -75,13 +75,14 @@ def model_ClaimsProvider_no_sign_out(model_ClaimsProvider):
...
 @pytest.fixture
-def model_EnrollmentFlow():
+def model_EnrollmentFlow(model_TransitAgency):
     flow = EnrollmentFlow.objects.create(
         system_name="Test Flow",
         selection_label_template="eligibility/includes/selection-label.html",
         label="Test flow label",
         group_id="group123",
         enrollment_success_template="enrollment/success.html",
+        transit_agency_id=model_TransitAgency.id,
```
angela-tran, line 85: Here I think we should use `transit_agency = model_TransitAgency` instead since Django's ORM allows us to work from an object perspective rather than thinking about the database representation.

I actually didn't even know accessing this `_id`-appended field was a thing! I found [some docs](https://docs.djangoproject.com/en/5.1/ref/models/fields/#database-representation) on it.

tests/pytest/conftest.py
```diff
@@ -112,6 +113,7 @@ def model_EnrollmentFlow_with_scope_and_claim(model_EnrollmentFlow, model_Claims
 @pytest.fixture
 def model_EnrollmentFlow_with_claims_scheme(model_EnrollmentFlow_with_scope_and_claim):
     model_EnrollmentFlow_with_scope_and_claim.claims_scheme_override = "scheme"
+    model_EnrollmentFlow_with_scope_and_claim.transit_agency_id = 1
```
angela-tran, line 116: Same note here about using `transit_agency` instead of `transit_agency_id`. I guess this means you'll need to bring in the `model_TransitAgency` fixture to assign it.

tests/pytest/core/test_models.py
```diff
@@ -168,8 +168,8 @@ def test_EnrollmentFlow_supports_expiration(model_EnrollmentFlow_supports_expira
 
 
 @pytest.mark.django_db
-def test_EnrollmentFlow_enrollment_index_template():
-    new_flow = EnrollmentFlow.objects.create()
+def test_EnrollmentFlow_enrollment_index_template(model_TransitAgency):
+    new_flow = EnrollmentFlow.objects.create(transit_agency_id=model_TransitAgency.id)
```
angela-tran, line 172: Same note here about using `transit_agency` instead of `transit_agency_id`.

tests/pytest/core/test_models.py
```diff
@@ -180,15 +180,15 @@ def test_EnrollmentFlow_enrollment_index_template():
 
 
 @pytest.mark.django_db
-def test_EnrollmentFlow_enrollment_success_template():
-    new_flow = EnrollmentFlow.objects.create()
+def test_EnrollmentFlow_enrollment_success_template(model_TransitAgency):
+    new_flow = EnrollmentFlow.objects.create(transit_agency_id=model_TransitAgency.id)
```
angela-tran, line 184: Same note here about using `transit_agency` instead of `transit_agency_id`.

tests/pytest/core/test_models.py
```diff
@@ -180,15 +180,15 @@ def test_EnrollmentFlow_enrollment_index_template():
...
 
     assert new_flow.enrollment_success_template == "enrollment/success.html"
 
 
 @pytest.mark.django_db
-def test_EnrollmentFlow_supported_enrollment_methods():
-    new_flow = EnrollmentFlow.objects.create()
+def test_EnrollmentFlow_supported_enrollment_methods(model_TransitAgency):
+    new_flow = EnrollmentFlow.objects.create(transit_agency_id=model_TransitAgency.id)
```
angela-tran, line 191: Same note here about using `transit_agency` instead of `transit_agency_id`.

tests/pytest/core/test_views.py
```diff
@@ -134,16 +130,9 @@ def test_agency_card_with_multiple_eligibility_api_flows(
     new_flow.label = "New flow"
     new_flow.system_name = "new"
     new_flow.pk = None
+    new_flow.transit_agency_id = model_TransitAgency.id
```
angela-tran, line 133: Same note here about using `transit_agency` instead of `transit_agency_id`.

tests/pytest/core/test_views.py
```diff
@@ -156,9 +145,8 @@ def test_agency_card_with_multiple_eligibility_api_flows(
 def test_agency_card_without_eligibility_api_flow(
     client, model_TransitAgency, model_EnrollmentFlow_with_scope_and_claim, mocked_session_update, mocked_session_reset
 ):
-    model_TransitAgency.enrollment_flows.clear()
-    model_TransitAgency.enrollment_flows.add(model_EnrollmentFlow_with_scope_and_claim)
-    model_TransitAgency.save()
+    model_EnrollmentFlow_with_scope_and_claim.transit_agency_id = model_TransitAgency.id
```
angela-tran, line 148: Same note here about using `transit_agency` instead of `transit_agency_id`.

angela-tran (changes requested):
The overall changes look great! There are just a few code clean-ups I think we should do.

angela-tran:
Comments for lines [351-354](https://github.com/cal-itp/benefits/pull/2403/files#diff-4d37ea08e0da7b3a7ea86f5fa29f0a86a74acb8089d2c9bd881173c430fad818R351-R354):

https://github.com/cal-itp/benefits/blob/4e7cdbe357ac4f268b5e6f8a6af16f605fb0f436/benefits/core/models.py#L351-L354

Do we consider transit agency staff users as "end-users"? I'm wondering if we should remove the part that says `not shown to end-users`.

benefits/core/models.py
```diff
@@ -235,6 +381,7 @@ class EnrollmentFlow(models.Model):
         default=[EnrollmentMethods.DIGITAL, EnrollmentMethods.IN_PERSON],
         help_text="If the flow is supported by digital enrollment, in-person enrollment, or both",
     )
+    transit_agency = models.ForeignKey(TransitAgency, on_delete=models.PROTECT)
```
angela-tran, line 384: The migration is failing for me; we need to update this field to allow nulls.

### Steps I'm using to test migration:
- Checkout `main` and reset database using sample fixtures -- this makes our database use the same schema as that on our Azure instances (dev, test, and prod)
   - `git checkout main`
   - `./bin/reset_db.sh`
- Checkout `refactor/transit-agency-enrollment-flow`
- Run `./bin/init.sh` to apply migration

Migration fails with:
```
django.db.utils.IntegrityError: NOT NULL constraint failed: new__core_enrollmentflow.transit_agency_id
```

`EnrollmentFlow.transit_agency` needs to allow nulls, i.e. pass in [`null=True`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#null) when declaring it as a `ForeignKey` field. We should also pass in `blank=True` so that from the admin interface, the model form allows us to set a blank value and pass form validation.

angela-tran (changes requested):
The updates look good! Just one more change I see as needed to make sure the migration works.

machikoyasuda:
Testing the new helptext in Admin: 

image


This text makes way more sense.

benefits/core/migrations/0028_remove_transitagency_enrollment_flows_and_more.py
```diff
@@ -1,12 +1,11 @@
-# Generated by Django 5.1.1 on 2024-09-25 16:47
+# Generated by Django 5.1 on 2024-09-30 23:25
```
machikoyasuda, line 1: The 5.1.1/5.1 change is a weird diff!

tests/pytest/conftest.py
```diff
@@ -5,14 +5,16 @@
...
+# from pytest_socket import disable_socket
 
 from benefits.core import session
 from benefits.core.models import ClaimsProvider, EnrollmentFlow, TransitProcessor, PemData, TransitAgency
 
 
 def pytest_runtest_setup():
-    disable_socket()
+    # disable_socket()
+    pass
```
thekaveman, line 17: This change should be reverted, we _do_ want to disable sockets in our tests.

If a test is failing because of this (some kind of SocketError or whatever) -- that is an indication that the test is using a service that should be mocked (e.g. analytics or sentry or something).

thekaveman (changes requested):

tests/pytest/conftest.py
angela-tran: I think some of us have seen an issue where _all the tests_ raise an error about sockets, and to prevent that, I've added `"python.experiments.optOutFrom": ["pythonTestAdapter"],` to my VS Code user settings.

angela-tran: Agreed we should revert this change. @machikoyasuda Can you see if adding that user setting helps for you?

machikoyasuda: @angela-tran Not sure why but adding that line to my settings.json doesn't get the tests to _run_ - still get the same sockets error. I can't run any tests without disabling sockets.

This is what I have in my settings.json now:
```
  "[python]": {
    "editor.defaultFormatter": "ms-python.black-formatter"
  },
  "python.languageServer": "Pylance",
  "python.testing.pytestArgs": ["tests/pytest"],
  "python.testing.pytestEnabled": true,
  "python.testing.unittestEnabled": false,
```

machikoyasuda: Got it working with these instructions: https://github.com/microsoft/vscode-python/issues/22383#issuecomment-1949198679 and then re-starting container! 

image

thekaveman: Yeah this is related to that thread we come back to from time to time: https://github.com/microsoft/vscode-python/issues/22383. Thought it seems like this should have been fixed in a VS Code version released in May?

The opt-out has to go in your _user_ settings, which is not the file that is checked into the repo unfortunately. See that thread for more details.

angela-tran: I'm glad it's working @machikoyasuda! All good to revert this change (and the change on line 9)?

machikoyasuda: https://github.com/cal-itp/benefits/pull/2403/commits/4d7f09fc48a1a9b2898a6c494215547bc820bc86

angela-tran (approved):
Migration worked, tests pass, app launches locally, admin interface shows new field / help text - looks great ✅ 

Thanks @machikoyasuda !

PR #2409, opened 2024-09-26
Admin: Use "log" not "sign"

### LINKED ISSUE #2380: Admin: Use "Log" over "Sign"
The Administrator app (which encompasses everything under `/admin/` and `/in_person/`) should use the term `Log` in over the term (`Sign`). 

Currently, Sign is used in two places in the Admin app:

- the agency-base header override uses `Sign`

image

- the Google SSO button uses `Sign`
image

This ticket is to change it in both of these places.

- Django SSO offers a variable to do it: https://megalus.github.io/django-google-sso/customize/#the-button-text
- And the header can be changed in this template file here: https://github.com/cal-itp/benefits/blob/feb06d22c04b6a60cf26c6b1e8202a1d8005d706/benefits/templates/admin/agency-base.html#L51

image

Why?

It's good to be consistent, and the Admin app is using the two terms interchangeably currently. For example, when I click _Sign_ out, it takes me to this page: 
image

---

Note: The Benefits app itself uses Sign and Log interchangeably, but this PR is only about Admin.

machikoyasuda:
closes #2380 

The Administrator app should use "Log in" and "Log out" in app copy. This PR changes this in:

- Google SSO button copy
- Admin header

Test this by logging in and out as a super user and a transit agency user.


angela-tran (approved):

angela-tran:
Tested this as super user and transit agency staff user

PR #2410, opened 2024-09-26
Admin: Update admin page titles

### LINKED ISSUE #2381: Admin: Update all Admin app view titles
This ticket is still in Needs shaping, and is a placeholder for @indexing to update the titles in Miro, so that devs can update the titles throughout the Adminstrator app.

---

Proposed title structure: `Agency name | View title | Name of app`

Examples: 

- /admin, not logged in: `Log in | Cal-ITP Benefits Administrator`
- /admin, logged in: `California State Transit | Dashboard | Cal-ITP Benefits Administrator`
- /admin, in-person enrollment flow: `California State Transit | In-person enrollment | Cal-ITP Benefits Administrator`

machikoyasuda:
closes #2381 

## What this PR does
- Updates all Administrator app page titles according to what's on https://miro.com/app/board/uXjVKwXyUy0=/ - check the text above the bordered page sections
- Uses Django admin variables or Django template default titles for as much as possible: https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#adminsite-attributes
- Updated anywhere we could use above variables

### DIFF
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -8,6 +8,10 @@ from benefits.core import session
 
 class BenefitsAdminSite(admin.AdminSite):
 
+    site_title = "Cal-ITP Benefits Administrator"
+    site_header = "Administrator"
+    index_title = "Dashboard"
+
     def index(self, request, extra_context=None):
         """
         Display the main admin index page if the user is a superuser or a "staff_group" user.
@@ -42,6 +46,7 @@ class BenefitsAdminSite(admin.AdminSite):
                     {
                         "has_permission_for_in_person": has_permission_for_in_person,
                         "transit_processor_portal_url": transit_processor_portal_url,
+                        "title": f"{agency.long_name} | {self.index_title} | {self.site_title}",
                     }
                 )
 
--- a/benefits/in_person/views.py
+++ b/benefits/in_person/views.py
@@ -142,7 +142,12 @@ def enrollment(request):
 
 def reenrollment_error(request):
     """View handler for a re-enrollment attempt that is not yet within the re-enrollment window."""
-    context = {**admin_site.each_context(request)}
+
+    agency = session.agency(request)
+    context = {
+        **admin_site.each_context(request),
+        "title": f"{agency.long_name} | In-person enrollment | {admin_site.site_title}",
+    }
 
     flow = session.flow(request)
     context["flow_label"] = flow.label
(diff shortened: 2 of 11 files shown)


benefits/templates/admin/login.html
```diff
@@ -4,10 +4,6 @@
 {% block dark-mode-vars %}
 {% endblock dark-mode-vars %}
 
-{% block title %}
-    Log in | Cal-ITP Benefits Administrator
-{% endblock title %}
-
```
machikoyasuda, line 10: This is not necessary here, because the `{% extends "admin/login.html" %}` default Django login page's title uses `Log in | site_title`.

benefits/templates/registration/logged_out.html
```diff
@@ -2,7 +2,7 @@
 {% load static %}
 
 {% block title %}
-    Logged out | Cal-ITP Benefits Administrator
+    Logged out | {{ site_title }}
 {% endblock title %}
```
machikoyasuda, line 6: This `logged_out` page extends from `agency-base`, and _not_ the default Django admin log out template, so it's necessary to explicitly spell out `Logged out`

lalver1:
This looks good! I like how the titles are organized now 👍
While testing I did notice that `server_error.html` and `success.html`, for example, did not have the title set. So I'm wondering, would we want to add `title` to the context for the views: `reenrollment_error`, `retry`, `system_error`, `server_error`, `success`. I think this would set the title for these pages since they already inherit from `error-base.html` or `agency-base.html`.

benefits/in_person/views.py
```diff
@@ -130,6 +134,7 @@ def enrollment(request):
             "form_server_error": tokenize_server_error_form.id,
             "form_success": tokenize_success_form.id,
             "form_system_error": tokenize_system_error_form.id,
+            "title": f"{agency.long_name} | In-person enrollment | {admin_site.site_title}",
```
angela-tran, line 137: I agree with @lalver1 that we want to add this to the context for the error view functions below (`reenrollment_error`, etc.).

Currently, they have no page title:
image

angela-tran (changes requested):
1 requested change about success / error page titles. The other 2 comments are just notes

benefits/admin.py
```diff
@@ -8,6 +8,10 @@
 
 class BenefitsAdminSite(admin.AdminSite):
 
+    site_title = "Cal-ITP Benefits Administrator"
+    site_header = "Administrator"
+    index_title = "Dashboard"
```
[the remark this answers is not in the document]
machikoyasuda: Thanks for documenting this!

machikoyasuda:
@angela-tran @lalver1 Thanks for the review.
I added code, test updates to cover all those pages I forgot (sys error, server error, re-enrollment error, retry). And I tested them!

I wanted to do the simplest fix for now, as it's the last day of the sprint. But we could make a ticket in the future to refactor like @angela-tran suggested above at a later time.

angela-tran (approved):
This looks great - thank you @machikoyasuda !

PR #2420, opened 2024-10-01
Eligibility: Filter enrollment flows by supported method

### LINKED ISSUE #2396: Use `supported_enrollment_methods` field to filter flows in digital/in-person views

machikoyasuda:
closes #2396 

## What this PR does
- Only display supported enrollment flows (in-person, digital or both) on Eligibility forms for Digital flow and In-Person flow.
- Update tests: Adds a spec case to both in-person and digital specs, where a form must filter out the 2 supported flows out of 3 total flows.

## How to test
- Set some enrollment flows to be In-Person only or Digital only via the Admin tool
- Check the digital flow's eligibility page
- Log in as a CST staff user on Admin and check the in-person flow page

### DIFF
--- a/tests/pytest/eligibility/test_views.py
+++ b/tests/pytest/eligibility/test_views.py
@@ -73,7 +73,7 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
 
 
 @pytest.mark.django_db
-def test_index_filtering_fows(mocker, model_TransitAgency, client):
+def test_index_filtering_flows(mocker, model_TransitAgency, client):
     digital = models.EnrollmentFlow.objects.create(
         transit_agency=model_TransitAgency,
         supported_enrollment_methods=[models.EnrollmentMethods.DIGITAL],
--- a/tests/pytest/eligibility/test_views.py
+++ b/tests/pytest/eligibility/test_views.py
@@ -80,6 +80,7 @@ def test_index_get_agency_multiple_flows(mocker, model_TransitAgency, model_Enro
     mock_manager = mocker.Mock()
     mock_manager.all.return_value = [model_EnrollmentFlow, model_EnrollmentFlow]
     type(mock_agency).enrollment_flows = mocker.PropertyMock(return_value=mock_manager)
+    type(mock_agency).enrollment_flows.filter.return_value = [model_EnrollmentFlow, model_EnrollmentFlow]
 
     mock_agency.index_url = "/agency"
     mock_agency.eligibility_index_template = "eligibility/index.html"
@@ -103,6 +104,7 @@ def test_index_get_agency_single_flow(mocker, model_TransitAgency, model_Enrollm
     mock_manager = mocker.Mock()
     mock_manager.all.return_value = [model_EnrollmentFlow]
     type(mock_agency).enrollment_flows = mocker.PropertyMock(return_value=mock_manager)
+    type(mock_agency).enrollment_flows.filter.return_value = [model_EnrollmentFlow, model_EnrollmentFlow]
 
     mock_agency.index_url = "/agency"
     mock_agency.eligibility_index_template = "eligibility/index.html"
--- a/tests/pytest/eligibility/test_views.py
+++ b/tests/pytest/eligibility/test_views.py
@@ -2,6 +2,7 @@ from django.urls import reverse
 
 import pytest
 
+from benefits.core import models
 from benefits.routes import routes
 from benefits.core.middleware import TEMPLATE_USER_ERROR
 import benefits.core.session
@@ -69,6 +70,36 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
     model_EnrollmentFlow.save()
     mocker.patch("benefits.eligibility.views.session.flow", return_value=model_EnrollmentFlow)
     return model_EnrollmentFlow
+
+
+@pytest.mark.django_db
+def test_index_filtering_fows(mocker, model_TransitAgency, client):
+    digital = models.EnrollmentFlow.objects.create(
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.DIGITAL],
+        label="Digital",
+        selection_label_template="eligibility/includes/selection-label.html",
+    )
+    in_person = models.EnrollmentFlow.objects.create(
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.IN_PERSON],
+        label="In-Person",
+        selection_label_template="eligibility/includes/selection-label.html",
+    )
+    both = models.EnrollmentFlow.objects.create(
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.DIGITAL, models.EnrollmentMethods.IN_PERSON],
+        label="Both",
+        selection_label_template="eligibility/includes/selection-label.html",
+    )
+    mocker.patch("benefits.core.session.agency", autospec=True, return_value=model_TransitAgency)
+
+    path = reverse(routes.ELIGIBILITY_INDEX)
+    response = client.get(path)
+    filtered_flow_ids = [choice[0] for choice in response.context_data["form"].fields["flow"].choices]
+
+    assert digital.id, both.id in filtered_flow_ids
+    assert in_person.id not in filtered_flow_ids
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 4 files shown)


tests/pytest/eligibility/test_views.py
```diff
@@ -71,6 +72,39 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
     return model_EnrollmentFlow
 
 
+@pytest.mark.django_db
+def test_eligibility_logged_in_filtering_flows(mocker, model_TransitAgency, client):
```
angela-tran, line 76: I really like that this single test covers that 1) digital flows are filtered "in" and 2) in-person flows are filtered out, as opposed to having two separate tests as you had pointed out in https://github.com/cal-itp/benefits/pull/2399#discussion_r1777666206. Nice job!

I think there is some code clean-up that can be done in this test. Some lines are redundant; more details are below in my comments on specific lines.

tests/pytest/eligibility/test_views.py
```diff
@@ -71,6 +72,39 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
...
+        selection_label_template="eligibility/includes/selection-label.html",
+    )
+    both = models.EnrollmentFlow.objects.create(
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.DIGITAL, models.EnrollmentMethods.IN_PERSON],
+        label="Both",
+        selection_label_template="eligibility/includes/selection-label.html",
+    )
+    model_TransitAgency.save()
+    model_TransitAgency.enrollment_flows.add(digital, in_person, both)
```
angela-tran, line 98: We don't need this line because when the `EnrollmentFlow`s are being created, they are already setting `model_TransitAgency` as their `transit_agency` (lines 80, 86, and 92).

tests/pytest/eligibility/test_views.py
```diff
@@ -71,6 +72,39 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
     return model_EnrollmentFlow
 
 
+@pytest.mark.django_db
+def test_eligibility_logged_in_filtering_flows(mocker, model_TransitAgency, client):
+    model_TransitAgency.pk = None
+    model_TransitAgency.save()
```
angela-tran, line 78: For this test, I don't think we need a new transit agency; it's the `EnrollmentFlow`s that we needed to have new instances of. Also, the flows are just using the existing `model_TransitAgency` anyways (`transit_agency=model_TransitAgency,`).

I think we can remove lines 77, 78, 97, and 98.

tests/pytest/eligibility/test_views.py
```diff
@@ -71,6 +72,39 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
...
+    )
+    model_TransitAgency.save()
+    model_TransitAgency.enrollment_flows.add(digital, in_person, both)
+    mocker.patch("benefits.core.session.agency", autospec=True, return_value=model_TransitAgency)
+
+    path = reverse(routes.ELIGIBILITY_INDEX)
+    response = client.get(path)
+
+    assert model_TransitAgency.enrollment_flows.count() == 3
+    assert len(response.context_data["form"].fields["flow"].choices) == 2
```
angela-tran, line 105: > I would have liked to have been able to have an assertion on the last line of these specs that are more like "The response's form should have a field that is supported by digital", but the `context_data["form"].fields["flow"].choices` only returns an id of the flow, not name of the flow or anything more. So I decided to test it like this instead. I'm open to other ideas on how to test this!
(_from https://github.com/cal-itp/benefits/pull/2399#discussion_r1777666206)_)

I think we could do some assertions on the IDs.

You could get a list of the flow IDs with a list comprehension:
```python
flow_ids = [choice[0] for choice in response.context_data["form"].fields["flow"].choices]
```
and then check if `digital.id` and `both.id` are in that list. And check that `in_person.id` is not.

tests/pytest/eligibility/test_views.py
angela-tran, line 76: I think the test name should be something like `test_index_filtering_flows`? 

My thinking is that it makes a request to `routes.ELIGIBILITY_INDEX`/the `index` view, and "logged in" doesn't apply to the `eligibility` app.

tests/pytest/in_person/test_views.py
```diff
@@ -37,6 +37,36 @@ def test_view_not_logged_in(client, viewname):
     assert response.url == "/admin/login/?next=" + path
 
 
+@pytest.mark.django_db
+def test_eligibility_logged_in_filtering_flows(mocker, model_TransitAgency, admin_client):
+    model_TransitAgency.pk = None
+    model_TransitAgency.save()
```
angela-tran, line 43: We can remove these lines as well for the same reason as the other test

tests/pytest/in_person/test_views.py
```diff
@@ -37,6 +37,36 @@ def test_view_not_logged_in(client, viewname):
...
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.IN_PERSON],
+        label="In-Person",
+    )
+    both = models.EnrollmentFlow.objects.create(
+        transit_agency=model_TransitAgency,
+        supported_enrollment_methods=[models.EnrollmentMethods.DIGITAL, models.EnrollmentMethods.IN_PERSON],
+        label="Both",
+    )
+    model_TransitAgency.enrollment_flows.add(digital, in_person, both)
```
angela-tran, line 58: We can remove this line as well for the same reason as the other test

tests/pytest/in_person/test_views.py
```diff
@@ -37,6 +37,36 @@ def test_view_not_logged_in(client, viewname):
...
+    mocker.patch("benefits.core.session.agency", autospec=True, return_value=model_TransitAgency)
+
+    path = reverse(routes.IN_PERSON_ELIGIBILITY)
+    response = admin_client.get(path)
+
+    assert model_TransitAgency.enrollment_flows.count() == 3
+    assert len(response.context_data["form"].fields["flow"].choices) == 2
+    assert response.context_data["form"].fields["flow"].choices[0][1] == "In-Person"
+    assert response.context_data["form"].fields["flow"].choices[1][1] == "Both"
```
angela-tran, line 67: Nice!

angela-tran (changes requested):
The code changes to application code look great! Noticed some code clean-ups for the test code, but otherwise, what they cover looks great.

tests/pytest/in_person/test_views.py
angela-tran: If you wanted to add assertions on the IDs, that would be fine too. These assertions seem precise enough since they're going further than just looking at the count.

tests/pytest/eligibility/test_views.py
```diff
@@ -71,6 +72,36 @@ def model_EnrollmentFlow_with_form_class(mocker, model_EnrollmentFlow):
     return model_EnrollmentFlow
 
 
+@pytest.mark.django_db
+def test_index_filtering_fows(mocker, model_TransitAgency, client):
```
angela-tran, line 76: Small typo
```suggestion
def test_index_filtering_flows(mocker, model_TransitAgency, client):
```

tests/pytest/in_person/test_views.py
```diff
@@ -49,6 +49,31 @@ def test_eligibility_logged_in(admin_client):
     assert response.template_name == "in_person/eligibility.html"
 
 
+@pytest.mark.django_db
+def test_index_filtering_fows(mocker, model_TransitAgency, admin_client):
```
angela-tran, line 53: This one was fine / more correct as `test_eligibility_logged_in_filtering_flows` since it's testing the `eligibility` view function.

```suggestion
def test_eligibility_logged_in_filtering_flows(mocker, model_TransitAgency, admin_client):
```

angela-tran (commented):
This is about ready to merge! The tests look so good - just some typos

machikoyasuda:
@angela-tran fows 😆

angela-tran:
Should we try to get #2403 done and approved / merged before merging this one in?

machikoyasuda:
image
The big enrollmentflow/agency refactor is now merged to `main` and deployed to `dev`, so this PR is now unblocked and ready to be reviewed + merged.

angela-tran (approved):
🎉

Issue #2384, opened 2024-09-19
Admin: Decide what to do with "Change password" page


angela-tran:
When logged in as a native Django user account, you'll see a "Change password" link in the usertools section.

![image](https://github.com/user-attachments/assets/08ac4b50-0579-4870-92d8-961edb8521b7)   

That link takes us to a page that we have not overridden / that doesn't use our `extrastyles`, so it uses all the default Django styles. 

![image](https://github.com/user-attachments/assets/2c376c35-4904-4905-94ad-4262c99e60e8)

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/pull/2371#discussion_r1767602802_

We should decide if we're keeping this page in and if we should style it like the rest of the agency-facing pages.

angela-tran:
I wanted to see what this page would look like if I made some bare minimum changes, like 
- import our admin styles and Bootstrap into the page
- remove sections like the breadcrumbs and sidebar so it matches our other pages

This is the result:

![Image](https://github.com/user-attachments/assets/b325b8ea-e9ae-4798-bc80-1950e020a8b1)

I'll open a draft PR that shows the few code changes that were needed to get this.

There's a template that's shown after the user successfully changes their password (`registration/password_change_done.html`), so I'd need to do the same thing for it: 

![Image](https://github.com/user-attachments/assets/b575359e-d29d-4a52-ba90-b4e10456e6eb)

angela-tran:
Here's what `password_change_done.html` looks like with the same type of minimal code changes:

![Image](https://github.com/user-attachments/assets/a751e6fa-f024-464f-b9d5-8980ad401b0a)

thekaveman:
@indexing Do we need to figure out what to do with this page before giving MST users access?

indexing:
We are addressing this issue in #2940, so I'm closing this one. cc// @ohmegasquared @thekaveman

PR #2423, opened 2024-10-03
Style: templates related to changing password in admin interface

angela-tran:
This is just an exploratory PR for #2384.

### DIFF
--- a/benefits/templates/registration/password_change_done.html
+++ b/benefits/templates/registration/password_change_done.html
@@ -1,0 +1,66 @@
+{% extends "registration/password_change_done.html" %}
+{% load static %}
+
+{% block extrastyle %}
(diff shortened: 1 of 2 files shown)


angela-tran:
Side note / thought: I think we could benefit from extracting out some of the common pieces that we've been copy-pasting around (like the `usertools` block, Bootstrap import, etc.) as include templates.

angela-tran:
If you look through the commits on this, you can see this was really just copy-pasting stuff from other templates.

thekaveman:
> Side note / thought: I think we could benefit from extracting out some of the common pieces that we've been copy-pasting around (like the `usertools` block, Bootstrap import, etc.) as include templates. This can be a separate issue / PR.

This was my exact thought :sweat_smile:

PR #2430, opened 2024-10-07
Chore: extract common Admin overrides

### LINKED ISSUE #2425: Extract out the common Admin overrides
Side note / thought: I think we could benefit from extracting out some of the common pieces that we've been copy-pasting around (like the `usertools` block, Bootstrap import, etc.) as include templates. This can be a separate issue / PR.

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/issues/2423#issuecomment-2392390753_

lalver1:
Closes #2425

This PR has a few cleanups. It fixes the indentation of the `usertools` block in `agency-base`; extracts `favicon`, `Bootstrap`, and `styles.css` for the admin app; and extracts  the `login` and `logged_out` branding.

### DIFF
--- a/benefits/templates/admin/agency-base.html
+++ b/benefits/templates/admin/agency-base.html
@@ -28,30 +28,7 @@
 {% endblock branding %}
 
 {% block usertools %}
-    {% if has_permission %}
-        <div id="user-tools">
-            {% block welcome-msg %}
-                <span class="text-uppercase text-white">
-                    Welcome,
-                    <span class="fw-bold">{% firstof user.get_short_name user.get_username %}</span>.
-                </span>
-            {% endblock welcome-msg %}
-            {% block userlinks %}
-                {% if user.is_active and user.is_staff %}
-                    {% url 'django-admindocs-docroot' as docsroot %}
-                    {% if docsroot %}<a href="{{ docsroot }}">Documentation</a> /{% endif %}
-                {% endif %}
-                {% if user.has_usable_password %}
-                    <a href="{% url 'admin:password_change' %}">Change password</a> /
-                {% endif %}
-                <img class="icon" width="15" height="15" src="{% static "img/icon/box-arrow-right.svg" %}" alt="" />
-                <form id="logout-form" method="post" action="{% url 'admin:logout' %}">
-                    {% csrf_token %}
-                    <button type="submit" class="border-0">Log out</button>
-                </form>
-            {% endblock userlinks %}
-        </div>
-    {% endif %}
+    {% include "admin/includes/usertools.html" %}
 {% endblock usertools %}
 
 {% block coltype %}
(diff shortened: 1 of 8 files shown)


lalver1:
I could group `bootstrap.html`, `favicon.html` and `style-admin.html` into one `html` file but I was having trouble finding a short name for it and also thought that breaking them out made the `{% include %}` statements clearer, but I'm leaning either way.

angela-tran:
> I could group `bootstrap.html`, `favicon.html` and `style-admin.html` into one `html` file but I was having trouble finding a short name for it and also thought that breaking them out made the `{% include %}` statements clearer, but I'm leaning either way.

I think it's better to have them as separate includes 👍

benefits/templates/admin/agency-base.html
```diff
@@ -36,45 +33,46 @@ <h1 class="p-0 m-0 text-white">{{ site_header }}</h1>
...
+                {% endif %}
+                <img class="icon" width="15" height="15" src="{% static "img/icon/box-arrow-right.svg" %}" alt="" />
+                <form id="logout-form" method="post" action="{% url 'admin:logout' %}">
+                    {% csrf_token %}
+                    <button type="submit" class="border-0">Log out</button>
+                </form>
+            {% endblock userlinks %}
+        </div>
+    {% endif %}
+{% endblock usertools %}
```
angela-tran, line 55: I think we should extract this `usertools` block out as an include template as well. Your thoughts @lalver1?

lalver1: Yep, I was also wondering about this @angela-tran. Ah, because `usertools` will be used in a few places in PR #2423  it would be good to move it to an include template 👍

benefits/templates/admin/includes/favicon.html
```diff
@@ -0,0 +1,3 @@
+{% load i18n static %}
+
+<link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
```
machikoyasuda, line 3: There is no `img/favicon.ico`. Maybe need a future ticket to make a new favicon? I don't think it should be the same favicon as https://benefits.calitp.org/.

cc @indexing

benefits/templates/admin/agency-base.html
lalver1: Moved `usertools` to an include template in 6ee7e3cc222092ecb9ef21d92f3470768aa851b9

benefits/templates/admin/includes/favicon.html
lalver1: Thanks @machikoyasuda! I'm looking into this to see where favicon gets set in the `index` admin page that we see when we log in as a superuser.

indexing: Yes @machikoyasuda. Let's use the same favicon.

lalver1: Sounds good! I'll create a separate ticket to fix this favicon issue.

angela-tran (approved):
This looks good to me. Thanks!

PR #2446, opened 2024-10-09
Fix: add to CSP_FRAME_SRC

### LINKED ISSUE #2445: Blocked 'frame-src' from 'verify.littlepay.com'
Sentry Issue: [BENEFITS-90](https://sentry.calitp.org/organizations/sentry/issues/77636/?referrer=github_integration)

```
{"csp-report":{"effective_directive":"frame-src","blocked_uri":"https://verify.littlepay.com","document_uri":"https://benefits.calitp.org/in_person/enrollment/","original_policy":"connect-src 'self' https://api.amplitude.com/; frame-src https://www.google.com; script-src 'self' https://cdn.amplitude.com/libs/ https://cdn.jsdelivr.net/ https://*.littlepay.com https://code.jquery.com/jquery-3.6.0.min.js https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/releases/ 'nonce-rTHCLTlH5svycrZpbo6qMA=='; font-src 'self' https://california.azureedge.net/ https://fonts.gstatic.com/; base-uri 'none'; object-src 'none'; img-src 'self' data: https://*.googleusercontent.com; style-src 'self' 'unsafe-inline' https://california.azureedge.net/ https://fonts.googleapis.com/css https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/; frame-ancestors 'none'; default-src 'self'; report-uri https://sentry.calitp.org/api/3/security/?sentry_key=684f0b010b09422f9c5f4ad6b1694ec7","referrer":"https://benefits.calitp.org/in_person/eligibility/","status_code":0,"violated_directive":"frame-src","column_number":1,"disposition":"enforce"}}

Blocked 'frame-src' from 'verify.littlepay.com'
```

angela-tran:
Closes #2445 

#2445 revealed that in production, the `frame-src` directive only includes `google.com` and therefore blocks the transit processor iframe.

This is because in `settings.py`, the logic was to throw away the originally assigned value of `CSP_FRAME_SRC` and instead use whatever is in `env_frame_src`.

https://github.com/cal-itp/benefits/blob/8cce5addeaff897220fd97e8a7eb25f3cbe0b285/benefits/settings.py#L310-L315

This logic was fine when the originally assigned value was `'none'` since the directive only allows [`'none'` or a space separated list of values](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src#syntax), but in [a23b51e2](https://github.com/cal-itp/benefits/commit/a23b51e26511110af8528bdcd4c70027ad6635ba#diff-d71405aa44f1e899a9cd920e5187aa1d02affdbd0da472eca95f814b0a8c1210L302), we changed it to be `*.littlepay.com` since we know we will always need to allow that as an iframe source.

This PR fixes it so we add to `CSP_FRAME_SRC` instead of re-assigning it.

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -312,7 +312,7 @@ env_frame_src = _filter_empty(os.environ.get("DJANGO_CSP_FRAME_SRC", "").split("
 if RECAPTCHA_ENABLED:
     env_frame_src.append("https://www.google.com")
 if len(env_frame_src) > 0:
-    CSP_FRAME_SRC = env_frame_src
+    CSP_FRAME_SRC.extend(env_frame_src)
 
 CSP_IMG_SRC = [
     "'self'",


lalver1 (approved):
This looks good 👍 
I saw that before this PR, `*.littlepay.com` only appeared in `script-src` in the CSP header but now it also appears in `frame-src`.

thekaveman (approved):
Yep, looks good :+1:

Now that we're initializing the frame src with the Littlepay domain, no need to replace.

Issue #2433, opened 2024-10-08
Setting up SBMTD in-person for October 17th


thekaveman:
- [x] Meet with @indexing to go over what is needed
- [x] Confirm SBMTD `EnrollmentFlow`s are configured correctly
   - [x] Check `label` is set to correct display value
   - [x] Check over other config for OAuth, eligibility-server API, transit processor   
- [x] Configure `TransitProcessor.portal_url`
- [x] Enable login fields
- [x] Set up accounts for transit agency staff members
   - [x] Check `Staff` checkbox for each user
   - [x] Prep for sharing (initial) passwords securely
- [x] Create `Group`s to set as `staff_group` and `customer_service_group` on SBMTD `TransitAgency`, and add users to groups
- ~Do an internal test enrollment (if we can delete the EnrollmentEvent) (or maybe it's ok if we see the user was one of us?)~ - we decided to hold off on doing this, see [this Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1728597005282719?thread_ts=1728592143.795789&cid=C037Y3UE71P)
- [x] Prep to do a test run of in-person enrollment with the agency
   - Customer success will facilitate
   - Will run through it with a real eligible user so we don't have to go back and delete any test EnrollmentEvents
 - [x] Fix #2445 - #2446 
 - [x] Remove the "Change password" link so it doesn't distract from this beta test - #2447

indexing:
cc// @o-ram

angela-tran:
Prep work done

PR #2447, opened 2024-10-09
Chore: remove "Change password" link from admin interface for now

angela-tran:
Part of #2433 

@indexing mentioned we don't need the "Change password" feature for the in-person event on October 17th and that it would be better to not show it.


lalver1 (approved):
Looks good! 👍

angela-tran:
Related to #2384

PR #2456, opened 2024-10-15
Enrollment Index: Remove space after Littlepay

### LINKED ISSUE #2428: Enrollment page: Space after Littlepay
Problem: There should not be an extra whitespace after the Littlepay icon link on the Eligibility confirmation page:

Image
Image

----

Should look like this:
Image

The linter doesn't allow this text to be on one line, but it needs to be on one line so the `whitespace` is eliminated:

I noticed that the rendered code in the text after the image has some whitespace built in: Image

Need to remove this whitespace: Image
Removing that whitespace manually removes the extra space, and makes the page look correct.

This ticket will involve investigating why this whitespace get rendered and figure out how to fix it.

---

These are similar kinds of links, and don't have the extra space:
Image

Possible fix:

I noticed that not using the modal trigger includes and using a djLint off ignore fixes it: 

```
            <!-- djlint:off -->
            <a href="modal--littlepay" class="{{ classes }}" data-bs-toggle="modal" data-bs-target="#modal--littlepay">Littlepay</a>{% translate ", to enter your contactless card details." %}<!-- djlint:on -->
```

machikoyasuda:
closes #2428 

## What this PR does
- Make a fix to the Enrollment Index page that shows up on the Agency Card, Veterans, Senior, Medicare flows (all but CalFresh).
- Remove the space that appears after the Littlepay link by *not* using the includes and using the `a href` link code directly, allowing the link code to be right next to the copy, which allows the HTML to be rendered w/o any extra whitespaces.
- The only drawback to this is that if there is ever a style change to `modal-trigger`, we need to make sure the change is made here too. I found this drawback to be acceptable, b/c I think most changes if any that happen to the modal-trigger will be done with CSS classes, and it wouldn't be hard to add the CSS classes in both places. I'm open to the alternatives considered though.

image

image

image

## Alternatives considered
- Tried removing line breaks from the `modal-trigger.html` template, from the `media-item` template, but didn't fix the bug.
- I considered trying to add conditional code to the `modal-trigger.html` includes, something like _If no_space=True, then render `{% translate ", to enter your contactless card details." %}`_ - but it felt a little over the top, doesn't scale if this ever happened on another page. Would this be preferred though? I can do that.

### DIFF
--- a/benefits/enrollment/templates/enrollment/includes/media-item--bankcardcheck--index.html
+++ b/benefits/enrollment/templates/enrollment/includes/media-item--bankcardcheck--index.html
@@ -17,8 +17,7 @@
     <div class="media-body--details">
         <p>
             {% translate "You will be directed to our partner, " %}
-            {% include "core/includes/modal-trigger.html" with modal="modal--littlepay" text="Littlepay" period=False %}
-            {% translate ", to enter your contactless card details." %}
+            <!-- djlint:off --><a href="modal--littlepay" class="{{ classes }}" data-bs-toggle="modal" data-bs-target="#modal--littlepay">Littlepay</a>{% translate ", to enter your contactless card details." %}<!-- djlint:on -->
             {% translate "We don’t store your information, and you won’t be charged." %}
         </p>
         <p>{% translate "Please use a debit or credit card by Visa or Mastercard." %}</p>


thekaveman (approved):
Yep, I agree your approach is simpler and fine to maintain compared to the alternatives.

Modal link works  :white_check_mark: 

<details><summary>English :white_check_mark: </summary>

| `dev` | `local` |
| ------- | --------- |
| image | image |

</details>

<details><summary>Spanish  :white_check_mark:</summary>

| `dev` | `local` |
| ------- | --------- |
| image | image |

</details>

PR #2460, opened 2024-10-15
Footer: Add Compiler line

### LINKED ISSUE #2449: Add "Powered by" link to application footer
We'd like visitors to the Benefits application to understand who is behind the planning, design, development, and maintenance of the software. We can add a simple link to the application footer, visible on all screens in the app.

I suggest placement at the far right side of the footer, within the overall container, e.g.

Image

Use the following language:

> Powered by [Compiler LLC](https://compiler.la)

## Acceptance Criteria

- [x] Copy is translated to Spanish
- [x] Design for an additional footer link with the above text is done and handed-off to Dev
- [ ] Dev implements the link in Benefits app footer

## Additional context

Related to #2448

machikoyasuda:
closes #2449 

image
image
image

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -111,6 +111,9 @@
           <li>
             <a class="footer-link m-0 p-0" href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer">{% translate "Privacy Policy" %}</a>
           </li>
+          <li class="ms-auto">
+            <a class="footer-link m-0 p-0" href="https://compiler.la" target="_blank" rel="noopener noreferrer">{% translate "Powered by Compiler LLC" %}</a>
+          </li>
         </ul>
       </div>
 
@@ -128,6 +131,16 @@
         <ul class="col-12 footer-links ps-0 mb-0">
           <li>
             <a class="footer-link" href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer">{% translate "Privacy Policy" %}</a>
+          </li>
+        </ul>
+      </div>
+      <div class="d-block d-lg-none container-fluid p-0">
+        <hr class="border border-white border-1 p-0 m-0 w-100 opacity-100">
+      </div>
+      <div class="d-block d-lg-none container">
+        <ul class="col-12 footer-links ps-0 mb-0">
+          <li>
+            <a class="footer-link" href="https://compiler.la" target="_blank" rel="noopener noreferrer">{% translate "Powered by Compiler LLC" %}</a>
           </li>
         </ul>
       </div>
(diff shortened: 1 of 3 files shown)


benefits/core/templates/core/base.html
```diff
@@ -111,6 +111,9 @@
           <li>
             <a class="footer-link m-0 p-0" href="https://cdt.ca.gov/privacy-policy/" target="_blank" rel="noopener noreferrer">{% translate "Privacy Policy" %}</a>
           </li>
+          <li class="ms-auto">
```
machikoyasuda, line 114: `ms-auto` means "margin _start_: auto", and "start" means `left` if the page is set to a language that reads from left to right.

benefits/core/templates/core/base.html
```diff
@@ -111,6 +111,9 @@
...
+            <a class="footer-link m-0 p-0" href="https://compiler.la" target="_blank" rel="noopener noreferrer">{% translate "Powered by Compiler LLC" %}</a>
```
thekaveman, line 115: Sorry, I didn't notice this in the design. I think the link text should not include the "Powered by" copy, that should come before/outside the link text.

This was what I wrote in the [original ticket](https://github.com/cal-itp/benefits/issues/2449#issue-2579250422):

> Use the following language:

>> Powered by [Compiler LLC](https://compiler.la)

thekaveman (changes requested):

benefits/core/templates/core/base.html
srhhnry: Updated to this: 
Screenshot 2024-10-17 at 6 06 21 PM

I think it's a cleaner look to have the entire phrase be a link, and I don't see that being an issue with screenreaders either. HOWEVER, either way is fine by me.

machikoyasuda: Hrm.. this is actually tricky because the footer links change color after they've been visited, but the "Powered by" would stay light blue. So the Powered by always stay light blue, and the Compiler LLC part changes to purple after it's been clicked:

image

Are we okay with this?

thekaveman: I would expect the "Powered by" to be... white text? Making it the same color as the link feels strange.

thekaveman: If this is too complicated, leaving the entire text as the link is fine.

machikoyasuda: This is how it would look with white/purple visited link or white/blue unvisited link:

image
image
image
image

machikoyasuda: Here's the commit for the white text + regular link change: https://github.com/cal-itp/benefits/pull/2460/commits/7d20a5be4ee11a6fd67be970d28ce4c0485e3387

thekaveman (approved):

thekaveman:
Thanks for the last-minute updates.

Issue #2355, opened 2024-09-11
Medicare cardholder: distinguish between Older Adult and Disabled riders


thekaveman:
## Background

[Medicare](https://en.wikipedia.org/wiki/Medicare_(United_States)) is available to generally two different types of people:

* Those who are 65 years of age and older
* Those who are under 65, but have some sort of qualifying disability

This distinction between different types of Medicare recipients is important for the Benefits application enrollment analytics and reporting purposes.

## Current state

The Medicare Cardholder flow will rely on the OIDC exchange of scopes and claims with the CDT Identity Gateway (IdG). The code that implements this exchange is found in: 

* https://github.com/cal-itp/benefits/blob/main/benefits/oauth/client.py#L44 (configuring the scope)
* https://github.com/cal-itp/benefits/blob/main/benefits/oauth/views.py#L89 (parsing the claim)

The [`EnrollmentFlow` model](https://github.com/cal-itp/benefits/blob/main/benefits/core/models.py#L105) contains all the information needed for this exchange: the scope(s), the claim, _AND_ the value eventually needed for analytics (the system name).

We know that the IdG supports receiving multiple scopes with a given request; we also know that the IdG supports sending back multiple associated claims.

## Proposal

The dev team discussed various approaches on 10/9 and here's where we landed:

* We'll use the pre-calculated `medicare` and `disabled` claims coming out of IdG rather than doing any calculation on the Benefits side
* We'll represent the "main" claim under verification (e.g. `medicare`) AND any additional claims (e.g. `disabled`) on the `EnrollmentFlow` model

## Acceptance criteria

- [ ] `EnrollmentFlow.claims_claim` is renamed to something like `EnrollmentFlow.eligibility_claim`
- [ ] New (optional) field `EnrollmentFlow.extra_claims` -- space separated string of any additional claims for that flow
- [ ] The `userinfo` object returned in the IdG token is parsed for all claims on the given flow
- [ ] Session is updated to store _one or more_ verified OAuth claims
- [ ] `eligibility.verify.eligibility_from_oauth` modified to check only the "main" claim of the flow
- [ ] _TBD analytics is updated with all claims that came back as verified_

thekaveman:
@indexing @o-ram we discussed an approach for this that will work given what we know IdG is capable of supporting.

The last question is, how should this distinction be represented in analytics? 

* We currently send the `enrollment_flows` property with relevant events (this was recently renamed, from `eligibility_types`)
[...]

PR #2463, opened 2024-10-18
Feat: process multiple claims on a flow

lalver1:
Part of #2355

This PR implements a way to distinguish between older adult and disabled Medicare sessions. `EnrollmentFlow` is modified so that it now stores a "main" claim and an optional set of "extra" claims. An `@property` method on the model takes care of arranging them into one list of claims. The code now handles this list of claims everywhere that multiple claims are involved. `session` has been modified to store and process this list of claims to make it available to verify and analytics. Updating analytics will probably be part of a separate PR.


benefits/core/models.py
```diff
@@ -282,9 +282,12 @@ class EnrollmentFlow(models.Model):
         blank=True,
         help_text="A space-separated list of identifiers used to specify what access privileges are being requested",
     )
-    claims_claim = models.TextField(
+    claims_eligibility_claim = models.TextField(
         null=True, blank=True, help_text="The name of the claim (name/value pair) that is used to verify eligibility"
     )
+    claims_extra_claims = models.TextField(
+        null=True, blank=True, help_text="A space-separated list of any additional claims (name/value pair)"
```
thekaveman, line 289: I think this help text is a little misleading (I see you just grabbed it from the item above).

This isn't a name/value pair right? It's just the name of the claim. Can we edit these help texts?

thekaveman (changes requested):

benefits/core/models.py
lalver1: Yep, I'm sure I was thinking about scopes and thought that name/value made sense but these are claims, thanks for pointing this out.

I implemented the change in 1de70c6da2ef2cfaedd3e20cc68aed71d4201dd1 and to avoid data migration operations, used two steps in the Django migration, a `RenameField` and an `AlterField`. Otherwise, Django thinks `claims_eligibility_claim` is a brand new field and that would require migrating data.

benefits/core/session.py
```diff
@@ -148,17 +148,17 @@ def logged_in(request):
 
 def logout(request):
     """Reset the session claims and tokens."""
-    update(request, oauth_claim=False, oauth_token=False, enrollment_token=False)
+    update(request, oauth_claims=False, oauth_token=False, enrollment_token=False)
```
thekaveman, line 151: I know you didn't introduce this, but I'm wondering if `False` makes sense here, especially now with the list approach? Should we set this to an empty list `[]` instead?

benefits/oauth/views.py
```diff
@@ -123,27 +123,29 @@ def authorize(request):
...
+                # the claim comes back in userinfo like { "claim": "1" | "0" }
+                claim_value = int(claim_value) if claim_value else None
+                if claim_value is None:
+                    logger.warning(f"userinfo did not contain: {claim}")
+                elif claim_value == 1:
+                    # if userinfo contains our claim and the flag is 1 (true), store the *claim*
+                    stored_claims.append(claim)
+                elif claim_value >= 10 and claim == flow.claims_eligibility_claim:
+                    # error_claim is only set if claim is the eligibility claim
+                    error_claim = claim_value
```
thekaveman, line 146: This can be saved for the follow-up PR that implements analytics with the new `extra_claims`...

I'm thinking we should send all the errors with the analytics event, instead of just the error on the `eligibility_claim`.

E.g. make `error_claim` a `dict` for `{"claim": "error_code"}` and send that with the analytics.

benefits/core/models.py
```diff
@@ -459,6 +460,12 @@ def claims_scheme(self):
             return self.claims_provider.scheme
         return self.claims_scheme_override
 
+    @property
+    def claims_all_claims(self):
+        claims = [self.claims_eligibility_claim]
+        claims.extend(self.claims_extra_claims.split())
```
thekaveman, line 466: I like this helper to calculate the full list in the model.

However this call to `.split()` will throw an error if `self.claims_extra_claims` is `None` (the default).

We should either set the default to an empty string, or handle when this is `None`.

thekaveman (changes requested):

benefits/core/models.py
angela-tran: I agree and want to add that we should have a unit test that covers this.

benefits/oauth/views.py
angela-tran: Yeah, I agree we haven't discussed how and which error codes should be sent, so we can save it for the other PR. Good to note this though

tests/pytest/core/test_session.py
```diff
@@ -269,12 +269,12 @@ def test_reset_enrollment(app_request):
 @pytest.mark.django_db
 def test_reset_oauth(app_request):
     app_request.session[session._OAUTH_TOKEN] = "oauthtoken456"
-    app_request.session[session._OAUTH_CLAIM] = "claim"
+    app_request.session[session._OAUTH_CLAIMS] = "claim"
```
angela-tran, line 272: Similar comment as above: I think this should be a list like `app_request.session[session._OAUTH_CLAIMS] = ["claim"]` to be more aligned with what the value would actually be set to.

benefits/core/session.py
lalver1: Yep, makes sense to change it to `[]`, changed in c0a2516db02df7dea3467262321d145e5117d8b4

benefits/core/models.py
lalver1: Ah, thanks both for pointing this out 👍
In f9eaa59d68b1dfb9eccb791bfc0bfe018095fde3 we now have a handler to make sure we don't get an error and a test for when we have `self.claims_extra_claims` set to `None`, one claim, or two claims separated by a space.

tests/pytest/core/test_session.py
```diff
@@ -199,16 +199,16 @@ def test_logged_in_True(app_request):
 
 @pytest.mark.django_db
 def test_logout(app_request):
-    session.update(app_request, oauth_claim="oauth_claim", oauth_token="oauth_token", enrollment_token="enrollment_token")
+    session.update(app_request, oauth_claims="oauth_claim", oauth_token="oauth_token", enrollment_token="enrollment_token")
```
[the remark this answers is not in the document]
lalver1: Definitely agree, thanks! Changed it in c0a2516db02df7dea3467262321d145e5117d8b4

tests/pytest/core/test_session.py
lalver1: Definitely agree, thanks! Changed it in c0a2516db02df7dea3467262321d145e5117d8b4

benefits/oauth/views.py
lalver1: Good points, noted for the follow-up PR 👍

angela-tran (approved):
Looks good to me 👍

thekaveman (approved):
:+1:

PR #2465, opened 2024-10-18
Feat: add Slack thread to Release issues

thekaveman:
Adds a new question / section for the the Release process issue template to make it easier to paste in the Slack thread in a standard location.

This is not required to create a Release issue. Also removed the requirement for smoke tester, as that is not always known when the Release issue is created.


lalver1 (approved):
Looks good!

PR #2471, opened 2024-10-22
Refactor: field defaults for optional fields, templates

thekaveman:
This PR is another step along the way of making it easier and faster to configure new `TransitAgency` and `EnrollmentFlow` instances.

### `TransitAgency`

- Make the `slug` field a Django `SlugField` to enforce the URL slugness
- Make all Eligibility API fields optional
- Calculate sane defaults for templates from the `slug` following our established pattern, e.g. `core/index--{slug}.html` and `eligibility/index--{slug}.html`
- Retain ability to override templates if needed

## `EnrollmentFlow`

- Make all template fields optional with a default of `None`
- Calculate sane defaults for templates from the `system_name` or `slug` following our established pattern, e.g. `eligibility/start--{system_name}.html` and `enrollment/success--{slug}.html`
- Calculate defaults specific to Agency Card flows establishing a new pattern, e.g. `eligibility/start--{slug}-agency-card.html` (essentially removing the specific name of the agency card from the template, like `courtesy-card`).
- Retain ability to override templates if needed


lalver1 (dismissed):
I tested running through these flows and they all looked good:

- Medicare ✅ 
- Older Adult ✅ 
- Calfresh ✅ 
- Veteran ✅ 
- Agency Card ✅ 

These admin panel actions also behaved as expected:
1. New `TransitAgency` instances can be created with only the slug field provided ✅ 
2. Saving an active `TransitAgency` requires certain info fields to have values ✅ 
3. Saving an active `TransitAgency` requires templates to exist (and the defaults are calculated from the `slug`) ✅ 
4. Saving an `EnrollmentFlow` with an active `TransitAgency` requires templates to exist, and defaults are calculated from the `system_name` or `TransitAgency.slug` ✅ 

At first I was a bit confused about action 4 above, because to review I created a new `EnrollmentFlow` without claims information for an active agency and so what I got in the eligibility form was another `Agency Cardholder` radio input identical to the existing one. But this is completely normal and basically shows that for a case like this, overriding the selection label template would be needed.

image

lalver1:
Yep, now that I look at it again, I feel it should be an error. I think that if we add something like

```python
if self.claims_provider and not bool(self.claims_scope):
    errors.update(claims_scope=ValidationError("Missing scope"))

if self.claims_provider and not bool(self.claims_eligibility_claim):
    errors.update(claims_eligibility_claim=ValidationError("Missing eligibility claim"))
 ```
 
 to `clean()` anywhere before 
 
 ```python
if errors:
    raise ValidationError(errors)
 ```
 
then the form will give a validation error if a `claims_provider` has been selected but no scope/claim has been entered. Then let's suppose that the scope/claim info is entered but templates have not been created, then the existing `template` validation section of `clean()` (starting on [line 624](https://github.com/cal-itp/benefits/blob/b413ddc919e67ea821c914975ca1186b411d97d0/benefits/core/models.py#L624)) will correctly return that the template was not found.

I think a similar check should happen if it is an Eligibility API flow. Currently if `self.claims_provider ` is blank, we can still save the flow and another `Agency Cardholder` radio input identical to the existing one is displayed.

**Comment by angela-tran** [comment]
Reviewing this

(just realized I hadn't indicated that here yet :sweat_smile:)

File: benefits/core/models.py [hunk 1]
```diff
@@ -251,6 +306,34 @@ def transit_processor_client_secret(self):
     def enrollment_flows(self):
         return self.enrollmentflow_set
 
+    def clean(self):
```
angela-tran, line 309: Are we ok with this behavior:

- Create a new TransitAgency by providing a slug and leaving the rest as defaults
   - (agency is inactive)
- Create a new EnrollmentFlow by providing a system name and configure its `transit_agency` as the new TransitAgency, leave all the other fields as defaults
   - (flow has no configuration for claims or API verification)
- go back to the TransitAgency, fill out the required info fields, and make it active by checking the "Active" checkbox


Result: the agency is saved and made active, but the user might not realize the invalid EnrollmentFlow is connected to the active TransitAgency.

I'm not necessarily arguing for one way or the other, but just wanted clarity on if this is intentional.

angela-tran (changes requested):
Lots of nice refactors in this PR @thekaveman!

I thought these were really great changes:
- removing `null=True` for fields where it was not truly needed and default empty string worked just as well / is better
- making more fields optional (basically allowing for saving an empty, inactive object that can be filled out later)
- fixing things like `group_id` and `label` not allowing blank

I have two pieces of feedback:
- a requested change for the default values for TransitAgency's info fields
- a question on validation behavior

Let me know if you want to talk through any of these. Thanks!

benefits/core/models.py
thekaveman: Ah, that is interesting... 

I have the `EnrollmentFlow.clean()` calling its `TransitAgency.clean()` -- but maybe it makes more sense the other way around?

benefits/core/models.py
```diff
@@ -251,6 +306,34 @@ def transit_processor_client_secret(self):
...
+            message = "This field is required for active transit agencies."
+            needed = dict(
+                short_name=self.short_name,
+                long_name=self.long_name,
+                phone=self.phone,
+                info_url=self.info_url,
+                logo_large=self.logo_large,
+                logo_small=self.logo_small,
+            )
+            field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
```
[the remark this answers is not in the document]
thekaveman: Yes, I agree with you -- the behavior you describe is more desirable than having the CST defaults.

benefits/core/models.py
angela-tran: Oh I see. Yeah, since it's upon saving the TransitAgency to be active that we would want to make sure everything is valid, I think it makes more sense the other way around. 

I think this means we'd remove `EnrollmentFlow.clean()` calling its `TransitAgency.clean()`, and instead `TransitAgency.clean()` would call `EnrollmentFlow.clean()` for all its related `EnrollmentFlow`s.

angela-tran: I'm curious what that validation would end up looking like in the UI. The user is looking at a `TransitAgency` form that does validation for some other objects :thinking: Hopefully it shows up in a reasonable way

thekaveman: I'm curious too, I hadn't thought of a way to test this. But I think inverting the `clean()` gives us a way to test as you outline.

thekaveman: Hmm well, not good news

I inverted the `clean()` and now we can trigger the condition, but, it's an error in Django Admin since the fields from `EnrollmentFlow` don't exist on `TransitAgency`:

image

So I think we have to remove the `clean()` entirely.

benefits/core/models.py
```diff
@@ -206,15 +255,13 @@ class TransitAgency(models.Model):
...
-        null=True,
+        default="agencies/cst-lg.png",
         blank=True,
         upload_to=agency_logo_large,
         help_text="The large version of the transit agency's logo.",
     )
     logo_small = models.ImageField(
-        default=None,
-        null=True,
+        default="agencies/cst-sm.png",
```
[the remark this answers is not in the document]
thekaveman: Fixed this in a new combined commit [`2b03147`](https://github.com/cal-itp/benefits/pull/2471/commits/2b03147deae3a1b4ea6ceda9d00bc57a7ae4e807).

benefits/core/models.py
```diff
@@ -145,41 +166,70 @@ class TransitAgency(models.Model):
...
     long_name = models.TextField(
-        help_text="The user-facing long name for this agency. Often the short_name acronym, spelled out."
+        default="California State Transit",
+        help_text="The user-facing long name for this agency. Often the short_name acronym, spelled out.",
+    )
+    info_url = models.URLField(
+        default="https://www.agency-website.com",
+        help_text="URL of a website/page with more information about the agency's discounts",
+    )
+    phone = models.TextField(default="1-800-555-5555", help_text="Agency customer support phone number")
```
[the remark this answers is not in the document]
thekaveman: Fixed this in a new combined commit [`2b03147`](https://github.com/cal-itp/benefits/pull/2471/commits/2b03147deae3a1b4ea6ceda9d00bc57a7ae4e807).

benefits/core/models.py
angela-tran: Hmm I see. Seems like what's happening is `EnrollmentFlow.clean()` raises field-level `ValidationError`s which then `TransitAgencyForm` doesn't have fields for, so understandably the `ValueError` is raised.

I started thinking if perhaps we could catch the error from `EnrollmentFlow.clean()` and have logic to handle it in `TransitAgency.clean()`, so something like this:
```python
        for flow in self.enrollment_flows.all():
            try:
                flow.clean()
            except Exception:
                raise ValidationError("There was an error with a related EnrollmentFlow.")
```

We could do more to make the message more helpful to the user (like somehow including the `EnrollmentFlow` system name), but to avoid scope creep, this simple message would at least help the user not activate a TransitAgency with invalid EnrollmentFlows.

Here's what it looks like when you try to activate the `TransitAgency`:
Screenshot from 2024-11-26 17-05-18

Reloading the page shows that the agency was not activated. (I reloaded the page by selecting the URL bar in my browser and hitting Enter to make sure it's a GET request)

Screenshot from 2024-11-26 17-14-36

thekaveman:
Rebased with new commit [`6f0ed51`](https://github.com/cal-itp/benefits/pull/2471/commits/6f0ed5140738fe2256793ec3c5aef226770eb4da) to address @angela-tran's comments: https://github.com/cal-itp/benefits/pull/2471#discussion_r1859389877

`TransitAgency.clean()` calls `clean()` on each of its `EnrollmentFlow`, catches any `ValidationError`, and rethrows with a  message indicating an invalid flow.

This makes it less likely that an agency could go live with a misconfigured flow. 

Thanks for continuing to push on this one Angela, I think this is definitely the better outcome. 

## Test this scenario

1. Login to the Admin
2. Create a `TransitAgency` with all fields necessary to be `active=True`, but don't make it active yet. Save.
3. Create an `EnrollmentFlow` with only a `claims_provider`; select the `TransitAgency` from above. Save.
4. Go back to the `TransitAgency` and check to make it `active=True`. Save.
5. :boom:

benefits/core/models.py
thekaveman: [`a8c40bb`](https://github.com/cal-itp/benefits/pull/2471/commits/a8c40bb6e9f95146790f6bccf7d871f3037d2190)

lalver1:
I really like these latest changes where `TransitAgency.clean()` calls `clean()` on each of its `EnrollmentFlow` 👍
One thing I'm wondering is, if we should remove `and self.transit_agency.active` from 

https://github.com/cal-itp/benefits/blob/a8c40bb6e9f95146790f6bccf7d871f3037d2190/benefits/core/models.py#L633

so that the user can still see the particular validation errors when an `EnrollmentFlow` is being configured. 

Following the testing steps, the validity of an `EnrollmentFlow` is only checked when setting a `TransitAgency` to `active` and attempting to save, and if there is an error, the message is only that the flow is invalid. So then we would go back to the flow, but we'd have no hint of what is invalid on the flow since the `TransitAgency` is inactive (because we were not allowed to activate it from the previous step) and the `field` and `template` validations are not ran because of the condition referenced above.

The only downside I guess of making this change is that if the `EnrollmentFlow` is associated with a `TransitAgency`, the user needs to have all the information for an `EnrollmentFlow` before being able to save it, regardless of the `active` status of the `TransitAgency`. But maybe this isn't too bad since if you can choose to not associate a `TransitAgency` with an `EnrollmentFlow`, and this would allow you to temporarily save an `EnrollmentFlow` that is not completely configured yet.

thekaveman:
Ahh @lalver1 that's a really good point!! :sob: 

Yes, I agree with you -- I'm going to update that commit one more time :crossed_fingers:

thekaveman:
@lalver1 great suggestion! [`b48b7ac`](https://github.com/cal-itp/benefits/pull/2471/commits/b48b7aca1285d6b165db3b31fa248f635c2d4758)

angela-tran (approved):
Nice work on all this @thekaveman!

lalver1 (approved):
This looks great  @thekaveman and @angela-tran!

PR #2479, opened 2024-10-28
Feat: analytics for multiple claims on a flow

### LINKED ISSUE #2355: Medicare cardholder: distinguish between Older Adult and Disabled riders

lalver1:
Closes #2355
This PR is a follow-up to #2463. It includes the following changes:
- [ ] Updating the `returned enrollment` analytics event to include the "extra claims" that came back as verified from IdG
- [ ] Storing the error codes returned from IdG in a dictionary since multiple claims may be returned
- [ ] Adding a column to `EnrollmentEvent` that holds the "extra claims"

Note that even though the coverage report for https://github.com/cal-itp/benefits/pull/2463 looked ok, this PR updates a few tests associated with https://github.com/cal-itp/benefits/pull/2463 because they felt a bit incomplete. For example
 https://github.com/cal-itp/benefits/blob/43492c614a504a697be0b3b4a7a2aa4b13009503/tests/pytest/oauth/test_views.py#L252

and

https://github.com/cal-itp/benefits/blob/43492c614a504a697be0b3b4a7a2aa4b13009503/tests/pytest/oauth/test_views.py#L283 
## Reviewing

First run the database migrations since a new field was added to `EnrollmentEvent`. Since the changes involve the `enrollment` application and the claims that are returned from IdG, it's a bit tricky to review this PR. The tests seem to capture the behavior that has changed due to this PR, but you can also hard code a few lines to review the error code changes and `EnrollmentEvent` changes.

To review the error codes part, set `claim_value = "10"` in https://github.com/cal-itp/benefits/blob/43492c614a504a697be0b3b4a7a2aa4b13009503/benefits/oauth/views.py#L136 and confirm that the error codes show up in the `finished sign in` event.

To see a new instance of an `EnrollmentEvent` that has multiple claims you can set, for example, `extra_claims = "claim_1, claim_2"` in https://github.com/cal-itp/benefits/blob/727a4c13c41fea77047dce0005bee66ce25426ed/benefits/enrollment/views.py#L74 to see how multiple claims are stored in the local database.


angela-tran:
@lalver1 Thanks for the thorough notes in the PR description. I'm just now starting to review this PR and am a little confused by this note:

> To see a new instance of an `EnrollmentEvent` in the local database, remove the `EligibleSessionRequired` decorator ...

Shouldn't we be able to just do a successful enrollment locally in order to see a new instance of an `EnrollmentEvent` in the local database? I'm not following why we need to do all these steps to fake it.

lalver1:
Thanks for the review @angela-tran, yep, if you go through a successful enrollment you'll see the new instance of an `EnrollmentEvent`, but I think it only covers the case when IdG returns one claim. In the 'Reviewing' notes I was more thinking about being able to see how multiple claims would be shown on the local database in the new column since currently we don't really have a way to see this path when we go through a successful enrollment, so that's why I was faking it but there may be a cleaner way to test it out.

angela-tran:
@lalver1 Oh ok, I see. So in the steps for faking it, should there also be a step where we make it so the `EnrollmentEvent` will have some `extra_claims`? I see that it's around [line 74](https://github.com/cal-itp/benefits/pull/2479/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R74) in `enrollment.views.index` that we read extra claims.

lalver1:
Exactly! I'm re-reading the `EnrollmentEvent` section of 'Reviewing' and actually all that stuff isn't really necessary @angela-tran. Like you mentioned, you really you only need to overwrite [line 74](https://github.com/cal-itp/benefits/pull/2479/files#diff-b67985b5acabad774a30a1fccbb912e611d4056d38316cd6614ce601f6930335R74) and set, for example, `extra_claims = "claim_1, claim_2"` to see how multiple claims are stored in the local database. Thanks for the comment, I'll update the description.

lalver1:
Update on this PR:

- Updated the function https://github.com/cal-itp/benefits/blob/44bbb761456b55e30ad35cc32819a9d93033193b/benefits/core/session.py#L164 that returns only the "extra claims", now it checks for the `claims_eligibility_claim` instead of assuming it is the first element of the list
- Used existing pattern of calling helper function to simplify this PR's original code in https://github.com/cal-itp/benefits/blob/44bbb761456b55e30ad35cc32819a9d93033193b/benefits/enrollment/analytics.py#L22

Currently addressing the missing coverage due to the new code in `session.py`

benefits/core/session.py
```diff
@@ -157,10 +157,20 @@ def oauth_token(request):
...
 
 
+def oauth_extra_claims(request, eligibility_claim, oauth_claims):
+    """Get the extra oauth claims from the request's session, or None"""
+    if oauth_claims is not None:
+        # since .remove works in-place, make a copy
+        all_claims = oauth_claims[:]
+        return all_claims.remove(eligibility_claim)
+    else:
+        return None
```
thekaveman, line 171: This is starting to feel a little weird. As written, this doesn't feel like it belongs in `session`.

If it is in `session`, it should just use the other helpers here to get the data. E.g. `.flow()` to get the `eligibility_claim` and `.oauth_claims()` to get those.

benefits/enrollment/views.py
```diff
@@ -71,15 +71,25 @@ def index(request):
                 agency = session.agency(request)
                 flow = session.flow(request)
                 expiry = session.enrollment_expiry(request)
+                if session.oauth_extra_claims(request, flow.claims_eligibility_claim, session.oauth_claims(request)):
+                    extra_claims = ", ".join(
+                        session.oauth_extra_claims(request, flow.claims_eligibility_claim, session.oauth_claims(request))
```
thekaveman, line 76: Get the value once and reuse it -- avoid calling the method multiple times.

tests/pytest/enrollment/test_analytics.py
```diff
@@ -1,10 +1,28 @@
...
+
+
+@pytest.mark.django_db
+def test_ReturnedEnrollmentEvent_without_error(app_request, mocker):
+
+    key1 = "enrollment_flows"
+    key2 = "extra_claims"
+    mock_flow = mocker.Mock()
+    mock_flow.system_name = "flow_1"
+    mocker.patch("benefits.core.session.flow", return_value=mock_flow)
```
thekaveman, line 20: Does the existing `mocked_session_flow` fixture not work here?

thekaveman (changes requested):
It's getting closer, but still could use a bit of code cleanup

benefits/core/session.py
thekaveman: To be clear, I think it's fine to leave this in `session` -- but in that case, actually make use of the session! The `request` parameter isn't even used here.

lalver1: I agree, I ended up simplifying `oauth_extra_claims()` and realized that `request` is not used, so it feels like it doesn't belong in `session`. As I wrote it out, I realized that what this function needs (`eligibility_claim` and `oauth_claims`) should be passed into it because they are already available in (typically) the view that calls it. Then, the function is only taking out `eligibility_claim` from `oauth_claims`. 

If this makes sense, I'm struggling a bit to decide where `oauth_extra_claims()` would be declared.

thekaveman: I think leave it in `session` since it is calculated from values in the session. 

But actually calculate it, don't pass anything in. This is similar to e.g. how the reenrollment date is calculated: https://github.com/cal-itp/benefits/blob/main/benefits/core/session.py#L107

tests/pytest/enrollment/test_analytics.py
thekaveman: Not worrying about this for now.

lalver1:
One more update on this PR, we worked with @thekaveman on cleaning up the code some more earlier today and I ran through older adult with the debugger (mostly to make sure types made sense when making calls to `.remove()` and `.join()` for example) one last time just to feel a bit more comfortable with the changes introduced.

thekaveman:
Nice job :tada:

PR #2496, opened 2024-10-31
Feat: `benefits` command line interface

thekaveman:


thekaveman:
We're moving away from heavy reliance on many different templates. One of the main ideas with the Benefits CLI was to help scaffold out e.g. a new agency's templates or a new flow's templates.

Abandoning this PR.

PR #2497, opened 2024-10-31
Fix: show favicon on the Superuser's admin

### LINKED ISSUE #2443: favicon doesn't show up on the Superuser's admin
Maybe need a future ticket: The favicon doesn't show up on the superuser homepage, but it does for the other admin pages: log in, log out, in-person flow pages. 

image

_Originally posted by @machikoyasuda in https://github.com/cal-itp/benefits/pull/2430#discussion_r1792263899_

The favicon is not showing up in the Superuser's admin page, but it seems like we can use the `extrahead` block of the `admin` app's `base.html` template to include it. So we may need to override the `admin/index.html` template and include the icon in the `extrahead` block.

lalver1:
Closes #2443

This PR overrides the `admin/base_site.html` template to show the favicon in the Superuser's admin page by [extending an overridden template](https://docs.djangoproject.com/en/5.1/howto/overriding-templates/#extending-an-overridden-template).

image

## Reviewing

Log in as a Superuser to the admin page, click around the different models and verify that the favicon appears on the pages.

### DIFF
--- a/benefits/templates/admin/base_site.html
+++ b/benefits/templates/admin/base_site.html
@@ -1,0 +1,5 @@
+{% extends "admin/base_site.html" %}
+
+{% block extrahead %}
+    {% include "admin/includes/favicon.html" %}
+{% endblock extrahead %}


lalver1:
Oh and I didn't want to forget that this PR uses [the same favicon as Benefits](https://github.com/cal-itp/benefits/pull/2430#discussion_r1792552494) as @indexing suggested but we could change it to [this one](https://github.com/cal-itp/benefits/issues/2443#issuecomment-2415265289) too in the future if we decide to have a different icon.

thekaveman (approved):
Looks good :+1:

PR #2500, opened 2024-11-04
Feat: Playwright setup

angela-tran:
This PR adds a service called `playwright` to `compose.yml`. Developers can use this container to run Playwright tests.

The base image of this service is a [Microsoft-built image](https://playwright.dev/python/docs/docker) which includes Playwright and its dependencies already installed, thereby avoiding a long installation step for those.

This PR also includes a working test for running through the CST agency card flow on dev-benefits.calitp.org. You'll notice that we do install `pytest-playwright` into the devcontainer ([line 41 of `pyproject.toml`](https://github.com/cal-itp/benefits/pull/2500/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R41)) so that when viewing the test code in the devcontainer, we still have a nice developer experience.

Just to make this super clear: you should write Playwright tests in the devcontainer just as you would with writing application code and unit tests, but **run Playwright tests** from the `playwright` service container. The devcontainer _does not_ have Playwright's dependencies and browsers installed in it.

### Trying it out
(this is all outside of the devcontainer)

Run the tests with:
```bash
docker compose run --rm playwright
```

You can also start a terminal session in the `playwright` service by running:
```bash
docker compose run --rm playwright /bin/bash
```
and then you can run the tests and/or run other Playwright commands using that session.

(The tests will run as headless. This PR does not contain the setup for running in headed mode. That will come in a later PR.)

The console output will hopefully tell you the tests passed, and you can view more output under `tests/playwright/test-results`.

### Small note about the test

I'm still learning the Playwright API, so note that there may be a better way to do what I'm doing with `wait_for_timeout`.

### Helpful docs
- [Running tests using Pytest CLI](https://playwright.dev/python/docs/test-runners) 
   - this is what the helper script is doing
   - note that with the current setup, we can only run in headless mode
- The [Page](https://playwright.dev/python/docs/api/class-page) class
- Locator (element) [assertions](https://playwright.dev/python/docs/api/class-locatorassertion)


angela-tran:
I just realized / re-remembered that in the file structure, we have
```
tests/cypress
tests/pytest
```
so if it feels better to move everything in `playwright` in this PR to instead be under
```
tests/playwright
```
I could do that.

thekaveman:
It would be nice if you could simply run:

```
docker compose run playwright
```

And have that run the tests and output the results. In other words, overriding the `entrypoint/cmd` to call the `run.sh` script you created.

playwright/test_agencycard_flow.py
```diff
@@ -0,0 +1,70 @@
+import os
+
+from playwright.sync_api import Page, expect
+
+
+def test_agency_card_flow(page: Page):
+    page.goto("https://dev-benefits.calitp.org/")
```
thekaveman, line 7: Maybe you meant `page.goto("/")`?

playwright/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+base_url="https://dev-benefits.calitp.org"
```
thekaveman, line 5: Or maybe take this from the command line?

angela-tran:
I've finished cleaning up this PR. I think it'd be good to go over this PR during our next dev workshop so will leave this in draft.

angela-tran:
Closing this in favor of #2554

Issue #2489, opened 2024-10-29
Make agency logos configurable via Admin


thekaveman:
Currently the agency logo files are stored in the repository, e.g. for MST: [`mst-lg.svg`](https://github.com/cal-itp/benefits/blob/main/benefits/static/img/agencies/mst-lg.svg) and [`mst-sm.svg`](https://github.com/cal-itp/benefits/blob/main/benefits/static/img/agencies/mst-sm.svg).

This means to update an agency logo (and to create a new agency), a code change / release is needed.

Let's change this so that agency logo files are properties of the `TransitAgency` model, and therefor configurable via the Admin interface. For storage we can:

* Just implement this on the local filesystem for now, since there aren't that many, they are small files, and they don't change that often
* Think about moving to Azure Storage buckets in a future effort

## Acceptance Criteria

- [ ] Agency logos are used via properties on the `TransitAgency` model in e.g. templates
- [ ] Agency logos are not maintained in the source code

## Additional context

Prompted by #2434 and #2484

PR #2514, opened 2024-11-07
Feat: agency logos configurable via Admin

lalver1:
- **feat: add the pillow library that's required for ImageField**
- **feat: add settings for the storage that manages files**
- **feat: add a property for a small and large agency logo**


benefits/settings.py
```diff
@@ -234,14 +234,21 @@ def RUNTIME_ENVIRONMENT():
...
             "DJANGO_STATICFILES_STORAGE", "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
         )
-    }
+    },
 }
 STATIC_ROOT = os.path.join(BASE_DIR, "static")
 
+#  User-uploaded files
+
+MEDIA_ROOT = os.environ.get("DJANGO_DB_DIR", BASE_DIR) + "/uploads/"
```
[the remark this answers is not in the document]
thekaveman: And circling back to @angela-tran's point from our discussion earlier this week:

It looks like `DATABASE_DIR` is a custom variable we defined, not one of the built-in Django settings. So that means it can be renamed to whatever we want -- to be more descriptive now that we're using it for more than the database.

To see which settings have been changed from the default and which are custom, run:

```console
$ python manage.py diffsettings
ALLOWED_HOSTS = ['localhost']  <--- modified from the Django default
ANALYTICS_KEY = None  ### <--- a custom setting not in the defaults
...
```

See https://docs.djangoproject.com/en/5.1/ref/django-admin/#django-admin-diffsettings

thekaveman (changes requested):
I think this is a good start, and I realize you're focused on getting this field into the model so it is configurable via the admin. This does that, but it also feels incomplete.

Based on my reading of https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.FileField.storage it sounds like there are just a few more steps:

1. Define [`MEDIA_URL`](https://docs.djangoproject.com/en/5.1/ref/settings/#std-setting-MEDIA_URL) as the base public URL of the `MEDIA_ROOT` directory
2. Map this URL and directory in nginx, similar to `/static`
3. Use the convenience [`url`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.fields.files.FieldFile.url) attribute provided by Django on [`ImageField`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.ImageField), to get the absolute path to the image in a template e.g. `{{ object.logo.url }}`

Are you thinking one, two, three... more PRs to implement this? I'd be in favor of just going all the way in this PR, making it a complete feature. We can't even view the image that we upload right now, kind of hard to validate that it is working correctly.

benefits/settings.py
```diff
@@ -234,14 +234,23 @@ def RUNTIME_ENVIRONMENT():
...
-    }
+    },
 }
 STATIC_ROOT = os.path.join(BASE_DIR, "static")
 
+#  User-uploaded files
+
+MEDIA_ROOT = os.path.join(STORAGE_DIR, "uploads/")
+
+MEDIA_URL = "/uploads/"
```
thekaveman, line 252: Can we call these `media` instead of `uploads`? We can be a little obscure that there is a way to upload files into the site :wink: 

Sorry I know this is going to impact a bunch of other changes.

lalver1:
Thanks for the comments @thekaveman and for walking through this earlier today, I think this is ready for a re-review, just 2 things to note:

We thought that adding
```
uploads/
!uploads/agencies/cst-*
```
to `.gitignore` in the root of the repo would make git ignore all files inside `uploads` except the 2 CST logos, but it didn't seem to be working, everything inside `uploads` was being ignored. From what I read, since `uploads/` is ignored, the `agencies` folder is not visible to git so it won't be able to re-include the 2 CST logos. The changes in `.gitignore` in this PR work but I feel like they can be simplified in case you know of a more straightforward way to do this.

Also, [L43 in urls.py](https://github.com/cal-itp/benefits/pull/2514/files#diff-be22b5797b2ebbb4724790795fd7f2576f998677d8f49e3fc3656cbc3dfc0b58R43-R43) is not covered, but it seems ok since we would only be testing `.extend` and `static` which I guess are already tested.

.gitignore
```diff
@@ -7,6 +7,10 @@
 *.tmp
 *.egg-info
 static/
+uploads/*
+!uploads/agencies/
+uploads/agencies/*
+!uploads/agencies/cst-*
```
thekaveman, line 13: I feel like these can be simplified and we should only need 2: ignore the root folder, and then allow anything matching the `agencies/cst-*` pattern?

thekaveman (changes requested):

.gitignore
thekaveman: There are 2 changes that I think will make this work:

1. Remove `*` from the first line: `uploads/` -- this says, ignore the whole directory (with the `*` it means, ignore everything _inside_ the directory, but not the directory itself)
2. Add `.png` to the last line: `!uploads/agencies/cst-*.png` -- this is specifically PNG files in that subdirectory matching the name

Then the other 2 shouldn't be necessary.

benefits/settings.py
```diff
@@ -176,11 +176,11 @@ def RUNTIME_ENVIRONMENT():
 
 WSGI_APPLICATION = "benefits.wsgi.application"
 
-DATABASE_DIR = os.environ.get("DJANGO_DB_DIR", BASE_DIR)
+STORAGE_DIR = os.environ.get("DJANGO_STORAGE_DIR", BASE_DIR)
```
thekaveman, line 179: Please also update the docs and `.env.sample` with this change.

thekaveman (changes requested):

thekaveman:
I think we can say this will close #2489 at this point :grin:

.gitignore
lalver1: Ah, sorry for getting stuck on this, I didn't think this would be as complicated since I was also thinking that the simplified version should work:
```
uploads/
!uploads/agencies/cst-*.png
```
but I'm guessing that there's something about having that intermediate `agencies` folder:

Without any additions to the `gitignore`(`uploads` not ignored, all files inside it untracked)


With our simplified `gitignore` (all ignored, even CST's logos unfortunately):


With the current PR's `gitignore` (all ignored except CST's logos):

lalver1: We decided to go with a separate `.gitignore` file in `uploads` instead of putting everything in the root `.gitignore`

lalver1:
Thanks for looking at this this morning @thekaveman, I think it's ready for a review. 

Also, once this PR is merged and I've updated the logos in `dev`, I'll probably ask Andy in Slack to confirm that the new logos look ok. I've tested on my screen but it'll be good to get confirmation from someone else too.

appcontainer/nginx.conf
```diff
@@ -72,6 +72,13 @@ http {
       add_header Cache-Control public;
     }
 
+    # path for uploaded files
+    location /uploads/ {
```
thekaveman, line 76: Should be `/media/` after the change

thekaveman (changes requested):

uploads/.gitignore
```diff
@@ -0,0 +1,4 @@
+*
+!.gitignore
```
thekaveman, line 2: Sorry to belabor this further, but we shouldn't need to unignore a `.gitignore` file??

I tried it with:

```
*
!*/
!agencies/cst-*.png
```

And everything still works as expected, including changes to the `.gitignore` file:

image

image

appcontainer/nginx.conf
lalver1: Ah yes, sorry about that!

thekaveman (changes requested):

uploads/.gitignore
thekaveman: I might be wrong about this, because the `.gitignore` was already committed?

lalver1: Not at all, thanks for the suggestion. I just tried it and I also see that it works as expected, I'll make this update 👍

thekaveman (approved):
Great work on this @lalver1!

PR #2518, opened 2024-11-12
Chore: update copy on Eligiblity Index for SacRT

### LINKED ISSUE #2505: Copy change request for SacRT: edit transportation type and add context regarding timeline
Make 2 copy changes to SacRT's benefits enrollment home page (page titled `Choose the transit benefit you would like to enroll in`), as per requests from SacRT conveyed to the payments team.

1. The transportation type on the page says that the reduced fare is available on light rail. However, SacRT's launch is going to be on bus only in early 2025, with light rail following later in the year.
2.  They would like to make it clear that the new system with the discount won't be available until early 2025.

## Acceptance Criteria
- [x]  The transportation type is changed to bus so it reads `All SacRT transit benefits reduce fares by 50% for bus service.`
- [ ] Add context to the first paragraph that tells the user this will be available in early 2025. Something like `All SacRT transit benefits will reduce fares by 50% for bus service starting in early 2025.` _See comment section for requested copy._
- [ ] All copy changes that are made are translated into Spanish as well

## Additional context

Task/request number one is the priority, so the transportation type is correct when SacRT announces this to riders (They are planning to announce this to the public on Thursday). The second request can follow after if need be.

Relevant slack thread: https://cal-itp.slack.com/archives/C013W8RUAMU/p1730828646175419

angela-tran:
Closes #2505 

### Screenshots

**English copy**
Screenshot from 2024-11-12 12-44-15

**Spanish copy**
image


lalver1 (approved):
Thanks @angela-tran! Looks good!

PR #2523, opened 2024-11-13
Feat: use pre-loaded WebP background image on homepage

angela-tran:
Closes #2490 

This PR switches out our JPEG background image to a WebP image. It also adds HTML that will trigger preloading of the appropriate image based on media queries. 

Lastly, to handle the case where the user's browser does not support WebP, we will set the background-color of the page to have sufficient contrast with the text on the page. (PR #2524)

See https://github.com/cal-itp/benefits/pull/2523#issuecomment-2477524721 for details on the observed performance gains.

PR #2524, opened 2024-11-14
Feat: fallback background color

angela-tran:
Part of #2490 

This PR will be merged in to #2523.

I think it makes sense for the changes from both PRs to be merged into `main` together, but I also wanted to have separate code reviews.

### Fallback background color

I took the color from the left side of the background image and used that as the `background-color` for the main section of the landing page.

The WebAIM contrast checker shows this contrast ratio [passes WCAG AAA requirements](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=1A1B16).

### Screenshots of background-color in use (background image does not load)

| Desktop | Mobile |
| --- | --- |
| image | localhost_11369_(Moto G Power) |


thekaveman (approved):
:+1:

PR #2554, opened 2024-12-02
Feat: set up Playwright

### LINKED ISSUE #2533: Figure out a way for developers to run Playwright easily
### Acceptance Criteria

- [ ] Developers are able to run Playwright tests from a Docker container
- [ ] Installation of Playwright tooling should not take a long time

angela-tran:
Closes #2533 

This PR sets us up to be able to run and write Playwright tests inside containers.
(_This is basically a re-do of #2500 with cleaner commits and smaller scope, i.e. this just focuses on getting the needed tooling in place._)

### Running
The `playwright` service added in this PR is to be used for running Playwright tests. 

The service uses an [image built by Microsoft](https://playwright.dev/python/docs/docker) that has Playwright + dependencies already installed as the base image.

<details><summary><b>Trying it out</b></summary>
<br/>
Outside of the devcontainer, run this in a terminal:

```bash
docker compose run --rm playwright
```

You can also start a terminal session in the `playwright` service by running:
```bash
docker compose run --rm playwright /bin/bash
```

and then you can run the tests and/or run other Playwright commands using that session.

The test will run as headless. A later PR will contain the setup for running in headed mode.

The console output will hopefully tell you the tests passed, and you can view more output under `tests/playwright/test-results`.

</details>

### Writing

We do install `pytest-playwright` into the devcontainer so that when viewing the test code in the devcontainer, we get syntax-highlighting, code completion, etc.

<details><summary><b>Docs on writing tests</b></summary>

- [Running tests using Pytest CLI](https://playwright.dev/python/docs/test-runners) 
   - this is what the helper script is doing
- The [Page](https://playwright.dev/python/docs/api/class-page) class
- Locator (element) [assertions](https://playwright.dev/python/docs/api/class-locatorassertion)
</details>

### Healthcheck test as a starting point

A test called `test_dev_healthcheck` is included here just so we have something to run. We can decide later if there's any value in keeping it once other tests are added.

### DIFF
--- a/tests/playwright/Dockerfile
+++ b/tests/playwright/Dockerfile
@@ -1,0 +1,11 @@
+# https://playwright.dev/docs/docker
+FROM mcr.microsoft.com/playwright/python:v1.48.0-jammy
+
+WORKDIR /playwright
+
+COPY tests/playwright/requirements.txt requirements.txt
+
+RUN python -m pip install --upgrade pip && \
+   pip install -r requirements.txt
+
+USER pwuser
--- a/compose.yml
+++ b/compose.yml
@@ -41,3 +41,13 @@ services:
       - "8000"
     volumes:
       - ./.devcontainer:/.devcontainer
+
+  playwright:
+    build:
+      context: .
+      dockerfile: tests/playwright/Dockerfile
+    image: benefits_client:playwright
+    volumes:
+      - ./tests/playwright:/playwright
+    # https://code.visualstudio.com/docs/remote/create-dev-container#_use-docker-compose
+    entrypoint: sleep infinity
--- a/compose.yml
+++ b/compose.yml
@@ -49,5 +49,4 @@ services:
     image: benefits_client:playwright
     volumes:
       - ./tests/playwright:/playwright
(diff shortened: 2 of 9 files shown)


tests/playwright/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# https://playwright.dev/docs/docker
+FROM mcr.microsoft.com/playwright/python:v1.48.0-jammy
```
angela-tran, line 2: @thekaveman tried running this setup and ran into an error about needing to use the latest image which is now `mcr.microsoft.com/playwright/python:v1.49.0-jammy`.

Is there a `latest` tag we could use? Or maybe it has something to do with the version of `pytest-playwright` (try to pin it)?

angela-tran: Fixed in https://github.com/cal-itp/benefits/pull/2554/commits/0dd03850329d79c1333053da5e1eebcb22094b5c

machikoyasuda (approved):
image
Tests passing!

machikoyasuda:
Very much appreciate how small and scoped this PR is 💯

Issue #2504, opened 2024-11-05
[TRACKING] Automated smoke testing


angela-tran:
As a part of our release process, we perform smoke tests after deploying to the production app instance (see [the step](https://github.com/cal-itp/benefits/blob/main/.github/ISSUE_TEMPLATE/release.yml?plain=1#L99) in our release issue template: `Smoke Test the app in prod`). 

It would be nice to automate this, and [Playwright](https://playwright.dev) seems like a good tool for doing so.

## Acceptance Criteria

- [ ] Figure out a way for developers to run Playwright easily
- [ ] Write a Playwright test for each test listed in the [Benefits Smoke Testing](https://docs.google.com/document/d/1ADXWW4mfLJwLKVqtVemPNX9yTqYKqIWfCz1DipAU02g/edit?usp=sharing) Google Doc
   - [ ] If it is difficult or not possible to automate certain tests, document the reason
- [ ] Set up GitHub workflow to run Playwright against dev, test, or prod when manually triggered
- [ ] (Optional) Figure out a way for developers to use Playwright tools (trace viewer, inspector, headed mode)

## Related, but separate

- [ ] Set up GitHub workflow to run Playwright against local instance for pull requests
- [ ] Update Benefits Smoke Testing Google Doc to include SacRT flows
- [ ] Move Benefits Smoke Testing Google Doc contents into a page on our documentation site

PR #2558, opened 2024-12-04
Feat: Playwright headed mode

angela-tran:
Part of #2504 

Builds off #2554 

This PR makes it so we can run Playwright in headed mode and use its GUI tools like trace viewer, inspector, debugger, etc.

See documentation: https://github.com/cal-itp/benefits/pull/2558/files#diff-02ef96f880a015712e37e12d3d02053fe947546b30eb532f0b2361e7da57cc45R37


machikoyasuda:
Nice to view and use the docs live preview: https://benefits-2558--cal-itp-previews.netlify.app/tests/automated-tests/#playwright

machikoyasuda (approved):
Got it running on MacOS. Tested Chrome, Firefox, Webkit browsers.

PR #2563, opened 2024-12-04
Chore: add back values for CST local fixture

angela-tran:
This is a follow-up to #2471 to add back values for the local CST TransitAgency fixture's `info_url`, `phone`, `logo_large`, and `logo_small` fields.

In that PR, those values were removed from the local fixtures because the idea at one point was to use them as defaults for those fields. 

We decided to default to the empty string instead after discussion in https://github.com/cal-itp/benefits/pull/2471#discussion_r1858988519 and https://github.com/cal-itp/benefits/pull/2471#discussion_r1859137740.

Currently in `main`, if you run `./bin/reset_db.sh` and run the app, you will see missing values:

**Missing logo**
Screenshot from 2024-12-04 15-32-55

**Missing `info_url` and `phone`**
Screenshot from 2024-12-04 15-33-13

### DIFF
--- a/benefits/core/migrations/local_fixtures.json
+++ b/benefits/core/migrations/local_fixtures.json
@@ -53,6 +53,8 @@
       "slug": "cst",
       "short_name": "CST (local)",
       "long_name": "California State Transit (local)",
+      "info_url": "https://www.agency-website.com",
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
Thanks!

PR #2570, opened 2024-12-05
Feat/2539 recaptcha eligibility

machikoyasuda:


benefits/static/css/styles.css
```diff
@@ -558,27 +556,12 @@ footer .footer-links li a.footer-link:visited {
...
-  }
 }
 
 .radio-label {
   cursor: pointer;
 }
 
-.radio-input-group {
-  gap: var(--radio-input-gap);
-}
```
machikoyasuda, line 576: This is a refactor that I made by:

- Bumping up the 15px gap on desktop to 16px,
- Allows us to get to place where we are only using 8pt values for both desktop (16px) and mobile (24px)
- Which then allows us to use `gap-4` (4 is 24px) and `gap-md-3` (3 is 16px) in the template code.

Now all the radio button CSS styles are around who the element looks, and all the positioning (margin, padding) are done with classes on the page the elements are on.

benefits/core/templates/core/widgets/flow-radio-select.html
```diff
@@ -2,9 +2,9 @@
 
 {% with id=widget.attrs.id %}
 
-  <label for="{{ id }}" class="pb-4 h2">{% translate "Which transit benefit would you like to enroll in?" %}</label>
+  <label for="{{ id }}" class="h2">{% translate "Which transit benefit would you like to enroll in?" %}</label>
 
-  <div {% if id %}id="{{ id }}"{% endif %} class="ps-3 ps-md-0 ps-lg-0 offset-md-1 offset-lg-1 radio-container">
```
machikoyasuda, line 7: I didn't see the `radio-container` class name used in any CSS or JS, so I deleted it. I can't remember why I made this class name, perhaps purely for naming/reading purposes(!?!)

benefits/core/templates/core/widgets/flow-radio-select.html
```diff
@@ -13,7 +13,7 @@
           {% endif %}
 
           {% for option in options %}
-            <div class="radio-input-group d-flex">{% include option.template_name with widget=option %}</div>
+            <div class="radio-input-group gap-4 gap-md-3 d-flex pt-4">{% include option.template_name with widget=option %}</div>
```
angela-tran, line 16: I don't see `radio-input-group` in any CSS either. The way I'm looking for it is in the devtools Inspector:

Screenshot from 2024-12-05 17-54-37

benefits/static/css/styles.css
angela-tran: Bullet points make sense :+1: 

Not sure I'm following this part:
> Now all the radio button CSS styles are around who the element looks

benefits/core/templates/core/widgets/flow-radio-select.html
angela-tran: I suggest we remove `radio-input-group` too

angela-tran (changes requested):
Glad to see the spacing implementation is simpler. Noticed a class that can probably be removed?

benefits/static/css/styles.css
machikoyasuda: _how_ the element _looks_, rather than how the elements as a whole are positioned onto the page. sorry typo!

angela-tran: Ah, got it

angela-tran (approved):

PR #2578, opened 2024-12-06
CI: update littlepay switch syntax

thekaveman:
Should not be merged until after https://github.com/cal-itp/littlepay/pull/72

But _needs_ to be merged after that PR for this check to continue to work.

### DIFF
--- a/.github/workflows/check-api.yml
+++ b/.github/workflows/check-api.yml
@@ -32,8 +32,7 @@ jobs:
 
       - name: Run littlepay to get access token
         run: |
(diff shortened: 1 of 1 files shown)


angela-tran:
@thekaveman Should we release a new version of `calitp-littlepay`?

thekaveman:
@angela-tran we probably should!

But this CI check just _checks out the repo directly @ `main`_ so we actually want this update whether or not we release!

angela-tran:
Oh I see, got it!

angela-tran (approved):

PR #2579, opened 2024-12-06
reCAPTCHA: Enrollment Success, Logged Out

### LINKED ISSUE #2544: reCAPTCHA: Enrollment success - Widen text + Logged out
Image

This is not a 6-col wide design and has text starting mid-column. I suggest a 9-col width on Desktop like this and change the text from centered to left align:

Image

machikoyasuda:
closes #2544 

### Enrollment Success
image

This is now the only page in the flow that _isn't_ `col-lg-6`. Though, as a reminder, the home page and the Help page aren't either. The Help page will remain `col-lg-8`. This page doesn't by any columns on Figma, both the headline text and the image/text combo. On the app, we are using a `col-lg-9` as our closest approximate.

### Logged out
image

A `col-lg-6` page with the now standard, 72px between header and headline, and 24px between icon and graf.

### DIFF
--- a/benefits/core/templates/core/logged-out.html
+++ b/benefits/core/templates/core/logged-out.html
@@ -7,12 +7,10 @@
 
 {% block main-content %}
   <div class="container">
-    <div class="row justify-content-lg-center">
+    <div class="row justify-content-center">
       <div class="col-md-6">
-        <h1 class="h2 text-center">
-          <span class="d-block pb-lg-8 pb-5">{% include "core/includes/icon.html" with name="happybus" %}</span>
-          {% translate "You have successfully logged out. Thank you for using Cal-ITP Benefits!" %}
-        </h1>
+        <div class="py-4 mt-4 text-center">{% include "core/includes/icon.html" with name="happybus" %}</div>
+        <h1 class="h2 pt-0">{% translate "You have successfully logged out. Thank you for using Cal-ITP Benefits!" %}</h1>
       </div>
     </div>
   </div>
--- a/benefits/core/templates/core/logged-out.html
+++ b/benefits/core/templates/core/logged-out.html
@@ -9,7 +9,7 @@
   <div class="container">
     <div class="row justify-content-center">
       <div class="col-md-6">
-        <div class="py-4 mt-4 text-center">{% include "core/includes/icon.html" with name="happybus" %}</div>
+        <div class="py-4 mt-5 text-center">{% include "core/includes/icon.html" with name="happybus" %}</div>
         <h1 class="h2 pt-0">{% translate "You have successfully logged out. Thank you for using Cal-ITP Benefits!" %}</h1>
       </div>
     </div>
--- a/benefits/enrollment/templates/enrollment/success.html
+++ b/benefits/enrollment/templates/enrollment/success.html
@@ -24,7 +24,7 @@
 {% endblock headline %}
 
 {% block inner-content %}
-  <div class="col-12 col-sm-12 col-lg-9">
+  <div class="col-12 col-lg-9">
     <div class="row flex-column-reverse flex-lg-row">
       <div class="col-12 col-lg-7 pe-1">
         {# djlint:off #}
@@ -58,7 +58,7 @@
 {% block call-to-action %}
   {% if authentication and authentication.logged_in and authentication.sign_out_button_template %}
     <div class="row d-flex justify-content-start justify-content-lg-center">
-      <div class="col-12 col-lg-8 pt-lg-5 mt-lg-0 pt-4 mt-2">
+      <div class="col-12 col-lg-9 pt-lg-5 mt-lg-0 pt-4 mt-2">
         <p>
           {% translate "If you are on a public or shared computer, don’t forget to sign out of " %}
           {% include authentication.sign_out_button_template %}
--- a/benefits/enrollment/templates/enrollment/success.html
+++ b/benefits/enrollment/templates/enrollment/success.html
@@ -12,7 +12,7 @@
 {% endblock page-title %}
 
 {% block headline %}
-  <div class="col-lg-7">
+  <div class="col-lg-9">
     <h1 class="pb-lg-4 mb-lg-3 pb-4">
       {% block headline-message %}
         {% blocktranslate trimmed %}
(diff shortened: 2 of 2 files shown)


benefits/core/templates/core/logged-out.html
```diff
@@ -7,12 +7,10 @@
 
 {% block main-content %}
   <div class="container">
-    <div class="row justify-content-lg-center">
+    <div class="row justify-content-center">
       <div class="col-md-6">
-        <h1 class="h2 text-center">
-          <span class="d-block pb-lg-8 pb-5">{% include "core/includes/icon.html" with name="happybus" %}</span>
```
machikoyasuda, line 13: Putting a `span` with a `d-block` to wrap an `img`...... so weird!

benefits/enrollment/templates/enrollment/success.html
```diff
@@ -52,18 +52,13 @@ <h2 class="h3 mt-lg-3 mb-1">{% translate "Your benefit will expire on" %} {{ enr
...
+    {% if authentication and authentication.logged_in and authentication.sign_out_button_template %}
+      <div class="row justify-content-start justify-content-lg-center">
+        <p class="pt-lg-5 mt-lg-0 pt-4 mt-2">
           {% translate "If you are on a public or shared computer, don’t forget to sign out of " %}
           {% include authentication.sign_out_button_template %}
         </p>
       </div>
-    </div>
-  {% endif %}
-{% endblock call-to-action %}
```
machikoyasuda, line 69: The current file has the `If you are on a public or shared computer..` line in the `call-to-action` area. But I moved it to the parent block container, `inner-content` instead. In the future, this will allow us to be more specific with what the `call-to-action` block has in it (if for example we ONLY use it for the majority case where there's just 1 big button).

machikoyasuda:
@angela-tran This PR is ready for review. (Goes a little out of order from what I had listed! It's a small PR though, compared with the others).

angela-tran:
@machikoyasuda I see that this PR contains some changes that do make the code cleaner, namely the `<span>` cleanup. But I'm a little confused on how these changes are related to the re-designs for reCAPTCHA. I looked over the [notes from the design review on 2024-11-25](https://docs.google.com/document/d/1vActZALSbRwO0HxDvm-YQK37p22SPbL4XuMdcRx-BI8/edit?tab=t.0#heading=h.5jqgllcfo0tg) and don't see anything about making the content on this page wider.

I'm not sure how to proceed with my code review of this.

machikoyasuda:
@angela-tran Hrm.. I'm just going by what alignment changes were made on Figma: un-centering the H1s, making sure the H1 is 72px from the top, 24px between grafs. The most visibly noticeable change is the un-centering the H1s.

benefits/core/templates/core/logged-out.html
```diff
@@ -7,12 +7,10 @@
...
-          {% translate "You have successfully logged out. Thank you for using Cal-ITP Benefits!" %}
-        </h1>
+        <div class="py-4 mt-4 text-center">{% include "core/includes/icon.html" with name="happybus" %}</div>
```
angela-tran, line 12: I suggest `mt-5` here instead of `mt-4` to get 72px of spacing. This currently has us at 48px of spacing.

| Before (70px from `h1` rule) | Figma | This branch (48px) | Using `mt-5` |
| --- | --- | --- | --- |
| image | image | Screenshot from 2024-12-10 13-11-46 | Screenshot from 2024-12-10 13-12-00 |

angela-tran:
> @angela-tran Hrm.. I'm just going by what alignment changes were made on Figma: un-centering the H1s, making sure the H1 is 72px from the top, 24px between grafs. The most visibly noticeable change is the un-centering the H1s.

Ah yeah, the headline becoming un-centered and the 24px spacing change are all from the reCAPTCHA design changes :+1: I was getting caught up in the Figma design not following the grid and calling for the content to be wider for some reason, but I guess it's fine.

Thanks for clarifying and sending me the link to the Figma diff -- it was really helpful!

benefits/enrollment/templates/enrollment/success.html
angela-tran: Ok so I think the idea here with moving it up to `inner-content` was so it can be inside the `<div>` defined on line 27 (`<div class="col-12 col-sm-12 col-lg-9">`) and therefore not have to duplicate the `col-` classes. 

Ultimately the spacing / alignment looks correct :white_check_mark: :

| Before | This branch |
| --- | --- |
| image | image |


Could you explain more what you mean by:
> In the future, this will allow us to be more specific with what the `call-to-action` block has in it (if for example we ONLY use it for the majority case where there's just 1 big button). 

? I'm not fully following and am curious what it means to be more specific with the `call-to-action` block.

angela-tran (changes requested):
Just a small change requested and a question

benefits/enrollment/templates/enrollment/success.html
machikoyasuda: Sorry that line was confusing 😆 , as I'm wondering to myself what it means 😆 
I'm gonna retract that statement entirely. https://github.com/cal-itp/benefits/pull/2583#discussion_r1878742909 achieved what I meant in that confusing sentence. 

Instead I'm going to say, although we _can_ put the `If you are on a public computer` sentence in _either_ `inner-content` or `call-to-action`, I think we should keep it in `inner-content` b/c it's not really the same design/intent as the `call-to-action` block, which is really meant to be the container of the big CTA buttons. So it's more of a semantics decision rather than a code change. 

Just in case in the future, there's some other new re-design to the `call-to-action` button / parent area, and we want to apply it to all the big buttons - it would be easier to do if we're only using this block *only* had the big buttons.

benefits/core/templates/core/logged-out.html
machikoyasuda: Fixed: https://github.com/cal-itp/benefits/pull/2579/commits/279fc00f459b02effa6bb0316c937713d2948564

angela-tran (approved):

benefits/core/templates/core/logged-out.html
thekaveman: Yeah, weird also that the image/icon was _inside_ the `h1`. This makes way more sense :+1:

PR #2584, opened 2024-12-10
reCAPTCHA: Help page

### LINKED ISSUE #2545: reCAPTCHA: Help, Error pages - Button width
- Help page: Widen the button at the bottom
- Error page: Widen button at the bottom

machikoyasuda:
closes #2545 (Part 1) / Error pages to come in separate PR

## What this PR does
- Changes text alignment to `col-lg-8`
- Changes button alignment
- Squashes some of the divs into 1 div

### DIFF
--- a/benefits/core/templates/core/help.html
+++ b/benefits/core/templates/core/help.html
@@ -9,7 +9,7 @@
 {% block main-content %}
   <div class="container">
     <div class="row justify-content-center">
-      <div class="col-12 col-lg-8">
+      <div class="col-lg-8">
         <h1 class="text-center">{% translate "Help" %}</h1>
         <h2 class="h2-sm pt-4 pt-lg-8" id="what-is-cal-itp">{% translate "What is Cal-ITP Benefits?" %}</h2>
         <p class="pt-2 pt-lg-4">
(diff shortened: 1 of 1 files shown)


benefits/core/templates/core/help.html
```diff
@@ -152,8 +143,8 @@ <h2 class="h2-sm pt-4 pt-lg-8" id="questions">{% translate "Questions?" %}</h2>
       </div>
     </div>
 
-    <div class="row pt-4 pt-lg-8">
-      <div class="col-lg-2 offset-lg-10 col-8 offset-2">
+    <div class="row justify-content-center pt-4 pt-lg-8">
+      <div class="col-12 col-lg-6">
         {% translate "Go Back" as button_text %}
         {% include "core/includes/button--origin.html" with button_text=button_text %}
       </div>
```
machikoyasuda, line 150: I tried using the `call-to-action` or `call-to-action-button` blocks instead, but I can't b/c I'm overriding the parent `main-content` block entirely. I'm okay with it though, b/c this page is a different alignment and design from the rest of the pages... so it feels appropriate to override the `base` layout.

thekaveman (approved):
Thanks for the divitis cleanup! (And for sharing that fantastic term :grin:)

Just noting that this is now the only page where the heading is center aligned, which is fine.

PR #2585, opened 2024-12-11
reCAPTCHA: Error pages

### LINKED ISSUE #2545: reCAPTCHA: Help, Error pages - Button width

machikoyasuda:
closes #2545 

## What this PR does
- Realign all the error pages
- Refactors error pages so they all inherit from 1 base file, now called `error-base.html` (previously was `error.html`). Now both Admin and Benefits have an `error-base`. The `base` naming is used like this now:
image, to let the developer know that this is a template that inherits from `base.html`, and other templates then inherit it.
- `error-base.html` now makes it so that there is one place to make these changes, making it easier to review also. When making a new error page that is based on this template/design, the developer should not have to write any padding/margin/centering/alignment classes or code. They should only have to write copy (headline-text), create a new or use an existing `include`s for an icon, add any `<p>` of text, and use an existing button, or create a new one (the button might have existing CSS classes, like `login`). But there shouldn't be a need for any new `row`s or `justify-content-centers` or `col-lg-6`. That is the responsibility of the `error-base` template.
 
## How `error-base` works:

```html
{% extends "core/base.html" %}  /* inherits from base, so any blocks in base can be used - even if not specified here */
{% load i18n %}

{% block page-title %}
  {% block title %}
  {% endblock title %}
{% endblock page-title %}

/* inherits from base, so you could put a block for nav-buttons here to have a nice area to put the log out button */

{% block inner-content %}   /* inherits from base, `inner-content` has a `<div class="row justify-content-center`"> outside of it, so we can omit it. you have 2 options here: override inner-content entirely, OR, onlly override the children: icon, headline-text, paragraphs. all of our error pages now do this, but in the future if there is some weird new one-off error design where these elements are in a different order, we can override inner-content directly*/
  <div class="col-lg-6">
    <div class="mt-5 pt-4 text-center">
      {% block icon %} /* put the icon includes in here directly, no wrapper div needed */
      {% endblock icon %} /* Default: sadbus icon. If sadbus, then don't need to include this block at all */
    </div>
    <h1 class="h2 py-4">
      {% block headline-text %} /* put the text in here directly, no wrapper div needed */
      {% endblock headline-text %}
    </h1>
    {% block paragraphs %} /* put paragraph elements and other includes (like agency-links) in here directly, no wrapper div needed */
    {% endblock paragraphs %}
  </div>
{% endblock inner-content %}

{% block call-to-action-button %} /* the logic here is, "show the button to home page by default if a template doesn't specify one. or override the button by putting in your own block. */
  {% include "core/includes/button--origin.html" %}
{% endblock call-to-action-button %}
```

## Error pages

1. 500
2. 404
3. 400
4. 200 user error
5. OAuth system error
6. Enrollment system error
7. Enrollment retry
8. Re-enrollment CalFresh
9. Eligibility unverified

## Side note outside of the scope of this PR/Milestone
- The transit agency name is supposed to be bold on the `agency-links`, but it's not. Not part of this PR.
- I noticed an inconsistency in button text casing. There's `Start Here` and `Return home` and `Try again`. Not part of this PR.
- Some files are set to 4 space tabs, while most are 2: `reenrollment-error--calfresh` and `start--medicare` at least. Not sure why. Not part of this PR.
- Testing these pages is kinda cumbersome. Perhaps we could have some sort of helper file / method that makes it easier to access all of them.


benefits/static/css/variables.css
machikoyasuda, line 1: I didn't change or touch this file at all.... the linter did this!

benefits/templates/error-base.html
```diff
@@ -0,0 +1,26 @@
+{% extends "core/base.html" %}
+{% load i18n %}
```
machikoyasuda, line 2: Maybe this is unnecessary???

benefits/templates/error-base.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block page-title %}
+  {% block title %}
+  {% endblock title %}
+{% endblock page-title %}
```
machikoyasuda, line 7: Is this necessary? Test.

thekaveman: I wouldn't think so. I would think the inheriting error page e.g. `404.html` or whatever could just use the
`{% block page-title %}` directly, right?

benefits/templates/error-base.html
thekaveman: Seems like it, there are no `{% translate %}` tags on this base page.

benefits/templates/200-user-error.html
```diff
@@ -1,21 +1,25 @@
...
+{% extends "error-base.html" %}
 {% load i18n %}
 
 {% block title %}
   {% translate "Start over" %}
 {% endblock title %}
 
-{% block headline %}
+{% block icon %}
+  {% include "core/includes/icon.html" with name="sadbus" %}
```
thekaveman, line 9: I love this refactor to have all error pages inherit from `error-base.html` :heart_eyes: 

I think this icon include could be moved there as the "default" icon. Which means it wouldn't have to be repeated on all these regular error pages.

That doesn't prevent us from overriding it on error pages that do need a custom icon.

benefits/templates/error-base.html
```diff
@@ -0,0 +1,26 @@
...
+
+{% block inner-content %}
+  <div class="col-lg-6">
+    <div class="mt-5 pt-4 text-center">
+      {% block icon %}
+      {% endblock icon %}
```
thekaveman, line 13: See [comment above](#) about a "default" icon here.

thekaveman (changes requested):

benefits/templates/200-user-error.html
machikoyasuda: Ooooh yes!

machikoyasuda: Default sad bus.

benefits/templates/error-base.html
machikoyasuda: So this was not necessary, but I had to then go back to each `40X` error page and change `page-title` to `title.

This anti-pattern is also exists on `landing`/`agency-index` (which btw is confusing naming. `landing-base`/`agency-landing` might be less confusing.

benefits/templates/error-base.html
machikoyasuda: Done

benefits/templates/200-user-error.html
machikoyasuda: Done

machikoyasuda:
I noticed that the transit agency name bolding on agency links is a regression on `main`. See more here: https://github.com/cal-itp/benefits/issues/2590 So I made a bug ticket and I can work on it after reCAPTCHA is done, in this sprint or immediately in slush.

thekaveman (approved):
Really nice clean ups :+1:

machikoyasuda:
Addressing some side notes:
- @thekaveman figured out the spacing issue. Ticket created #2591 
- I figured out the agency links agency not being bolded issue. Related to CA State Web Template. Ticket created #2590

benefits/enrollment/templates/enrollment/retry.html
```diff
@@ -1,37 +1,30 @@
...
+{% endblock icon %}
 
-    <div class="row justify-content-center">
-      <div class="col-lg-8 pt-4">
-        <p>{% translate "Please check the details on your card and try again, or contact your transit agency for help." %}</p>
-      </div>
-    </div>
+{% block headline-text %}
+  {% translate "The card information may not have been entered correctly." %}
+{% endblock headline-text %}
```
angela-tran, line 20: It looks like this is not being used? but we want it to be.

Noticed this while on the `feat/2541-recaptcha-add-link-hide-flag` branch:

The headline is incorrect, and the button is supposed to say "Try again."
image

Currently on prod:
image

Figma:
image

https://www.figma.com/design/SeSd3LaLd6WkbEYhmtKpO3/Benefits-(Full-Application)?node-id=15064-11562&m=dev

angela-tran: Created https://github.com/cal-itp/benefits/issues/2596

PR #2555, opened 2024-12-02
Fix: Use same Bootstrap version for both Admin and Benefits

machikoyasuda:
closes #2551

The #2491 PR made a regression on Admin, by changing this line in the settings CSP: https://github.com/cal-itp/benefits/pull/2526/files#diff-d71405aa44f1e899a9cd920e5187aa1d02affdbd0da472eca95f814b0a8c1210R357, which then made the Bootstrap CSS version used on Admin inaccessible.

This PR:
- Moves the Bootstrap includes from Admin into Core
- Upgrades the Bootstrap includes version to 5.3.3
- Uses that includes in both Admin and Benefits apps

QA:
- QA the app flow on Benefits
- QA a in-person verification flow on Admin
- QA logging in and out of Admin

---

I've now realized that #2491 introduced a few small changes (mostly deleting style declarations no longer necessary, that were overriding CA State Web Template) to the Button CSS in the Benefits app stylesheet, but not Admin's Button CSS - which had copied Benefits' CSS. This PR is not attempting to bring in those changes from Benefits CSS to Admin CSS. It's still a little weird that the two files' Button CSS are now diverging, and I think that can be done in a future clean up ticket.

Issue #2493, opened 2024-10-30
Web perf: Audit and improve blocking resources


machikoyasuda:
This ticket is part of a group of web performance tickets tracked here https://github.com/cal-itp/benefits/issues/2487.

## Acceptance Criteria

- [ ] Read + follow recommendations from https://www.webpagetest.org/result/241029_BiDcSX_ARH/2/experiments/ and https://pagespeed.web.dev/analysis/https-benefits-calitp-org/rqzk6itbxm?form_factor=mobile - Using `async` `defer` `pre-connect` for our various JS/CSS third-party resources.
- [ ] Consider compressing/minifying files from server-side.

## Additional context

- More Django-specific context/info: https://openfolder.sh/django-faster-speed-tutorial

angela-tran:
After looking over the recommendations from the WebPageTest report linked above and then looking through our code, I think there are a few efforts here that are independent of one another. Each effort alone requires a decent amount of work, so I think it'd be best to split them out into separate tickets.

<details><summary><b><i>Small note about the WebPageTest report</i></b></summary>

The report is from before #2526  and #2523, so some of the recs in the report are a little out-of-date. I ran [another test today](https://www.webpagetest.org/result/241212_BiDcV2_CN6/2/experiments/), but note that we haven't released those changes to prod yet, so the report is the same. 

I can use Lighthouse reports to analyze current performance and improvements, and the point still stands that we need to reduce the number of resources that block page rendering. 

</details>

Here's what I think should be split out:

## Eliminating / improving block resources

Problems:

_From WebPageTest report on benefits.calitp.org_
![Image](https://github.com/user-attachments/assets/8441d225-7315-4f21-9dbd-d33f77d5da66)

![Image](https://github.com/user-attachments/assets/ea329e53-7ad7-4549-951f-75212afda5da)


_From Lighthouse report on my localhost_
<img src="https://github.com/user-attachments/assets/7e817425-d0e1-4ccd-af9a-fcf64559be59" width=400>

### Improvements for loading Javascript

- **Load external Javascript in a non-blocking manner**
   - i.e. with either `async` or `defer`, depending on whether we need it to load in a determined order or not
   - this requires determining if there are inline scripts that depend on external Javascript (e.g. jQuery) and ensuring they do not break
- **Identify inline Javascript that should be moved into their own files, and load them in a non-blocking manner**
[...]

PR #2586, opened 2024-12-11
reCAPTCHA: Elig Index, Elig Confirm - Add reCAPTCHA text, hide flag

### LINKED ISSUE #2541: reCAPTCHA: Add reCAPTCHA text and hide button
- Add reCAPTCHA text to Eligibility page and Agency Card form page. Spanish translation is ready. 
- Text links should have an appropriately-sized external link icon. Ensure no regressions on the other link icons.
- And hide reCAPTCHA icon via CSS (https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed).
Image

machikoyasuda:
closes #2541 

## What this PR does
This ticket _actually_ adds the reCAPTCHA text to 2 forms (Eligibility Confirm, Eligibility Index) and hides the badge.

- Creates a `recaptcha-text` includes with the English and Spanish text
- Includes `recaptcha-text` in `form.html` within the `if request.recaptcha` block, so that it shows up automatically on where recaptcha is used
- Adds CSS to: hide the badge via visibility hidden, updates `form-text` class to fix the form field helper letter-spacing, and use that same class for the recaptcha text, adds styling so that the external link icon is smaller for the smaller text
- Minor div/spacing refactors on form and template pages.

## Testing notes
- Use https://github.com/cal-itp/benefits/pull/1071 instructions to test in Dev. The reCAPTCHA test environment key/secret are in LastPass, under Benefits reCAPTCHA configuration.
- Post-merge test plan: When this PR is merged into `dev`, we can first test that the reCAPTCHA message does *not* show up. It should not show up b/c reCAPTCHA is not enabled. Once we _do_ turn reCAPTCHA on, then we can test that the reCAPTCHA text turns on and the feature is enabled. 

## Screenshots

| Desktop     | Mobile     |
| --------  | -------- |
|Screenshot 2024-12-16 at 16-48-34 Choose benefit option Cal-ITP Benefits|Screenshot 2024-12-16 at 16-49-02 Choose benefit option Cal-ITP Benefits|
|Screenshot 2024-12-16 at 16-48-44 Agency card information Cal-ITP Benefits|Screenshot 2024-12-16 at 16-48-54 Agency card information Cal-ITP Benefits|


benefits/eligibility/templates/eligibility/confirm.html
```diff
@@ -26,7 +26,8 @@ <h1>{{ form.headline }}</h1>
...
 
 {% block inner-content %}
-  <div class="container">
-    <div class="row">{% include "core/includes/form.html" with form=form %}</div>
-  </div>
+  {% include "core/includes/form.html" with form=form %}
 {% endblock inner-content %}
+
+{% block call-to-action %}
+{% endblock call-to-action %}
```
machikoyasuda, line 33: Why is this here?

It's here to ensure that more spacing is not created. If this is _not_ here, then the `base.html` will put this HTML on the page, which includes `pt-8` -- which we do *not* want: 

```html
            <div class="row d-flex justify-content-center pt-8">
              <div class="col-12 col-lg-6">
                {% endblock call-to-action-button %}
              </div>
            </div>
```

There's an alternate way to achieve this. I could have put this code on both `eligibility/confirm.html` and `eligibility/index.html`:

```html
                    {% include "core/includes/recaptcha-text.html" %}
                {% endblock call-to-action-button %}
```

But it felt semantically incorrect to put _text_ into code that was meant to be for a button only. And I didn't want to come up with a more generic name to rename `call-to-action-button` or rename all those on the other templates... So I went with this method instead.

benefits/static/css/variables.css
```diff
@@ -11,8 +11,9 @@
   --medium-font-weight: 500;
   --bold-font-weight: 700;
   --bs-body-font-size: calc(18rem / 16);
-  --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
-    "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  --bs-font-sans-serif: Roboto, system-ui, -apple-system, "Segoe UI",
+    sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol",
+    "Noto Color Emoji";
```
machikoyasuda, line 16: Wait why does `pre-commit` do this to me T_T

benefits/static/css/styles.css
```diff
@@ -556,7 +564,7 @@ footer .footer-links li a.footer-link:visited {
 .form-text {
   font-size: var(--font-size-14px);
   line-height: var(--form-field-text-line-height);
-  letter-spacing: var(--letter-spacing-5);
+  letter-spacing: calc(var(--font-size-14px) * var(--letter-spacing-5));
```
machikoyasuda, line 567: image
This letter-spacing update affects not just reCAPTCHA text, but the form field helper text as well. It was not updated on Production.

The way this is calculated: In Figma, click on the text. In the Dev resources sidebar, ensure Properties is enabled (instead of Code). Look for the letter spacing percentage:
image

Multiply the percentage with the font size.

benefits/static/css/styles.css
```diff
@@ -517,26 +533,18 @@ footer .footer-links li a.footer-link:visited {
 }
 
 /* Forms: Text Inputs */
-/* Form Container: Must use .form-container parent to use these styles */
+/* Form Field Container: Must use .form-container parent to use these styles */
```
machikoyasuda, line 536: Renamed this for clarity. Each `.form-field-container` wraps one of these as a unit
image

`.form-container` made it seem like it was wrapping the entire form, which is kinda true for Elig Index but not on Elig Confirm.

benefits/core/templates/core/includes/form.html
```diff
@@ -3,41 +3,39 @@
...
-              {% endif %}
-              {# djlint:on #}
-
-              {{ field }}
-
-              {% if field.help_text %}<small class="d-block mt-2 pt-1 form-text text-body">{{ field.help_text }}</small>{% endif %}
-            </div>
-          </div>
-        {% endfor %}
-      </div>
```
machikoyasuda, line 27: All of these `row justify-content-center`, `row`, `col-12` are unnecessary.

benefits/core/templates/core/includes/form.html
```diff
@@ -3,41 +3,39 @@
...
+          {{ field }}
+
+          {% if field.help_text %}<small class="d-block mt-2 pt-1 form-text text-body">{{ field.help_text }}</small>{% endif %}
+        </div>
+      {% endfor %}
     </div>
 
     {% if form.submit_value %}
-      <div class="row d-flex justify-content-center pt-8">
-        <div class="col-lg-6">
```
machikoyasuda, line 32: Now longer needed.

benefits/static/css/styles.css
```diff
@@ -285,6 +297,10 @@ footer {
   margin-top: 64px;
 }
 
+.eligibility-confirm footer {
+  margin-top: 24px;
+}
+
```
machikoyasuda, line 303: image

The Elig Index page's reCAPTCHA text has 64px padding both above and below the text, but the Elig Confirm's has 64px above and 24px below. So this class allows us to get that. This is the only page on the whole app, aside from the Index page, to *not* have that 64px buffer above the footer.

benefits/static/css/styles.css
```diff
@@ -517,26 +533,18 @@ footer .footer-links li a.footer-link:visited {
 }
 
 /* Forms: Text Inputs */
-/* Form Container: Must use .form-container parent to use these styles */
+/* Form Field Container: Must use .form-field-container parent to use these styles */
```
machikoyasuda, line 536: Renamed this b/c this div is a container around one form field, rather than the entire form.

benefits/static/css/styles.css
```diff
@@ -517,26 +533,18 @@ footer .footer-links li a.footer-link:visited {
...
 
-:root {
-  --form-input-gap: calc(24rem / 16);
-}
-
-.form-container .form-group:not(:first-child) {
-  padding-top: var(--form-input-gap);
-}
```
machikoyasuda, line 528: Use the `d-flex flex-column gap-4` CSS class to achieve this instead.

machikoyasuda:
This is ready for review.

benefits/eligibility/templates/eligibility/confirm.html
thekaveman: This sounds reasonable :+1: I think a comment above the empty block in this file just indicating that it is there for a reason (maybe a link to this thread here?) would be helpful for the future.

benefits/static/css/styles.css
```diff
@@ -14,6 +14,11 @@ html {
   scroll-padding-top: 101px; /* Header Height (80px) + Skip Nav Height (21px) = 101px */
 }
 
+/* reCAPTCHA: Hide the badge */
+.grecaptcha-badge {
+  visibility: hidden;
```
thekaveman, line 19: Just a question: why not `display: none;` instead?

benefits/static/css/styles.css
```diff
@@ -125,6 +132,11 @@ a[target="_blank"]::after {
   vertical-align: text-bottom;
 }
 
+.form-text a[target="_blank"]::after {
+  width: 18px;
+  height: 18px;
```
thekaveman, line 137: What are these rules doing? Do we want to hardcode this pixel value?

thekaveman (commented):
Looking really good. Just a couple of questions.

machikoyasuda:
How to test reCAPTCHA locally: https://github.com/cal-itp/benefits/pull/1071

benefits/static/css/styles.css
machikoyasuda: Honestly, I am not sure. It's what the Google reCAPTCHA docs FAQ recommends: https://developers.google.com/recaptcha/docs/faq#id-like-to-hide-the-recaptcha-badge.-what-is-allowed But I would _prefer_ `display: none;` so it's not on the DOM and doesn't take up any space.

benefits/static/css/styles.css
machikoyasuda: This is to make the external link icon smaller for the reCAPTCHA sentence. It's saying, any external link with the `form-text` class (which makes the font size 18px), the external link image should also be 18px. I can variable-ize it to make it easier to understand.

machikoyasuda: Added variables and made a small CSS fix to get the icon to be the right size. https://github.com/cal-itp/benefits/pull/2586/commits/a4e9b6f15b0631970830be1c3766c9352f8909a6 to make the number meaning more explicit.

image

benefits/eligibility/templates/eligibility/confirm.html
machikoyasuda: Added comments in both templates to clarify what is going on there

benefits/core/templates/core/includes/form.html
```diff
@@ -94,6 +92,8 @@
       {% endcomment %}
```
machikoyasuda, line 92: @angela-tran Would it make sense to remove the recaptcha code from `base.html` and put it in this `if request.recaptcha` block in `form.html` instead?

```
    {% if request.recaptcha %}
      <script nonce="{{ request.csp_nonce }}" src="{{ request.recaptcha.script_api }}"></script>
    {% endif %}
```

I could do it in this PR.

angela-tran: @machikoyasuda I am planning to do that as a PR for #2493. Will have it up soon

I think this PR has a good number of changes in it to test and review already, so probably best to not lump in the reCAPTCHA includes refactor. Even if it turns out to be a small code change, it feels like a separate PR to me.

benefits/static/css/styles.css
thekaveman: Let's do that

benefits/core/templates/core/includes/form.html
machikoyasuda: Good idea! Thank you :)

benefits/static/css/styles.css
machikoyasuda: Done: https://github.com/cal-itp/benefits/pull/2586/commits/b21f8060b17ae006aba526dcf42e0771519cdc85

benefits/eligibility/templates/eligibility/confirm.html
angela-tran, line 1: I'm seeing some space under the footer on this page when the window is long enough:

image

It seems to only affect this page. For example, here is the `confirm.html` page vs. the `start.html` page:
| Confirm | Start |
| --- | --- |
| image | Screenshot from 2024-12-17 13-43-14 |

Possible CSS regression?

benefits/static/css/styles.css
angela-tran: I narrowed down the issue I mentioned above for `confirm.html` to this style... somehow this is causing the footer to have spacing on the bottom?

Screenshot from 2024-12-17 13-51-25

Screenshot from 2024-12-17 13-51-43

angela-tran (changes requested):
Changes are looking good. I noticed an odd CSS issue though with how we're trying to override the 64px top spacing for the footer with 24px.

angela-tran:
@machikoyasuda can you check on this alignment issue on Enrollment Index?

image

machikoyasuda:
@angela-tran Oh wow I don't know how those non-fixes on in there. For now, I just added those good ole divs on divs back again. But they'll all be removed in [#2588](https://github.com/cal-itp/benefits/pull/2588/)

benefits/core/templates/core/includes/form.html
```diff
@@ -3,41 +3,39 @@
...
-  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
+  <form id="{{ form.id }}"
+        class="{{ form.classes }}"
+        action="{{ form_action }}"
+        method="{{ form.method | default:"post" | upper }}"
+        role="form">
     {% csrf_token %}
 
-    <div class="row form-container justify-content-center">
-      <div class="{{ form.classes }}">
```
angela-tran, line 10: I think there's a regression in Benefits Admin coming from this change.

We have some styles in `admin/styles.css` so that descendant checkboxes will be on the left side: https://github.com/cal-itp/benefits/blob/dce769c917469c7e9f1fb4f781e339005437e772/benefits/static/css/admin/styles.css#L88-L103

`InPersonEligibilityForm` has `checkbox-parent` in its `classes`, which would end up on this line, but now it's on the `<form>` element. 

| main | this branch (with #2555 merged in locally... *)|
| --- | --- |
| image | image  |

Not sure if we should put this `<div>` back or try to tweak the CSS?

(*When we're back in the new year, it'd be good to get all this reCAPTCHA work caught up with the fixes that are in main.)

angela-tran: @machikoyasuda How about we fix the admin styles as a follow-up after we merge this and #2588?

angela-tran: Created https://github.com/cal-itp/benefits/issues/2608

machikoyasuda: Thank you for reminding me to test Admin. I am going through Admin on this PR now, and comparing it with dev.

I am finding even more visual regressions T_T!!!!!

machikoyasuda:
> @machikoyasuda can you check on this alignment issue on Enrollment Index? I see it only in this branch.
> 
> image
> 
> _edit_: maybe you've fixed it in [#2588](https://github.com/cal-itp/benefits/pull/2588/files#diff-13b30474536f391a8f3f76ace48017731f8543d0dd40e3a69ff2ec4d11b7ac4e)? I haven't checked yet

image
I am not getting this on this branch.

angela-tran (approved):

PR #2526, opened 2024-11-19
Refactor: Remove CA State Web Template

machikoyasuda:
closes #2491 

## What this PR does
- Remove CA State Web Template CSS, which will then remove the call to the CA State Web Template font
- Remove CA State Web Template domain from CSP
- Makes any CSS fixes necessary

## Changelog
- fix:`text-left` to `text-start`
- fix: add css to fix skip nav
- fix: button focuses
- slight refactor: buttons in general. could be greatly refactored to mostly be written with css variables instead, reducing the number of new classes/modified classes.

## QA
- Trigger the skip nav
- Go through the whole app via keyboard, paying attention to the way the modals behave with the radio buttons
- Manually enable the JavaScript/Cookies warnings via HTML (remove the no display class) and compare it visually with the current one
- Check the app on responsive widths

PR #2599, opened 2024-12-19
Fix: Fix agency link regression

### LINKED ISSUE #2590: Bug: Regression - Transit name is not bolded in agency links
Image
Image

The `agency-links` includes is used on the Help page and some of the agency-specific Error pages, as shown above in Figma screenshots.

The transit agency names are currently _unbold_ on Dev, but they are bolded correctly on Prod. So we have a regression somewhere.
Image

machikoyasuda:
closes #2590 

A regression was introduced in #2526. When I removed the CA State Web Template, it removed this piece of CSS:

image

Since the agency name area uses a `<label>` element, the element was getting `font-weight: 700` from there.

But I actually read that using [`<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) for something that is not a form is not semantic!

So in this PR, I:

- Changed the `label` to a regular `p`
- Created a `.h4` CSS class, with the same exact attributes of the `h4` element. And used the `h4` class on this agency name, to get the h4 styles.
- Added `mt-1` to get the lines to be 4px away from each other.


thekaveman (approved):
LGTM

PR #2609, opened 2025-01-03
Refactor: split models into multiple files

thekaveman:
Take all models defined in `benefits/core/models.py` and split into distinct modules grouped by domain:

* `benefits/core/models/common.py` -- common fields/models, helper functions
* `benefits/core/models/claims.py` -- `ClaimsProvider` model
* `benefits/core/models/enrollment.py` -- `EnrollmentFlow`, `EnrollmentEvent` models
* `benefits/core/models/transit.py` -- `TransitProvider`, `TransitAgency` models

Maintain existing imports via top-level `benefits/core/models/__init__.py`


thekaveman:
I'm opening this for review, but I just wanted to state explicitly here:

This was an idea from the GitHub Copilot LLM (to refactor `models.py` into individual files grouped by "domain").

I am totally fine _not_ going this direction if we don't like it.

thekaveman:
I put this back in draft because I thought I should add some tests for the refactor to `SecretNameField.secret_value()`. 

This is a new method that encapsulates the usage of `benefits.secrets.get_secret_by_name()` into a single place, rather than spreading usage around the various models that have a `SecretNameField`.

angela-tran:
Starting my review of this :eyes:

tests/pytest/core/models/test_claims.py
```diff
@@ -0,0 +1,20 @@
...
+    assert model_ClaimsProvider.supports_sign_out
+    assert str(model_ClaimsProvider) == model_ClaimsProvider.client_name
+
+
+@pytest.mark.django_db
+def test_model_ClaimsProvider_client_id(mock_secret_name_field, secret_value, model_ClaimsProvider):
+    mock_field = mock_secret_name_field(model_ClaimsProvider)
+
+    assert model_ClaimsProvider.client_id == secret_value
+    mock_field.secret_value.assert_called_once_with(model_ClaimsProvider)
```
angela-tran, line 15: In these tests that use `mock_secret_name_field`, it's a bit unexpected to me that `mock_field` could be any field. An example of what I mean is that, if you go to [`benefits.core.models.claims` line 25](https://github.com/cal-itp/benefits/pull/2609/files#diff-87ae2f4b5261c3b8c191ed3bfec7f8b24fb63f4ad8ab81d5f6b7e5543ebb727fR25)
```
secret_name_field = self._meta.get_field("client_id_secret_name")
```

and replace `"client_id_secret_name"` with any other string, this test still passes.

This current test seems to be more focused on testing that the `secret_name` method is being called, so maybe we could have another test to cover that the right field is being called? I'm thinking of something like this:

```
    mock_field = model_ClaimsProvider._meta.get_field("client_id_secret_name")

    assert model_ClaimsProvider.client_id == mock_field.secret_value(model_ClaimsProvider)
```

angela-tran:
The splitting up of models into separate files looks fine to me! :+1:

tests/pytest/core/models/test_claims.py
thekaveman: Yeah I struggled with this a little. Some context:

* When all the models were in one file, we imported [`benefits.secrets.get_secret_by_name`](https://github.com/cal-itp/benefits/blob/main/benefits/secrets.py#L44) and used it directly in each model. It was easy to mock this specific function in this single file.
* With going to a package of models, this felt like an area for cleanup -- we already had the `SecretNameField` imported whenever a model needed, so wrapping `benefits.secrets.get_secret_by_name` in that class made sense
* As for testing... I originally had them written more like you show. Reflecting again, I think it's much more readable, and probably makes more sense as a test. I didn't like the test having to know the `model_ClaimsProvider._meta.get_field("")` details from the model, but... it does seem like a better test.

thekaveman: I refactored the tests in [`93d1a7a`](https://github.com/cal-itp/benefits/pull/2609/commits/93d1a7a5ec1837d2cd84b9dc125a9a194579d44b), LMK what you think.

angela-tran: Thanks for adding more context, and that cleanup makes sense to me.

The refactor you made to the mocking looks great! It's more readable to me for sure.

angela-tran (approved):

PR #2572, opened 2024-12-05
Refactor: reCAPTCHA copy + redesign

angela-tran:
closes #2537

### Actual feature
- Adds reCAPTCHA text to the 2 form pages on Benefits user-facing flow
- Removes reCAPTCHA icon from the 2 form pages

### Design changes
- Removes icons from the main flow. Icons remain on Error pages.
- Makes the main flow plus error pages (minus Enrollment Success) all 6 columns.
- Makes all the headers left-aligned, across desktop and mobile. The design is now much more "mobile-first" now, with zero alignment/column-width or font size changes between mobile and desktop.
- Buttons are 6-col wide. 
- Get started with Login.gov button is now the same height as the other CTAs.
- Agency Card form: The form helper text now has correct letter-spacing.
- Updates the external link icon CSS to allow for the icon to change sizes when font size changes

### Under the hood refactors
- Removes `media-item` CSS and templates. Now there's an `eligibility-item` that is only used for that purpose.
- Refactors `base.html` to be simpler, and have fewer things to override. Fewer empty divs delivered to the code (should be none). Details: https://github.com/cal-itp/benefits/pull/2588
- Error page refactors: https://github.com/cal-itp/benefits/pull/2585
- Naming conventions around `base` templates: All templates inherit from `base`, which may inherit from another intermediary base (like `error-base`). If a template is inheriting from `base` and meant for other pages to inherit, put the term `base` in the filename.

## How to test locally
- Use https://github.com/cal-itp/benefits/pull/1071 instructions to test in Dev. The reCAPTCHA test environment key/secret are in LastPass, under Benefits reCAPTCHA configuration.
- Make sure to test CalFresh flow
- No regressions on Admin In Person flow

## How to test on `dev`
@angela-tran is adding the reCAPTCHA env vars to dev, so we will be able to test on dev.

PR #2613, opened 2025-01-07
Fix: reCAPTCHA-related - Fix Admin form style regression

### LINKED ISSUE #2608: reCAPTCHA: fix style regressions in Benefits Admin
Once #2586 and #2588 are merged in to `refactor/recaptcha-copy`, let's get that branch caught up with changes from `main`, #2555 being one of those changes.

Then we can go through Benefits Admin and clean up any regressions that came from template refactors / clean-up.

One that we know about is this one: https://github.com/cal-itp/benefits/pull/2586#discussion_r1894469315

## Acceptance Criteria

- [x] All pages in Benefits Admin have been checked for style regressions
- [x] Regressions are fixed

machikoyasuda:
fixes #2608 

## What this PR does
- Form class: Passing in a class declares a class on `div.form-field-container`
- CSS refactor: Instead of adding styles on the last field, use `.form-group:has(input[type="checkbox"])` to select any parent div that has a checkbox in it.  
- A11y: Makes an Aria fix

## Code design goals
- It is confusing to have to read both Admin's CSS and Transit User's CSS style pages to figure out what is going on. We want to make it easier for the developer to know what exact form they are working on, and where the styles are coming from.
- I removed all Bootstrap utility classes in `form.html`. The form container now only has: `form-field-container` (which has all of our app style _defaults_), and an optional `{{ form.classes }}` that is delcared from the Python form class per form.
- The convention here is to only pass in 1 class, and the class be named something extremely specific so we don't get confused between Admin vs. Not Admin. For now, I chose: `in-person-eligibility-form` and `eligibility-verification-form` (perhaps the latter could be even more specific to indicate it's for transit users? Very open to better name suggestions!).
- Declare all styles in this class, in the proper styles.css file for whichever app you are on.
- Even though we are passing in form-specific classes, still strive to write code that is as general and agnostic as possible. Try not to rely on helpers like `last-of` or `first` etc. to select your target.

This branch vs. Dev
image

### DIFF
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -3,7 +3,7 @@
 
   {% url form.action_url as form_action %}
 
-  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
+  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}">
     {% csrf_token %}
 
     <div class="form-field-container d-flex flex-column {{ form.classes }}">
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -6,7 +6,7 @@
   <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}">
     {% csrf_token %}
 
-    <div class="form-field-container d-flex flex-column {{ form.classes }}">
+    <div class="form-field-container {{ form.classes }}">
       {% for field in form %}
         <div class="form-group mb-0">
           {# djlint:off #}
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -3,14 +3,10 @@
 
   {% url form.action_url as form_action %}
 
-  <form id="{{ form.id }}"
-        class="{{ form.classes }}"
-        action="{{ form_action }}"
-        method="{{ form.method | default:"post" | upper }}"
-        role="form">
+  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
     {% csrf_token %}
 
-    <div class="form-field-container d-flex flex-column gap-4">
+    <div class="form-field-container d-flex flex-column {{ form.classes }}">
       {% for field in form %}
         <div class="form-group mb-0">
           {# djlint:off #}
--- a/benefits/eligibility/forms.py
+++ b/benefits/eligibility/forms.py
@@ -28,6 +28,7 @@ class EnrollmentFlowSelectionForm(forms.Form):
         super().__init__(*args, **kwargs)
         flows = agency.enrollment_flows.filter(supported_enrollment_methods__contains=models.EnrollmentMethods.DIGITAL)
 
+        self.classes = "gap-4"
         # second element is not used since we render the whole label using selection_label_template,
         # therefore set to None
         flow_field = self.fields["flow"]
@@ -108,6 +109,7 @@ class EligibilityVerificationForm(forms.Form):
         self.title = title
         self.headline = headline
         self.blurb = blurb
+        self.classes = "gap-4"
 
         sub_widget = widgets.FormControlTextInput(placeholder=sub_placeholder)
         if sub_pattern:
--- a/benefits/eligibility/forms.py
+++ b/benefits/eligibility/forms.py
@@ -28,7 +28,6 @@ class EnrollmentFlowSelectionForm(forms.Form):
         super().__init__(*args, **kwargs)
         flows = agency.enrollment_flows.filter(supported_enrollment_methods__contains=models.EnrollmentMethods.DIGITAL)
 
-        self.classes = "gap-4"
         # second element is not used since we render the whole label using selection_label_template,
         # therefore set to None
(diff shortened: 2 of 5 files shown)


benefits/static/css/admin/styles.css
```diff
@@ -86,20 +86,12 @@ input[type="password"]:focus,
 }
 
 /* Eligibility Page */
-.checkbox-parent .form-group:last-of-type .col-12 {
+/* Checkbox form fields should show the checkbox first, then label */
+.form-field-container .form-group:has(input[type="checkbox"]) {
```
machikoyasuda, line 90: The current selector is selecting for the last `.form-group` div in the container. This code will break if we ever add more form fields below the checkbox. I wanted to write CSS that didn't require adding a new class (like `checkbox-parent`), and also, did not require knowing that the checkbox field is last. So I came up with this:

"Look for all `.form-group` classes within the `.form-field-container` parent, that *has* an `input` that is a `checkbox`."

Learn how to use the `has` selector here: https://css-tricks.com/the-css-has-selector/

benefits/core/templates/core/includes/form.html
```diff
@@ -3,14 +3,10 @@
 
   {% url form.action_url as form_action %}
 
-  <form id="{{ form.id }}"
-        class="{{ form.classes }}"
-        action="{{ form_action }}"
-        method="{{ form.method | default:"post" | upper }}"
-        role="form">
+  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}" role="form">
```
thekaveman, line 6: Do you want to fix that `role=form` here?

thekaveman (commented):
Are we OK specifying important display classes in Python code? It feels a little weird to me.

Since `gap-4` is on all the forms, and only `flex-reverse` is added for the Admin form, couldn't this be done with an `#id` selector for that specific form needing override (e.g. purely in CSS)?

benefits/core/templates/core/includes/form.html
thekaveman: RE: https://github.com/cal-itp/benefits/pull/2572#discussion_r1904679166

machikoyasuda: Oh yes! Forgot about this.

machikoyasuda: https://github.com/cal-itp/benefits/pull/2613/commits/c44320cd707008fdd6ab46adff693723f2be369a

machikoyasuda:
> Are we OK specifying important display classes in Python code? It feels a little weird to me.
> 
> Since `gap-4` is on all the forms, and only `flex-reverse` is added for the Admin form, couldn't this be done with an `#id` selector for that specific form needing override (e.g. purely in HTML/CSS)?

Hrm.. you are right. I'm gonna put this back into draft and think on this a little more.

benefits/core/templates/core/includes/form.html
```diff
@@ -3,14 +3,10 @@
...
 
-  <form id="{{ form.id }}"
-        class="{{ form.classes }}"
-        action="{{ form_action }}"
-        method="{{ form.method | default:"post" | upper }}"
-        role="form">
+  <form id="{{ form.id }}" action="{{ form_action }}" method="{{ form.method | default:"post" | upper }}">
     {% csrf_token %}
 
-    <div class="form-field-container d-flex flex-column gap-4">
```
machikoyasuda, line 13: Decided to remove `d-flex flex-column gap-4` from the Form template. The Form template should be as generic as possible and not assume a design.

benefits/eligibility/forms.py
```diff
@@ -108,6 +108,7 @@ def __init__(
         self.title = title
         self.headline = headline
         self.blurb = blurb
+        self.classes = "eligibility-verification-form"
```
machikoyasuda, line 112: I did not add a class for the Enrollment Flow Selection form, because it doesn't need one. This form actually did _not_ need the gap declaration at all!

benefits/in_person/forms.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, agency: models.TransitAgency, *args, **kwargs):
         super().__init__(*args, **kwargs)
         flows = agency.enrollment_flows.filter(supported_enrollment_methods__contains=models.EnrollmentMethods.IN_PERSON)
 
-        self.classes = "checkbox-parent"
+        self.classes = "in-person-eligibility-form"
```
machikoyasuda, line 29: I decided to add the `in-person` here because I want to be able to know for sure I'm working in Admin. It can get confusing when both files are named `style.css` and have a classes that have similar names.

machikoyasuda:
@thekaveman Updated the code + wrote out my code design goals/thoughts here https://github.com/cal-itp/benefits/pull/2613#issue-2771696329

benefits/core/templates/core/includes/form.html
```diff
@@ -3,14 +3,10 @@
 
   {% url form.action_url as form_action %}
 
-  <form id="{{ form.id }}"
-        class="{{ form.classes }}"
-        action="{{ form_action }}"
-        method="{{ form.method | default:"post" | upper }}"
-        role="form">
```
machikoyasuda, line 10: Removes the redundant ARIA declaration. HTML element is a form so it doesn't need anything more than that.

angela-tran (approved):
This fix looks good to me!

thekaveman (approved):
Yep, me too! :+1:

PR #2619, opened 2025-01-09
Fix: have Django serve media files even when DEBUG=False

### LINKED ISSUE #2618: Dev experience: agency logos returning 404 when running in DEBUG=False
## To Reproduce

1. Launch the app in VS Code debugger with [DEBUG=False](https://github.com/cal-itp/benefits/blob/3ecb24f/.vscode/launch.json#L19-L30)
2. Click on "Choose your Provider"
3. See missing agency logos. Network tab in dev tools shows 404 is being returned.

## Expected behavior

Agency logos are served successfully.

## Screenshots
Image

## Additional context
https://github.com/cal-itp/benefits/pull/2572#issuecomment-2576254898

angela-tran:
Closes #2618 

The reason why the images were not being served is because when DEBUG was false, there was no URL mapping from `/media` to the `django.views.static.serve` view.

If we simply move `urlpatterns.extend(static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT))` outside of the `if settings.DEBUG` block, the URL mapping still is not added because the `static` helper function only works when [DEBUG is true](https://github.com/django/django/blob/main/django/conf/urls/static.py#L23-L25). So to get around this, we can just add the URL mapping ourselves.

This URL mapping does not affect the way media images are served by the `client` container because we configured NGINX to [handle the `/media` path](https://github.com/cal-itp/benefits/blob/main/appcontainer/nginx.conf#L75-L80).

### DIFF
--- a/benefits/urls.py
+++ b/benefits/urls.py
@@ -30,15 +30,16 @@ urlpatterns = [
     path("in_person/", include("benefits.in_person.urls")),
 ]
 
-# serve user-uploaded media files
-#
-# the helper function `django.conf.urls.static.static` mentioned in
-# https://docs.djangoproject.com/en/5.1/howto/static-files/#serving-files-uploaded-by-a-user-during-development
-# only works when settings.DEBUG = True, so here we add the URL pattern ourselves so it works regardless of DEBUG.
-prefix = settings.MEDIA_URL
-urlpatterns.extend(
-    [re_path(r"^%s(?P<path>.*)$" % re.escape(prefix.lstrip("/")), serve, {"document_root": settings.MEDIA_ROOT})]
-)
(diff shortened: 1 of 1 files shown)


thekaveman (commented):
We have `benefits.settings.RUNTIME_ENVIRONMENT()` which has a value for `local`.

Would it make sense to check that here, just to be extra sure this is only for development?

angela-tran:
Great idea @thekaveman! I added a check for that here: https://github.com/cal-itp/benefits/pull/2619/commits/76e914e887067b150f27b215c0b5b96427bd6adc

thekaveman (approved):
Looks great

image

lalver1 (approved):
Thank you for the clear explanation @angela-tran! It works great on my end too 👍

PR #2621, opened 2025-01-09
Fix: use correct storage directory for client service

### LINKED ISSUE #2620: Dev experience: client service fails to come up
## To Reproduce

1. Set `DJANGO_STORAGE_DIR` in your .env file to `/home/calitp/app/data` (it has to be this path so that nginx can serve static and media files correctly -- see our [nginx.conf](https://github.com/cal-itp/benefits/blob/main/appcontainer/nginx.conf#L68-L80))
2. Outside the devcontainer, run `docker compose up -d client`
3. The container fails to start because the database cannot be opened

## Expected behavior

The container starts, and the app is running in similar way to how it runs in production using NGINX.

angela-tran:
Closes #2620 

This PR adds volume mappings to bring in `django.db` and the `uploads` folder into `/home/calitp/app/data` inside the `client` service. 

### Explanation
It needs to be this specific path is because the client container needs `DJANGO_STORAGE_DIR` to be `/home/calitp/app/data` so that nginx can serve media files correctly. But then, the client container fails to start because there is no database file at `/home/calitp/app/data`.

This volume mapping gives the container access to the `django.db` file and the `uploads/` directory from the host side of the mapping.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -10,7 +10,8 @@ services:
     ports:
       - "${DJANGO_LOCAL_PORT:-8000}:8000"
     volumes:
-      - ./:/home/calitp/app/data # matches the path in nginx.conf for media files. set your DJANGO_STORAGE_DIR to `/home/calitp/app/data` so the container can access `django.db` and `uploads/`.
+      - ./django.db:/home/calitp/app/data/django.db # the app container expects the database to already exist
+      - ./uploads/:/home/calitp/app/data/uploads
 
   dev:
     build:
--- a/compose.yml
+++ b/compose.yml
@@ -9,6 +9,8 @@ services:
     env_file: .env
     ports:
       - "${DJANGO_LOCAL_PORT:-8000}:8000"
+    volumes:
+      - ./:/home/calitp/app/data # matches the path in nginx.conf for media files. set your DJANGO_STORAGE_DIR to `/home/calitp/app/data` so the container can access `django.db` and `uploads/`.
 
   dev:
     build:


thekaveman (commented):
I'm confused why this is required? What problem are we trying to solve here?

I though `DJANGO_STORAGE_DIR` was more of a [deployment config](https://docs.calitp.org/benefits/configuration/environment-variables/#django_storage_dir) (i.e. not really for  devtime). I don't even have this setting in my local `.env`, so wouldn't this volume mapping... not work for me?

angela-tran:
Yeah, so you're right that when running `client` locally, it's not strictly necessary to set `DJANGO_STORAGE_DIR`. When that env variable isn't set, `settings.py` will use `BASE_DIR`, i.e. the working directory, as `STORAGE_DIR`.

If you run `docker compose -d client` in that state, the app will come up just fine and use `django.db` at your project root. But then you'll notice that the media files return 404s. And that's the main problem that needed to be solved here. 

It was kind of a chain of problems, so I totally understand the need for more clarification.

angela-tran:
I just edited the title of #2620 to be more accurate / descriptive :sweat_smile:

thekaveman:
OK thank you -- I did see a missing agency logo on the modal selector before this PR, and now I see the agency logo :+1: 

One comment, the effect of this is to have the entire source directory mounted under `/home/user/calitp/app/data`. Is there any way to reduce this? It takes us away from `client` which is supposed to really be a clean runtime. I.e. could we mount in just the `./uploads` directory or some other subset?

image

angela-tran:
> One comment, the effect of this is to have the entire source directory mounted under `/home/user/calitp/app/data`. Is there any way to reduce this? It takes us away from `client` which is supposed to really be a clean runtime. I.e. could we mount in just the `./uploads` directory or some other subset?

Great suggestion. Turns out we can map files too, so I updated it to only bring in `django.db` and the `uploads` directory: [ea3e18a](https://github.com/cal-itp/benefits/pull/2621/commits/ea3e18a1f443c01b27714c1d3f8c7c5e47602cbf)

image

thekaveman (changes requested):
I was worried this would break if `django.db` did not exist. But that doesn't appear to be the case, the `client` service still comes up OK. 

However, if `django.db` does not exist, a local _directory_ called `django.db` is created upon service startup:

image

So that isn't what we want.

angela-tran:
Closing this PR since we decided against an approach that involves volume mappings on `client`. #2620 documents the work that needs to be done, so no need to keep this PR open.

PR #2623, opened 2025-01-11
Feat: smoke tests

angela-tran:


angela-tran:
I was looking over the open pull requests just now and noticed this PR needed to be rebased. It was an easy one, so went ahead and did it.

angela-tran:
We had some notes for this PR in our dev workshop on [Jan 22 2025](https://docs.google.com/document/d/1YipXzRN4uAknBnc0VKOb0nJGhvoq_Y7C_pMCsANspuM/edit?tab=t.0) that mention improving how test config is done as well as adding documentation.

On the config piece, I think it would be well worth it to go ahead and implement a YAML config file, similar to what we do in `cal-itp/littlepay`, for the tests rather than the interim `.env` setup that's described by our workshop notes.

The blockers have been (1) not having time to spend on this and (2) uncertainty around what the schema of the YAML file should be.

This would be great to look at during our in-person workshop day.

thekaveman:
I had a thought related to our experience with timeouts not really working for some pages...

We're currently using the `playwright.sync_api` to e.g. import the `Page` object and `expect` helper:

```python
from playwright.sync_api import Page, expect


def test_dev_healthcheck(page: Page):
    page.goto("/healthcheck")

    expect(page.get_by_text("Healthy")).to_be_visible()
```

If we made the small change to use `playwright.async_api`, that might avoid the need for the `.wait_for_timeout()` calls:

```python
from playwright.async_api import Page, expect


async def test_dev_healthcheck(page: Page):
    await page.goto("/healthcheck")

    await expect(page.get_by_text("Healthy")).to_be_visible()
```

PR #2629, opened 2025-01-16
Fix: use headline-text for block

lalver1:
closes https://github.com/cal-itp/benefits/issues/2630

This PR fixes a small bug in `benefits/eligibility/templates/eligibility/start--sbmtd-agency-card.html` , where `block` should be `headline-text` instead of `headline` to render the title properly.

### DIFF
--- a/benefits/eligibility/templates/eligibility/start--sbmtd-agency-card.html
+++ b/benefits/eligibility/templates/eligibility/start--sbmtd-agency-card.html
@@ -5,9 +5,9 @@
     {% translate "Agency card overview" %}
 {% endblock page-title %}
 
-{% block headline %}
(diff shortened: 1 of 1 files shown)


machikoyasuda:
What this bug is -

Using `headline` instead of `headline-text` renders this view:
image

Why?

https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/templates/eligibility/start.html#L13 

This template inherits from `eligibility/start`, which sets a `headline-text` to have an `<h1>` wrapping the contents.

machikoyasuda (approved):

machikoyasuda:
Made a ticket for this for clarity, records, etc.

PR #2636, opened 2025-01-17
Refactor: web perf - load Bootstrap CSS asynchronously

angela-tran:
Part of #2493 

This PR loads the Bootstrap CSS resource asynchronously with the goal of reducing the time that page rendering is blocked.

We can use this PR to examine the code changes needed to do this and the effects of this change on other web performance metrics.

### Code changes needed

<details>

Trying out the approach in https://www.filamentgroup.com/lab/load-css-simpler, we simply need to add `media="print" onload="this.media='all'"` to the `link` element that is loading CSS.

However, then we see that Bootstrap CSS is not loading at all:

   
This is because our CSP `script-src` directive is blocking the inline Javascript in the `onload` event handler from executing:

Screenshot from 2025-01-17 12-09-48

#### Modifying our CSP to allow the `onload` code

Adding the `unsafe-inline` keyword to our `script-src` would be the simplest way to allow the event handler, but we see that is also blocked:

image

This is because:
> If a directive contains nonce or hash expressions, then the unsafe-inline keyword is ignored by browsers.

_from https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP#inline_javascript_

So then let's try out the `unsafe-hashes` keywords, which requires that we also provide the hash of the Javascript we're wanting to execute. We can get the hash by running:
```
>  echo -n "this.media='all'" | openssl dgst -sha256 -binary | openssl base64
MhtPZXr7+LpJUY5qtMutB+qWfQtMaPccfe7QXtCcEYc=
```
Then we add it to the `script-src` directive in `settings.py` -- see [lines 343 and 344](https://github.com/cal-itp/benefits/pull/2636/files#diff-d71405aa44f1e899a9cd920e5187aa1d02affdbd0da472eca95f814b0a8c1210R343-R344).

This works:

image

One final note: `unsafe-nonce` does not work with inline event handlers -- see https://stackoverflow.com/questions/56399872/content-security-policy-nonce-does-not-apply-to-event-handler-attributes.

Screenshot of attempt to use `nonce` to allow event handler to execute:

</details>

### Effects on web performance

Running Lighthouse reports on the app before and after the code changes, we see that `bootstrap.min.css` is no longer blocking page rendering. See the Diagnostic results below:

| Before | After |
| - | - |
|image  |image |

However, the overall Performance score is worse because loading the CSS asynchronously causes a layout shift.
image

**Lighthouse Performance metrics _before_ code changes**

| Run | Performance score | First Contentful Paint (s) | Largest Contentful Paint (s) | Total Blocking Time (ms) | Cumulative Layout Shift | Speed Index (s) |
| --- | ----------------- | -------------------------- | ---------------------------- | ------------------------ | ----------------------------- | --------------- |
| 1   | 97                | 1.9                        | 2.3                          | 10                       | 0                             | 1.9             |
| 2   | 93                | 1.7                        | 3.2                          | 0                        | 0                             | 1.7             |
| 3   | 93                | 1.9                        | 3.1                          | 0                        | 0                             | 1.9             |
| 4   | 97                | 1.7                        | 2.3                          | 0                        | 0                             | 1.7             |
| 5   | 94                | 2.3                        | 2.7                          | 0                        | 0                             | 2.3             |

<details><summary>Screenshots</summary>

| Run | Screenshot |
| - | - |
| 1 | Screenshot from 2025-01-17 13-35-38 |
| 2 | Screenshot from 2025-01-17 13-36-06 |
| 3 | Screenshot from 2025-01-17 13-36-30 |
| 4 | Screenshot from 2025-01-17 13-36-44 |
| 5 | Screenshot from 2025-01-17 13-37-00 |
</details>

**Lighthouse Performance metrics _after_ code changes**

| Run | Performance score | First Contentful Paint (s) | Largest Contentful Paint (s) | Total Blocking Time (ms) | Cumulative Layout Shift | Speed Index (s) |
| --- | ----------------- | -------------------------- | ---------------------------- | ------------------------ | ----------------------------- | --------------- |
| 1   | 76                | 1.7                        | 3.0                          | 0                        | 0.365                         | 1.7             |
| 2   | 78                | 2.3                        | 2.3                          | 30                       | 0.365                         | 2.3             |
| 3   | 76                | 2.3                        | 2.7                          | 60                       | 0.365                         | 2.3             |
| 4   | 80                | 1.8                        | 2.2                          | 40                       | 0.365                         | 1.8             |
| 5   | 78                | 2.3                        | 2.3                          | 60                       | 0.365                         | 2.3             |

<details><summary>Screenshots</summary>

| Run | Screenshot |
| - | - |
| 1| Screenshot from 2025-01-17 13-38-51 |
| 2 | Screenshot from 2025-01-17 13-39-01 |
| 3 | Screenshot from 2025-01-17 13-39-14 |
| 4 | Screenshot from 2025-01-17 13-39-29 |
| 5 | Screenshot from 2025-01-17 13-40-22 | 

</details>

### DIFF
--- a/benefits/core/templates/core/includes/bootstrap-css.html
+++ b/benefits/core/templates/core/includes/bootstrap-css.html
@@ -3,5 +3,4 @@
       integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH"
       crossorigin="anonymous"
       media="print"
-      onload="this.media='all'"
-      nonce="{{ request.csp_nonce }}">
+      onload="this.media='all'">
--- a/benefits/core/templates/core/includes/bootstrap-css.html
+++ b/benefits/core/templates/core/includes/bootstrap-css.html
@@ -1,4 +1,7 @@
 <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css"
       rel="stylesheet"
       integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH"
-      crossorigin="anonymous">
+      crossorigin="anonymous"
+      media="print"
+      onload="this.media='all'"
+      nonce="{{ request.csp_nonce }}">
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -340,6 +340,8 @@ if sentry.SENTRY_CSP_REPORT_URI:
 
 CSP_SCRIPT_SRC = [
     "'self'",
+    "'unsafe-hashes'",
+    "'sha256-MhtPZXr7+LpJUY5qtMutB+qWfQtMaPccfe7QXtCcEYc='",  # hash of 'this.media='all'
     "https://cdn.amplitude.com/libs/",
     "https://cdn.jsdelivr.net/",
     "*.littlepay.com",


angela-tran:
A good CLS score is < 0.1

https://web.dev/articles/cls#what-is-a-good-cls-score

angela-tran:
Before code changes to load CSS async:
before






After code changes to load CSS async:


after

angela-tran:
@machikoyasuda - do you have thoughts on if there's something we can do about the layout shift that is introduced by loading CSS asynchronously?

machikoyasuda:
First off, thank you for all this work and documenting it so well. It made it really easy for me to see what is going on and the tradeoffs.

One way to get around this layout shift is to write in-line CSS (like `<div styles="...">`) and put it all in there, for the big things on the page, like setting the font to be white or the height of the header. That wouldn't be too hard for font colors and sizes, for example, but it gets complex with CSS that needs media queries. 

Might be a good short (10 min) discussion topic for tomorrow's meeting - but I'm leaning towards No at this time.

machikoyasuda:
More readings:

https://www.speedcurve.com/web-performance-guide/using-critical-css-for-faster-rendering/

https://csswizardry.com/2022/09/critical-css-not-so-fast/

angela-tran:
Today in our dev workshop, we talked about how the solution to the flash of unstyled content would be to inline critical CSS that's needed "above the fold," but this is complex, and all our page content is above the fold. 

We also talked about how we're using Bootstrap in the simplest way possible with bringing in the whole library. If we had some CSS build step, we could reduce the amount of CSS that we load.

This is not something we want or need to introduce right now, so we've decided to leave the CSS loading as it is.

PR #2639, opened 2025-01-21
Perf: improve Google Web Font performance

### LINKED ISSUE #2492: Web perf: Audit and improve Google Web Font performance
This ticket is part of a group of web performance tickets tracked here https://github.com/cal-itp/benefits/issues/2487.

## Acceptance Criteria

- [ ] Double-check that we are using all 3 Roboto weights: 400 normal, 500 normal, 700 normal. Can we remove any weights?
- [ ] Use `font-display: swap` and `preload`. Follow this guide: https://web.dev/learn/performance/optimize-web-fonts 
- [ ] Ensure there is a good fall-back default font. This might have changed after removing CA State Web Template.

lalver1:
Closes #2492 
This PR audits and improves how the Google Web Font is loaded.

## Background
In [`base.html`](https://github.com/cal-itp/benefits/blob/933cd572b6b5731665f877b40da677ccad8c5ee1/benefits/core/templates/core/base.html#L21), the external font (Roboto) is loaded first, and the local font (Public Sans) is loaded after. Since there are no conflicts, Roboto takes precedence. Then, Roboto is set for [`--bs-font-sans-serif`](https://github.com/cal-itp/benefits/blob/933cd572b6b5731665f877b40da677ccad8c5ee1/benefits/static/css/variables.css#L14) so it is used throughout the site. The local font is only used for [login.gov buttons](https://github.com/cal-itp/benefits/blob/933cd572b6b5731665f877b40da677ccad8c5ee1/benefits/static/css/styles.css#L423).

## Audit
- [x] Double-check that we are using all [3 Roboto weights](https://github.com/cal-itp/benefits/blob/933cd572b6b5731665f877b40da677ccad8c5ee1/benefits/core/templates/core/base.html#L21): 400 normal, 500 normal, 700 normal. _We are using all 3 weights._
- [x] Use [`font-display: swap`](https://developer.chrome.com/docs/lighthouse/performance/font-display/?utm_source=lighthouse&utm_medium=devtools#google_fonts_example) and ~[`preload`](https://web.dev/learn/performance/optimize-web-fonts#preload). _Lighthouse reports an improvement in performance from 92 to 97 when using these attributes on `<link>`._~
- [x] Ensure there is a good fall-back default font. _The [first fallback font](https://github.com/cal-itp/benefits/blob/933cd572b6b5731665f877b40da677ccad8c5ee1/benefits/static/css/variables.css#L14) is `system-ui` which is a good fallback._

### DIFF
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -20,7 +20,7 @@
 
     <link rel="preconnect" href="https://fonts.googleapis.com">
     <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
-    <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet" type="text/css">
+    <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" rel="stylesheet" type="text/css">
     {% include "core/includes/bootstrap-css.html" %}
     <link href="{% static "css/styles.css" %}" rel="stylesheet">
     <link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
--- a/benefits/core/templates/core/base.html
+++ b/benefits/core/templates/core/base.html
@@ -18,6 +18,8 @@
     {% block preload %}
     {% endblock preload %}
 
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
     <link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700" rel="stylesheet" type="text/css">
     {% include "core/includes/bootstrap-css.html" %}
     <link href="{% static "css/styles.css" %}" rel="stylesheet">


thekaveman:
I've been looking at this. Have a few questions, writing them up as a review comment.

thekaveman (commented):
I noticed in my browser console (FF 134.0.1 64-bit on Linux) these warnings related to preloading:

image


I'm looking at the MDN docs for [`rel="preload"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload), and specifically [_What types of content can be preloaded?_](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#what_types_of_content_can_be_preloaded):

> * `font`: Font file.
> * ...
> * ...
> 
> **Note:** `font`... preloading requires the `crossorigin` attribute to be set; see [CORS-enabled fetches](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#cors-enabled_fetches) below.

Then if we look at that _CORS-enabled fetches_ section:

> When preloading resources that are fetched with [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) enabled (e.g. [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch), [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) or [fonts](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face)), special care needs to be taken to setting the [crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#crossorigin) attribute on your [<link>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) element. The attribute needs to be set to match the resource's CORS and credentials mode, even when the fetch is not cross-origin.
>
> As mentioned above, one interesting case where this applies is font files. Because of various reasons, these have to be fetched using anonymous-mode CORS (see [Font fetching requirements](https://drafts.csswg.org/css-fonts/#font-fetching-requirements)).

I would _hope_ Google allows CORS requests for fonts since this seems like a common use-case and clear-cut requirement, but I'm not sure? 

It is worth adding the `crossorigin="anonymous"` attribute to the tag and seeing if it gets rid of this warning. We use this elsewhere e.g. for [Bootstrap CSS](https://github.com/cal-itp/benefits/blob/main/benefits/core/templates/core/includes/bootstrap-css.html#L4)

machikoyasuda:
Getting the same warnings as @thekaveman on Mac OS/Firefox and Mac OS/Chrome:

image

Safari, renders the same (fallback font) but no error/warning message in console.

lalver1:
Thanks for the comments @machikoyasuda and @thekaveman!

I re-tested using macOS/Firefox, macOS/Chrome, and macOS/Safari and found a few things (in addition to the `crossorigin="anonymous"` attribute) that I had overlooked. Here's a summary of the changes:

- Replaced `type="text/css"` with `as="font"` following the [optimization docs](https://web.dev/learn/performance/optimize-web-fonts#preload) in 2b25225e7cfd148ef5864e95b913a7f1875ccf4f. Previously, I had mistakenly kept both attributes.
- Moved the Google font url from the CSP `style-src` directive to a (new) `font-src` directive in 9ae7973a04a86ce30368c83f94000bfd09eef68e (since the Roboto font tag changed from a `text/css` type to a `font` type) to avoid this issue reported by Chrome and Safari (not sure why Firefox didn't show an issue):
image
- All 3 browsers reported this same issue (`The resource https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap was preloaded using link preload but not used within a few seconds from the window's load event.`) but the description in Chrome was the most specific:
image
Maybe this means that currently we aren't benefiting too much from `preload`, but I would guess it's still a good optimization to have.

angela-tran:
> I'm looking at the MDN docs for [`rel="preload"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload), and specifically [_What types of content can be preloaded?_](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#what_types_of_content_can_be_preloaded):
> 
> > * `font`: Font file.
> > * ...
> > * ...
> > 
> > **Note:** `font`... preloading requires the `crossorigin` attribute to be set; see [CORS-enabled fetches](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload#cors-enabled_fetches) below.

> - Replaced `type="text/css"` with `as="font"` following the [optimization docs](https://web.dev/learn/performance/optimize-web-fonts#preload) in 2b25225e7cfd148ef5864e95b913a7f1875ccf4f. Previously, I had mistakenly kept both attributes.

https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap is a CSS file that loads font files, so I don't think using `as="font"` is correct.

angela-tran:
The [documentation](https://web.dev/learn/performance/optimize-web-fonts#preload) linked by #2492 is pretty general and assumes you're loading actual font files.

I found some resources more specific to loading resources from the Google Fonts API, which is what we're doing to load Roboto.

- https://stackoverflow.com/questions/50824181/preloading-google-fonts
- https://www.cdnplanet.com/blog/faster-google-webfonts-preconnect

I think it could help to look at the network waterfall chart in the devtools before and after making changes to see if changes are working as intended.

lalver1:
Thanks @angela-tran!
That's a good point about the `as` attribute. Reading this [mdn doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#as) I thought that setting `as="font"` was appropriate (since it mentions it applies to `CSS @font-face`) but I agree with you, other docs such as this [web.dev doc](https://web.dev/learn/performance/optimize-web-fonts#preload) show that `as="font"` is meant for preloading font files, not CSS. Also, setting `as="style"` has the added benefit that I don't need to make any changes to our Django-CSP configuration 🙂 and I think that's a cleaner solution.

I'm still looking at this because even though the `GET` Google font was successful (I can see it in the Network tab of devtools), the page isn't rendering with Roboto, it's falling back to the System font. However, if I implement this:

```html
<link rel="preload" as="style" href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" crossorigin="anonymous">
<link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" rel="stylesheet" type="text/css">
```

the page renders using Roboto. I just want to make sure that having both `<link>` tags is correct, it's just that from the documentation I had the impression that only the first `<link>` tag was needed.

thekaveman:
> I just want to make sure that having both `<link>` tags is correct

We should only need a single tag here right? This doesn't make sense to me.

angela-tran:
> > I just want to make sure that having both `<link>` tags is correct
> 
> We should only need a single tag here right? This doesn't make sense to me.

The first `link` tells the preload scanner about the resource but doesn't actually download it yet. From MDN docs on [`preload`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload):

> Even though the name contains the term load, it doesn't load and execute the script but only schedules it to be downloaded and cached with a higher priority.

The second `link` actually causes the download of the resource.

thekaveman:
Doh, thank you @angela-tran :pray:

angela-tran:
@lalver1 What you're seeing makes sense with the docs on `preload` and also with these two StackOverflow answers to the question I linked in https://github.com/cal-itp/benefits/pull/2639#issuecomment-2610938611 :
- https://stackoverflow.com/a/62219966
- https://stackoverflow.com/a/56845961

lalver1:
After more testing (running scenarios 10 times and calculating the median) and reading more about the difference between `preconnect` and `preload`, I decided to make some changes to how I initially implemented this performance improvement.

In summary, we now warm up the connection to the Roboto web font using `preconnect` and still use `display=swap` to display the content immediately without waiting for the web font to download:

```html
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
<link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap"
rel="stylesheet" type="text/css">
```
The Largest Contentful Paint (LCP) (sec) is 0.13 and we see no warnings in the browser's console.

There are several reasons (informed by [article 1](https://sia.codes/posts/making-google-fonts-faster/), and [article 2](https://nitropack.io/blog/post/resource-hints-performance-optimization)) why I ended up going this route:

- The [Google Fonts HTML snippet](https://fonts.google.com/selection/embed) uses `preconnect` (I made slight modifications to not use their v2 API update)
image

- I noticed that the LCP was actually worse using `preload`:

```html
<link rel="preload" as="style" href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" crossorigin="anonymous">
<link href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&display=swap" rel="stylesheet" type="text/css">
```
LCP (sec): 0.17

- Both Chrome and Firefox reported several warnings when using `preload`
image
image

- The Google font stylesheet is loaded twice when using `preload` (although it seems that concurrently)
image

My only remaining concern is that Chrome's Lighthouse analysis gives `main` a score of 97 and this branch a score of 91. But this contradicts what I am seeing using other metrics

| branch | Chrome (LCP in Performance tool, sec) | Firefox (Network tool, finish time, msec) |
|---|---|---|
| `main` | 0.15| 762  |
| `perf/web-fonts` | 0.13| 676.5 |

angela-tran (approved):
These changes look good to me! Thanks for the links to those articles.

machikoyasuda (approved):
Tested this locally on Mac/Safari, Chrome, Firefox. 👍

PR #2644, opened 2025-01-23
Fix: Set important to override bootstrap h4

### LINKED ISSUE #2601: Bug: Chrome - h4 class not working on Chrome
Compare the agency name on agency links (Help page is easily accessible https://dev-benefits.calitp.org/help) on Chrome vs Firefox, Safari

## To Reproduce

1. Go to https://dev-benefits.calitp.org/help
1. Scroll down to bottom
4. The agency name should be size 18px font.

## Screenshots

Image

- OS: OSX
- Browser: Chrome

## Additional context

Need to disable https://github.com/twbs/rfs/tree/v10.0.0 I think.
https://getbootstrap.com/docs/5.3/getting-started/rfs/#what-is-rfs

machikoyasuda:
closes #2601 

Bootstrap has a class called `h4`, which has different sizing than what we want.

I could have done a few things:
- Rename the `h4` class in our styles
- Use this `disable-rfs` class that Bootstrap gives you
- Or add `important` to the existing `h4` size declaration

I decided to go with the latter, so that any future use of `h4` will be covered (and not require the dev to know to use `disable-rfs`). I didn't want to rename all the hX classes.

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -238,9 +238,11 @@ h3,
 
 h4,
 .h4 {
-  font-size: var(--h4-font-size);
-  letter-spacing: calc(var(--h4-font-size) * var(--h4-letter-spacing-percent));
-  line-height: var(--h4-line-height);
+  font-size: var(--h4-font-size) !important; /* override Bootstrap class */
+  letter-spacing: calc(
+    var(--h4-font-size) * var(--h4-letter-spacing-percent)
+  ) !important;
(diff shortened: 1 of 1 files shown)


angela-tran:
Did you find out anything about why it only happens on Chrome?

angela-tran (approved):
I'm not able to reproduce the issue on Linux, so I can't test this myself. From that and the ticket description, I'm guessing that this is an macOS / Chrome - specific issue.

The reasoning for using `!important` sounds good to me.

Issue #2201, opened 2024-07-04
Admin: Cal-ITP Benefits staff user can add new models


angela-tran:
## Acceptance Criteria

- [x] The [Benefits admin configuration](https://docs.google.com/spreadsheets/d/1lPp_LJTfJoWCyOkRC4RaljABGG_jPWOgm4gaAHJigMU/edit) is updated 
  - [x] Current model and field definitions
  - [x] Confirm read/write permissions for each model field
- [x] Cal-ITP Benefits staff users can create new instances of relevant models
  - [x] `ClaimsProvider` - #2671 
  - [x] `EnrollmentFlow` - #2654 
  - [x] `TransitAgency` - #2667 

## Additional context

Part of #2001

angela-tran:
> Cal-ITP Benefits staff users can create new instances of EligibilityTypes, EligibilityVerifiers, AuthProviders, PaymentProcessors, and TransitAgencys

Some of these models have required fields that a Cal-ITP user wouldn't be able to set. 

Cal-ITP users also currently wouldn't be able to add the templates / form classes for a new transit agency without engineers implementing those.

thekaveman:
> > Cal-ITP Benefits staff users can create new instances of EligibilityTypes, EligibilityVerifiers, AuthProviders, PaymentProcessors, and TransitAgencys
> 
> Some of these models have required fields that a Cal-ITP user wouldn't be able to set.
> 
> Cal-ITP users also currently wouldn't be able to add the templates / form classes for a new transit agency without engineers implementing those.

#2471 addressed some of the concern with required fields and templates.

thekaveman:
Let's confirm what if anything needs to be done to complete this. I think it might already be mostly done?

angela-tran:
Completed by PRs mentioned above

Issue #2278, opened 2024-08-06
Admin: Transit agency staff should access configuration for only their transit agency


angela-tran:
Assuming the work in #2243 lays the foundation of associating users to transit agencies, this ticket covers protecting the model configuration views using that association. What this means from the code-perspective is that there are implementations of `has_view_permissions`, `has_change_permissions`, etc. on the `ModelAdmin`s for our models.

## Acceptance Criteria

- [ ]

thekaveman:
Goal for `2025 - Sprint 01`: Shape this ticket with acceptance criteria.

angela-tran:
I met with @indexing and @thekaveman yesterday to discuss what this ticket represents more specifically.

We came up with a list of new Benefits Admin pages that would be for agency staff users (read-only configuration page and prior enrollment table) and some minimal UX things we'll need to start thinking about.

We'll eventually create tickets from the notes we took yesterday. Closing this ticket.

PR #2654, opened 2025-01-30
Feat: Cal-ITP staff can add EnrollmentFlow

### LINKED ISSUE #2664: Admin: Cal-ITP Benefits staff user can add `EnrollmentFlow` models
This is for breaking up the part of #2638 that lets staff members add `EnrollmentFlow` models.

## Acceptance Criteria

- Cal-ITP staff member can add a new `EnrollmentFlow`:
   - [ ] can add an `EnrollmentFlow` with claims verification to a transit agency
   - [ ] can start creation of an `EnrollmentFlow` with API verification; they have to ask an engineer to finish out configuration of the Eligibility API fields
- [ ] Cal-ITP staff member should not be able to add incomplete `EnrollmentFlow` to a transit agency

angela-tran:
Part of #2201

Closes #2664 

Allows Cal-ITP staff members to add new `EnrollmentFlow` objects.

### Staff member
Login into Benefits Admin as `calitp-user` from our LastPass shared credentials.

- Cal-ITP staff member can add a new `EnrollmentFlow`:
   - [ ] can add an `EnrollmentFlow` with claims verification to a transit agency
   - [ ] can start creation of an `EnrollmentFlow` with API verification; they have to ask an engineer to finish out configuration of the Eligibility API fields before the flow can be added to a transit agency
- Cal-ITP staff member should not be able to add incomplete `EnrollmentFlow` to a transit agency
   - [ ] incomplete - none of the fields for claims or API verification have been filled out
   - [ ] incomplete - some but not all fields for claims verification have been filled out
   - [ ] incomplete - some but not all fields for expiration have been filled out
   - ~~[ ] incomplete - some but not all fields for API verification have been filled out~~ - not possible for staff member to do by themselves

### Some screenshots

>  Cal-ITP staff member should not be able to add incomplete `EnrollmentFlow` to a transit agency

<details><summary>Incomplete - none of the fields for claims or API verification have been filled out</summary>

image

</details>

<details><summary>Incomplete - some but not all fields for expiration have been filled out</summary>

image

</details>

### DIFF
--- a/benefits/core/admin/enrollment.py
+++ b/benefits/core/admin/enrollment.py
@@ -53,16 +53,6 @@ class SortableEnrollmentFlowAdmin(SortableAdminMixin, admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
         fields = []
 
-        if not is_staff_member_or_superuser(request.user):
-            fields.extend(
-                [
-                    "claims_scope",
-                    "claims_eligibility_claim",
-                    "claims_scheme_override",
-                    "eligibility_api_url",
-                    "eligibility_form_class",
-                ]
-            )
         if not request.user.is_superuser:
             fields.extend(
                 [
@@ -80,16 +70,6 @@ class SortableEnrollmentFlowAdmin(SortableAdminMixin, admin.ModelAdmin):
     def get_readonly_fields(self, request, obj=None):
         fields = []
 
-        if not is_staff_member_or_superuser(request.user):
-            fields.extend(
-                [
-                    "system_name",
-                    "transit_agency",
-                    "supported_enrollment_methods",
-                    "claims_provider",
-                    "supports_expiration",
-                ]
-            )
         if not request.user.is_superuser:
             fields.extend(
                 [
--- a/benefits/core/admin/enrollment.py
+++ b/benefits/core/admin/enrollment.py
@@ -1,4 +1,6 @@
+from django import forms
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.contrib import admin
 from django.http import HttpRequest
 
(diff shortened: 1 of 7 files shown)


benefits/core/admin/enrollment.py
```diff
@@ -47,36 +47,70 @@ def has_view_permission(self, request: HttpRequest, obj=None):
...
+        if not is_staff_member_or_superuser(request.user):
+            fields.extend(
+                [
+                    "claims_scope",
+                    "claims_eligibility_claim",
+                    "claims_scheme_override",
+                    "eligibility_api_url",
+                    "eligibility_form_class",
+                ]
+            )
```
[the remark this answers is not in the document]
angela-tran: Ok, yeah I think this was just from when there was a "regular" user test case. The [`"staff"` test case](https://github.com/cal-itp/benefits/pull/2654/files#diff-a85571e835f3bacb4969112185accb9e0cf2f7879974ee928d66a8459c5c3646R95-R103) passes regardless of this code.

I'll remove these. They will probably be part of #2278

benefits/core/admin/enrollment.py
```diff
@@ -47,36 +47,50 @@ def has_view_permission(self, request: HttpRequest, obj=None):
...
+                    "eligibility_api_auth_header",
+                    "eligibility_api_auth_key_secret_name",
+                    "eligibility_api_public_key",
+                    "eligibility_api_jwe_cek_enc",
+                    "eligibility_api_jwe_encryption_alg",
+                    "eligibility_api_jws_signing_alg",
+                ]
+            )
+
+        return fields or super().get_exclude(request, obj)
```
angela-tran, line 68: Hmm, isn't this always going to return `fields`?

benefits/core/admin/enrollment.py
```diff
@@ -47,36 +47,50 @@ def has_view_permission(self, request: HttpRequest, obj=None):
...
+                ]
+            )
+
+        return fields or super().get_readonly_fields(request, obj)
+
+    def has_add_permission(self, request: HttpRequest, obj=None):
+        if settings.RUNTIME_ENVIRONMENT() != settings.RUNTIME_ENVS.PROD:
+            return True
+        elif request.user and is_staff_member_or_superuser(request.user):
+            return True
```
angela-tran, line 94: I just noticed this is a divergence from how we've been setting up permissions for the Cal-ITP staff group. The pattern so far has been to add `Permission` objects to the Cal-ITP group in migrations -- e.g. in [this migration](https://github.com/cal-itp/benefits/blob/feat/staff-add-enrollmentflow/benefits/core/migrations/0014_staff_group_view_permissions.py#L6-L21) we gave the Cal-ITP group view permissions.

I'm not opposed to changing to this pattern if we like this better.

The original reason for using the `Permission` objects that Django provides is so that this UI isn't totally worthless or misleading:

image

benefits/core/admin/enrollment.py
angela-tran: Just realized with Python being falsy, this is fine :smile:

benefits/core/admin/enrollment.py
angela-tran: Maybe we leave this implementation, and create a new follow-up ticket to:
 - refactor the other permissions to be code-driven as well
 - remove this view (if that's possible?)

benefits/core/admin/enrollment.py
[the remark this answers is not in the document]
lalver1: Yep, I think that implements the "regular" user logic, I agree with removing it in 7bc668770ee090e34173702eca0cfd4726ef731b

benefits/core/admin/enrollment.py
thekaveman: Yeah I do not like migration-based permissions... it feels like the wrong way to do it, having the permission be encoded in transient database state vs. explicitly defined in the code. 

It also spreads permissions out across (many) different files. Would rather they are consolidated in the admin-related code.

thekaveman: Is there a way to do "both" e.g. encode them in the admin models, and also have that UI from your screenshot make sense?

angela-tran: > Would rather they are consolidated in the admin-related code. Easier to test that way I think too.

Yeah, I'm good with changing over to defining them in admin-related code.

> Is there a way to do "both" e.g. encode them in the admin models, and also have that UI from your screenshot make sense?

No, I think if we're overriding `has_add_permission`, `has_view_permission`, etc., we're essentially opting out of using the built-in [permissions system](https://docs.djangoproject.com/en/5.1/topics/auth/default/#permissions-and-authorization), so it'd be better to just remove that UI so no one is misled to think that using it will do anything.

The built-in permission system really just consists of the `Permission` objects and and those [`has_X_permission` methods](https://github.com/django/django/blob/main/django/contrib/admin/options.py#L559-L614) checking if the user is associated with those `Permission` objects or not, either directly or through the groups they belong to.

angela-tran: The `User` admin page also has a UI that we'd want to remove as well:

image

thekaveman: What are your thoughts on doing it one way over the other?

angela-tran: I think it's a good idea to consolidate these permissions as code. As we add more and more configuration to `get_exclude` and `get_readonly_fields` for each `ModelAdmin`, it'll be helpful to have the `has_add_permission`, etc. logic right there too, rather than it being far away in Django code that reads from the database.

I had us start out with trying the built-in permissions system so that we just use what they've already built. Now that we've gotten further along with using `ModelAdmin`s, it seems it would be better to more explicitly code out what permissions we want for certain groups and types of users.

The built-in permissions system gives you the flexibility to change permissions very easily, but we really don't have any need for that I don't think.

angela-tran: I created https://github.com/cal-itp/benefits/issues/2658 for this

angela-tran:
I just have one more test to fix :muscle: :triumph:

It's [`test_TransitAgency_clean_dirty_flow`](https://github.com/cal-itp/benefits/blob/30cafde46811907a6a5abe6978bc54703ed4228a/tests/pytest/core/models/test_transit.py#L198) which asserts that saving a `TransitAgency` linked to an incomplete `EnrollmentFlow` will fail validation.

The test needs to be refactored to go through forms rather than models so that it more accurately captures the validation process.

angela-tran:
I'm working to add back the test coverage for validation. I removed tests that covered model validation and hadn't put in equivalent tests for the form.

https://github.com/cal-itp/benefits/pull/2654#issuecomment-2625647526
image

angela-tran:
Just realized I still needed to push my rebase so that you have the Cal-ITP staff user fixture

benefits/core/models/enrollment.py
```diff
@@ -306,14 +276,12 @@ def clean(self):
...
 
-        if field_errors:
-            raise ValidationError(field_errors)
         if template_errors:
             raise ValidationError(template_errors)
 
     def eligibility_form_instance(self, *args, **kwargs):
         """Return an instance of this flow's EligibilityForm, or None."""
-        if not bool(self.eligibility_form_class):
+        if not self.uses_api_verification:
```
lalver1, line 284: This is nice, it's more explicit and also checks on the value of `self.eligibility_api_url`

lalver1 (approved):
This looks good to me and thanks for the step by step explanation in the commit history @angela-tran! It was also nice to see how to handle field validation on the `model` and on the `form` associated with the model.
I tested that as a Staff member I
- [x] can add an `EnrollmentFlow` with claims verification to a transit agency
- [x] can start creation of an `EnrollmentFlow` with API verification; but I'd have to ask an engineer to finish out configuration of the Eligibility API fields before the flow can be added to a transit agency
- [x] can't add an incomplete `EnrollmentFlow` to a transit agency
   - [x] none of the fields for claims or API verification have been filled out
   - [x] some but not all fields for claims verification have been filled out
   - [x] some but not all fields for expiration have been filled out

PR #2656, opened 2025-01-31
Fix: port forwarding in VS Code

### LINKED ISSUE #2642: Port Forwarding in VS Code is unstable
When debugging the app, port forwarding in VS Code is unstable and in some situations it prevents the user from being able to access the application in their browser.

## To Reproduce

1. In VS Code, close all ports, reopen the folder locally, and rebuild the container to start with a fresh dev container, notice the two forwarded ports 
Image
2. Run the VS Code debugger using the `Django: Benefits Client` launcher configuration, notice that a third port was added
Image
3. In your web browser, navigate to the app running on `localhost:DJANGO_LOCAL_PORT` where `DJANGO_LOCAL_PORT` is set in your `.env` file.
4. Select a Transit Agency and submit the form `Choose this benefit` with `Agency cardholder` selected. Notice that a new port has been added
Image
5. The web browser hangs
6. Stop the debugger
7. Restart the debugger. Notice that the forwarded addess has changed, the browser is still hangingImage
 
## Expected behavior

The web browser should still be able to reach the application on `localhost:DJANGO_LOCAL_PORT`

## Screenshots

Similar behavior occurs with VS Code running on Linux although the `Forwarded Address` and `Running Process` columns look slightly different from VS Code running in macOS.

Image

## Desktop

- OS: macOS Sonoma
- Browser: Firefox
- Version: 134.0.1 

## Additional context

A possible solution is to add `"otherPortsAttributes": { "onAutoForward": "ignore" }` to `devcontainer.json` as suggested by https://stackoverflow.com/a/74177846. A description of the `otherPortsAttributes` property is provided at https://containers.dev/implementors/json_reference/#general-properties and the options that the property can take are shown at https://containers.dev/implementors/json_reference/#port-attributes.
While adding `"portsAttributes": { "onAutoForward": "ignore" }` could also be a solution, adding this did not make any difference in contrast to adding `"otherPortsAttributes": { "onAutoForward": "ignore" }`.

lalver1:
Closes #2642

This PR changes the behavior of the dev container to only auto forward the ports associated with the `docs` and `server` services in the compose file. By setting `"otherPortsAttributes": { "onAutoForward": "ignore" }`, [any other ports should not be auto-forwarded at all](https://containers.dev/implementors/json_reference/#port-attributes).

## Reviewing
1. Exit your current dev container
2. Switch to this branch
3. Rebuild and reopen the dev container
4. Start the debugger
5. Verify that the app is always reachable from your browser at `localhost:DJANGO_LOCAL_PORT/` (open the benefits app, submit the eligibility form, navigate to the admin, etc.) and that the `PORTS` tab in VS Code always only shows 2 ports, the `server` and `docs` ports

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -5,6 +5,7 @@
   "service": "dev",
   "runServices": ["dev", "docs", "server"],
   "forwardPorts": ["docs:8001", "server:8000"],
+  "otherPortsAttributes": { "onAutoForward": "ignore" },
   "workspaceFolder": "/calitp/app",
   "postStartCommand": ["/bin/bash", "bin/reset_db.sh"],
   "postAttachCommand": ["/bin/bash", ".devcontainer/postAttach.sh"],


thekaveman:
I'm still not quite sure about the underlying issue here. I've never experienced the situation described in #2642, and it doesn't seem straightforwardly reproducible (e.g. does it happen that way every single time going through an Agency Card flow?)

In any case, on Linux at least, this change appears to have no effect. I'm still seeing all the various ports on the ports tab:

image

machikoyasuda:
I tested this locally on my macOS.

My ports tab looks like this:
image


I am able to access the server and healthcheck:
image

And the docs: http://localhost:8001/benefits/
image

But not the Benefits client app at http://localhost:11369/ after running the debugger. Am I mising something?

machikoyasuda (approved):
Confirmed on VS Code Insiders (pre-release) `1.97.0-insider`. Ran the app in container, opened docs, server and the client. Went through client flow.

thekaveman:
@machikoyasuda:

> But not the Benefits client app at http://localhost:11369/ after running the debugger. Am I mising something?

I'm guessing, but from the screenshot it looks like maybe `DJANGO_LOCAL_PORT` was not set to `11369`?

image

I would expect that very last entry to have `0.0.0.0:11369` in the `Forwarded Address` column.

lalver1:
I added `dev:8000` to `forwardPorts`
 
```yaml
  "forwardPorts": ["docs:8001", "server:8000", "dev:8000"],
  "otherPortsAttributes": { "onAutoForward": "ignore" },
```
but did not quite get the behavior I was expecting:
image
VS Code did not open extra ports when debugging (that's good) but `dev` was available at both ports 11369 and 51422. This may be confusing for the dev experience, so I'm leaning to not making changes to this branch.

My guess was that `dev` would have behaved just like `docs` and `server`,  but there are 2 differences that may be the reason for this behavior:

1.  `dev` is the dev container service in `devcontainer.json` (`"service": "dev"`), somehow making port allocation in the Ports tab different compared to the other services
2.  In the `compose` file, `dev` is assigned a specific port on the host but `docs` and `server` get allocated any unassigned port of the host, and this creates a different behavior in the Ports tab of VS Code

angela-tran:
So I noticed in your screenshots @lalver1 that you don't have any "Statically forwarded" ports.

In mine, I have two.

image

I've realized they correspond with the `ports` being published in `compose.yml` for [`dev`](https://github.com/cal-itp/benefits/blob/main/compose.yml#L23-L24).

If I comment out `ports` for `dev`, they go away. Then, if I start my app through the debugger, I see I can not reach it when I go to `http://localhost:11369` in my browser, which makes sense because I'm not forwarding the port from the devcontainer to my host.

image

(Side note: I'm not sure why I have two since we're just publishing one port, but I guess they are for the same port number, just with different hosts.)

I'm not sure why your VS Code isn't showing you the ports corresponding to the published ports for `dev`, but clearly they are being forwarded for you since you can reach the app at `http://localhost:11369`.

angela-tran:
I was also curious about the "Auto Forwarded" ports that y'all have, and I've learned that "auto-forwarding" is referring to a feature in VS Code where the terminal will detect ports being allocated and set up forwarding for that port.

I didn't find this in any official documentation, but rather from piecing together various comments online:
- Here, someone is asking how the feature works: https://github.com/microsoft/vscode/issues/143958
- Here's another comment confirming that it's a thing: https://github.com/microsoft/vscode/issues/108346#issuecomment-706049303
- Someone asking how to turn off auto-forwarded ports: https://stackoverflow.com/questions/65496520/vscode-remote-dev-django-avoiding-forwarding-ports
- Some more info on configuring auto-forward behavior: https://github.com/microsoft/vscode/issues/121780

I think auto-forwarding is what's happening when we click the `Starting development server at http://0.0.0.0:8000/` link in the VS Code terminal and see a new port in the Ports tab.

Maybe the extra port that would show up and cause hanging is from auto-forwarding?

angela-tran:
I noticed I don't get auto-forwarded ports :thinking:  and then I found that in my User Settings, I have it turned off:

Screenshot from 2025-02-07 11-29-42

image

I don't remember doing this, but in my Firefox history, I see I searched for "vs code keeps adding auto forwarded port" on June 5, 2024, so maybe I did change it? :rofl: 

image

thekaveman:
Oh this seems relevant too... from that same settings pane:

image

angela-tran:
With what we know now, does @lalver1's solution seem ok? I guess there are multiple places where the behavior can be changed (Settings vs. `devcontainer.json`)

thekaveman:
> With what we know now, does @lalver1's solution seem ok? I guess there are multiple places where the behavior can be changed (Settings vs. `devcontainer.json`)

Oh yeah I think it seems OK, but also maybe we should disable this automatic forwarding? Seems like it doesn't matter (you've had it disabled) and it also seems like it could be the source of the some of the problems @lalver1 and @machikoyasuda experience on Mac.

lalver1:
Thanks for all the testing and helpful comments @thekaveman @angela-tran @machikoyasuda !

PR #2659, opened 2025-02-03
Test: introduce admin conftest to use for testing `add` permissions

angela-tran:
This PR extracts out the `conftest.py` from #2654. This is so other PRs can be worked on to add and test the `add` permissions for `ClaimsProvider` and `TransitAgency` without having to wait for #2654.

Some guidance for the other PRs:
- Continue using `model_AdminUser`. We can look into switching to `admin_user` as a separate PR after these PRs that add the permissions.
- No need to add or test any `"regular"` user scenarios


thekaveman (approved):
:+1:

PR #2667, opened 2025-02-05
Feat: field/add permissions for Transit models

### LINKED ISSUE #2665: Admin: Cal-ITP Benefits staff user can add `Transit` models
This is for breaking up the part of #2638 that sets visible/nonvisible and editable/readonly permissions on the fields of `TransitAgency` and `TransitProcessor` models for `superuser` and `Cal-ITP` group members, and lets `Cal-ITP` group members add `TransitAgency` models.

### `superuser`

- [ ] Confirm `TransitAgency` and `TransitProcessor` models can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `TransitAgency` model can be added AND saved

lalver1:
Closes #2665

This PR sets visible/nonvisible and editable/readonly permissions on the fields of `TransitAgency` and `TransitProcessor` models for `superuser` and `Cal-ITP` group members, and lets `Cal-ITP` group members add `TransitAgency` models.

## Reviewing

Set `DJANGO_ALLOWED_HOSTS=benefits.calitp.org,localhost` in your `.env` file (and don't forget to `source .env` in the dev container's terminal) so that you can see the add `TransitProcessor` and `TransitAgency` model behavior as it would run on `prod`. Otherwise, you will always be able to add the `TransitProcessor` model because of

```python
def has_add_permission(self, request):
        if settings.RUNTIME_ENVIRONMENT() != settings.RUNTIME_ENVS.PROD:
            return True
```

### `superuser`

- [ ] Confirm all fields of `TransitAgency` and `TransitProcessor` models are visible and editable as before
- [ ] Confirm `TransitAgency` and `TransitProcessor` models can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `TransitAgency` model can be added AND saved

### DIFF
--- a/benefits/core/admin/transit.py
+++ b/benefits/core/admin/transit.py
@@ -1,48 +1,74 @@
+from django.conf import settings
 from django.contrib import admin
 
 from benefits.core import models
+from .users import is_staff_member_or_superuser
 
 
 @admin.register(models.TransitAgency)
-class TransitAgencyAdmin(admin.ModelAdmin):  # pragma: no cover
+class TransitAgencyAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
+        fields = []
+
         if not request.user.is_superuser:
-            return [
-                "eligibility_api_private_key",
-                "eligibility_api_public_key",
-                "transit_processor_client_id",
-                "transit_processor_client_secret_name",
-                "transit_processor_audience",
-            ]
-        else:
-            return super().get_exclude(request, obj)
+            fields.extend(
+                [
+                    "eligibility_api_private_key",
+                    "eligibility_api_public_key",
+                    "sso_domain",
+                    "transit_processor_audience",
+                    "transit_processor_client_id",
+                    "transit_processor_client_secret_name",
+                ]
+            )
+
+        return fields or super().get_exclude(request, obj)
 
     def get_readonly_fields(self, request, obj=None):
+        fields = []
+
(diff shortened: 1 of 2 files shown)


benefits/core/admin/transit.py
```diff
@@ -1,48 +1,74 @@
...
 
 
 @admin.register(models.TransitProcessor)
-class TransitProcessorAdmin(admin.ModelAdmin):  # pragma: no cover
-    def get_exclude(self, request, obj=None):
-        if not request.user.is_superuser:
-            return []
-        else:
-            return super().get_exclude(request, obj)
+class TransitProcessorAdmin(admin.ModelAdmin):
```
thekaveman, line 52: I was reviewing this model in the Admin UI and thought... a Cal-ITP Staff probably doesn't need to see the fields we currently have as read-only for them (`card_tokenize_*` and `api_base_url`). It's more visual noise that they cannot act on anyway.

I updated the spreadsheet, can we exclude these for all but `superuser`?

thekaveman (changes requested):
One minor request

benefits/core/admin/transit.py
lalver1: Yep, that makes sense, I updated the permissions in 6bc5f536c583ae7d0e1e44fa5f1c97325778a6b6

angela-tran:
Part of #2201

PR #2671, opened 2025-02-06
Feat: field/add permissions for ClaimsProvider model

### LINKED ISSUE #2663: Admin: Cal-ITP Benefits staff user can add `Claims` models
This is for breaking up the part of #2638 that sets visible/nonvisible and editable/readonly permissions on the fields of `ClaimsProvider` for `superuser` and `Cal-ITP` group members, and lets `Cal-ITP` group members add instances of the `ClaimsProvider` model.

### `superuser`

- [ ] Confirm instances of `ClaimsProvider` can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `ClaimsProvider` model can be added AND saved

lalver1:
Closes #2663

This PR sets visible/nonvisible and editable/readonly permissions on the fields of `ClaimsProvider` for `superuser` and `Cal-ITP` group members, and lets `Cal-ITP` group members add instances of the `ClaimsProvider` model.

## Reviewing

Set `DJANGO_ALLOWED_HOSTS=benefits.calitp.org,localhost` in your `.env` file (and don't forget to `source .env` in the dev container's terminal) so that you can see the add `ClaimsProvider` model behavior as it would run on `prod`. Otherwise, you will always be able to add the `ClaimsProvider` model because of

```python
def has_add_permission(self, request):
        if settings.RUNTIME_ENVIRONMENT() != settings.RUNTIME_ENVS.PROD:
            return True
```

### `superuser`

- [ ] Confirm all fields of `ClaimsProvider` are visible and editable as before
- [ ] Confirm `ClaimsProvider` can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `ClaimsProvider` model can be added AND saved

### DIFF
--- a/benefits/core/admin/claims.py
+++ b/benefits/core/admin/claims.py
@@ -2,7 +2,6 @@ from django.conf import settings
 from django.contrib import admin
 
 from benefits.core import models
-from .users import is_staff_member_or_superuser
 
 
 @admin.register(models.ClaimsProvider)
@@ -33,7 +32,7 @@ class ClaimsProviderAdmin(admin.ModelAdmin):
     def has_add_permission(self, request):
         if settings.RUNTIME_ENVIRONMENT() != settings.RUNTIME_ENVS.PROD:
             return True
-        elif request.user and is_staff_member_or_superuser(request.user):
+        elif request.user and request.user.is_superuser:
             return True
         else:
             return False
--- a/benefits/core/admin/claims.py
+++ b/benefits/core/admin/claims.py
@@ -1,23 +1,39 @@
+from django.conf import settings
 from django.contrib import admin
 
 from benefits.core import models
+from .users import is_staff_member_or_superuser
 
 
 @admin.register(models.ClaimsProvider)
-class ClaimsProviderAdmin(admin.ModelAdmin):  # pragma: no cover
+class ClaimsProviderAdmin(admin.ModelAdmin):
     def get_exclude(self, request, obj=None):
+        fields = []
+
         if not request.user.is_superuser:
-            return ["client_id_secret_name"]
-        else:
-            return super().get_exclude(request, obj)
+            fields.extend(["client_id_secret_name"])
+
+        return fields or super().get_exclude(request, obj)
 
     def get_readonly_fields(self, request, obj=None):
+        fields = []
+
         if not request.user.is_superuser:
-            return [
-                "sign_out_button_template",
-                "sign_out_link_template",
-                "authority",
-                "scheme",
-            ]
+            fields.extend(
+                [
+                    "sign_out_button_template",
+                    "sign_out_link_template",
+                    "authority",
+                    "scheme",
+                ]
+            )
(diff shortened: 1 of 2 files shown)


thekaveman:
Small comment as I'm reviewing these:

> Set `DJANGO_ALLOWED_HOSTS=benefits.calitp.org,localhost` in your `.env` file (and don't forget to `source .env` in the dev container's terminal) so that you can see the add `ClaimsProvider` model behavior as it would run on `prod`. 

Thanks for this reminder! Just wanted to point out another way to do this for local reviews: edit your `.vscode/launch.json` file

```jsonc
{
  "name": "Django: Benefits Client",
  "type": "python",
  "request": "launch",
  "program": "${workspaceFolder}/manage.py",
  "args": ["runserver", "--insecure", "0.0.0.0:8000"],
  "django": true,
  "env": {
    "DJANGO_DEBUG": "true",
    "PYTHONWARNINGS": "default",
    "DJANGO_ALLOWED_HOSTS": "benefits.calitp.org,localhost"  //<--- add any env overrides here
  }
}
```

Then just `F5` to have the new environment on launch.

thekaveman (commented):
As I'm reviewing, I'm having second thoughts on these permissions as well (see also https://github.com/cal-itp/benefits/pull/2667#discussion_r1943841586).

I don't know if it makes sense for Cal-ITP staff to create/edit these models? Given they can't edit any of the important fields, creating a nearly empty instance seems a little meaningless? 

Read access seems fine. But it will always take a superuser to fully configure these model instances.

lalver1:
I agree @thekaveman, since  Cal-ITP staff can only edit `client_name`, there's not much use in allowing them to create instances. So I removed the add permission, only superusers can add fad041a377af69ee0c7a515dd1b2b4a43fe7c0a2
However, once a superuser adds a `ClaimsProvider`, Cal-ITP staff can still edit `client_name`. This could be useful, for example, if the list of `claimsProvider`s were to become longer and staff want to group `ClaimsProvider`s by adopting some naming convention. Read permissions were kept the same.

angela-tran:
### `superuser`

- [x] Confirm all fields of `ClaimsProvider` are visible and editable as before
- [x] Confirm `ClaimsProvider` can be added as before

### `Cal-ITP` group member

- [x] Confirm correct fields are visible/nonvisible, and editable/readonly
- [x] Confirm new instances of the `ClaimsProvider` model can be added AND saved

angela-tran:
> I don't know if it makes sense for Cal-ITP staff to create/edit these models? Given they can't edit any of the important fields, creating a nearly empty instance seems a little meaningless?

Great catch here. I think what happened is #2201 was written when `scope` and `claim` were still on `ClaimsProvider`, so at the time, it did make some sense to allow staff members to add this model. Then when we moved the fields in https://github.com/cal-itp/benefits/issues/2237, we never revisited the acceptance criteria, so I'm glad you thought about this!

PR #2673, opened 2025-02-06
Feat: field/add permissions for PemData model

### LINKED ISSUE #2666: Admin: Cal-ITP Benefits staff user cannot add `PemData` models
This is for breaking up the part of https://github.com/cal-itp/benefits/pull/2638 that ensures staff members cannot add `PemData` models. See https://github.com/cal-itp/benefits/pull/2638/commits/0995c0e1fbb90ce8816126a1b87e6924cd775241.

Based on the [spreadsheet](https://docs.google.com/spreadsheets/d/1lPp_LJTfJoWCyOkRC4RaljABGG_jPWOgm4gaAHJigMU/edit?gid=2013104124#gid=2013104124), we do not want staff members to see anything from `PemData`s or be able to add them.

Let's make it explicit by implementing `has_view_permission` and `has_add_permission`.

### `superuser`

- [ ] Confirm instances of `PemData` can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `PemData` model cannot be added

lalver1:
Closes #2666

This PR sets visible/nonvisible and editable/readonly permissions on the fields of `PemData` for `superuser` and `Cal-ITP` group members, and does not allow `Cal-ITP` group members to add instances of the `PemData` model.

## Reviewing

### `superuser`

- [ ] Confirm all fields of `PemData` are visible and editable as before
- [ ] Confirm `PemData` can be added as before

### `Cal-ITP` group member

- [ ] Confirm new instances of the `PemData` model cannot be added

### DIFF
--- a/benefits/core/admin/common.py
+++ b/benefits/core/admin/common.py
@@ -1,8 +1,20 @@
 from django.contrib import admin
+from django.http import HttpRequest
 
 from benefits.core import models
 
 
 @admin.register(models.PemData)
 class PemDataAdmin(admin.ModelAdmin):
-    list_display = ("label",)
+
+    def has_view_permission(self, request: HttpRequest, obj=None):
+        if request.user and request.user.is_superuser:
+            return True
+        else:
+            return False
+
+    def has_add_permission(self, request: HttpRequest, obj=None):
+        if request.user and request.user.is_superuser:
+            return True
+        else:
+            return False
--- a/tests/pytest/core/admin/test_common.py
+++ b/tests/pytest/core/admin/test_common.py
@@ -1,0 +1,43 @@
+import pytest
+
+from django.contrib import admin
+
+from benefits.core import models
+from benefits.core.admin.common import PemDataAdmin
+
+
+@pytest.fixture
+def admin_model():
+    return PemDataAdmin(models.PemData, admin.site)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "user_type,expected",
+    [("staff", False), ("super", True)],
+)
+def test_has_view_permission(
+    admin_user_request,
+    admin_model,
+    user_type,
+    expected,
+):
+    request = admin_user_request(user_type)
+
+    assert admin_model.has_view_permission(request) == expected
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "user_type,expected",
+    [("staff", False), ("super", True)],
+)
+def test_has_add_permission(
+    admin_user_request,
+    admin_model,
+    user_type,
+    expected,
+):
+    request = admin_user_request(user_type)
+
+    assert admin_model.has_add_permission(request) == expected


tests/pytest/core/admin/test_common.py
```diff
@@ -0,0 +1,56 @@
...
+
+    readonly = admin_model.get_readonly_fields(request)
+
+    assert set(readonly) == set(expected)
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize(
+    "user_type,expected",
+    [("super", True)],
```
thekaveman, line 46: We should probably have at least one `False` test case here right?

thekaveman (changes requested):

angela-tran:
Some of us already know this, but just to state it here for reference: 

Before this PR, it was already true that staff members could not add or even see `PemData`s. This is because the Cal-ITP group does not have the `Can add pem data` or `Can view pem data` permission objects assigned to it.

The value of this PR is that

1. we make this behavior more explicit in code, rather than relying on the group having the correct permissions in the database
2. we have unit tests to assert the expected behavior

So before and after this PR, there is actually no difference to the Cal-ITP staff member's user experience.

benefits/core/admin/common.py
```diff
@@ -1,8 +1,31 @@
...
+
+        return fields or super().get_exclude(request, obj)
+
+    def get_readonly_fields(self, request, obj=None):
+        fields = []
+
+        if not request.user.is_superuser:
+            fields.extend(["label"])
+
+        return fields or super().get_readonly_fields(request, obj)
```
angela-tran, line 25: Based on the spreadsheet, only Admins should be able to see anything from `PemData`s: 

image

So I don't think we should be specifying any fields as read-only. And then if all fields are just going to be excluded, it's probably better to not show the objects at all to non-superusers.

I suggest we replace this with an implementation of `has_view_permission` that makes explicit that non-superusers should not have view permission to this model.

angela-tran (changes requested):

angela-tran:
What a Cal-ITP staff member sees on `main` currently:

No `PemData`s are visible.

image

tests/pytest/core/admin/test_common.py
lalver1: Yep, I overlooked having a `staff` `user_type` yet we need to test that case too, thanks for the reminder! It's now here https://github.com/cal-itp/benefits/blob/3b315b26d5c672323cdefe4e912525e0c5eb8eb4/tests/pytest/core/admin/test_common.py#L33

benefits/core/admin/common.py
```diff
@@ -1,8 +1,20 @@
 from django.contrib import admin
+from django.http import HttpRequest
 
 from benefits.core import models
 
 
 @admin.register(models.PemData)
 class PemDataAdmin(admin.ModelAdmin):
-    list_display = ("label",)
```
angela-tran, line 8: _(This is nitpicky, but important for consistency)_

Django admin by default uses `label` in the table, so the table will display the same either way.

What superusers see:

| With this line | Without this line |
| --- | --- | 
|image  |image |

But since we're adopting a philosophy of being explicit even if it's redundant, I think it'd be better to keep this line.

angela-tran (changes requested):

benefits/core/admin/common.py
lalver1: Thanks for the suggestion! I agree that it is more explicit and simpler to implement `has_view_permission` instead of using both `get_exclude` and `get_readonly_fields`.

~I also removed `list_display = ("label",)` from `class PemDataAdmin(admin.ModelAdmin)` since `PemData` already implements `__str__` in https://github.com/cal-itp/benefits/blob/main/benefits/core/models/common.py#L72, so it seemed redundant to still have it per the [docs](https://docs.djangoproject.com/en/5.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display).~ Ah, yes, I agree about consistency, I'll make this quick change following https://github.com/cal-itp/benefits/pull/2673#discussion_r1953399793

benefits/core/admin/common.py
angela-tran: Hmm, just checked through our various `ModelAdmin`s... not all of them define a `list_display`. It makes sense to be explicit even if redundant for the permissions since that logic is complex and varies across user types, but maybe for display it's not as important.

lalver1: That sounds good @angela-tran! If we decide otherwise, I'm happy to bring it back and make it explicit.

angela-tran:
:smile: Thanks for tolerating the back and forth on `list_display` @lalver1. I'm fine with either way. Since @thekaveman will need to approve this PR, maybe he will have an opinion on this that can help us decide.

PR #2674, opened 2025-02-07
Refactor: simplify Agency Index templates

angela-tran:
This PR refactors the `agency_index` view so that it uses a single template. It uses a `context` dictionary from the `agency` to provide the template with copy.

### Testing locally
- Check out this branch - `git checkout refactor/agency-index-context`
   - (Part of the `0034` migration is to populate the new field with each transit agency's slug)
- **Benefits**: Try out the Agency Index in English and Spanish - i.e. go to `/cst`
- **Benefits Admin**: Login in as a superuser, and notice that on your `TransitAgency`s, you can choose what "Index copy" to use
   - You can try out choosing a different option from the dropdown, saving the transit agency, and then going to `/cst` again to see that the copy has changed. (You can change the slug too if that makes the scenario feel more realistic.)
- **Benefits Admin**: Login in as a Cal-ITP staff member, and notice that on your `TransitAgency` you can see what copy it is using -- (maybe we should change this to be editable by staff members)

### Screenshots

**Benefits**

- English
Screenshot from 2025-02-06 20-27-16

- Spanish
Screenshot from 2025-02-06 20-27-21

**Benefits Admin**

benefits - slug choices


angela-tran:
@thekaveman @machikoyasuda @lalver1 -- curious on your early thoughts and feedback. I see a number of things that could be done slightly differently, but rather than me continuing to tweak at it, I thought I'd just see what you think about this as-is.

angela-tran:
> a number of things that could be done slightly differently

To be a little more specific, I see some UX things, naming, and DevX things that I might want to change

thekaveman (commented):
Overall I'm liking this approach! 

Some different thoughts:

* Defining the `dict(self)` method on `AgencyIndex` to simply wrap and return  `asdict(self)` seems a little... extra? I see wanting to avoid the `from dataclasses import asdict` everywhere that uses this context. Maybe a base context class to implement `dict(self)` that can be shared for all of these? Not a huge concern, just a note.

* I was looking at the docs for `dataclasses`; for these cases where we are essentially hard-coding our config, maybe we look at the [`frozen=True`](https://docs.python.org/3/library/dataclasses.html#frozen-instances) parameter. I assume we don't want callers modifying these once they are setup.

* We started calculating template names using the `slug`... I wonder if we can do the same thing here. E.g. rather than new model fields for `*_context_key`, we just use the `slug`? Maybe `slug` becomes the `choices` field and everything just flows out of that. Would allow us to delete a bunch of fields if we keep going with this approach.

angela-tran:
> * We started calculating template names using the `slug`... I wonder if we can do the same thing here. E.g. rather than new model fields for `*_context_key`, we just use the `slug`? Maybe `slug` becomes the `choices` field and everything just flows out of that. Would allow us to delete a bunch of fields if we keep going with this approach.

I was thinking the same thing! I could see it working really nicely that way

thekaveman:
> > * We started calculating template names using the `slug`... I wonder if we can do the same thing here. E.g. rather than new model fields for `*_context_key`, we just use the `slug`? Maybe `slug` becomes the `choices` field and everything just flows out of that. Would allow us to delete a bunch of fields if we keep going with this approach.
> 
> I was thinking the same thing! I could see it working really nicely that way

Can we prioritize this small change for this PR, and maybe (or not) work on the other points in follow-up work?

lalver1:
I like this approach to managing copy! 👍 

Just two thoughts that came to my mind:

1. Related to the template 'override' capability we currently have: I don't think we'd be in a situation where this would be a requirement for the app, but if we need to change the copy _and_ layout of a section of a page, the approach in this PR would still work. We could insert a conditional in the `return` of a view and hard code the overriding template (we'd want to avoid this). Or we could add an extra template in between `index-base` and `index--agency` with the layout code inside a `block`. Either way, it would only be one extra template file, so we still achieve the goal of this effort which is to reduce the overall number of templates.
2. Related to how the context may get built: Again, I'm not sure if we have this situation, but I guess we could have a more complex context object that needs data from different models. This is doable for sure, just a recognition that it seems that we'd be trading off complexity of logic (in some `context.py` files) vs number of templates, which is totally fine.

angela-tran:
@thekaveman 

> Can we prioritize this small change for this PR, and maybe (or not) work on the other points in follow-up efforts?

Done in [fa8d60f](https://github.com/cal-itp/benefits/pull/2674/commits/fa8d60f281e58b3bc63b8d7811859283a03945f0) with [0b60cfc](https://github.com/cal-itp/benefits/pull/2674/commits/0b60cfcf0296495bc4e72da190556ef139df5553) as a clean-up to naming / display values.

Updated the screenshot with a GIF in the [PR description](https://github.com/cal-itp/benefits/pull/2674#issue-2837046157)

benefits/core/context.py
```diff
@@ -0,0 +1,32 @@
...
+class AgencySlug(models.TextChoices):
+    # raw value, display value
+    CST = "cst", "cst"
+    MST = "mst", "mst"
+    NEVCO = "nevco", "nevco"
+    SACRT = "sacrt", "sacrt"
+    SBMTD = "sbmtd", "sbmtd"
+
+
+index_context = {
```
thekaveman, line 24: Some thoughts on naming (_can be for later_):

The module name of `context` does not feel quite descriptive enough when considered with the existing `context_processors` module and the wide use of the term in Django. 

For example, there are some [views where we construct a `context` variable](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L31) for the template, a name that would clash with this module name in e.g. `from benefits.core import context`

A module name that includes a component like `copy` or `msgids` or similar, could help by being more descriptive about what this specific set of context is for, and to avoid clashes elsewhere.

Separately, there's an argument for not duplicating `_context` in these module-level variable names. If the pattern is going to be something like:

```py
from benefits.module import context

# ...

context.page_context[key]
```

The extra `_context` instead of the shorter `context.page[slug]` seems unnecessary.

benefits/core/models/transit.py
```diff
@@ -160,8 +159,9 @@ def __str__(self):
         return self.long_name
 
     @property
-    def index_template(self):
-        return self.index_template_override or f"core/index--{self.slug}.html"
+    def index_context(self):
+        key = context.AgencySlug(self.slug)
+        return context.index_context[key].dict()
```
thekaveman, line 164: This construction of the key for the context lookup seems a little strange. I would expect `self.slug` to be used directly?

Looking at https://docs.djangoproject.com/en/5.1/ref/models/fields/#enumeration-types

> Note that using `YearInSchool.SENIOR`, `YearInSchool['SENIOR']`, or `YearInSchool('SR')` to access or lookup enum members work as expected...
> ...
> If you don’t need to have the human-readable names translated, you can have them inferred from the member name (replacing underscores with spaces and using title-case)

A couple of small refactors that could help simplify:

```python
# simpler choices definition
class AgencySlug(models.TextChoices):
    # display value = raw value
    CST = "cst"
    MST = "mst"
    NEVCO = "nevco"
    SACRT = "sacrt"
    SBMTD = "sbmtd"

# use value for context key
index_context = {
  AgencySlug.CST.value: _("Content..."),
  # others...
}

# later...

# lookup is easier
context.index_context[agency.slug]
```

thekaveman (approved):
Additional thoughts, but non-blocking. Nice start of this idea!

angela-tran:
Thanks for those notes @thekaveman! I agree with those points and that we can work on them as follow-up.

PR #2688, opened 2025-02-13
Admin: Front-end for Flow pages

machikoyasuda:
closes #2634 

WIP

## General idea
- Extract `flow-base` as a template for the whole flow
- Move all pages into that, including error page
- Use default Bootstrap color class namings as much as possible
- Then, after the page is all laid out and all pages in flow are aligned, then edit the Bootstrap CSS Variables to make the secondary, secondary-subtle colors be the right grays.
- Ensure the other pages are not affected

Goals: Do not create any new classes and try to do default overrides as much as possible. Do not declare styles on existing classes.

## Notes
- Re-doing the header to the new design is out of scope for this ticket.
- Re-doing the button CSS is also out of scope for this ticket.

image

Issue #2553, opened 2024-12-02
Go live checklist for Nevada County


thekaveman:
This task documents a two-phase go live for Nevada County.

## Phase 1: Validation testing

This configuration occurs in `test`.

- [x] [Link additional products to sample group](https://github.com/cal-itp/benefits/issues/2553#issuecomment-2515814813)
- [x] Create a new `TransitProcessor` instance with details for the _live_ system
- [x] Ensure the Nevada County `nevco-payment-processor-client-secret` is updated
- [x] Ensure the Nevada County agency is up to date with this `TransitProcessor` reference and config data
- [x] Ensure a flow for `Older Adult` is configured for Nevada County, see [group details below](https://github.com/cal-itp/benefits/issues/2553#issuecomment-2515814813)
- [x] Confirm that you can successfully go through the flows with sample user and _real payment card details_

## Phase 2: Production

This configuration occurs in `test`.

- [x] Update the Nevada County `nevco-payment-processor-client-secret`, go back to the QA value
- [x] Disassociate the _live_ `TransitProcessor` from the Nevada County agency, go back to the original QA `TransitProcessor`
- [x] Update the `Older Adult` flow to remove the sample group ID, go back to group ID from QA `1ad3c030-70b6-4948-b417-40d332c58a96`

This configuration occurs in `prod`.

- [x] Create the Nevada County agency; do not mark as Active yet
- [x] Ensure a flow for `Older Adult` is configured for Nevada County, [group details below](https://github.com/cal-itp/benefits/issues/2553#issuecomment-2522117799)
- [x] Ensure a flow for `Medicare` is configured for Nevada County, [group details below](https://github.com/cal-itp/benefits/issues/2553#issuecomment-2522117799)
- [x] Activate Nevada County

o-ram:
For testing purposes, a `Test Group` has been set up in the Littlepay production portal and the "senior" product suite will be associated to the group.

**Test Group ID**: `1ab565da-a617-4cd0-ac5c-6167540bc274`

The following Product IDs below should be associated to it.

**Discount Product ID**: `09d8bef8-3f44-48ab-8e60-e8a5b1f3d2e2`
**Discount Time cap ID**: `f61eb087-5b4e-447f-ba8e-48852e4e826f`
**Discount Day cap ID**: `2ff9ea8c-73c3-4667-bc33-81b8ee1a2bd5`
**Discount Week cap ID**: `e485db41-9150-43d7-b573-bc14d4eb0244`
[...]

PR #2689, opened 2025-02-13
Feat: in-person eligibility policies

### LINKED ISSUE #2632: Admin: In-Person - View policy
This is a functionality ticket.

## Acceptance criteria

- [ ] After clicking on an in-person enrollment eligibility pathway, the correct logic appears
- [ ] Nothing is clicked by default
- [ ] Changing pathway changes the logic that appears

Image

## Definition of "Done"

An admin user must get a logic warning and confirm it before proceeding. Do not hard-code anything, must be coded to work with any new agencies and flows.

Design does not matter yet. Get the functionality going first.

## Links to user research, designs and resources

[](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/Cal-ITP-Admin?node-id=154-3052&p=f&m=dev)

Please research/test for any a11y implications from adding code to the way the radio buton functions.

angela-tran:
Closes #2632 

This PR implements the functionality where, when verifying eligibility in-person, the user must agree they have used the specific eligibility criteria that is shown.

## Screen recordings of behavior

_(GIFs play only once. You can refresh the page to replay I think.)_

### User-facing behavior (for transit agency staff user)

<details><summary>Checkbox changes based on selected flow</summary>

benefits-in-person-dynamic-checkbox-edited

</details>

<details><summary>Checkbox resets if user changes selected flow</summary>

benefits-in-person-checkbox-reset-edited

</details>

<details><summary>Browser validation messages</summary>

benefits-in-person-browser-validation-edited

</details>

### User-facing behavior (for superusers and Cal-ITP staff users)

- We currently only have in-person policies for the Older Adult, Medicare Cardholder, and Courtesy Card flows, so any other flows that were marked as supporting `In-person` enrollment will be automatically updated to have that option unchecked.
- When attempting to add `In-person` as a supported enrollment method, that will only work for the three flows mentioned above. For any other flows, an error message will be shown:

<details><summary>Error message when attempting to add `In-person` to a flow without a policy</summary>

Screenshot from 2025-02-21 15-12-59

</details>

### [Data migration](https://github.com/cal-itp/benefits/pull/2689/files#diff-567073f571cde41a26ac0d8ade9a4217f9dc1d0c0f436fdcd7e173f4e82113e9)

- Start on `main` branch, load sample fixtures with `./bin/reset_db.sh`
   - Launch app, go to `/admin`, and log in as `cst-user`. At `/in_person/eligibility`, you'll see the 5 in-person flows for CST.
   - Log out and log back in as a superuser. Look over your `EnrollmentFlow`s and notice the `In-person` checkbox is checked.
- Checkout this branch (`git checkout feat/in-person-policies`) and run the migration with `./bin/init.sh`
   - Launch app, go to `/admin`, and log in as `cst-user`. At `/in_person/eligibility`, you'll see only "Older Adult" and "Medicare Cardholder".
   - Log out and log back in as a superuser. Looking at the `EnrollmentFlow`s, only the "Older Adult" and "Medicare Cardholder" flows have `In-person` checked.

### [`EnrollmentFlow` validation](https://github.com/cal-itp/benefits/pull/2689/files#diff-396f633e76d6f72ae9c1c006143cd627138a10b7773649831c0db29c8dabf627R286-R293) enforces that in-person flows have a policy
- Try to check `In-person` for a flow and save it. If it is not one of the flows with an in-person eligibility policy, you will get an error message.

### a11y

I tested going through the form with only keyboard interaction, and the tab order / interactions worked as I expected. 

I also inspected the accessibility tree and observed that when the checkboxes go from hidden to shown, the accessibility tree makes sense, i.e. it only shows the checkbox that is currently displayed.


benefits/core/models/enrollment.py
```diff
@@ -10,6 +10,8 @@
 from .common import PemData, SecretNameField, template_path
 from .claims import ClaimsProvider
 from .transit import TransitAgency
+from benefits.core.context import SystemName
+from benefits.in_person.context import eligibility
```
angela-tran, line 14: I know we're wanting to discuss the naming for this copy context stuff. For now, I am just following the existing naming pattern but am aware that we don't necessarily like it.

benefits/in_person/context.py
```diff
@@ -0,0 +1,26 @@
...
+    def dict(self):
+        return asdict(self)
+
+
+eligibility = {
+    SystemName.OLDER_ADULT.value: EligibilityIndex(
+        policy_details="I confirmed this rider’s identity using a government-issued ID and verified they are age 65 or older."
+    ),
+    SystemName.MEDICARE.value: EligibilityIndex(
+        policy_details="I confirmed this rider’s identity using a government-issued ID and verified they posses a valid "
```
angela-tran, line 19: Copied this from Figma. Just noticed the spelling error in `posses` - should be `possess`

benefits/core/models/enrollment.py
angela-tran: We discussed this today in dev workshop and came up with a plan! See #2697

benefits/core/models/enrollment.py
```diff
@@ -257,8 +258,12 @@ def enrollment_success_template(self):
         else:
             return self.enrollment_success_template_override or f"{prefix}--{self.transit_agency.slug}.html"
 
+    @property
+    def in_person_eligibility_context(self):
+        return in_person_context.eligibility_index[self.system_name].dict()
```
angela-tran, line 263: @thekaveman I rewrote this so that it is non-defensive and simply looks up the key. Consumers of this property are responsible for knowing what to do, whether that's to let the exception bubble up or to handle it.

Here's the diff: https://github.com/cal-itp/benefits/pull/2689/commits/21e0f69d52d7331e5cce87d1be15b2b2e9fce5a2

That commit shows how `InPersonEligibilityForm` consumes the property. [`EnrollmentFlow.clean`](https://github.com/cal-itp/benefits/pull/2689/commits/d27a53e995573b6213412f8f63711b5d102c3ebf#diff-396f633e76d6f72ae9c1c006143cd627138a10b7773649831c0db29c8dabf627R286-R290) consumes it in a similar way.

benefits/core/models/enrollment.py
```diff
@@ -276,10 +281,19 @@ def clean(self):
...
+                    errors.append(ValidationError(f"Template not found: {t}"))
+
+            if EnrollmentMethods.IN_PERSON in self.supported_enrollment_methods:
+                try:
+                    in_person_eligibility_context = self.in_person_eligibility_context
+                except KeyError:
+                    in_person_eligibility_context = None
+
+                if not in_person_eligibility_context:
+                    errors.append(ValidationError(f"In-person eligibility context not found for: {self.system_name}"))
```
angela-tran, line 293: Open to suggestions on this error message :sweat_smile: 

I guess it would help if it told the user some sort of action to take, like to contact a dev or to uncheck `In-person`.

thekaveman: Hmmm... yeah, this is a technically accurate error message! But since we show it to end-users, I think we probably want something more friendly here :sweat_smile: 

I agree with your action-oriented take:

> {system_name} not configured for In-person. Please uncheck to continue."

benefits/core/models/enrollment.py
thekaveman: Ah, I see... yeah I don't think this is the specific use-case I was thinking about, and I didn't understand the difference in what we were talking about. I agree this change isn't as clean.

The use-case I was thinking about (related to https://github.com/cal-itp/benefits/pull/2698#issuecomment-2672430934): Why was it possible to load a fixture with a bad `system_name`? I expected this to have failed on import: https://github.com/cal-itp/benefits/pull/2698#issuecomment-2672297744

But I think you described why this is: https://github.com/cal-itp/benefits/pull/2698#issuecomment-2672384526. And then it was fixed anyway by adding `agency_card` to the enum...

Sorry, I mixed up a few different things in that comment. But really I was wondering: why should we even be able to import bad data like that? And does that mean it is somehow possible to get a bad value in there from the Admin? I think both of those are probably answered by your explanation.

benefits/core/models/enrollment.py
lalver1: I just tested this out and I agree with y'all, adding an action to take and saying that the flow is not configured for In-person, something like Kegan's message, works well for the user. And because the code is explicit that this error occurs when `not in_person_eligibility_context`, it's still easy for a dev to see what exactly is missing in the configuration.

thekaveman: Yeah don't worry too much about the specific message here @angela-tran, anything action-oriented for the user and @indexing can always give feedback via the design process as we evolve the interface.

benefits/core/models/enrollment.py
thekaveman: Just to resolve this, the way this is implemented is just fine if you don't want to worry about it anymore :+1:

angela-tran: Thanks, yeah I think this is fine too

angela-tran:
Ready for review again

lalver1 (dismissed):
I went through the code and tested locally and everything looked good, this is a really nice new feature! 👍

This is only a note, that Nevada County Connects is set up in `test` with 2 in-person flows (`disability` and `youth`) with no `policy_details` yet. Seems like the migration in this PR will reset these to `digital` so we just need to remember to reconfigure these 2 flows and add `policy_details` to them after we do an `rc` release. If this makes sense, I can create a ticket for this chore.

thekaveman:
@lalver1 good call! Want to update this ticket? #2553

indexing:
@lalver1 Quick note: we don't have `policy_details` for either `disability` or `youth` yet; those won't be pathways we offer NevCo with the initial release of in-person enrollments.

thekaveman:
@indexing @lalver1 

Ah, so it must have been used by us testing the interface. In any case, sounds like there is nothing to do here at all :+1:

angela-tran:
After rebasing onto `main` which has had #2688 merged into it:

image

angela-tran:
> Seems like the migration in this PR will reset these to `digital` ...

The migration only removes `in_person` from flows and does not set any flow to `digital`.

thekaveman (changes requested):
One minor change: the `local_fixtures.json` should be updated so that when starting from scratch (e.g. no database, like a fresh local environment), we don't load flows marked for in-person that aren't in the context. 

This happens after the database migration here, so they are loaded in an invalid state.

angela-tran:
> One minor change: the `local_fixtures.json` should be updated so that when starting from scratch (e.g. no database, like a fresh local environment), we don't load flows marked for in-person that aren't in the context.
> 
> This happens after the database migration here, so they are loaded in an invalid state.

Thanks for catching that!

thekaveman:
> Thanks for catching that!

NP. But I guess I also didn't realize, we aren't supporting Agency Card here for CST? Would we want that for testing/demoing?

thekaveman (approved):
:identification_card: :mag: 

:white_check_mark:

Issue #2679, opened 2025-02-11
OAuth refactor: remove `benefits.oauth` and replace with `django-cdt-identity`


thekaveman:
[`django-cdt-identity`](https://github.com/Office-of-Digital-Services/django-cdt-identity) will be the new codebase for common CDT Identity Gateway integration and claims verification implementation.

This issue documents the tasks necessary to replace this repository's usage of the [`benefits.oauth`](https://github.com/cal-itp/benefits/tree/main/benefits/oauth) module with `django-cdt-identity`

## Additional context

Related: https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/issues/19

Plan for how to make `django-cdt-identity` extensible for usage in e.g. Benefits: Office-of-Digital-Services/django-cdt-identity#13

thekaveman:
Task for `2025 - Sprint 02`: shape this ticket

angela-tran:
We've finished all the tickets on this! 🙌

PR #2690, opened 2025-02-14
Fix: OAuth exception is not overwritten

lalver1:
ClaimsProvider (oauth_client.authorize_redirect throws an exception)
is not overwritten and is the one that gets bubbled further.


thekaveman (approved):
Normally I would say we should add a test for this.

But since we'll be deleting most of this code as part of #2679, this is fine :+1:

PR #2691, opened 2025-02-14
Feat: add Discover as an accepted bank card

lalver1:
Closes #2680

This PR needs to be merged when we are cleared to do so by Product, after Littlepay's Discover compatibility begins. It updates English and Spanish copy across the user-facing flow and the Help page to include Discover as an accepted bank card. The copy is located at the following locations:
- Help page
- Eligibility start
- Contactless modal
- Enrollment index
- CalFresh alert

PR #2696, opened 2025-02-20
Chore: run pre-commit on all files

angela-tran:
We're not sure what changed to cause the prettier pre-commit hook to fail on `styles.css` and `variables.css`. It keeps getting re-run on our PRs like #2691 and #2689 and [even on merges into `main`](https://results.pre-commit.ci/run/github/285501392/1739894951.TJKhBB2dS7e70pbofhCPNA), so let's just apply the re-formatting so it stops re-running.

Ran `pre-commit run --all-files`:

### DIFF
--- a/benefits/static/css/styles.css
+++ b/benefits/static/css/styles.css
@@ -122,10 +122,12 @@ a[target="_blank"]::after {
   content: " ";
   background-color: currentColor;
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
omg thank you :sob:

PR #2697, opened 2025-02-20
Refactor: context module into a package

### LINKED ISSUE #2682: Follow-up refinement to copy context classes and module
These are items suggested in #2674. Creating this ticket so we don't forget about them.

~~We can split this up into multiple tickets whenever we get to working on these. For now, just going to list them:~~

- ~~[Data class suggestions](https://github.com/cal-itp/benefits/pull/2674#pullrequestreview-2602281159)~~
- [Naming of the module and dict variables](https://github.com/cal-itp/benefits/pull/2674#discussion_r1952059138)
- [Key for the context dict](https://github.com/cal-itp/benefits/pull/2674#discussion_r1952080068)

angela-tran:
Closes #2682 

This PR addresses:

> Naming of the module... 
 - by introducing a `context` folder that has a [package initialization](https://realpython.com/python-modules-packages/#package-initialization) file (`__init__.py`) and moving the agency-specific context code into an `agency` module within the package

> ... and [naming of the] dict variables
 - by removing `_context` from the module-level variables
 
> Key for the context dict
- by using `.value` when constructing the dictionary


benefits/core/context/agency.py
```diff
@@ -21,12 +21,12 @@ class AgencySlug(models.TextChoices):
     SBMTD = "sbmtd", "sbmtd"
 
 
-index_context = {
-    AgencySlug.CST: AgencyIndex(headline=_("Get a reduced fare on CST public transportation when you tap to ride")),
-    AgencySlug.MST: AgencyIndex(headline=_("Get a reduced fare on MST public transportation when you tap to ride")),
-    AgencySlug.NEVCO: AgencyIndex(
+index = {
```
thekaveman, line 24: Ugh, I guess I'm still a little torn on this variable name. 

We were thinking/talking about something more along the lines of:

```py
from benefits.core import context as core_context

# ...

core_context.agency.index[key]
```

But yeah that extra `.agency.` in the middle obviously won't work. 

What does everyone think about calling this variable `agency_index`? That actually aligns more closely to the route/view name where this is used.

```py
from benefits.core import context as core_context

# ...

core_context.agency_index[key]
```

thekaveman (approved):
This looks great 👍

thekaveman:
I think this can close #2682 since we decided to ignore frozen data classes for now.

PR #2698, opened 2025-02-20
Refactor: choices for EnrollmentFlow.system_name

angela-tran:
Needed by #2689 for #2632

This PR introduces a `SystemName`  choices enum to be used as the possible values for `EnrollmentFlow.system_name`. This will be useful for defining flow-specific `context` dicts .

It's helpful to split this out from #2689 so that I can fix unit tests that are broken by the introduction of `choices` to `EnrollmentFlow.system_name`.

### Benefits Admin user-facing change

The `System name` field on `EnrollmentFlow`s now is a dropdown with these values:

System name dropdown in Benefits Admin

System name dropdown with helper text in Benefits Admin


angela-tran:
Added a screenshot above of user-facing change. (Also, Kegan was correct that Django uses the enum name to generate the display value.)

thekaveman:
@lalver1 

> I don't think we need to do anything here, in our local_fixtures we use `agency_card` as the system name

Wouldn't this cause an exception when that slug is looked up in the context dictionary?

angela-tran:
> ... in our `local_fixtures` we use `agency_card` as the system name for the `Agency Cardholder` `EnrollmentFlow`. 

Ahh, that's a good catch @lalver1 

> This name isn't in the `SystemName` enum but this doesn't cause any trouble when you run `bin/reset_db.sh`, for example. The database still populates with `agency_card` as the system name even though the admin shows dashed lines:

What you're pointing out here is that Django only enforces the `choices` when saving via a `ModelForm` (see `django.forms.ChoiceField`'s [`validate`](https://github.com/django/django/blob/87c5de3b7f2316aa17353d74f54e6ff19013d049/django/forms/fields.py#L899-L907)). I think it would be best practice to create objects in the same way that `ModelForm`s would, so that means our sample fixtures should abide by what the model's fields say. Same goes for the objects and fixtures we create in our pytests.

I suppose we could add logic to the model's `clean` method to mirror the validation that Django does with `ModelForm`s, but not sure if that's actually worth our time and maintenance.

angela-tran:
@thekaveman 

> Wouldn't this cause an exception when that slug is looked up in the context dictionary?

It would depend on how the look-up is written. 

For example, in my WIP PR, the look-up of `in_person_eligibility_context` is defensive and [will return an empty dictionary](https://github.com/cal-itp/benefits/pull/2689/files#diff-396f633e76d6f72ae9c1c006143cd627138a10b7773649831c0db29c8dabf627R264) if the key was not found. 

However, the look-up of `agency_index` assumes that [every `slug` will have an entry](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/transit.py#L163), so that would throw an exception. I think that's desired because we would want to be alerted that we forgot to add copy (I mean, hopefully we would've caught it earlier than that, but still).

Definitely open to a differing opinion on how this should work.

angela-tran:
I'm going to add `agency_card` to our `SystemName` choice enum.

thekaveman:
I think my initial take was, an exception should be thrown because we have a mismatch / invalid config in this case. E.g. a non-defensive lookup.

lalver1:
Thanks @angela-tran and @thekaveman for elaborating on this.

I agree about adding `agency_card` to the  `SystemName` enum, so that

> ... sample fixtures should abide by what the model's fields say

angela-tran:
Ready for review again

angela-tran:
> I think my initial take was, an exception should be thrown because we have a mismatch / invalid config in this case. E.g. a non-defensive lookup.

Makes sense. I'll think about this more in #2689

Issue #2678, opened 2025-02-10
Refactor the sign out with IdG to use client_id instead of id_token_hint


thekaveman:
See details in this Slack thread: https://cal-itp.slack.com/archives/C037Y3UE71P/p1738976616340329

TLDR; we're replacing `id_token_hint` with `client_id` in the `deauthorize_redirect` helper.

This is the relevant section: https://github.com/cal-itp/benefits/blob/main/benefits/oauth/redirects.py#L26

## Acceptance Criteria

- [ ] `id_token_hint` is no longer sent in the `end_session_endpoint` params
- [ ] `client_id` is sent in the `end_session_endpoint` params 

## Additional context

This functionality is deployed to IdG `dev` so we need to coordinate with @johnatstate on deploying this to further environments.

lalver1:
This functionality got deployed to IdG `UAT` and `prod`, and to Benefits in [2025.03.1](https://github.com/cal-itp/benefits/releases/tag/2025.03.1). It is documented in this [release process issue](https://github.com/cal-itp/benefits/issues/2733) and this [Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1741712884309019).

PR #2700, opened 2025-02-21
Refactor: IdG sign out to use client_id

lalver1:
Part of #2678

Due to sign out changes in the IdG, in this PR we're replacing `id_token_hint` with `client_id` in the `deauthorize_redirect` helper that is used as part of the sign out process.

**~We can review this PR but we should hold off from merging it to `main` since this functionality has been deployed to IdG `dev`, but~ we should coordinate with the IdG team before deploying this to the `test` and `prod` environments.**

## Reviewing
Using this branch, set up a login.gov flow, such as Older Adult, and go through the flow until eligibility is confirmed

image

Click on `Sign out of Login.gov` and confirm that you can successfully sign out

image

Compare this behavior with [`dev`](https://dev-benefits.calitp.org/) (or `main`, where this fix has not been deployed yet) and notice these extra pages provided by the IdG

image

image

### DIFF
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -27,7 +27,7 @@ _ENROLLMENT_EXP = "enrollment_expiry"
 _FLOW = "flow"
 _LANG = "lang"
 _OAUTH_CLAIMS = "oauth_claims"
-_OAUTH_TOKEN = "oauth_token"
+_OAUTH_AUTHORIZED = "oauth_authorized"
 _ORIGIN = "origin"
 _START = "start"
 _UID = "uid"
@@ -59,7 +59,7 @@ def context_dict(request):
         _ENROLLMENT_TOKEN: enrollment_token(request),
         _ENROLLMENT_TOKEN_EXP: enrollment_token_expiry(request),
         _LANG: language(request),
-        _OAUTH_TOKEN: oauth_token(request),
+        _OAUTH_AUTHORIZED: oauth_authorized(request),
         _OAUTH_CLAIMS: oauth_claims(request),
         _ORIGIN: origin(request),
         _START: start(request),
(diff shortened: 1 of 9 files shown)


thekaveman:
@lalver1 for this PR, I think we should make sure there isn't any other `token` stuff laying around the codebase, since it was purely used for this sign out. 

E.g. there is no need to keep this in the session anymore: https://github.com/cal-itp/benefits/blob/fix/idg-signout/benefits/oauth/views.py#L147

This clean up may have wider implications, but I think its more important to do it all at once vs. leaving some of it for later (e.g. potentially a different release).

lalver1:
Ah thanks for pointing that out @thekaveman! I agree, it makes sense to clean it up all at once in this PR so I'll work on updating it.

lalver1:
This is ready for another review (mostly session cleanups in 4b1a6a590ba090be64abd344f5885c2147687852 and f8c26bb006f150ce24c4a9e029dad704e42020c1) but if it's ok I'll leave it as `Draft` so that I don't accidentally merge it.

thekaveman (changes requested):
@lalver1 overall this is great! Exactly the kind of low-impact change we talked about :+1: 

One small request: the terminology in this PR now mixes the concepts of [_authentication_](https://en.wikipedia.org/wiki/Authentication) with [_authorization_](https://en.wikipedia.org/wiki/Authorization). 

IMO the role of this session variable is closer to _authorization_; i.e. the user has already proven their identity (_authentication_) by logging in. We are then (via e.g. the middleware) giving access (or not) to additional resources (e.g. views, the other parts of the flow).

So I think just a simple rename from `authentication` to `authorization` and this will be normalized.

lalver1:
Thanks @thekaveman! That's a good point, I'll change the session variable name to `oauth_authorized` and update any docstrings or comments to make sure we normalize to `authorization` 👍

thekaveman:
> > Exactly the kind of low-impact change we talked about
> 
> I take it from this that there is some reason for keeping the session variable as opposed to removing it. It's not clear to me what that reason would be though since we're not using `session.oauth_authorized` anywhere other than tests.

`session.logged_in` is used in a lot of places (middleware, views, templates, etc.), and that previously worked by converting `bool(session.oauth_token)`. 

Although we're removing the token, we still need the concept of `logged_in`.

angela-tran:
I see, it is used in `session.logged_in`. I missed that, thanks

thekaveman:
@lalver1 have you confirmed this timeline with John? We should coordinate the full schedule _before_ merging to `dev` so we aren't blocked on other releases if John isn't available or ready to release on his side.

lalver1:
Since we are planning on making a [release](https://github.com/cal-itp/benefits/issues/2733) today, I'll mark this PR as ready for review. But this PR won't get merged until we actually start the release process.

thekaveman (approved):
Nice job cleaning up all the old `token` stuff too :+1:

PR #2705, opened 2025-02-27
Refactor: simplify Eligibility Index templates

angela-tran:
This PR refactors the `eligibility.index` view so that it uses a single template. It uses a context dictionary from the agency to provide the template with copy.

The [`EligibilityIndex`](https://github.com/cal-itp/benefits/pull/2705/files#diff-861ca9c59ea52eff4e0160e8f87dbe2f97d8370b1649bcee9a6ced268e73a908R35-R43) class and [`eligibility/index.html`](https://github.com/cal-itp/benefits/pull/2705/files#diff-1e3c02710205416114dcc25094079f5cdd0a6d71b504924fe9c6d0d215612ef1R22) template support one or more paragraphs.

## Testing locally
- Checkout this branch - `git checkout refactor/eligibility-index-context`
- Check the Eligibility Index for each agency slug - `cst`, `mst`, `nevco`, `sacrt`, `sbmtd` (see notes below)

### CST

If you're using the sample fixtures, you'll have one `TransitAgency` with slug `cst` and see this Eligibility Index:

:white_check_mark: 

### MST

To see MST's Eligibility Index, you need a `TransitAgency` with slug `mst`. An easy way to do this is to change the slug of your existing `TransitAgency` in the admin interface.

image

<details><summary>The Eligibility Index shows MST copy :white_check_mark: </summary>

image

</details>

### NevCo

We can repeat the same process for the rest of the slugs, but notice that for some slugs, like `nevco`, an error is raised when we go to the Eligibility Index:

```
django.template.exceptions.TemplateDoesNotExist: eligibility/includes/selection-label--nevco-agency-card.html
```

It's trying to load a Nevada County agency card template, but we don't have that template because that agency-flow combination isn't actually supported, so we need to go remove that flow from the agency.
- Go to page for editing `Enrollment flows` 
image
- Click on the "Agency Cardholder" flow
- Set its `Transit agency` field to be blank
image
- Click "Save"

<details><summary>Now we see Nevada County's Eligibility Index :white_check_mark: </summary>

image

</details>

### SacRT

SacRT's Eligibility Index has its two paragraphs of copy :white_check_mark: 

### SBMTD

:white_check_mark:


angela-tran:
## Annoying issue with `%` characters in copy

Some reminders:
- [`makemessages`](https://docs.djangoproject.com/en/5.1/ref/django-admin/#django-admin-makemessages) scans our code and templates for translatable strings and generates PO files
- [`compilemessages`](https://docs.djangoproject.com/en/5.1/ref/django-admin/#django-admin-compilemessages) reads the PO files and compiles them into binary MO files which are used when the app runs

`makemessages` is called when we run the helper [`./bin/makemessages.sh`](https://github.com/cal-itp/benefits/blob/main/bin/makemessages.sh#L6) script. 
`compilemessages` is called whenever the [`./bin/init.sh`](https://github.com/cal-itp/benefits/blob/main/bin/init.sh#L10) script is run, which would be either when you run `./bin/reset_db.sh` or when the app container starts.

### Background
Before this PR, our copy containing `%` characters works fine.

In our templates we have the [translatable string](https://github.com/cal-itp/benefits/blob/2025.02.1/benefits/eligibility/templates/eligibility/index--cst.html#L7), e.g.:
```
# eligibility/index--cst.html
...

{% translate "Cal-ITP doesn’t save any of your information. All CST transit benefits reduce fares by 50% for bus service on fixed routes." %}

...
```

When the `makemessages` command generates a PO entry for this string, it marks the entry as `#, python-format` (because `%` in Python is the old-style [string-formatting operator](https://www.geeksforgeeks.org/string-formatting-in-python/)). It also realizes this `%` is not actually being used for interpolation, so it escapes the `%` character, which is done by putting [another `%`](https://www.geeksforgeeks.org/different-ways-to-escape-percent-in-python-strings/) in front. This results in the following PO entry (notice the `%%`):

```
#, python-format
msgid ""
"Cal-ITP doesn’t save any of your information. All CST transit benefits "
"reduce fares by 50%% for bus service on fixed routes."
msgstr ""
```

Then, in the Spanish translation, we also use `%%`.
```
#, python-format
msgid ""
"Cal-ITP doesn’t save any of your information. All CST transit benefits "
"reduce fares by 50%% for bus service on fixed routes."
msgstr ""
"Cal-ITP no almacena su información. Todos los beneficios de tránsito de CST "
"reducen las tarifas de autobuses en un 50%%."
```

`compilemessages` is able to compile the PO files, and the copy is correct when the template is rendered:

image

### Moving the translatable string to Python code
With our `context` approach, we need to move the string from the template into [Python code](https://github.com/cal-itp/benefits/pull/2705/commits/29b66df6ac8138020d39a290b480898ad87dd76e#diff-861ca9c59ea52eff4e0160e8f87dbe2f97d8370b1649bcee9a6ced268e73a908R46-R49), e.g.:

```python
...

eligibility_index = {
AgencySlug.CST.value: EligibilityIndex(form_text=_("Cal-ITP doesn’t save any of your information. All CST transit benefits reduce fares by 50%% for bus service on fixed routes."))
...
}

...
```

Notice we use an escaped percent sign (`%%`).

`makemessages` sees this and generates the [same PO entries](https://github.com/cal-itp/benefits/pull/2705/commits/29b66df6ac8138020d39a290b480898ad87dd76e#diff-8b77e09357e0ac2a23a88239a579cb8218a1c212a5d816af33125cf3690e7a16R33-R37) as before. `compilemessages` is able to compile the PO files.

But the copy is not correct when the template is rendered. It shows `%%`.

image


Why?? If you think about the template's perspective, the difference from before is that the string to be translated now has `%%` in it. So we need to somehow get it to have just `%` instead.

angela-tran:
I'll need to rebase on top of changes from #2714 and add a test to check `eligibility_index` contains all `AgencySlug`s.

angela-tran:
> Just wanted to check if we would want to merge this PR until after #2733 later today?

I think this question is outdated since we finished the release containing #2733 yesterday.


> One more thing, I was thinking that depending on how the [Simplify Templates tickets](https://github.com/cal-itp/benefits/milestone/54) go, we could name the files inside the `context` directory of each Django app following the naming of the files inside `core/models`. For example, `core/context/agency.py` could be renamed to `core/context/transit.py` since `agency_index` in `core/context/agency.py` is used by `core/context/transit.py`. I'm not sure how much difference this would make for readability/quickly finding where a context is used, but thought about mentioning it.

Yeah, I would be fine with that rename for sure; the consistency would be nice. I'll leave it to be done in a separate PR.

PR #2713, opened 2025-02-28
Feat: prevent agency from being saved if it doesn't have copy for index

angela-tran:
While working on #2705, I realized that it would help prevent errors if `TransitAgency.clean` checks that the agency has all the copy it will need. That is, the `clean` method should try to call each `@property` method for getting the context dictionary for whatever page and make sure calling that method doesn't raise an error. 

This is similar to how [`EnrollmentFlow.clean`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/enrollment.py#L292) checks that the flow is able to get its `in_person_eligibility_context` if it is an in-person flow.

Basically if somehow we've forgotten to add copy for some slug, or if we accidentally remove copy from one of the mappings, then upon saving, we'll realize it.

### DIFF
--- a/benefits/core/models/transit.py
+++ b/benefits/core/models/transit.py
@@ -197,7 +197,7 @@ class TransitAgency(models.Model):
 
     def clean(self):
         field_errors = {}
-        template_errors = []
+        non_field_errors = []
 
         if self.active:
             message = "This field is required for active transit agencies."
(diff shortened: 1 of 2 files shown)


angela-tran:
I realized while writing the PR description that this doesn't fully protect us from errors. For example, if we removed an entry for an existing `TransitAgency`, an error could still happen.

It would be better to have unit tests that can be more comprehensive.

PR #2714, opened 2025-02-28
Test: ensure core_context.agency_index contains all AgencySlugs

angela-tran:
This will protect us from forgetting to add Agency Index copy.

Other PRs (like #2705) can add their own tests.

### DIFF
--- a/tests/pytest/core/context/test_agency.py
+++ b/tests/pytest/core/context/test_agency.py
@@ -1,0 +1,8 @@
+import pytest
+
+from benefits.core import context as core_context
+
+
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
So the parameter sends all the possible values of `AgencySlug` into the text and then just verifies they are in the `agency_index` dict?

I like it!

angela-tran:
> So the parameter sends all the possible values of `AgencySlug` into the test and then just verifies they are in the `agency_index` dict?
> 
> I like it!

Yep! Thanks

PR #2715, opened 2025-02-28
Chore: remove unused parameter in eligibility_from_oauth

lalver1:
While working on https://github.com/cal-itp/benefits/pull/2700, I noticed that we were not using this parameter anymore. This PR makes this small cleanup.


thekaveman (approved):
Looks like a good cleanup :+1:

PR #2717, opened 2025-02-28
docs: roadmap updates

indexing:
- moved enhanced eligibility checks for veterans to Q2
- moved single eligibility check to H2
- modified order of Q2 items

### DIFF
--- a/docs/enrollment-pathways/README.md
+++ b/docs/enrollment-pathways/README.md
@@ -46,18 +46,17 @@ timeline
 
           Q1<br>Next
         : Front-end enhancements and optimization
-        : Benefits admin tool (user management)
         : Deploy in-person enrollments to all agencies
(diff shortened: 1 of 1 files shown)


docs/enrollment-pathways/README.md
```diff
@@ -46,18 +46,17 @@ timeline
 
           Q1<br>Next
```
thekaveman, line 47: Maybe this should be `Current` or something?

thekaveman (changes requested):

docs/enrollment-pathways/README.md
indexing: Good catch. I use `Now` for current quarter and `Next` for the following quarter. I've updated those labels in a new commit.

thekaveman (approved):

PR #2726, opened 2025-03-07
Refactor: simplify Eligibility Start templates

### LINKED ISSUE #2728: Simplify templates: Eligibility Start

angela-tran:
Closes #2728 

This PR refactors the `eligibility.start` view so that it uses a single template. It uses a context dictionary from the flow to provide the template with copy.

The `EligibilityStart` class models the pieces of copy that are needed for the page. I noticed enough duplication with the Login.gov-based instances and also with the agency-card instances of `EligibilityStart`, so I decided to create subclasses - `LoginGovEligibilityStart` and `AgencyCardEligibilityStart` - to consolidate the copy that was being duplicated.

## Testing locally
- Launch the app
- Check that the Eligibility Start page for each distinct flow looks exactly the same as before ( :grin: )


lalver1 (approved):
This looks good to me! I liked the organization/cleanup in 0b3c428d6fd640c8f2b01208dcab3f73afe2d51a and 6e067f9d3a6a6c41dbb6ed11ab7c405efbdb5fec

Also tested locally and compared it on the browser with https://dev-benefits.calitp.org/ and the Eligibility Start page looked identical 👍

benefits/eligibility/context/flow.py
```diff
@@ -0,0 +1,80 @@
...
+    route: str
+    fallback_text: Optional[str] = None
+    extra_classes: Optional[str] = None
+
+
+@dataclass
+class EligibilityStart:
+    page_title: str
+    headline_text: str
+    eligibility_item_template: str
```
thekaveman, line 22: Maybe we can take a follow-up item to refactor this field away from using these templates, and to encode the item strings directly in context.

benefits/eligibility/templates/eligibility/start.html
```diff
@@ -5,26 +5,34 @@
...
   <ul class="d-flex flex-column gap-4 list-unstyled ps-0 mb-0">
-    {% block eligibility-item %}
-    {% endblock eligibility-item %}
+    {% include eligibility_item_template %}
     {% include "eligibility/includes/eligibility-item--contactless-card--start.html" %}
   </ul>
 {% endblock inner-content %}
 
 {% block call-to-action-button %}
+  {% translate call_to_action_button.text as button_text %}
```
thekaveman, line 30: This is something I was wondering about when working on #2753:

Since we already use the `gettext()` helper (aliased as `_()`) setting up the context, to mark the string as translatable -- do we _also_ need to translate the string in the template with this `{% translate %}` (or `{% blocktranslate %}` tag?

I think the answer, from #2753 anyway, seems to be: no. You can just render the variable directly in the template and it will be translated as necessary. E.g. this template change: https://github.com/cal-itp/benefits/pull/2753/files#diff-70ed6fd5fa2321e83db0d63df8312f680cd40428de4ada50dab22e93f55db7c7R109

<details>
<summary>Translation still works</summary>
image
</details>

angela-tran: I agree that we can remove `translate` from the templates :+1:

benefits/eligibility/context/flow.py
angela-tran: Yep, great idea! Created https://github.com/cal-itp/benefits/issues/2759 for it

angela-tran: I put it in the sprint, but feel free to change that if you disagree

benefits/eligibility/templates/eligibility/start.html
angela-tran: I'm thinking to merge this PR as-is, and a follow-up PR can remove the unnecessary `translate`s in the templates

PR #2735, opened 2025-03-10
docs: marking nevada county live

indexing:
- removed target and test language
- added check icons for live pathways


thekaveman (approved):
LGTM :rocket:

Issue #2672, opened 2025-02-06
Set up Benefits administrator user accounts for MST


o-ram:
MST staff has identified **9 staff members** that they wish to have Benefits admin user accounts for. This includes the 6 CSRs on staff, 1 Mobility services staff member, Michael, and Beronica.

## Acceptance Criteria

- [x] Accounts are set up for all identified MST users

## Additional context

- Potential to reduce this number if it would help, in which case MST would likely start with only half of the CSRs having accounts 
- Users will need some training, and documentation needs to be updated, but the scope of this issue is just for the account setup.

o-ram:
@indexing here's the issue for MST accounts we discussed, with the number of people having been confirmed with Beronica earlier today.

thekaveman:
Blocking this for:

- [ ] #2625 

And maybe:

- [ ] #2384

We need a plan for how to create and share the initial account details with MST, and any initial steps they should take to complete account setup.

thekaveman:
We're going to meet on Tuesday 3/11 to do some of the initial setup as a group. 

@indexing and @thekaveman decided not to worry about the "Change password" page for now.

o-ram:
MST uses Bitwarden for a password manager. @indexing @thekaveman @angela-tran

thekaveman:
This is now unblocked as we've released reCAPTCHA protection for the Admin login page: https://github.com/cal-itp/benefits/releases/tag/2025.03.2

o-ram:
- Accounts have been created for all 8 users.
- Credentials were shared via email (one secret was used for passwords)
- A handful of users have already logged in and confirmed that they faced no issues!

PR #2746, opened 2025-03-12
Feat: reCAPTCHA on the Admin login

### LINKED ISSUE #2625: Add reCAPTCHA protection to Benefits Admin login page
While discussing #2537 and #2608 as a team, we had the idea that it would be good to protect the Benefits Admin login page with reCAPTCHA as well. 

This is a separate effort from #2537 because it is introducing reCAPTCHA to a page that does not have it currently.

## Acceptance Criteria

- [ ] Benefits Admin login page is protected by reCAPTCHA
- [x] We inform the user that the login form is protected by reCAPTCHA using the same copy as #2494

thekaveman:
Closes #2625 

## Testing

* Checkout this branch in the devcontainer
* Add `DJANGO_RECAPTCHA_SITE_KEY` and `DJANGO_RECAPTCHA_SECRET_KEY` to your `.vscode/launch.json`
* `F5` to launch the app
* Browse to the Admin page, observe the reCAPTCHA text
* Log in with a local account
* Confirm login success
* Log out
* Update `benefits.core.recaptcha.verify()` to return `False`
* Attempt to login in again with a local account
* Confirm reCAPTCHA error message

## What this PR does

* Adds reCAPTCHA protection to the Admin login form
* Reports reCAPTCHA error as standard form validation error

## Still TODO

In the interest of unblocking #2672, this PR skips some of more nuanced work to make this feature complete. This can come in follow-up PR(s):

* Fully match the Figma design (text placement under the form, link styles, etc.)
* DRY the reCAPTCHA code (using includes, form fields/widgets, etc.)
* DRY the styles


angela-tran (approved):
This looks good to me. I see the same as what your screenshots show

benefits/static/css/admin/styles.css
```diff
@@ -193,6 +193,62 @@ iframe.card-collection {
...
+  height: var(--external-link-icon-size);
+  margin-left: 2px;
+  vertical-align: text-bottom;
+}
+
+/* reCAPTCHA: Do not display badge */
+.grecaptcha-badge {
+  display: none;
+}
+
```
machikoyasuda, line 251: This is how I would recommend fixing the link color issue.

The link is getting the white color from this variable, `--link-fg`. Our `styles.css` file is setting `--link-fg` to white for the whole app (`:root`). We can set the variable to a separate value _just_ for the `login` page with:


image

benefits/static/css/admin/styles.css
```diff
@@ -193,6 +193,62 @@ iframe.card-collection {
   display: none;
 }
 
+/* Login Page link styles */
+.login a:not(.btn):not(.card) {
+  color: var(--primary-color);
+  text-decoration: underline;
+  text-decoration-style: solid;
```
machikoyasuda, line 200: Change this to `text-decoration: none;`

machikoyasuda: I think I'd recommend changing the `not(.btn)` to `not(.google-btn-label)`

machikoyasuda (changes requested):
Added CSS suggestions

benefits/static/css/admin/styles.css
thekaveman: This doesn't address all the other styles though, right? Like the hover, focus rings, external link icon... we still need that right?

benefits/static/css/admin/styles.css
thekaveman: `not(.google-btn-label)` doesn't work since that class is on the `<div>` that wraps the `<a>`.

I used `:not(.btn-content)` which is the class applied to the Google SSO link, and that worked [`58148dd`](https://github.com/cal-itp/benefits/pull/2746/commits/58148ddb5ea8f053858b525539f0468b360f226e)

benefits/static/css/admin/styles.css
machikoyasuda: Yeah we still need it. This is just merely adding those 3 lines to address the white color:

```.login {
   --link-fg: var(--bs-body-color-rgb);
 }
 ```

machikoyasuda: I have not done a full full review of this CSS, just trying to get the button and link in a normal state. I suspect the button CSS code can be substantially improved but that would require a lot more thought/time.

machikoyasuda (dismissed):
LGTM!

I did notice this regression on log out:
image

But I can make a bug ticket and fix/improve this button CSS code at a later time.

thekaveman:
@machikoyasuda Oh no, you're right that was a bad bug -- the hover style made the text totally unreadable! 

Fixed it.

machikoyasuda (approved):
💯  + thanks for the nice PR write-up

PR #2748, opened 2025-03-13
Admin: Disable sidebar

### LINKED ISSUE #2740: Admin: Turn off sticky sidenav
## Acceptance criteria
- Sidebar nav should not show up on any page, including password reset
- Make sure the design looks okay without it
- Any jQuery/JS only used for this sidenav can be unrequired

More info: https://runebook.dev/en/articles/django/ref/contrib/admin/index/django.contrib.admin.AdminSite.enable_nav_sidebar

I am referring to this:
Image

Image

machikoyasuda:
closes #2740 

We do not want the sidebar to show up across the Django admin app.

This is what the sidebar looks like and how it functions, when open and closed on the Reset Password page:
image
image

This PR sets the `enable_nav_sidebar` setting to False, so that it does not appear at all:

image

### DIFF
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -32,10 +32,6 @@ class BenefitsAdminSite(AdminSite):
     def login(self, request, extra_context=None):
         return super().login(request, extra_context)
 
-    def __init__(self, name="admin"):
-        super().__init__(name)
-        self.enable_nav_sidebar = False
-
     def index(self, request, extra_context=None):
         """
         Display the main admin index page if the user is a superuser or a "staff_group" user.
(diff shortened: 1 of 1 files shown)


lalver1:
Nice! I see that the sidebar does not appear 👍

I was just wondering if we'd want to implement the `enable_nav_sidebar = False` setting as an override on the class attribute, like we do with `site_title`, `site_header`, and `index_title`. In the end I think it's the same result, the sidebar won't appear, but maybe we could take this alternate approach for consistency? I think it would look like
```python
class BenefitsAdminSite(admin.AdminSite):

    site_title = "Cal-ITP Benefits Administrator"
    site_header = "Administrator"
    index_title = "Dashboard"
    enable_nav_sidebar = False
    
    def index(self, request, extra_context=None):
    ...
```

machikoyasuda:
@lalver1 Thank you for the suggestion. It worked!

lalver1 (approved):
This looks great! Tested locally and I don't see the sidebar anymore 👍

PR #2750, opened 2025-03-13
Docs: remove reference to `index_template` and `eligibility_index_template`

angela-tran:
Forgot to remove these in #2674 and #2705.


machikoyasuda (approved):
🧹🧹🧹

Issue #2723, opened 2025-03-07
OAuth refactor: create hooks implementation


thekaveman:
## Currently

- Usage of the [`benefits.oauth.analytics`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/analytics.py) module functions are sprinkled throughout the [`benefits.oauth.views`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/views.py#L12) , [`benefits.oauth.redirects`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/redirects.py#L7) , and [`benefits.oauth.middleware`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/middleware.py) modules.
- Usage of `sentry_sdk.capture_exception()` is also sprinkled throughout the same files

## Refactor

- [`DefaultHooks`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/hooks.py#L36) provides definitions to plug into specific points in the integration with IdG
- `cdt_identity` implements the views and redirects necessary for this integration, we won't need these in `benefits` anymore
- `benefits` can provide a custom hooks implementation

## Acceptance criteria

- [x] Define a custom hooks class e.g. `benefits.oauth.hooks.OAuthHooks` inheriting from `cdt_identity.hooks.DefaultHooks`
- [x] All calls to `benefits.oauth.analytics` functions happen within the corresponding custom hooks classmethod(s)
- [x] All calls to `sentry_sdk.capture_exception()` happen within the corresponding custom hooks classmethod(s)
- [x] All overridden classmethods call their corresponding `DefaultHooks` classmethod to start

angela-tran:
#2767 was the last PR for this ticket and was merged earlier today

Issue #2724, opened 2025-03-07
OAuth refactor: wire up `cdt_identity` URLs


thekaveman:
## Currently

- `benefits.oauth` defines a [URLConf](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/urls.py)
- The top-level `benefits.urls` URLConf [includes the `benefits.oauth` URLConf](https://github.com/cal-itp/benefits/blob/main/benefits/urls.py#L29) under the prefix `oauth/`

## Refactor

* `cdt_identity` defines a [URLConf](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/urls.py)
* `benefits` should simply register the routes from `cdt_identity`

## Acceptance criteria

- [ ] Routes from `cdt_identity.urls` are included in the top-level `benefits.urls` under the prefix `oauth/`
- [ ] The custom hooks defined earlier are injected into the views at this stage
- [ ] Any middleware wrappers needed for individual view functions in `cdt_identity` are applied at this stage

angela-tran:
This has been split up into #2786, #2787, and #2788

PR #2757, opened 2025-03-17
Feat: OAuth hooks for `login` view

angela-tran:
Part of #2723 

This PR implements the hook methods that are called in the `django-cdt-identity` [`login`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L117) view.

Here are the lines in the existing Benefits `login` view that correspond with each hook:
- https://github.com/cal-itp/benefits/blob/af0ec2cb2980978820856149cb374cb8b6e1af1f/benefits/oauth/views.py#L66 --> [`pre_login`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L130) hook 
- https://github.com/cal-itp/benefits/blob/af0ec2cb2980978820856149cb374cb8b6e1af1f/benefits/oauth/views.py#L81-L83 --> [`system_error`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L151) hook
- none --> [`post_login`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L153) hook

### DIFF
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -1,0 +1,5 @@
+from cdt_identity.hooks import DefaultHooks
+
+
+class OAuthHooks(DefaultHooks):
+    pass
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -1,5 +1,10 @@
 from cdt_identity.hooks import DefaultHooks
 
+from . import analytics
+
 
 class OAuthHooks(DefaultHooks):
-    pass
+    @classmethod
+    def pre_login(cls, request):
+        super().pre_login(request)
+        analytics.started_sign_in(request)
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -1,5 +1,8 @@
 from cdt_identity.hooks import DefaultHooks
+from django.shortcuts import redirect
+import sentry_sdk
 
+from benefits.routes import routes
 from . import analytics
 
 
@@ -8,3 +11,10 @@ class OAuthHooks(DefaultHooks):
     def pre_login(cls, request):
         super().pre_login(request)
         analytics.started_sign_in(request)
+
+    @classmethod
+    def system_error(cls, request, exception):
+        super().system_error(request, exception)
+        analytics.error(request, message=str(exception), operation="authorize_redirect")
+        sentry_sdk.capture_exception(exception)
+        return redirect(routes.OAUTH_SYSTEM_ERROR)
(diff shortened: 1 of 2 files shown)


benefits/oauth/hooks.py
```diff
@@ -0,0 +1,20 @@
...
+class OAuthHooks(DefaultHooks):
+    @classmethod
+    def pre_login(cls, request):
+        super().pre_login(request)
+        analytics.started_sign_in(request)
+
+    @classmethod
+    def system_error(cls, request, exception):
+        super().system_error(request, exception)
+        analytics.error(request, message=str(exception), operation="authorize_redirect")
```
angela-tran, line 18: @thekaveman I noticed we have this `operation` parameter in the analytics call, and the value is different depending on when the system error occurred.

The 3 different values I see are:
- [`"init"`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/views.py#L40)
- [`"authorize_redirect"`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/views.py#L81)
- [`"authorize_access_token"`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/views.py#L116)

I think we need to modify this hook slightly to handle that, and I have some ideas on how it could change, but I'm curious if you have any immediate thoughts?

angela-tran: The easiest thing to do I think is to have the hook take an `operation` parameter, and `django-cdt-identity` would call the `system_error` hook with the correct value for `operation` at the appropriate places. It's pretty specific to this analytics use case, but maybe that's ok.

thekaveman: Hmmm good catch, and I like your straightforward approach 🙌 The operations seem to roughly correspond to the different methods the OAuth client class we use, so maybe we stick to that as our guide for what values to pass to `operation`.

angela-tran: Thanks @thekaveman. https://github.com/Office-of-Digital-Services/django-cdt-identity/pull/49 is ready for review

thekaveman (changes requested):
Starting with the simpler `login` view hooks means we can have the class definition in place and able to expand -- with no impacts to the existing app! Nice!

For now I think we should revert cleaning up the tests though, since the code paths they cover are still live in the app until we cut-over the routes in #2724.

angela-tran:
> Starting with the simpler `login` view hooks means we can have the class definition in place and able to expand -- with no impacts to the existing app! Nice!
> 
> For now I think we should revert cleaning up the tests though, since the code paths they cover are still live in the app until we cut-over the routes in #2724.

Makes sense, will do!

angela-tran:
Ugh, sorry - kept rebasing onto an outdated `main`. This is ready for review

thekaveman (approved):
:+1:

Issue #2721, opened 2025-03-07
OAuth refactor: update models


thekaveman:
## Currently

- [`ClaimsProvider`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/claims.py) holds configuration related to the IdG connection
- [`EnrollmentFlow`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/enrollment.py) holds configuration related to the specific claims request to IdG

## Refactor

- [`IdentityGatewayConfig`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/models.py#L53) holds configuration related to the IdG connection
- [`ClaimsVerificationRequest`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/models.py#L6) holds configuration related to the specific claims request to IdG

## Acceptance criteria

- [ ] Use the new models in all cases where the existing models are used but there is overlap in behavior/fields
- [ ] Create data migrations to avoid loss of configuration data

lalver1:
@thekaveman or @angela-tran , curious if you have any thoughts about how to go about the refactor of `EnrollmentFlow`: 

We could approach this refactor just like the `ClaimsProvider` refactor, with `ClaimsVerificationRequest.system_name` playing the same role of `IdentityGatewayConfig.client_name` as in the [PR](https://github.com/cal-itp/benefits/pull/2758#issue-2926854579) associated with this ticket.

But the difference is that for the `EnrollmentFlow` refactor, `EnrollmentFlow.system_name` is a `models.SlugField` with `choices=benefits.core.context.SystemName` whereas `ClaimsProvider.client_name` was just a `models.TextField`. `ClaimsProvider.client_name` being a `models.TextField` allowed us to 1) populate `IdentityGatewayConfig.client_name` with a data migration (and we could still do this for this refactor) and 2) delete `ClaimsProvider.client_name`. `ClaimsProvider.client_name` did not depend on a set of choices defined in Benefits so it was just moved to `IdentityGatewayConfig`. For the `EnrollmentFlow` refactor, it seems that it would make more sense to somehow make `ClaimsVerificationRequest.system_name` aware that it can only choose from the choices in `benefits.core.context.SystemName`.

[...]

PR #2758, opened 2025-03-18
Refactor: use `IdentityGatewayConfig` model from `cdt_identity`

lalver1:
Part of #2721 

This PR  implements the `IdentityGatewayConfig` refactor.

There are 2 ways to proceed with this refactor, this PR implements option 1:

1. Completely remove `ClaimsProvider` and move `sign_out_button_template` and `sign_out_link_template` from `ClaimsProvider` to `EnrollmentFlow`. Add a foreign key to `IdentityGatewayConfig` in `EnrollmentFlow` to deal with the removal of `ClaimsProvider`.
Even though there will be some data duplication in `sign_out_button_template` and `sign_out_link_template` across` EnrollmentFlow`s, our medium term goal is to remove template fields and simplify models / Admin config, and this option achieves this goal better than option 2.

2. Keep `ClaimsProvider` and add an `IdentityGatewayConfig` foreign key to it. This avoids duplicating `sign_out_button_template` and `sign_out_link_template` in multiple` EnrollmentFlow`s. Instead, an `EnrollmentFlow` refers to one `ClaimsProvider` which in turn refers to one `IdentityGatewayConfig`. There are `on_delete` protections in place to prevent deletion of `IdentityGatewayConfigs` that are in use by a `ClaimsProvider`.

This schematic shows how the fields for `EnrollmentFlow` ended up:

| `EnrollmentFlow`                      | 
| ------------------------------ | 
|         `id`                                    |  
|          `system_name`               |   
|          `...`                                   |   
| `sign_out_button_template`   |    
| `sign_out_link_template`        |
| `oauth_config` (`IdentityGatewayConfig fk`)                  |               
|     `...`                                       |

## To review

- Checkout `main`
- Set `DJANGO_DB_FIXTURES` to `local_fixtures.json` and run `bin/reset_db.sh` to reset your database
- Switch to this PR's branch
- Run `bin/init.sh` to run the migrations in this branch and test that the data was migrated successfully from `ClaimsProvider` to `IdentityGatewayConfig` and to `EnrollmentFlow`
  - Note: If you are using the `test` database to verify this step, make sure to add (and export) an `auth_provider_client_id` environment variable (with its appropriate value) to your `.env` file since this is what the `test` environment's database will look for.
- Start a debugger 
- Log in to the admin with your Google SSO account and enter valid values for any flow that uses the IdG (you can use `Older Adult` since it will test that the logout button works):
  - Identity gateway configs -> Client id 
    - Use the value that used to be a secret and associated with `claims_provider_client_id` in your `.env` file. This can be set in the database fixtures file since at the moment the fixtures have a sample value, but I wasn't 100% comfortable yet putting the actual value in the fixtures 😅 
  - Identity gateway configs -> Authority
  - Enrollment flows -> Claims scope
  - Enrollment flows -> Claims eligibility claim
- Test the user facing flow that you configured and also test the admin using `calitp-user`
- Test the local developer environment, ensure that the fixtures have been updated and that they successfully load by running `bin/reset_db.sh`


benefits/core/migrations/local_fixtures.json
```diff
@@ -28,22 +28,12 @@
...
   {
     "model": "core.claimsprovider",
-    "pk": 2,
-    "fields": {
-      "client_name": "benefits-medicaregov",
-      "client_id_secret_name": "claims-provider-client-id",
-      "authority": "https://example.com",
-      "scheme": "dev-cal-itp_benefits"
-    }
+    "pk": 2
```
thekaveman, line 36: Let's make sure to add back the sample fixture(s) for `IdentityGatewayConfig`

thekaveman (changes requested):

angela-tran (approved):
I read through the code changes and tested locally, and this looks good to me. :clap: 

I had a little bit of feedback on the data migration in that I noticed a way we could do it in one loop instead of two, but I decided it doesn't matter, and this works :+1:

thekaveman (changes requested):
I initially had some trouble following the review steps... but I realized it was because I didn't have an UUID as the `claims_provider_client_id` value in my local `.env` file.

Thankfully we're getting rid of all of that in favor of configuration in the fixtures.

But this is an important point that was missed in the sample fixtures: the `id` property for the `IdentityGatewayConfig` (which is the primary key column) is an UUID type. In the sample fixtures we have integers, which results in the `00...1` and `00...2` UUIDs.

```json
{
  "model": "cdt_identity.identitygatewayconfig",
  "pk": 1, //<-- should be UUID
  "fields": {
  }
}
```

image

lalver1:
Ah good catch, thanks! 8da449e392512ccee9e62858292a6b7956ddac14 uses UUID for the `IdentityGatewayConfig`'s primary key.

thekaveman (approved):
:rocket:

Issue #2722, opened 2025-03-07
OAuth refactor: update session handling


thekaveman:
## Currently

- The user's [selected `flow`](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L39) is maintained in the session
- The users's verified claims are [stored in the session](https://github.com/cal-itp/benefits/blob/main/benefits/core/session.py#L276) - this is the only definition of this concept, and it is ill-defined at best! (What is the type? What are expected values?)
- ~_We can ignore `oauth_token` since it will be removed as part of #2700 and is already refactored in `cdt_identity`_~

## Refactor

- [`ClaimsResult`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/claims.py) is a dataclass structure produced by a parser class that encodes the algorithm and types more explicitly.
- The [`Session`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/session.py) wrapper is used internally by `cdt_identity`, and as the interface to communicate configuration and results between `cdt_identity` and `benefits`.

## Acceptance criteria

- [x] Benefits configuration from the new models is used to initialize the `cdt_identity` Session wrapper - #2795 
- [x] Benefits session handling incorporates usage of the `cdt_identity` Session wrapper (e.g. for claims, logged in status)

thekaveman:
Please see this thread for more context on the intended changes here: https://github.com/cal-itp/benefits/pull/2783#discussion_r2014560634

angela-tran:
@thekaveman I believe this ticket is done, but if there was further refactoring specific to OAuth that you were thinking of for the session, feel free to re-open this.

I opened #2800 for the more general refactor to `benefits.core.session`.

PR #2767, opened 2025-03-18
Feat: OAuth hooks for `authorize` view

angela-tran:
Part of #2723 

This PR implements the hook methods that are called in the `django-cdt-identity` [`authorize`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L53) view.

These are the hooks and the corresponding Benefits code:
- [`pre_authorize`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L67): none
- [`system_error`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L83): https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L116-L118 (already done by #2757 with [the last two commits](https://github.com/cal-itp/benefits/pull/2757/files/b320e98a699f4ce20f001c5827456eb2d5e4a221..72705baa4bbb74599807456a3f7088dc1d923155))
- [`post_authorize`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L85): none
- [`pre_claims_verification`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L90): none
- [`claims_verified_eligible`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L101): https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L148-L150 (the variant with no errors)
- [`claims_verified_not_eligible`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L107): https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L148-L150 (the variant with errors)

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -112,7 +112,7 @@ def confirm(request):
             return TemplateResponse(request, TEMPLATE_CONFIRM, context)
         # no type was verified
         elif not is_verified:
-            return redirect(reverse(routes.ELIGIBILITY_UNVERIFIED))
+            return redirect(routes.ELIGIBILITY_UNVERIFIED)
         # type was verified
         else:
             return verified(request)
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -78,16 +78,8 @@ def confirm(request):
     if request.method == "GET" and session.eligible(request):
         return verified(request)
 
-    unverified_view = reverse(routes.ELIGIBILITY_UNVERIFIED)
-
     agency = session.agency(request)
     flow = session.flow(request)
-
-    # GET for OAuth verification
-    if request.method == "GET" and flow.uses_claims_verification:
-        analytics.started_eligibility(request, flow)
-
-        return redirect(unverified_view)
 
     form = flow.eligibility_form_instance()
 
@@ -120,7 +112,7 @@ def confirm(request):
             return TemplateResponse(request, TEMPLATE_CONFIRM, context)
         # no type was verified
         elif not is_verified:
-            return redirect(unverified_view)
+            return redirect(reverse(routes.ELIGIBILITY_UNVERIFIED))
         # type was verified
         else:
             return verified(request)
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -87,12 +87,7 @@ def confirm(request):
     if request.method == "GET" and flow.uses_claims_verification:
         analytics.started_eligibility(request, flow)
 
-        is_verified = verify.eligibility_from_oauth(flow, session.oauth_claims(request))
-
-        if is_verified:
-            return verified(request)
-        else:
-            return redirect(unverified_view)
+        return redirect(unverified_view)
 
     form = flow.eligibility_form_instance()
 
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -1,7 +1,6 @@
 from cdt_identity.hooks import DefaultHooks
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator, decorator_from_middleware
-from django.urls import reverse
 import sentry_sdk
 
 from benefits.routes import routes
@@ -70,7 +69,7 @@ class OAuthHooks(DefaultHooks):
         flow = session.flow(request)
         eligibility_analytics.started_eligibility(request, flow)
 
-        return redirect(reverse(routes.ELIGIBILITY_UNVERIFIED))
+        return redirect(routes.ELIGIBILITY_UNVERIFIED)
 
     @classmethod
     def system_error(cls, request, exception, operation):
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -42,6 +42,12 @@ class OAuthHooks(DefaultHooks):
         return redirect(routes.ELIGIBILITY_CONFIRM)
 
     @classmethod
+    def claims_verified_not_eligible(cls, request, claims_request, claims_result):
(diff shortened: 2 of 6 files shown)


angela-tran:
I'm wondering about this line where the session is updated...

https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L147-L150

I think `cdt-identity` handles storing most information in the session, but maybe `oauth_authorized` is something specific to Benefits? Which would mean the `claims_verified_eligible` and `claims_verified_not_eligible` hooks need to handle updating the session with that.

I'll try to confirm this myself, but @thekaveman if you already know this is correct, let me know

angela-tran:
Ok, I figured out that `oauth_authorized` is specific to Benefits. It represents the concept of being "logged in" for purposes of eligibility verification -- see `session.logged_in`:

https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/core/session.py#L144-L146

Putting this PR into draft so I can add session-updating to the hooks.

thekaveman:
@angela-tran 

> Putting this PR into draft so I can add session-updating to the hooks.

Note we have #2722 as a specific task around session handling. Sort of relies on #2721 though.

angela-tran:
> @angela-tran
> 
> > Putting this PR into draft so I can add session-updating to the hooks.
> 
> Note we have #2722 as a specific task around session handling. Sort of relies on #2721 though.

Does that task cover updating `oauth_authorized`? I'm ok with either doing it in this PR or letting that PR handle it, so I'll leave it up to what you think is best.

The impetus for my thinking that the hooks should handle updating this specific `oauth_authorized` flag: I noticed the call to `session.logout` when working on https://github.com/cal-itp/benefits/pull/2769#issue-2932976891.

thekaveman:
> Does that task cover updating `oauth_authorized`?

I think in my head I had that task being completed _before_ the work on the hooks :sweat_smile: so in that case, it would have been handled there first. 

> The impetus for my thinking that the hooks should handle updating this specific `oauth_authorized` flag

You may well be correct that the hooks should be doing this. I would posit that it will become _more clear_ exactly what needs to be done after the models are updated and #2722 is done.

benefits/oauth/hooks.py
```diff
@@ -35,6 +35,18 @@ def post_logout(cls, request):
...
+        super().claims_verified_eligible(request, claims_request, claims_result)
+        analytics.finished_sign_in(request)
+        return redirect(routes.ELIGIBILITY_CONFIRM)
+
+    @classmethod
+    def claims_verified_not_eligible(cls, request, claims_request, claims_result):
+        super().claims_verified_not_eligible(request, claims_request, claims_result)
+        analytics.finished_sign_in(request, error=claims_result.errors)
+        return redirect(routes.ELIGIBILITY_CONFIRM)
+
```
thekaveman, line 49: Yeah these pair of hook methods were definitely meant to simplify the Eligibility view situation. The verification has already been done at this point, there's no reason to send through the `ELIGIBILITY_CONFIRM` view to do it again.

Related to https://github.com/cal-itp/benefits/pull/2783#discussion_r2014998270 but yeah, we can leave it for a follow-up if you'd rather.

angela-tran: Quick question @thekaveman --

In the `is_verified` branch of the Eligibility Confirm logic, it ends with returning the `verified` view:

https://github.com/cal-itp/benefits/blob/75ab4cf1ae94245d0cafbc8f676d0a8bfb804beb/benefits/eligibility/views.py#L93

which does this logic:

https://github.com/cal-itp/benefits/blob/75ab4cf1ae94245d0cafbc8f676d0a8bfb804beb/benefits/eligibility/views.py#L134-L144

Are you fine with the hook method (`claims_verified_eligible`) returning the `verified` method as well?

thekaveman: Yeah I think that makes sense, right?

benefits/oauth/hooks.py
```diff
@@ -35,6 +39,39 @@ def post_logout(cls, request):
...
+        ]
+    )
+    def claims_verified_not_eligible(cls, request, claims_request, claims_result):
+        super().claims_verified_not_eligible(request, claims_request, claims_result)
+        analytics.finished_sign_in(request, error=claims_result.errors)
+
+        flow = session.flow(request)
+        eligibility_analytics.started_eligibility(request, flow)
+
+        return redirect(reverse(routes.ELIGIBILITY_UNVERIFIED))
```
thekaveman, line 73: We don't need the call to `reverse` here; [`redirect`](https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#redirect) can handle an `app:endpoint` pattern and will call it for us.

thekaveman (changes requested):
Small nitpick but this looks good overall

benefits/eligibility/views.py
```diff
@@ -125,7 +112,7 @@ def confirm(request):
             return TemplateResponse(request, TEMPLATE_CONFIRM, context)
         # no type was verified
         elif not is_verified:
-            return redirect(unverified_view)
+            return redirect(reverse(routes.ELIGIBILITY_UNVERIFIED))
```
thekaveman, line 115: Same here

thekaveman (approved):
Amazing :broom:

PR #2768, opened 2025-03-18
Feat: OAuth hooks for `cancel` view

angela-tran:
Part of #2723 

This PR implements the `cancel_login` hook method that is called in the django-cdt-identity [`cancel`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L110) view.

- [`cancel_login`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L114): https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L157-L159

PR #2769, opened 2025-03-19
Feat: OAuth hooks for `logout` view

angela-tran:
Part of #2723 

This PR implements the hooks for the `logout` view.

- [`pre_logout`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L171) - https://github.com/cal-itp/benefits/blob/270492d228c0541e8952fe7719d971e61ad560c1/benefits/oauth/views.py#L177-L180
- [`system_error`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L187) - already handled by #2757

### DIFF
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -3,6 +3,7 @@ from django.shortcuts import redirect
 import sentry_sdk
 
 from benefits.routes import routes
+from benefits.core import session
 from . import analytics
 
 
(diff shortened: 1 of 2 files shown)


thekaveman (approved):
:+1:

thekaveman:
I think there's no way to avoid re-reviews if they are stacked branches:

* Either the first base branch PR is merged (e.g. like #2768), then then next one up gets a `Base automatically changed` (like this PR)
* OR the _last_ "base branch" is merged (e.g. #2770), but then that triggers a change of its target (this PR), which requires a review, and all over again

So, all good :smile:

thekaveman (approved):

PR #2772, opened 2025-03-19
docs: changes to agency adoption table

indexing:
- added veterans enrollment pathway for nevada county
- changed column order for enrollment pathways. instead of using the order in which they were released, they are now ordered by level of adoption.
- updated column heading to for veterans to match the benefit option label in the application


thekaveman (approved):
Looks really nice :+1:

thekaveman:
Reminder to assign yourself to these PRs @indexing

PR #2766, opened 2025-03-18
Admin: Header and Theme Toggle

machikoyasuda:
closes #2738 
closes #2739 

## What this PR does
- Creates a `base.html` that has a `header`. This `header` is used everywhere -- as in, on Agency Base and regular Django admin pages -- except for the Log in/Log out page, which has just a branding logo and title only.
- Removes dark mode variables in 1 spot, `base.html`
- Hides theme toggle button
- Makes some small font and color changes across all of admin, to enhance alignment/uniform-ness/close the gap between Figma and Admin. See below. 
- The Django admin app is now also using Bootstrap CSS. `Overrides.css` was created to write minimal CSS to get rid of any visual anomalies created by Bootstrap on Django admin. Now that this is in place, the whole Admin app, all the parts, are on Bootstrap.

## Overall note

I want to note that this PR is intending to get what I call the Universal header into place, file-wise and code-wise. This PR is not trying to improve the overall CSS/Variable situation - there are a lot of improvements/refactors that can be made, but not at this time. The Universal header was made with trying to use as much Bootstrap CSS utility classes as possible, and it was successful in this. After this PR and the #2742 PR are merged, the Admin users will have a visually aligned experience between the Log in/Log out page, Agency in person flow page, Reset password page, and Django Admin page (used only by Admins).

### Log in
image
- Remove the blue, as per Figma
- I looked into see how easy it would be to add the 1px black border that the Log in/Log out pages have on Figma. The div is not accessible from `login.html`, so I opted out of doing it. Would require class-based new CSS. (Which is fine. Just didn't want to include it in this PR).
Note: I do not have reCAPTCHA turned on locally

### Django Dashboard
image
- Uses universal header
- The h1 (`Dashboard`) matches the h1 from Agency Dashboard

### Django Dashboard - Class page
image
- The breadcrumb will be changed in the next PR. #2742 

### Django / Agency flow - Change password
image

### Agency Dashboard
image
- Uses universal header
- Change H1 to match Figma, also matches the H1 on Django Dashboard
- Button font size is smaller, to match Figma
- Font sizes match Figma

### Agency flow 
image
- Adds Change Password link to this header
- Clicking Change Password from Agency Flow or Agency Dashboard will take you back to Agency Dashboard

### Log out
image

Issue #2648, opened 2025-01-28
Admin app: Create primitives such as type, color, icons, etc. in Figma


indexing:

indexing:
@ohmegasquared When you've started this work, please update the issue status. I believe you are aiming to finalize the color palette by the end of Feb.?

PR #2773, opened 2025-03-20
Admin: Header theme, remove darkmode toggle & breadcrumb theme

### LINKED ISSUE #2738: Admin: Disable Dark Mode on Admin
Similar to what was done in the in-person flow here (https://github.com/cal-itp/benefits/commit/6de5abea6331b32cf4d634b33def675368aa3d43), disable the dark mode option from the Django admin.

## Accepance critera
- Dark mode toggle icon is hidden
- Dark mode toggling is not automatically triggered by a user's browser or operating system defaults, even if the icon is not there

## Notes

Explore if there is a way to code this without repeating https://github.com/cal-itp/benefits/commit/6de5abea6331b32cf4d634b33def675368aa3d43 code but disabling it higher up in 1 place. What's the most efficient way to achieve this across all of Admin?

### LINKED ISSUE #2739: Header: Create 1 header for all of Admin
## Acceptance criteria
- Add the Cal-ITP logo to the header
- Colors: Change the header background to black, text to white
- Text size, Text decoration, Font weight - match the comps. But do not worry about font.
- Both Django admin and In-Person flow use the same layout and code

Image

### LINKED ISSUE #2742: Admin: Password flow - Figure out the breadcrumb styles
Placeholder ticket. Eventually ticket will be used to have engineers implement a breadcrumb style.

machikoyasuda:
closes #2738 
closes #2739 
closes #2742 

## What this PR does, in order
- Delete unnecessary templates
- Removes **dark mode**, in 1 spot only: in `base.html`
- Favicon, Branding (logo icons), Bootstrap CSS, Styles (old code), Theme (new code), Analytics and are declared in 1 spot: in `base.html`
- **Universal header**: All admin pages have the same header. The _only_ new HTML necessary to make this happen is `branding.html`. It is used on _all_ pages, including login/log out, agency flows and Django admin pages, to add the logo images to the header. The rest of the header is the default. 
- **Breadcrumb colors** added to theme
- How overriding works, as demonstrated with `branding.html` across 2 different template styles: The `theme.css` file uses a body class, CSS variables to add per-page/per-template styling. For example, although _all_ pages have the logo image, the login/log out page's logo size is different. That is declared in the `theme.css` file.

## Notes

The goal of this overall effort is to have:
1. A `theme.css` file that gives the admin the look of the header that has been designed on Figma.
2. Demonstrate how to write the `theme.css` file in various scenarios.
3. Avoid writing HTML for Django header overrides like #2766 did. 

This PR does not achieve:
1. Pixel perfect comps from Figma, but gets pretty close within constraints (like not changing to Noto yet, not touching button CSS).
2. Some things are impossible to change - like, there is no way to add spacing after the `Welcome, username.` There is no way to remove the `View site` link (in a clean way - there are hacky ways to remove this with CSS, probably). 

The way I'd like to move forward after the primitives Design work (https://github.com/cal-itp/benefits/issues/2648) on Admin is done, is to slowly move code _from_ `styles.css` into `theme.css`. 

#### Logged out
image
- Entirely refactors `logged_out` template to be a lot simpler. 
- `logged_out` template now inherits from `login.html`. 

#### Log in / Logged out
image
- Adds border

#### Django admin
image
image

- Adds universal header.
- Fixes any weird Bootstrap regressions

### In-person flow
image
image

- Adds universal header
- Adds a few Bootstrap font utility classes to get the title font sizes right. 
- Future work necessary to get buttons sized right.

## Documentation of best practices

- Get the override blocks in order. Easier to read. Then all pages have the blocks in the same order. For example, the `title` block should come before the `body` block.
- When setting a variable override, comment how it is being used https://github.com/cal-itp/benefits/pull/2773/files#diff-cb8b26fb41fa3037fd9dc5ddd0f9cbea93abff15ad8e73277040466c189dd8f2R3-R18
-


machikoyasuda:
@thekaveman New PR is up, based on what we discussed today.

Started drafting but did not complete a README.md. I wanted to see how this PR lands first with everyone.

benefits/templates/admin/base_site.html
```diff
@@ -1,5 +1,5 @@
-{% extends "admin/base_site.html" %}
+{% extends "admin/base.html" %}
 
-{% block extrahead %}
-  {% include "admin/includes/favicon.html" %}
-{% endblock extrahead %}
```
machikoyasuda, line 5: Originally had intended to delete this file entirely, but all of the Admin app won't get this new branding unless we override this here as well.

benefits/static/css/admin/theme.css
```diff
@@ -0,0 +1,111 @@
...
+  --primary: #ffffff;
+  --bs-body-font-size: 1rem;
+  --bs-body-color: #212121; /* Body text color */
+  --bs-dark-rgb: 33, 33, 33; /* Background color for Header */
+  --bs-heading-color: #212121; /* Header text color */
+  --bs-secondary-bg-subtle: #f1f1f1; /* Background color for Admin, lighter than bs-secondary, used with bg-secondary-subtle */
+  --bs-secondary-rgb:
+    210, 210, 210; /* Border color for Admin, used with border-secondary */
+  --bs-secondary: #c5c4c4;
+}
```
machikoyasuda, line 14: I am leaving this as is, but wanted to note that I'm still noodling with how to format this better overall.

benefits/static/css/admin/theme.css
```diff
@@ -0,0 +1,111 @@
...
+/* Content */
+
+#content {
+  padding: 2rem;
+}
+
+/* Login page */
+
+.login .main {
+  border: 1px solid #000000;
```
machikoyasuda, line 110: Opted not to use a variable here, b/c it's not being repeated everywhere. Again, will be improved when #2648 is done.

benefits/templates/admin/base_site.html
thekaveman: I guess I don't fully understand the difference between admin/base.html and admin/base_site.html and why we need both?

But my main question is, if admin/base_site.html (this template) extends admin/base.html (our custom version of that template) -- then why do we need the same {% block branding %} here?

thekaveman: I.e. maybe it is enough to just have this template `{% extends admin/base.html %}` and that's all?

machikoyasuda: Deleting the `base_site.html` file entirely leads to this:
image

Keeping `base_site.html` but taking out the `branding` override behaves as expected. The `branding` override inherits from our `base.html`.

machikoyasuda: @thekaveman Fixed here: https://github.com/cal-itp/benefits/pull/2773/commits/1dca8396a9440c2dee46bb9e3bbf1ebdf739cdda

thekaveman (approved):
This looks great. We're getting closer to a true "admin theme".

Issue #2725, opened 2025-03-07
OAuth refactor: remove dead code


thekaveman:
## Acceptance criteria

- [ ] Code from `benefits.oauth` that is no longer used is removed: views, client, URLs, redirects, etc.
- [ ] `oauth:` routes from `benefits.routes` are removed #2821 
- [ ] Unused `tests.pytest.oauth` are removed
- [ ] Unused dependency (`Authlib`) removed from `pyproject.toml`

angela-tran:
This has been split up into #2786, #2787, and #2788

I'll leave this ticket open so we can take a look at what other code we might want to remove after those 3 tickets are done.

thekaveman:
Regarding:

- [ ] oauth: routes from benefits.routes are removed

I would have left this feedback on #2821 but we can consider it for this cleanup:

We don't actually need all those routes (re)defined in `benefits.routes` right? Just `OAUTH_LOGIN` and `OAUTH_SYSTEM_ERROR` maybe? We could clean up all the rest of those, and simplify the URLConf even more:

```diff
urlpatterns = [
-   path(Routes.login, decorator(cdt_identity_views.login), kwargs, name=routes.name(routes.OAUTH_LOGIN)),
+   path(Routes.login, decorator(cdt_identity_views.login), kwargs, name=Routes.login),
-   path(Routes.authorize, decorator(cdt_identity_views.authorize), kwargs, name=routes.name(routes.OAUTH_AUTHORIZE)),
+   path(Routes.authorize, decorator(cdt_identity_views.authorize), kwargs, name=Routes.authorize),
...
]
```

PR #2782, opened 2025-03-25
Refactor: use `ClaimsVerificationRequest` model from `cdt_identity`

lalver1:
Closes #2721 

This PR  implements the `ClaimsVerificationRequest` refactor. We first add a `claims_request` field (which is a foreign key to `ClaimsVerificationRequest`, a new model from `cdt_identity`) to `EnrollmentFlow`, migrate data from `EnrollmentFlow` to `ClaimsVerificationRequest`, and finally remove from `EnrollmentFlow`

-  `claims_scope`
- `claims_eligibility_claim`
- `claims_extra_claims`
- `claims_scheme_override`

since they now reside in `ClaimsVerificationRequest`. This schematic shows how the fields for `EnrollmentFlow` ended up:

| `EnrollmentFlow`                      | `ClaimsVerificationRequest`    |
| ------------------------------ | --------------------------- |
|         `id`                                    |               `id`                         |
| `system_name`                        |         `system_name`           |
| ~`claims_scope`~                       |                      `scopes`         |  
| ~`claims_eligibility_claim`~         |  `eligibility_claim`                 |  
| ~`claims_extra_claims`~             |  `extra_claims`                     |  
| ~`claims_scheme_override`~     |          `scheme`                    |  
| `claims_request` (`ClaimsVerificationRequest fk`)                  |
| `...`                                           |                               |

## To review

- Checkout `main`
- Set `DJANGO_DB_FIXTURES` to `local_fixtures.json` and run `bin/reset_db.sh` to reset your database
- Switch to this PR's branch
- Run `bin/init.sh` to run the migrations in this branch and test that the data was migrated successfully from `EnrollmentFlow` to `ClaimsVerificationRequest` and that `claims_request` in `EnrollmentFlow` points to the correct `ClaimsVerificationRequest`
- Start a debugger 
- Log in to the admin with your Google SSO account and enter valid values for any flow that uses the IdG (you can use `Older Adult`):
  - Identity gateway configs -> Client id
  - Identity gateway configs -> Authority
  - Claims verification requests -> Scopes
  - Claims verification requests -> Eligibility claim
- Test the user facing flow that you configured and also test the admin using `calitp-user`
- Test the local developer environment, ensure that the fixtures have been updated and that they successfully load by running `bin/reset_db.sh`

PR #2783, opened 2025-03-26
Refactor: use cdt_identity `authorize` view

angela-tran:
This PR refactors the `oauth_claims` function in `benefits.core.session` to read from `cdt_identity.session.Session(request).claims_result.verified` instead, as a part of #2722. 

In order to keep the Benefits app in a working state, we need OAuth claims to be _written_ to the `cdt_identity` session as well, so this means we also need to switch over to using the `cdt_identity` `authorize` view, which is technically a part of #2724.

With switching over to using the `cdt_identity` `authorize` view, many unit tests are no longer relevant and also now fail. The Benefits oauth `authorize` view is also unused. So the natural thing to do here is to delete them, which is technically part of #2725.


benefits/core/session.py
```diff
@@ -273,12 +275,13 @@ def update(
...
         request.session[_OAUTH_AUTHORIZED] = oauth_authorized
-    if oauth_claims is not None:
-        request.session[_OAUTH_CLAIMS] = oauth_claims
     if origin is not None:
         request.session[_ORIGIN] = origin
     if flow is not None and isinstance(flow, models.EnrollmentFlow):
         request.session[_FLOW] = flow.id
+        cdt_identity_session = Session(request)
+        cdt_identity_session.client_config = flow.oauth_config
+        # cdt_identity_session.claims_request = flow.claims_request
```
angela-tran, line 284: This line can be uncommented when we have the changes from #2782

benefits/core/session.py
```diff
@@ -158,7 +161,7 @@ def oauth_authorized(request):
 
 def oauth_claims(request):
     """Get the oauth claims from the request's session, or None"""
-    return request.session.get(_OAUTH_CLAIMS)
+    return [claim for claim, value in Session(request).claims_result.verified.items() if value] or None
```
thekaveman, line 164: As a one-liner this is a little harder to follow. 

Bigger question, what is this `session.oauth_claims()` function used for? Tracing through the code, it seems like it is used in only one spot outside of the `session` module itself: [`eligibility.views.confirm`](https://github.com/cal-itp/benefits/blob/refactor/session-oauth-claims/benefits/eligibility/views.py#L90) which sends the value to [`eligibility.verify.eligibility_from_oauth()`](https://github.com/cal-itp/benefits/blob/refactor/session-oauth-claims/benefits/eligibility/verify.py#L36)

`eligibility.verify.eligibility_from_oauth()` uses the value as a container (e.g. it does an `in` check):

```python
def eligibility_from_oauth(flow: models.EnrollmentFlow, oauth_claims):
    if flow.uses_claims_verification and flow.claims_eligibility_claim in oauth_claims:
        return True
    else:
        return False
```

So a `list` definitely works. But so would the `dict`-like value of `session.claims_result` directly. 

The `ClaimsResult` class in `cdt_identity` [implements some helpers](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/claims.py#L12) to make all this simpler, where you can do:

```python
# type is ClaimsResult
claims_result = Session(request).claims_result

# ClaimsResult.__contains__ implementation works like we expect
# so this "in" check looks for flow.claims_eligibility_claim in claims_result.verified
flow.claims_eligibility_claim in claims_result
```

angela-tran: Thanks for the example usage. It was easier for me to keep it as a list at least up to this point (allowed me to not worry about changing any tests), but now I do think it should be pretty easy to change the type for `oauth_claims`. I'll look into making the change

benefits/oauth/views.py
```diff
@@ -85,71 +85,6 @@ def login(request):
     return result
 
 
-@decorator_from_middleware(FlowUsesClaimsVerificationSessionRequired)
```
thekaveman, line 88: We lost the middleware here, and it wasn't re-applied in the URL registration. 

How should we handle that?

angela-tran: Thanks for catching. Here is a way to do it: https://github.com/cal-itp/benefits/pull/2783/commits/aabf6630b2f14182af28b25e2a9f5c7a17094a2f. Thoughts on it?

benefits/core/session.py
angela-tran: > One more thought here:
> 
> With the introduction of the hooks framework, we now have the `cdt_identity` library doing claims eligibility verification and the ability to take different actions via [`hooks.claims_verified_eligible`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L101) and [`hooks.claims_verified_not_eligible`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L107).
> 
> So we don't actually _need_ to keep the `eligibility.views.confirm` and `eligibility.verify.eligibility_from_oauth` concepts the same as they are today.

It's a valid point for sure that we could refactor out the claims-related verification logic in `eligibility.views.confirm` out to the hooks. I think that could be a really nice clean-up to do.

At the same time, it doesn't feel totally necessary to do in this PR, so does it sound ok to create a follow-up ticket for it?

> Do we even need `session.oauth_claims()` at all?

So aside from claims eligibility verification, `session.oauth_claims()` is also used by `session.oauth_extra_claims` to figure out what the non-eligibility claims are because we need those extra claims to be recorded on `EnrollmentEvent` in the database and `ReturnedEnrollmentEvent` for analytics.

That's not to say that we can't come up with a way that would allow us to remove `session.oauth_claims` and `session.oauth_extra_claims` methods, but what do we gain?

benefits/core/session.py
```diff
@@ -8,6 +8,8 @@
 import time
 import uuid
 
+from cdt_identity.claims import ClaimsResult
+from cdt_identity.session import Session
```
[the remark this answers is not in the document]
angela-tran: Rebased and edited some commits to do this aliasing - https://github.com/cal-itp/benefits/pull/2783/commits/710088575f42b4737f8495427262af16b0a76d58 and https://github.com/cal-itp/benefits/pull/2783/commits/9deac7566f20c8084cb7e8d1cee0fe1f02dba3c0

benefits/core/session.py
thekaveman: OK, agreed yeah we can come back to that.

angela-tran:
Still todo on this PR:
- change `session.oauth_claims` return type to `ClaimsResult`

Blockers:
- rebase on top of `main` once #2782 is reviewed/approved and merged in

Questions:
-  any objections to merging in #2767 before this PR? -- this PR needs those hooks implemented for the view to keep sending analytics and Sentry

benefits/core/session.py
angela-tran: I was able to refactor out the claims verification logic from `eligibility.views.confirm` to the hooks in https://github.com/cal-itp/benefits/pull/2767. 

I think this means it's not so much that we should refactor the return type of `session.oauth_claims`, but rather we should see if we can clean up `session.oauth_claims` and `session.oauth_extra_claims` to be simpler.

angela-tran: I re-wrote the one-liner as this: https://github.com/cal-itp/benefits/pull/2783/commits/7cf26a66926f0c700bfde0fa4068a1432f7afcd0#diff-8d7012c10c06c3b595e41efdc1b9ab823ebdf38756ce075be504e01d536336b7R162-R169

I'm not sure if that's what your feedback was getting at. I tried a different implementation that attempted to make use of the `eligibility_claim` and `extra_claims` on `cdt_identity.session.Session.claims_request`, but it was not making things simpler, so I abandoned it.

angela-tran: The list of claim-strings (`session._oauth_claims`) is used by [`session.context_dict`](https://github.com/cal-itp/benefits/pull/2783/files#diff-8d7012c10c06c3b595e41efdc1b9ab823ebdf38756ce075be504e01d536336b7R65) and by `session.oauth_extra_claims`.

I'm making an assumption that we want to keep that usage in `session.context_dict`. So seems like this is as simplified as I can get it.

benefits/oauth/views.py
angela-tran: There's going to be some duplication of the decorators and `{ "hook": ...` kwargs as we implement the various view tickets - #2786, #2787, #2788 - but I think that's ok to start with, and we can clean it up once it's all in.

angela-tran:
This is ready for re-review

benefits/core/session.py
thekaveman: Hmm do we want to re-think this? `session.context_dict()` is only used for the `context_processors.debug()`, it doesn't feel very critical to maintain.

I think my feedback wasn't so much about the one-liner / specific implementation. But the overall idea of our session maintaining stuff that the `cdt_identity` session now maintains, and wanting to get really narrow on the use-cases in Benefits. Do we absolutely need this old `oauth` related code in `session`?

I.e. how can we do more cleanup vs. maintaining stuff just because it is there.

thekaveman: (I realize we're discussing things maybe more related to #2722)

thekaveman (approved):
Approving this PR for the work on `authorize`.

I'd still like us to come back to cleaning up OAuth stuff in the `session` in #2722.

thekaveman (changes requested):
I'm getting caught in a redirect loop running this locally, after signing in with Login.gov (Older Adult flow).

Are you able to reproduce, or is that my local config somehow?

benefits/oauth/views.py
```diff
@@ -85,71 +85,6 @@ def login(request):
...
-                claim_value = int(claim_value) if claim_value else None
-                if claim_value is None:
-                    logger.warning(f"userinfo did not contain: {claim}")
-                elif claim_value == 1:
-                    # if userinfo contains our claim and the flag is 1 (true), store the *claim*
-                    stored_claims.append(claim)
-                elif claim_value >= 10:
-                    error_claim[claim] = claim_value
-
-    session.update(request, oauth_authorized=oauth_token_authorized, oauth_claims=stored_claims)
```
thekaveman, line 147: Since we lost this `session.update()` call with `oauth_authorized=True`, Benefits doesn't "know" the user is signed in with the IdG / IdP.

I think we'd need to update the hook(s) to set this value. Or like I've been saying, just generally fix our session :sweat_smile:

thekaveman (changes requested):

benefits/oauth/views.py
angela-tran: Oh man, I was 100% under the impression that I had added that to the hooks in #2767 because that's what I was going to do when writing https://github.com/cal-itp/benefits/pull/2767#issuecomment-2737873588. But now I see with the back and forth about dependent tickets, I never did make that change.

Thanks for catching that, and I can add it here.

I absolutely want to get the session refactored to follow your new pattern soon, but I also feel like it's too much to try and do while also introducing the change of switching over to `cdt_identity`. It would feel less overwhelming to me to get things working first with `cdt_identity` and then finish it out with refactoring existing behavior.

benefits/oauth/hooks.py
```diff
@@ -64,6 +65,7 @@ def claims_verified_eligible(cls, request, claims_request, claims_result):
     )
     def claims_verified_not_eligible(cls, request, claims_request, claims_result):
         super().claims_verified_not_eligible(request, claims_request, claims_result)
+        session.update(request, oauth_authorized=False)
```
thekaveman, line 68: Confusingly, this also needs to be `True`: it is not saying "user was verified as eligible", it is saying "user successfully signed in"

thekaveman (changes requested):

benefits/oauth/hooks.py
thekaveman: Notice in the old code, we didn't have a path for setting it to `False`. It was always `True`.

angela-tran: Ah, yep - thanks for the catch :pray:  Agreed that it's confusing - maybe `logged_in` would be a less confusing variable name since that's the method where it's ultimately used. Anyway, not going to worry about the naming in this PR.

benefits/core/session.py
angela-tran: > Hmm let's look at this again. `session.context_dict()` is only used for the `context_processors.debug()` (unused in `test` and `prod`), it doesn't feel very critical to maintain.

Yeah, I'm also feeling that it would be better to cut our losses with `session.context_dict()`. I was hesitating to make that executive decision, but thinking about it again, we're just "losing" the convenience of looking at the claims in the debug bar... we can still look at the claims in the `OAuthSession` with a debugger.

> 
> I think my feedback wasn't so much about the one-liner / specific implementation. But the overall idea of our session maintaining stuff that the `cdt_identity` session now maintains, and wanting to get really narrow on the use-cases in Benefits. Do we absolutely need this old `oauth` related code in `session`?
> 
> I.e. how can we do more cleanup vs. maintaining stuff just because it is there.

I'll take a look at what can be removed now that we're saying we don't need claims in `session.context_dict`

angela-tran: Alrighty, so in https://github.com/cal-itp/benefits/pull/2783/commits/37857e867b28b7cbf357eb5928fb74cd81fdbe3a I removed claims from `session.context_dict` and refactored `session.oauth_extra_claims` to use the methods on `ClaimsVerificationRequest` and `ClaimsResult` to calculate the extra claims. This allowed me to remove `session._oauth_claims` completely.

benefits/oauth/hooks.py
angela-tran: I ended up renaming the variable in https://github.com/cal-itp/benefits/pull/2783/commits/fc8e3cae9e5fff60c04f1fbef8b52cf4b2b80395 - was easy enough and cleaned up some code too

benefits/oauth/hooks.py
```diff
@@ -48,6 +48,7 @@ def post_logout(cls, request):
     )
     def claims_verified_eligible(cls, request, claims_request, claims_result):
         super().claims_verified_eligible(request, claims_request, claims_result)
+        session.update(request, logged_in=True)
```
angela-tran, line 51: I suppose the new name is not as technically precise (the user has to make it through `authorize` to have this status set), but I think it makes more sense from a conceptual, high-level perspective?

benefits/core/session.py
```diff
@@ -143,34 +142,24 @@ def language(request):
...
-    claims = oauth_claims(request)
-    if claims:
-        f = flow(request)
-        if f and f.uses_claims_verification:
-            claims.remove(f.claims_request.eligibility_claim)
-            return claims
-        raise Exception("Oauth claims but no flow")
+    oauth_session = OAuthSession(request)
+    eligibility_claim = oauth_session.claims_request.eligibility_claim
+    requested_extra_claims = [claim for claim in oauth_session.claims_request.claims_list if claim != eligibility_claim]
```
angela-tran, line 159: I wonder if it's worth adding methods to `ClaimsVerificationRequest` to give us the `extra_claims` list instead of us having to calculate it

benefits/oauth/hooks.py
thekaveman: Agree with you, this is more clear

benefits/core/session.py
thekaveman: Agreed this makes a lot of sense to move to the `cdt_identity` session wrapper

angela-tran:
Re: the redirect issue we saw due to [`LoginRequired`](https://github.com/cal-itp/benefits/pull/2783/commits/0d5bfeada9260f7ee3b2c34c113fab6407aebc41#diff-c208a75e3dacd5c461defbc8e7874d36d24ae3de196fbf8848ee3833b5b48700L46):

it makes me wonder how useful this middleware is... if somehow `session.logged_in` isn't set, the app will just loop the user around like this...? I guess it makes it very obvious during development if we have incorrect code... (that is, if you run the app)

thekaveman (approved):
It works :+1: :rocket:

thekaveman:
> it makes me wonder how useful this middleware is...

Looks like it is still used on `eligibility.views.verified` ... not sure why though? Since that view function is not routable i.e. there is not a URL corresponding to it. User shouldn't be able to get there without the app directing them.

benefits/core/session.py
angela-tran: Created https://github.com/Office-of-Digital-Services/django-cdt-identity/issues/50

angela-tran:
> > it makes me wonder how useful this middleware is...
> 
> Looks like it is still used on [`eligibility.views.verified`](https://github.com/cal-itp/benefits/pull/2783/files#diff-bd503c63c9a7b07070e81c9df074adafc509da84db6e270d3c91c54e282218b2R121) ... not sure why though? Or how useful that is. Since that view function is not routable i.e. there is not a URL corresponding to it. User shouldn't be able to get there without the app directing them.

Follow-up ticket to dump it? :broom: :wastebasket:

PR #2785, opened 2025-03-26
Fix: `EnrollmentFlowForm` test and validation

angela-tran:
@lalver1 pointed out that `test_EnrollmentFlowForm_clean_supports_expiration` started failing due to some model refactoring in #2721 and that it was not clear why the test still fails even after making the expected changes to reflect the refactor.

It turns out the test itself had a mistake, namely that the field for the eligibility claim was not correct. 

Correcting that revealed another bug which is that if you had both field-errors and non-field errors during form validation, you would lose the non-field errors.

Lastly, we also noticed that trying to re-use the same test for both `super` and `staff` parameters didn't actually make sense because `staff` users cannot actually POST some fields, like `reenrollment_error_template`. To address this, we decided to split the test out.

These changes are needed for Luis to finish out #2721.

### DIFF
--- a/benefits/core/admin/enrollment.py
+++ b/benefits/core/admin/enrollment.py
@@ -125,9 +125,11 @@ class EnrollmentFlowForm(forms.ModelForm):
                 non_field_errors.append(ValidationError(message))
 
         if field_errors:
-            raise ValidationError(field_errors)
+            for field_name, validation_error in field_errors.items():
+                self.add_error(field_name, validation_error)
         if non_field_errors:
-            raise ValidationError(non_field_errors)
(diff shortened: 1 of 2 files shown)


lalver1 (approved):
This looks great @angela-tran! The tests all pass, I tested the admin's behavior locally, and I got `field` and `non-field` validation errors where I was expecting them.
The lines
```python
for field_name, validation_error in field_errors.items():
    self.add_error(field_name, validation_error)
for validation_error in non_field_errors:
    self.add_error(None, validation_error)
```
also help with making the code read more clearly 👍

Issue #2788, opened 2025-03-26
OAuth refactor: use `post_logout` and `cancel` views from `cdt_identity.views`


angela-tran:
Combining these two into one task since they're relatively small.

## Acceptance Criteria

- [ ] Wire up the `post_logout` and `cancel` URL paths to the `cdt_identity.views.post_logout` and `cdt_identity.views.cancel` views (see notes on #2724)
- [ ] Delete unused code (see notes on #2725)

PR #2791, opened 2025-03-27
Refactor: use cdt_identity `post_logout` view

angela-tran:
Part of #2788 

Replaces Benefits oauth `post_logout` with cdt_identity `post_logout` view.

### Hook gets run

Run through a Login.gov flow, and click on any links for logging out. You should see that the `post_logout` hook is called.

### Middleware still works

Choose a non-Login.gov flow (e.g. Agency Card), and then attempt to go to `oauth/post_logout`. The `FlowUsesClaimsVerificationSessionRequired` middleware should get called and end up showing you the user error screen.

### DIFF
--- a/benefits/oauth/urls.py
+++ b/benefits/oauth/urls.py
@@ -19,6 +19,11 @@ urlpatterns = [
     ),
     path("cancel", views.cancel, name=routes.name(routes.OAUTH_CANCEL)),
     path("logout", views.logout, name=routes.name(routes.OAUTH_LOGOUT)),
-    path("post_logout", views.post_logout, name=routes.name(routes.OAUTH_POST_LOGOUT)),
+    path(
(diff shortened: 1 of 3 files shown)


angela-tran:
Putting in draft so I can look at the tests a little closer

angela-tran:
Was able to run through the scenarios in the app locally. Ready for review

thekaveman (approved):
> Run through a Login.gov flow, and click on any links for logging out. You should see that the post_logout hook is called.

I was able to do this, and I put a breakpoint and verified that the hook method was being called.

At first I didn't see any log output indicating the hook was called. Which surprised me... then I realized I had in my `.env`: `DJANGO_LOG_LEVEL=WARNING` and those are `DEBUG` level messages :sweat_smile:

PR #2793, opened 2025-03-27
Refactor: use cdt_identity `login` view

### LINKED ISSUE #2786: OAuth refactor: use `cdt_identity.views.login`
## Acceptance Criteria

- [ ] Wire up the `login` URL path to the `cdt_identity.views.login` view (see notes on #2724)
- [ ] Delete unused code (see notes on #2725)

lalver1:
Closes #2786 

As part of the oauth refactor, this PR replaces `benefits.oauth.views.login` with `cdt_identity.views.login`.

## Reviewing

Run through a login.gov flow and ensure you can confirm your eligibility. Also notice that when you click on the `Get started with LOGIN.GOV` CTA button, the `cdt_identity` `login` view is called.  In addition, a `started sign in` analytics event will be created as part of the `pre_login` hook.

### DIFF
--- a/benefits/urls.py
+++ b/benefits/urls.py
@@ -27,6 +27,7 @@ urlpatterns = [
     path("enrollment/", include("benefits.enrollment.urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("oauth/", include("benefits.oauth.urls")),
+    path("oauth/", include("cdt_identity.urls")),
     path("in_person/", include("benefits.in_person.urls")),
 ]
 
--- a/tests/pytest/oauth/test_views.py
+++ b/tests/pytest/oauth/test_views.py
@@ -11,7 +11,6 @@ from benefits.core.middleware import TEMPLATE_USER_ERROR
 from benefits.oauth.views import (
     TEMPLATE_SYSTEM_ERROR,
     _oauth_client_or_error_redirect,
-    login,
     logout,
     system_error,
 )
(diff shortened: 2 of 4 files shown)


benefits/urls.py
```diff
@@ -27,6 +27,7 @@
     path("enrollment/", include("benefits.enrollment.urls")),
     path("i18n/", include("django.conf.urls.i18n")),
     path("oauth/", include("benefits.oauth.urls")),
+    path("oauth/", include("cdt_identity.urls")),
```
angela-tran, line 30: Can you say more about what this is doing and why it's needed? I read the commit message on https://github.com/cal-itp/benefits/pull/2793/commits/f03facdc02e0ee31f47e099d1a6e16a8ee3a3fae but am not following :thinking:

tests/pytest/oauth/test_views.py
```diff
@@ -99,76 +98,6 @@ def test_oauth_client_or_error_oauth_client(
...
-    assert result.url == reverse(routes.OAUTH_SYSTEM_ERROR)
-    mocked_view_analytics_module.error.assert_called_once()
-
-
-@pytest.mark.django_db
-def test_login_no_session_flow(app_request):
-    result = login(app_request)
-
-    assert result.status_code == 200
-    assert result.template_name == TEMPLATE_USER_ERROR
```
angela-tran, line 117: I think we should keep these tests since what they cover is still valid. The first is covering what happens when the client is not configured correctly, and the second covers when there is no flow in the session.

The main thing here is to change these tests so that they aren't calling the view function directly. So we'd have them use `client.get` with the path instead. 

And then of course, their assertions need to be tweaked to reflect the fact that now it's the `analytics` module imported in `hooks` that will be called, etc.

I saw the same thing when refactoring the `authorize` view and the `post_logout` view, so I think you could reference these diffs:
- [tests for `authorize`](https://github.com/cal-itp/benefits/pull/2783/files#diff-b0a61b497a273f9f6798551e0de1ea95c6faac12281cddff3277b234b97222d8R174-R190)
- [tests for `post_logout`](https://github.com/cal-itp/benefits/pull/2791/files#diff-b0a61b497a273f9f6798551e0de1ea95c6faac12281cddff3277b234b97222d8R249)

tests/pytest/oauth/test_views.py
```diff
@@ -99,76 +98,6 @@ def test_oauth_client_or_error_oauth_client(
...
-    assert not session.logged_in(app_request)
-    mocked_oauth_client = mocked_oauth_client_or_error_redirect__client.return_value
-    # fake a permanent redirect response from the client
-    mocked_oauth_client.authorize_redirect.return_value.status_code = 301
-
-    login(app_request)
-
-    mocked_oauth_client.authorize_redirect.assert_called_with(app_request, "https://testserver/oauth/authorize")
-    mocked_view_analytics_module.started_sign_in.assert_called_once()
-    assert not session.logged_in(app_request)
```
angela-tran, line 132: This one I think is difficult to keep.

The test would be good to keep because it covers that our session `logged_in` variable should be updated when the `login` view is called, but the problem is that in order to get through the `login` view successfully, we'd have to mock internal implementation details of the cdt_identity `login` view.

There are unit tests for the `OAuthHook` methods that assert `session.logged_in` was updated, and cdt_identity itself has unit tests to assert that the hooks were called in the right places, so overall we have enough coverage to justify not having this test.

benefits/urls.py
thekaveman, line 30: I don't think we want this (yet) -- right?

tests/pytest/oauth/test_views.py
```diff
@@ -99,76 +98,6 @@ def test_oauth_client_or_error_oauth_client(
...
-):
-    mocked_oauth_client = mocked_oauth_client_or_error_redirect__client.return_value
-    mocked_oauth_client.authorize_redirect.return_value.status_code = status_code
-
-    result = login(app_request)
-
-    assert result.status_code == 302
-    assert result.url == reverse(routes.OAUTH_SYSTEM_ERROR)
-    mocked_view_analytics_module.error.assert_called_once()
-    mocked_sentry_sdk_module.capture_exception.assert_called_once()
```
angela-tran, line 169: These are also fine to remove with the same reasoning as above about not wanting to mock an external package's internal implementation.

angela-tran (changes requested):

benefits/urls.py
lalver1: Yep, if I don't include this line, when I run through the app I get this exception

image

when the `reverse` is evaluated on [line 132](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L132) of the `cdt_identity` `login` view.

This is what I see as I debug:
- On [line 132](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/views.py#L132) `Routes.routes_authorize` is `cdt:authorize`.
- To evaluate the `reverse`, Django goes through `urlpatterns` in the root [`urls`](https://github.com/cal-itp/benefits/blob/main/benefits/urls.py) but can't find a url with a name of `cdt:authorize` (I think that's the `'cdt' is not a registered namespace` error).

So the solution I tried was 
- First I added `path('oauth/', include(('cdt_identity.urls', 'cdt'), namespace='cdt'))` to `urlpatterns` in the root [`urls`](https://github.com/cal-itp/benefits/blob/main/benefits/urls.py) and it would fix the problem but I thought that the `namespace` parameter shouldn't really be needed since [`cdt_identity urls`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/urls.py) already defines `app_name = "cdt"` and this I think is the [namespace](https://docs.djangoproject.com/en/5.1/topics/http/urls/#url-namespaces).
- So I simply added `path("oauth/", include("cdt_identity.urls"))` after Benefits' `oauth` path (that will eventually get removed) so that `reverse` can actually evaluate to `oauth/authorize` which is what we need.

lalver1: I thought so too, that we wouldn't need it but I came across the exception described in this [comment](https://github.com/cal-itp/benefits/pull/2793#discussion_r2024897356)

tests/pytest/oauth/test_views.py
lalver1: Thanks! I updated the tests and closely followed the references since the testable logic of the `login` view is very similar. In `test_login_oauth_client_init_error`, test the `_client_or_error` when it has an error and in `test_login_no_session_flow` test when there is no flow in the session.

benefits/urls.py
angela-tran: Ok, I see. Thanks for the clarification.

So really the disconnect here is that `cdt_identity` is making an assumption of what route to use to redirect to the `authorize` view.

As we're seeing with Benefits, a consuming app can set up the routes to cdt_identity's views with some other namespace that is not `cdt`.

So the question is: are we wanting to keep `oauth` as the namespace? An argument for keeping it is so that this refactor is completely transparent to a user, and by that, I really just mean that the URL stays the same. I think that's the only thing the user sees that would be changed if we changed the namespace to `cdt`.

But if we want to keep it, I think we'll need to make a change to `cdt_identity` to allow consuming apps to override the route.

angela-tran: Also, if we're using `cdt_identity.urls` so that we can get to `cdt:authorize`, that means we're using a URLConf that doesn't use `OAuthHooks`, right?

angela-tran: > ... I don't know how exactly Django is resolving `cdt:authorize` to the URL pattern that passes in `{ "hooks": hooks.OAuthHooks }`, but looks like it is working.

I learned more about how this happens to work.

When you have duplicate URL patterns defined, this is the behavior:

> As documented, when you have two identical paths (or a path that matches multiple patterns) in your urlpatterns, Django will dispatch to the first one that matches.
> 
> When you have two identical names for those paths, the reverse method will return the _last_ such definition for that name.

according to [this Django forum post](https://forum.djangoproject.com/t/duplicate-urls-and-url-names/5999).

So I believe this URLconf is "working" because:
- cdt_identity calls `reverse("cdt:authorize")`
- Django looks at `benefits/urls.py` and evaluates the reversed route as matching the URL pattern with namespace `cdt`, and the URL is `oauth/authorize`:
```
    path("oauth/", include("benefits.oauth.urls")),
    path("oauth/", include("cdt_identity.urls")),
```
- Django makes a request (I guess?) to `oauth/authorize` which goes to the _first_ of the duplicated URL patterns

Anyways, all of this feels really convoluted and ultimately is because our root URLconf and `benefits.oauth` URLconfs are in an awkward, in-between state.

Maybe we say this is ok for this PR, and then immediately clean this up in #2813.

angela-tran (dismissed):

lalver1:
If I can get a re-review when you get a chance @angela-tran, thanks! I had to rebase and fix 2 conflicts so I put it back in draft while I was getting those fixes in.

angela-tran (approved):

PR #2794, opened 2025-03-27
Chore: enable debugging of external python modules

lalver1:
With this option set on the debugger's launch configuration we can set breakpoints in external modules and debug them.


angela-tran (dismissed):
I think we could add it to the `Debug=False` config as well, but not a blocker

lalver1:
Thanks! Added it to the `Debug=False` config also 👍

angela-tran (approved):

PR #2795, opened 2025-03-27
Feat(session): set config for cdt_identity when setting flow

angela-tran:
Part of #2722 

This PR adds the session flow's `oauth_config` and `claims_request` to the session as well.

### DIFF
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -8,6 +8,7 @@ import logging
 import time
 import uuid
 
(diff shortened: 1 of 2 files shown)


lalver1 (approved):
👍

angela-tran:
I realized it would be good to update the unit test for `update(flow=...)`. Ready for re-review @lalver1

lalver1 (approved):
Looks good to me 👍

PR #2796, opened 2025-03-28
Admin: Fixes to Log In/Out and Agency Dashboard page

machikoyasuda:
closes #2790 

This PR ended up getting a little bigger than the original issue. The theme of this PR is "small visual fixes" on these 2 pages that are now most prominent to current Agency users. This PR does _not_ touch any of the "flow" pages.

## How to test this PR
- Test this PR with reCAPTCHA turned _on_ to get the reCAPTCHA warning text

### Log In/Log Out
- CTA Button for Log In/Log Out and Sign in with Google: Border radius, font size, icon size and resulting total button height now match Figma - 3rem (48px) height buttons with 1rem (16px) font size text, centered with border-radius 2px.
image

- Make the external link icon 1rem, to match the font size
image
image
image

### Dashboard page
- CTA Button: Border radius, font size, and resulting total button height now match Figma
- Spacing above the transit agency name now matches Figma
- Spacing above the button now matches Figma
- Header nav text used to have a font weight too light. Now matches Figma.
image

PR #2798, opened 2025-03-28
Admin: Add Noto font

### LINKED ISSUE #2741: Admin: Use same font-family-primary
This ticket is an in-between until the full typeface/scale design work is complete. ~We are not switching to Noto yet.~ We *are* switching to Noto.

## Acceptance criteria
- Replace `--font-family-primary` on Django data theme with the same of the current Admin (Bootstrap's)

machikoyasuda:
closes #2741 

Currently, you might notice that when screensharing the Admin app with someone on a different operating system, the app looks different than yours. That's because the font is set to a user's system font. This can be confusing and delivers a disjointed experience for transit agency operators who may be using this app across various hardware (iPads, tablets, desktops) and browsers.

Originally, I had thought that it would be complicated to add Noto as a font for all the different parts of Admin - so I wanted to hold off on it. But now that all of Admin is on Bootstrap (#2766), I realized that we can easily make all of Admin use 1 font, with just changing one thing - Bootstrap's `--bs-font-family` variable in the `theme.css` file. I imported the 400, 500 and 700 font weights, because that is what Admin uses. Admin Figma uses only 400 and 700, but Django Admin uses 500 for the table headers. 

image
image
image
image

Note: This PR looks a lot better when you are looking at it as merged on top of #2796 - which these screenshots do _not_ reflect.

### DIFF
--- a/benefits/templates/admin/base.html
+++ b/benefits/templates/admin/base.html
@@ -15,6 +15,9 @@
 
 {% block extrahead %}
   <link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
+  <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;500;700&display=swap" rel="stylesheet">
 {% endblock extrahead %}
 
 {% block branding %}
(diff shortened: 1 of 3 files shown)


benefits/templates/admin/base.html
```diff
@@ -15,6 +15,9 @@
 
 {% block extrahead %}
   <link href="{% static "img/favicon.ico" %}" rel="icon" type="image/x-icon" />
+  <link rel="preconnect" href="https://fonts.googleapis.com">
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin="anonymous">
```
angela-tran, line 19: Nice :+1:

angela-tran (approved):
This looks good to me!

thekaveman (approved):
Looks so clean :star_struck:

PR #2819, opened 2025-04-03
Refactor: use cdt_identity `logout` view

lalver1:
Closes #2787

As part of the oauth refactor, this PR replaces `benefits.oauth.views.logout` with `cdt_identity.views.logout`.

## Reviewing

Run through a login.gov flow and ensure you can confirm your eligibility. On the eligibility confirmation page, click on `Sign out of Login.gov ` and check that the `cdt_identity` `logout` view is called.  In addition, a `started sign out` analytics event will be created as part of the `pre_logout` hook; and a `finished sign out` analytics event will be created as part of the `post_logout` hook.

PR #2821, opened 2025-04-04
Refactor: OAuth URLconf

### LINKED ISSUE #2813: OAuth refactor: clean up URLconf to cdt_identity views
> There's going to be some duplication of declaring the decorators and the `{ "hook": OAuthHooks }` kwargs as we implement the various view tickets - #2786, #2787, #2788 - but I think that's ok to start with, and we can clean it up once it's all in.

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/pull/2783#discussion_r2017723420_

angela-tran:
Closes #2813 

~Builds off #2819~

In https://github.com/cal-itp/benefits/pull/2793, we uncovered a need for the `oauth/authorize` route to have a namespace of `cdt`. See https://github.com/cal-itp/benefits/pull/2793#discussion_r2023782407 for more details. The temporary solution was to add a duplicate URL pattern that pointed to `cdt_identity.urls`.

@thekaveman, @lalver1, and I took a closer look at this today and figured out a cleaner solution, which is to make the URL patterns in `benefits.oauth.urls` have a namespace of `cdt`. Then there is no need for the duplicate.

We also cleaned up duplication in `benefits.oauth.urls` by extracting repeated values into variables.

### DIFF
--- a/benefits/oauth/urls.py
+++ b/benefits/oauth/urls.py
@@ -1,6 +1,7 @@
 from cdt_identity import views as cdt_identity_views
 from cdt_identity.routes import Routes
 from cdt_identity.urls import app_name as cdt_app_name
+
 from django.utils.decorators import decorator_from_middleware
 from django.urls import path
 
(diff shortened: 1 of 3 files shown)


angela-tran:
Putting this into draft until #2819 is merged in

lalver1 (approved):
This looks great! 👍 
Went through a login.gov flow, logged out, and tested `/oauth/error` ✅

PR #2824, opened 2025-04-04
Refactor: remove unnecessary usage of `translate`

### LINKED ISSUE #2760: Simplify templates: remove unnecessary `translate`
Since we already use the `gettext()` helper (aliased as `_()`) setting up the context, to mark the string as translatable -- do we _also_ need to translate the string in the template with this `{% translate %}` (or `{% blocktranslate %}` tag?

I think the answer, from #2753 anyway, seems to be: no. You can just render the variable directly in the template and it will be translated as necessary. E.g. this template change: https://github.com/cal-itp/benefits/pull/2753/files#diff-70ed6fd5fa2321e83db0d63df8312f680cd40428de4ada50dab22e93f55db7c7R109

<details>
<summary>Translation still works</summary>
image
</details>

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/2726#discussion_r1999642995_

angela-tran:
Closes #2760 

Removes unneeded `translate`s from Agency Index and Eligibility Start.

The [ones on Eligibility Index](https://github.com/cal-itp/benefits/pull/2824/commits/b246fb8d178c99f29954da8c9bd8e6a3a739e1fe#diff-1e3c02710205416114dcc25094079f5cdd0a6d71b504924fe9c6d0d215612ef1L24-L26) are needed because, for some reason, removing it causes translation to break (i.e. switching to Spanish does not show Spanish copy. The English copy sticks around.) Seems like something to do with the `%` character.

### DIFF
--- a/benefits/core/templates/core/index--agency.html
+++ b/benefits/core/templates/core/index--agency.html
@@ -6,7 +6,7 @@
 {% endblock title %}
 
 {% block headline %}
-  {% translate headline %}
+  {{ headline }}
 {% endblock headline %}
 
 {% block call-to-action %}
--- a/benefits/eligibility/templates/eligibility/index.html
+++ b/benefits/eligibility/templates/eligibility/index.html
@@ -21,9 +21,9 @@
 {% block inner-content %}
   {% for text in form_text %}
     {% if not forloop.last %}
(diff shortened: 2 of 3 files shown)


lalver1 (approved):
Looks good to me! 👍

PR #2834, opened 2025-04-08
Refactor: delete unneeded LoginRequired middleware

### LINKED ISSUE #2799: Look into removing `LoginRequired` middleware
> Re: the redirect issue we saw due to [`LoginRequired`](https://github.com/cal-itp/benefits/pull/2783/commits/0d5bfeada9260f7ee3b2c34c113fab6407aebc41#diff-c208a75e3dacd5c461defbc8e7874d36d24ae3de196fbf8848ee3833b5b48700L46):
>
>it makes me wonder how useful this middleware is... if somehow `session.logged_in` isn't set, the app will just loop the user around like this...? I guess it makes it very obvious during development if we have incorrect code... (that is, if you run the app)
>

_Originally posted by @angela-tran in https://github.com/cal-itp/benefits/issues/2783#issuecomment-2762576103_

Also, from https://github.com/cal-itp/benefits/pull/2783#issuecomment-2762596826:

> Looks like it is still used on [`eligibility.views.verified`](https://github.com/cal-itp/benefits/pull/2783/files#diff-bd503c63c9a7b07070e81c9df074adafc509da84db6e270d3c91c54e282218b2R121) ... not sure why though? Or how useful that is. Since that view function is not routable i.e. there is not a URL corresponding to it. User shouldn't be able to get there without the app directing them.

angela-tran:
Closes #2799 

Previously, the main purpose of this middleware was to guard the `confirm` view. Basically, if a user selected a claims-verification flow and then went directly to `eligibility/confirm`, this middleware would redirect them to `oauth/login` to force them to go through the flow correctly.

We also had `LoginRequired` guarding the `verified` view, but this was more just an extension of it guarding the `confirm` view -- it's not possible for the user to reach the `verified` view directly.

Since we removed all claims-verification logic from the `confirm` view in 56ff1a1f and e3011b24, this middleware isn't useful any more and can be removed.


thekaveman (approved):
Thank you for the detailed explanation in _both_ the PR description _and_ commit message!

lalver1:
Thanks for the explanation! I went through a few flows and tested that the app works as expected just to double check on my side. 
Just one question, would we want to remove the `LoginRequired` definition from https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L128 too? I did a search of the codebase and it looks like we don't use it anywhere else.

angela-tran:
> Thanks for the explanation! I went through a few flows and tested that the app works as expected just to double check on my side. Just one question, would we want to remove the `LoginRequired` definition from https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L128 too? I did a search of the codebase and it looks like we don't use it anywhere else.

I had meant to remove it but forgot, so thanks for the catch @lalver1!

lalver1 (approved):
Looks great!

Issue #2800, opened 2025-03-28
Refactor `session` to follow wrapper class pattern


angela-tran:
The Benefits app currently encapsulates interaction with the [Django session framework](https://docs.djangoproject.com/en/5.1/topics/http/sessions/) via the `benefits.core.session` module, which contains functions to retrieve session data and a single `update` function to update session data, as well as some other helper functions. This module serves as the single place where all session-handling logic is done, and we've noticed this is becoming harder to work with.

A new pattern was established in [Office-of-Digital-Services/django-cdt-identity](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/session.py) that follows a more class-based approach.

In the same spirit as https://github.com/cal-itp/benefits/issues/2778 and https://github.com/cal-itp/benefits/issues/2777, i.e. in the spirit of having smaller, more specific, self-contained and cohesive organization, we want to refactor the current `session` module into the new pattern. This could allow us to make use of inheritance or other language features to better organize the interaction with `request.session`. The [`Session`](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/blob/main/web/core/session.py) in [Office-of-Digital-Services/cdt-ods-disaster-recovery](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery) is an example of this.

## Acceptance Criteria

- [ ]

PR #2845, opened 2025-04-11
Fix: check if flow uses claims verification

lalver1:
This PR brings back a check to avoid trying to access `eligibility_claim` on a flow that does not use claims verification. Without this fix, an exception would be raised when trying to access `eligibility_claim` on an agency card flow, for example:

image

The exception would occur on [line 74 of enrollment index ](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/views.py#L74)when `extra_claims` would be requested from the `session`.
## Reviewing

Run through an agency card and a login.gov flow (all the way to the end) and ensure that you can successfully enroll.

### DIFF
--- a/benefits/core/session.py
+++ b/benefits/core/session.py
@@ -154,12 +154,14 @@ def logout(request):
 
 def oauth_extra_claims(request):
     """Get the extra oauth claims from the request's session, or None"""
-    oauth_session = OAuthSession(request)
-    eligibility_claim = oauth_session.claims_request.eligibility_claim
-    requested_extra_claims = [claim for claim in oauth_session.claims_request.claims_list if claim != eligibility_claim]
+    claims = [claim for claim, value in OAuthSession(request).claims_result.verified.items() if value]
 
-    if oauth_session.claims_result:
-        return [extra_claim for extra_claim in requested_extra_claims if extra_claim in oauth_session.claims_result]
+    if claims:
+        f = flow(request)
+        if f and f.uses_claims_verification:
+            claims.remove(f.claims_request.eligibility_claim)
+            return claims or None
+        raise Exception("Oauth claims but no flow")
     else:
         return None
 
--- a/tests/pytest/core/test_session.py
+++ b/tests/pytest/core/test_session.py
@@ -474,4 +474,11 @@ def test_oauth_extra_claims_no_claims(app_request, model_EnrollmentFlow_with_sco
     session.update(app_request, flow=model_EnrollmentFlow_with_scope_and_claim)
     OAuthSession(app_request).claims_result = ClaimsResult()
 
-    assert session.oauth_extra_claims(app_request) == []
+    assert session.oauth_extra_claims(app_request) is None
+
+
+@pytest.mark.django_db
+def test_oauth_extra_claims_claims_no_flow(app_request):
+    OAuthSession(app_request).claims_result = ClaimsResult(verified={"eligibility_claim": True, "extra_claim": True})
+    with pytest.raises(Exception, match="Oauth claims but no flow"):
+        session.oauth_extra_claims(app_request)


lalver1:
The tests are not covering all the PR, so still looking at that, the one I parametrized seems convoluted but I was trying to get full PR coverage with it.
The actual fix also smells because there is no `Exception("Oauth claims but no flow")` anymore, but maybe this is actually ok because the `if f and f.uses_claims_verification:` around all the code inside oauth_extra_claims is guarding us from this case.

lalver1:
**Update**: As soon as the [Pillow update PR](https://github.com/cal-itp/benefits/pull/2846) is merged, I'll rebase this and mark it as ready for review. I made a few updates to this PR since the last two comments so please disregard the previous comments (I didn't cross them out since some parts of them still apply and they may be helpful for context when reviewing).

benefits/core/session.py
```diff
@@ -154,12 +154,20 @@ def logout(request):
...
-    eligibility_claim = oauth_session.claims_request.eligibility_claim
-    requested_extra_claims = [claim for claim in oauth_session.claims_request.claims_list if claim != eligibility_claim]
+    f = flow(request)
 
-    if oauth_session.claims_result:
-        return [extra_claim for extra_claim in requested_extra_claims if extra_claim in oauth_session.claims_result]
+    if f and f.uses_claims_verification:
+        oauth_session = OAuthSession(request)
+        eligibility_claim = oauth_session.claims_request.eligibility_claim
+        requested_extra_claims = [claim for claim in oauth_session.claims_request.claims_list if claim != eligibility_claim]
```
thekaveman, line 162: I'm struggling with this logic a little bit.

We know this code is only used for the `EnrollmentEvent` and the `enrollment: returned success` analytics event.

In the previous iteration of this function, we had:

```py
def oauth_extra_claims(request):
    """Get the extra oauth claims from the request's session, or None"""
     claims = oauth_claims(request)
     if claims:
         f = flow(request)
         if f and f.uses_claims_verification:
             claims.remove(f.claims_request.eligibility_claim)
             return claims
         raise Exception("Oauth claims but no flow")
     else:
         return None
```

We can see this used the (now deleted) `session.oauth_claims()` function, which used to look like:

```py
def oauth_claims(request):
    """Get the oauth claims from the request's session, or None"""
    return request.session.get(_OAUTH_CLAIMS)
```

So a couple things:

1. This function only ever cared about _verified claims_, and that should continue to be the case
2. We can see a much simpler implementation from before: get the list of all claims, remove the eligibility claim, done.

thekaveman (changes requested):

benefits/core/session.py
thekaveman: See e.g. https://github.com/cal-itp/benefits/commit/7728119c7ded5b66a630f9e5292de5c93cb4cc66

angela-tran (approved):
This looks great! I tested this locally :white_check_mark: Thanks for catching and fixing this :pray: 

_edit_: just now seeing Kegan's comment

benefits/core/session.py
angela-tran: I think it's fair to request that we make this logic easier to follow, but just to check my understanding, this function _is_ getting us the _verified claims_, right? Because of the one-liner on lines 165-167](https://github.com/cal-itp/benefits/pull/2845/files/5f6d021155c4c93d35f1b58ea478187881248644#diff-8d7012c10c06c3b595e41efdc1b9ab823ebdf38756ce075be504e01d536336b7R165-R167)

lalver1: Yep, as I run through `oauth_extra_claims`, I think it still is returning the verified claims @angela-tran, and I also agree with making the logic easier to follow. To make the logic easier to follow I re-implemented the approach in https://github.com/cal-itp/benefits/commit/7728119c7ded5b66a630f9e5292de5c93cb4cc66 (thanks for the reference @thekaveman) with one small change
```python
if f and f.uses_claims_verification:
    claims.remove(f.claims_request.eligibility_claim)
    return claims or None # <- small change
raise Exception("Oauth claims but no flow")
```
because the function definition says it returns `None` if there are no extra claims, not an empty list. Even though when we use the function (and only in enrollment index) if it returns `[]` or `None` doesn't make a difference, I made this small change for clarity and to align with the docstring of the function.
I'm updating the tests and will push an updated branch for a re-review as soon as I'm done, thanks for the feedback!

thekaveman: > but just to check my understanding, this function is getting us the verified claims, right?

You are correct. My comment was more about, why are we even looking at the requested claims here? ("cared about") That is not the point of this function.

thekaveman: > Even though when we use the function (and only in enrollment index) if it returns [] or None doesn't make a difference, I made this small change for clarity and to align with the docstring of the function.

The code doesn't need to follow the comments, it should be the other way around! What is the correct thing to return here, `None` or `[]`?

thekaveman: The implementation looks good though :+1:

benefits/core/session.py
```diff
@@ -152,14 +152,22 @@ def logout(request):
     update(request, logged_in=False, enrollment_token=False)
 
 
+def oauth_claims(request):
+    """Get the oauth claims from the request's session, or None"""
+
+    return [claim for claim, value in OAuthSession(request).claims_result.verified.items() if value] or None
+
+
```
thekaveman, line 160: We did not want to reintroduce this function, it is not needed outside of the usage in `oauth_extra_claims`. We are trying to simplify usage of `session`

* #2800
* https://github.com/Office-of-Digital-Services/django-cdt-identity/issues/50

thekaveman (changes requested):

benefits/core/session.py
thekaveman: To be clear: keep the function deleted as it was already, simplify `session.oauth_extra_claims` as we talked about in the other thread. We'll come back to those session tickets later.

thekaveman: You can see we had a long PR discussion about this previously: https://github.com/cal-itp/benefits/pull/2783#discussion_r2014560634

lalver1: Ah, that makes sense. I removed `session.oauth_claims` and just inlined its behavior into `session.oauth_extra_claims`. The rest of `session.oauth_extra_claims` was simplified so I think this is probably ready.

thekaveman (approved):
:+1:

Issue #2838, opened 2025-04-10
Switchio integration: model changes to store info about Switchio APIs


angela-tran:
## Acceptance Criteria

- [ ] Fields for Switchio API credentials are introduced to our models
- [ ] ~~The `TransitProcessor` model conveys which system (Littlepay or Switchio) and environment (QA or prod) it is for~~
   - This will be done in #2840  - see https://github.com/cal-itp/benefits/pull/2854#issuecomment-2815831678

## Additional context

Currently we have a `TransitProcessor` model which stores info about APIs needed for Littlepay (API URL, tokenization Javascript URL). We also have some fields on `TransitAgency` that store the credentials for a particular agency to be used with its associated `TransitProcessor`.

We need to make some change to the models so that we can also store info about Switchio APIs. We also need the model to indicate which implementation to use for enrollment.

PR #2854, opened 2025-04-18
Refactor: Littlepay config

angela-tran:
Part of #2838 

Supersedes #2852 

## What this PR does

This PR extracts out the `transit_processor_` fields from `TransitAgency` out to a separate model called `LittlepayConfig`. This is needed so we can define a similar model called `SwitchioConfig`, which will be done in a separate PR.

Validation works similarly as before, where saving an active transit agency will validate that API credentials have been filled out.

## Post-merge actions
In `dev`,
- [ ] Set the `environment` for each `LittlepayConfig` to the correct value

For `test` and `prod`,
- [ ] Create configuration ticket to do the same whenever the next deployment / release happens

## Screenshots
<details>

Non-field error is shown if the `LittlepayConfig` is missing fields when saving back an active agency.

Screenshot from 2025-04-18 11-19-03

Both non-field errors and field errors will be shown, if any.

Screenshot from 2025-04-18 11-19-10

Validation also checks that the active agency has a `LittlepayConfig`.

image

If you try to clear out the fields on a `LittlepayConfig` currently being used by an active agency, that will also be validated and not allowed:
image


angela-tran:
Ready for review

benefits/core/models/transit.py
```diff
@@ -27,6 +27,67 @@ def agency_logo_large(instance, filename):
     return _agency_logo(instance, filename, "lg")
 
 
+class TransitProcessorConfig(models.Model):
```
thekaveman, line 30: I'm sort of confused what this gets us, other than not having to redefine the single `environment` field on each model? 

I.e. the `Environment` class can be defined at the top-level and used by both `*Config.environment` fields. Is there some other advantage with the ABC?

benefits/core/models/transit.py
```diff
@@ -27,6 +27,67 @@ def agency_logo_large(instance, filename):
...
+            needed = dict(audience=self.audience, client_id=self.client_id, client_secret_name=self.client_secret_name)
+            field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
+
+        if field_errors:
+            raise ValidationError(field_errors)
+
+    def __str__(self):
+        environment_label = self.Environment(self.environment).label if self.environment else "unknown"
+        agency_slug = self.agency_slug if self.agency_slug else "(no agency)"
+        return f"({environment_label}) {agency_slug}"
```
thekaveman, line 88: Maybe to answer my own [question above](), `agency_slug` and the `__str__` implementation could be moved to the ABC?

This too could be cleaned up if part / all of it was moved into the ABC

```python
environment_label = self.Environment(self.environment).label if self.environment else "unknown"
```

benefits/core/models/transit.py
```diff
@@ -208,17 +262,24 @@ def clean(self):
...
+                    message = "Littlepay configuration is missing fields that are required when this agency is active."
+                    message += f" Missing fields: {', '.join(e.error_dict.keys())}"
+                    non_field_errors.append(ValidationError(message))
+
+        all_errors = {}
         if field_errors:
-            raise ValidationError(field_errors)
+            all_errors.update(field_errors)
+        if non_field_errors:
+            all_errors.update({NON_FIELD_ERRORS: value for value in non_field_errors})
```
thekaveman, line 280: This is really cool :raised_hands:

tests/pytest/core/admin/test_users.py
```diff
@@ -148,6 +148,8 @@ def test_pre_login_user_add_transit_staff_to_transit_staff_group(mocker, setting
 
     transit_agency_staff_group = Group.objects.create(name="CST Staff")
     model_TransitAgency.pk = None
+    model_TransitAgency.littlepay_config.pk = None
+    model_TransitAgency.littlepay_config = model_TransitAgency.littlepay_config.save()
```
thekaveman, line 152: I'm confused by this? `.save()` doesn't return anything right?

tests/pytest/core/test_views.py
```diff
@@ -39,6 +39,8 @@ def test_index_multiple_agencies(model_TransitAgency, client):
     # https://stackoverflow.com/a/48149675/453168
     new_agency = TransitAgency.objects.get(pk=model_TransitAgency.id)
     new_agency.pk = None
+    new_agency.littlepay_config.pk = None
+    new_agency.littlepay_config = new_agency.littlepay_config.save()
```
thekaveman, line 43: Same here

tests/pytest/test_admin.py
```diff
@@ -34,6 +34,8 @@ def test_admin_override_agency_user_customer_service(model_AgencyUser, model_Tra
 
     # set up TransitAgency with staff_group and customer_service_group
     model_TransitAgency.pk = None
+    model_TransitAgency.littlepay_config.pk = None
+    model_TransitAgency.littlepay_config = model_TransitAgency.littlepay_config.save()
```
thekaveman, line 38: Same here

tests/pytest/test_admin.py
```diff
@@ -56,6 +58,8 @@ def test_admin_override_agency_user_not_customer_service(model_AgencyUser, model
 
     # set up TransitAgency with only a staff_group
     model_TransitAgency.pk = None
+    model_TransitAgency.littlepay_config.pk = None
+    model_TransitAgency.littlepay_config = model_TransitAgency.littlepay_config.save()
```
thekaveman, line 62: Same here

benefits/core/models/transit.py
angela-tran: I dropped the commit and force-pushed [just now](https://github.com/cal-itp/benefits/pull/2854#event-17339735842), and then I moved `Environment` to the top-level (https://github.com/cal-itp/benefits/pull/2854/commits/7154433bf667aa8b016c4fb69b74c84db98cbdaa). Let me know what you think!

thekaveman (approved):
:+1:

angela-tran:
Set the `environment` for each `LittlepayConfig` in dev:

before
Screenshot from 2025-04-21 18-02-40

after
Screenshot from 2025-04-21 18-03-26

Issue #2843, opened 2025-04-10
Switchio integration: implement Switchio template business logic


angela-tran:
Depends on #2838, #2839, and #2840 

This ticket includes implementing the Switchio template and any API endpoints it needs.

Switchio logic:
- request registration ID
- launch tokenization gateway using registration-ID-specific gateway URL
- ~~handle callback from Switchio~~ see note on next bullet point
-  ~~submit `CardTokenizeSuccessForm`~~ seems like this is not possible / relevant since the Promise is not resolving... we just get a GET request back from Switchio to our `eshopRedirectUrl`
- request information about the registration to get the card token

## Acceptance Criteria

- [x] When going through a flow for a transit agency that uses Switchio, the user is shown the Switchio tokenization gateway, which is where they enter their card details
- [x] Entering card details results in a card token

angela-tran:
This ticket needs https://github.com/cal-itp/benefits/issues/2839 to be finished before the subtasks on this ticket can be worked on.

angela-tran:
#2905 was the remaining sub-issue and has been done

PR #2865, opened 2025-04-29
Refactor: use enrollment index template from littlepay / switchio config

### LINKED ISSUE #2840: Switchio integration: refactor Enrollment Index to use specific template
Depends on #2838

The `benefits.enrollment.views.index` function currently uses `enrollment/index.html`. We'll need it to instead look at attributes on `LittlepayConfig`/`SwitchioConfig` and use a template specifically for that type of config.

This ticket is meant to introduce the template file with Switchio-specific copy and where Switchio business logic will go. #2843 will handle implementing that business logic.

## Acceptance Criteria

- [ ] No regressions to Littlepay enrollment
- [ ] Copy specific to transit processor is shown on Enrollment Index
- [ ] For this ticket, the CTA button on the Switchio template doesn't need to do anything

angela-tran:
Main part of #2840 

This PR makes it so there are separate templates for enrollment, one for Littlepay and another for Switchio. A base template defines the parts that are common between the two, but otherwise the two templates are free to do whatever they need to implement enrollment (within the `enrollment-form` block, that is).

Note that #2843 will handle implementing adding enrollment-related Javascript to the Switchio template.

## Testing
- [ ] No regressions to Littlepay enrollment
- [ ] When Switchio config is enabled on your transit agency, you see the Enrollment Index page with Switchio copy. The CTA button doesn't do anything, and says "Please wait..." forever.


lalver1 (approved):
Looks great! I also tested locally and confirmed that both templates looked like the screenshots 👍

PR #2882, opened 2025-05-01
Feat: CBV mixin enforces agency in session

### LINKED ISSUE #2873: Class-based views: Define AgencySessionRequired mixin

## Acceptance Criteria

- [ ] Implement logic in the `dispatch()` method
- [ ] Set `self.agency` on success

thekaveman:
Closes #2873 

Modeled after existing middleware class [`AgencySessionRequired`](https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L29), and will eventually replace this entirely once all dependent views are refactored to CBV.

### DIFF
--- a/benefits/core/mixins.py
+++ b/benefits/core/mixins.py
@@ -1,0 +1,23 @@
+import logging
+
+from . import session
+from .middleware import user_error
+
+logger = logging.getLogger(__name__)
+
+
+class AgencySessionRequiredMixin:
+    """Mixin intended for use with a class-based view as the first in the MRO.
+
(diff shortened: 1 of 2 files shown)


lalver1 (approved):
Looks good! 👍 
It was nice to see how to setup the tests, and how `SampleView` is required to test the mixin (I think because mixins can't stand alone, they need to have a view to inherit from in the MRO). I also like the setting up of `self.agency` since it will be used by other eligibility views that use `AgencySessionRequiredMixin`.

PR #2883, opened 2025-05-01
Feat: CBV mixin enforces flow in session

### LINKED ISSUE #2874: Class-based views: Define FlowSessionRequired mixin

## Acceptance Criteria

- [ ] Implement logic in the `dispatch()` method
- [ ] Set `self.flow` on success

thekaveman:
Closes #2874 

Modeled after existing middleware class [`FlowSessionRequired`](https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L85), and will eventually replace this entirely once all dependent views are refactored to CBV.


lalver1 (approved):
Looks good! 👍

PR #2884, opened 2025-05-01
Refactor: use Littlepay or Switchio enrollment class

### LINKED ISSUE #2839: Switchio integration: refactor enrollment views to use specific module
## Acceptance Criteria

- [ ] No regressions to existing enrollment

## Additional context

`benefits.enrollment.views` contains view functions `token` and `index` which are the main steps in enrollment logic. That is, you need a step to initiate the enrollment and another step to open the transit processor form / receive the card token.

Currently, these view functions call upon `benefits.enrollment.enrollment` to perform business logic. We should rename that module to `benefits.enrollment.littlepay` and introduce a new module named `benefits.enrollment.switchio` with functions of the same names.

Then, the view functions should look at the `TransitProcessor` in the session and use whichever module goes with that `TransitProcessor`.

Depends on #2838

angela-tran:
Closes #2839


benefits/core/session.py
```diff
@@ -23,8 +23,8 @@
 _DEBUG = "debug"
 _DID = "did"
 _ELIGIBLE = "eligibility"
-_ENROLLMENT_TOKEN = "enrollment_token"
-_ENROLLMENT_TOKEN_EXP = "enrollment_token_expiry"
+_LITTLEPAY_ENROLLMENT_TOKEN = "littlepay_enrollment_token"
+_LITTLEPAY_ENROLLMENT_TOKEN_EXP = "littlepay_enrollment_token_expiry"
```
angela-tran, line 27: Just a note here:

In a later PR for a different ticket (#2843), we will need to introduce Switchio session variables as well to hold the registration ID that the user is using.

I think eventually we will add new Django apps for `littlepay` and `switchio`, and those apps would have a `session` module that maintains these variables rather than them being in `core.session`.

benefits/enrollment/enrollment.py
angela-tran, line 1: Make sure to check out the changes in this file! 

GitHub is collapsing it because it sees this as a "large diff" (because of all the whitespace changes from indenting... :roll_eyes:)

benefits/enrollment/views.py
```diff
@@ -134,6 +120,56 @@ def index(request):
...
+def _index_for_switchio(request, enrollment):
+    agency = session.agency(request)
+    flow = session.flow(request)
+
+    context = {
+        "cta_button": "tokenize_card",
+        "enrollment_method": models.EnrollmentMethods.DIGITAL,
+    }
+    context.update(flow.enrollment_index_context)
+
```
angela-tran, line 132: This is just a bare-bones implementation to get a page showing up for Switchio and will be fleshed out in #2843 when we need to actually implement this view.

The view will need to handle 2 GET scenarios: 
- the user loading the page initially 
- the Switchio modal redirecting back to us (either from the Back button or from a submission)

benefits/core/session.py
thekaveman: Yeah I had the same thought, and I'd like to use this PR to start on that. We already have #2800 and I feel like this is just adding more and more technical debt to our session handling. I'd rather we avoid this kind of thing going forward, knowing that we want to clean all this up.

There is no reason for any of this to be in the `core` module right? No other parts of the code need to know about this except for `enrollment`?

thekaveman: (I'm not necessarily saying to implement a class-based session here, but let's at least keep this new stuff separate and where it is needed, vs. introducing new global/`core` level stuff when it isn't needed there).

benefits/enrollment/views.py
```diff
@@ -26,80 +26,66 @@
...
     session.update(request, origin=reverse(routes.ENROLLMENT_INDEX))
 
+    agency = session.agency(request)
+    enrollment = _get_enrollment_class(agency)()
+
+    # this is what we have to do until we introduce separate Django apps that have littlepay and switchio views
+    if isinstance(enrollment, LittlepayEnrollment):
+        return _index_for_littlepay(request, enrollment)
+    else:  # is instance of SwitchioEnrollment
+        return _index_for_switchio(request, enrollment)
```
angela-tran, line 75: At first, I tried to have a method on `LittlepayEnrollment`/`SwitchioEnrollment` that would return the right view function to use, but it was circular-import galore, so I resorted to this.

I think if we had `littlepay` and `switchio` apps, that might give us a cleaner way to invoke the appropriate view logic.

thekaveman: I think we should not introduce technical debt like this when you have a better idea/approach in mind already.

benefits/core/session.py
angela-tran: So are you thinking we should go ahead and introduce `littlepay` and `switchio` apps to hold their specific things (e.g. models, templates, session, etc.)? Or are you thinking of a smaller scope, like adding a `littlepay_session.py` and `switchio_session.py`? 

(You probably don't mean the latter since you said 
> There is no reason for any of this to be in the `core` module right? 

)

thekaveman: I think we should be doing the thing we think is the best long-term approach, not the thing we think is the way to get it done now and hopefully we'll have time to come back and clean it up. We are implementing new code/functionality now. Let's do it the cleanest way possible. Let's keep in mind all the stuff we've been talking about: smaller, more focused apps. Less global/shared `core` functionality. 

Not to mention, having all this code new in its own modules/apps/whatever, just imported and called in `enrollment`, would make reviews _so much easier_. I think testing would probably be simpler as well.

thekaveman: :point_up: that is a general comment for this entire PR, not just this specific session file.

thekaveman: Yeah I think we just don't have as much luxury of time anymore with all the other things going on, there's less chance we have to come back and clean up things.

angela-tran: That makes sense. Thanks for this feedback, and I'll look into how to organize things so that the code is cleaner.

angela-tran:
Closing this in favor of an approach where implementation is organized into separate Django apps, starting with #2889

PR #2885, opened 2025-05-01
Refactor: use Littlepay or Switchio enrollment module

angela-tran:
Was started for #2839 

I started with this approach, but then I remembered Kegan and I talked about having classes that encapsulate their enrollment rather than modules. So #2884 will probably be what I continue with. Just wanted to open this PR in case there was feedback and in case we need to go back to this approach.


thekaveman:
@angela-tran

> I started with this approach, but then I remembered Kegan and I talked about having classes that encapsulate their enrollment rather than modules.

These aren't mutually exclusive. We have can have `benefits/enrollment/littlepay.py` and `benefits/enrollment/switchio.py` etc. 

In fact, I see almost no reason to keep these in the same module/file. Your comment is a good example of why we shouldn't: https://github.com/cal-itp/benefits/pull/2884/files#r2071857010

angela-tran:
> @angela-tran
> 
> > I started with this approach, but then I remembered Kegan and I talked about having classes that encapsulate their enrollment rather than modules.
> 
> These aren't mutually exclusive. We have can have `benefits/enrollment/littlepay.py` and `benefits/enrollment/switchio.py` etc.
> 
> In fact, I see almost no reason to keep these in the same module/file. Your comment is a good example of why we shouldn't: https://github.com/cal-itp/benefits/pull/2884/files#r2071857010

Yeah, I was thinking to move the classes to their own modules as well.

thekaveman:
You aren't actually working on this PR right @angela-tran? #2884 is where the work is happening? Can we close whichever is the duplicate so we aren't having these conversations in multiple threads?

angela-tran:
Yes that's correct, and yep good idea

PR #2887, opened 2025-05-02
Feat: CBV mixin sets required reCAPTCHA settings

### LINKED ISSUE #2875: Class-based views: Define RecaptchaEnabled mixin

## Acceptance Criteria

- [ ] Implement logic in the `dispatch()` method and/or maybe just use context data instead?

lalver1:
Closes #2875

Modeled after existing middleware class [`RecaptchaEnabled`](https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L127) and will eventually replace this entirely once all dependent views are refactored to CBV.

### DIFF
--- a/benefits/core/mixins.py
+++ b/benefits/core/mixins.py
@@ -1,6 +1,8 @@
 import logging
 
-from . import session
+from django.conf import settings
+
+from . import recaptcha, session
 from .middleware import user_error
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


lalver1:
While looking at CBVs more closely, I decided to keep the reCAPTCHA configuration logic in the `dispatch()` method instead of putting it in `get_context_data()` since this mixin could potentially be used with a generic `View`, and `get_context_data` is not defined in `View`.

machikoyasuda:
Just realized CBV stands for class-based views 😅

benefits/core/mixins.py
```diff
@@ -39,3 +41,18 @@ def dispatch(self, request, *args, **kwargs):
         else:
             logger.warning("Session not configured with enrollment flow")
             return user_error(request)
+
+
+class RecaptchaEnabledMixin:
+    """Mixin intended for use with a class-based view as the first in the MRO.
```
machikoyasuda, line 47: I see that `MRO` is used in other parts of the app. But I am not sure what it is still. What does it stand for?

machikoyasuda (approved):

benefits/core/mixins.py
thekaveman: It stands for _Method Resolution Order_, which is the way that Python searches for a given method in a class hierarchy, and is especially important for multiple inheritance type situations (like we get with mixins).

In these mixin classes we are creating for class-based views, it will be important that these mixins come _first_ in the MRO, because they are acting as "gates" for the logic of the class-based view. We need the mixin `dispatch()` to be found in the hierarchy _before_ the `dispatch()` from the view. This is similar to how our middlewares work, where they process the request before it gets to the view function, and can decide to allow the request to continue or abort early with e.g. a user error.

Here's a really official and technical doc on the topic: https://docs.python.org/3/howto/mro.html

thekaveman:
> Just realized CBV stands for class-based views 😅

Also where this website got its domain name :grin: https://ccbv.co.uk/

PR #2889, opened 2025-05-05
Refactor: enrollment_littlepay app

angela-tran:
Part of #2839

This PR introduces an `enrollment_littlepay` app, which is where the Littlepay-specific implementation of enrollment (models, views, templates, session, etc.) should go.


lalver1:
This looks good to me! I went through an older adult flow (including enrollment) locally and everything worked (including setting up the database in `main` and running the migrations in this branch) 👍 

I only have one minor comment, in https://github.com/cal-itp/benefits/blob/refactor/enrollment_littlepay-app/benefits/core/models/__init__.py, would it be better to import `Environment` from `.common` directly? Even though everything works now (`Environment` is available in `transit.py` due to the import from `.common`) it may be more clear to import it directly from `.common`.

angela-tran:
Thanks for the review @lalver1 and for being thorough / checking the migrations. I agree that importing `Environment` from `common` is cleaner, so amended the last commit to do it that way. Ready for re-review

lalver1 (approved):
Looks great @angela-tran!

machikoyasuda (approved):
Ran this locally. Appreciate all the detailed help text and notes Angela!

angela-tran:
Thank you both!

PR #2893, opened 2025-05-07
Refactor(docs): apply Diataxis

angela-tran:


docs/index.md
```diff
@@ -1,5 +1,74 @@
 # Project overview
 
+## Current work
+
+We do sprint planning and track day-to-day work on our [Project Board][board].
+
+## Product roadmap
+
+Our product roadmap captures what we're currently building, what we've built, and what we plan to build in the future. We update it at the end of each quarter or when priorities change.
```
jgravois: this is subjective, but i think it would be less disorienting if 'Current work' and 'Product roadmap' were shuffled toward the bottom of the landing page. perhaps immediately above 'Practices'?

jgravois (commented):
this is already looking like a promising reorganization 🎉 

---

i'm not sure exactly how the TOC links in the sidebar are generated, but i noticed in 'Guides' that the 'inc' content that is injected inline is erroneously included in this branch.

Image

Scotchester (commented):
One question about missing content, but this otherwise LGTM.

jgravois (approved):
🌈🌈

PR #2908, opened 2025-05-13
Feat: enrollment switchio app

angela-tran:
Part of #2839 

This PR introduces an `enrollment_switchio` app, which is where the Switchio-specific implementation of enrollment (models, views, templates, session, etc.) should go.


angela-tran:
Closing this PR so that @lalver1 can finish this branch out with a new PR.

PR #2910, opened 2025-05-13
Refactor: move `enrollment` and `session` code to `enrollment_littlepay` app

### LINKED ISSUE #2907: Move Littlepay enrollment code into `enrollment_littlepay` app
## Acceptance Criteria

- [ ]

## Additional context

https://cal-itp.slack.com/archives/C037Y3UE71P/p1747095690036549

#2889 introduced the `enrollment_littlepay` app, and we moved the `LittlepayConfig` model into it, but we forgot that there was more code to move over. Let's do that in this ticket.

angela-tran:
Part of #2907

Moves Littlepay-specific code from `benefits.enrollment.enrollment` and `benefits.core.session` into the `enrollment_littlepay` app.

A subsequent PR will handle refactoring `benefits.enrollment.views.token` and `benefits.enrollment.views.index`.


angela-tran:
Paired with @machikoyasuda and @lalver1 on this earlier. 

This PR is still a WIP - many tests are failing because we deleted the `benefits.core.session` methods that had to do with an "enrollment token", i.e. the Littlepay card tokenization access token.

tests/pytest/core/test_session.py
```diff
@@ -201,15 +163,16 @@ def test_logged_in_True(app_request):
...
+    LittlepaySession(app_request, access_token="enrollment_token", access_token_expiry="3600")
     OAuthSession(app_request).claims_result = ClaimsResult(verified={"oauth_claim": True})
-    session.update(app_request, logged_in=True, enrollment_token="enrollment_token")
+    session.update(app_request, logged_in=True)
     assert session.logged_in(app_request)
 
     session.logout(app_request)
 
     assert not session.logged_in(app_request)
-    assert not session.enrollment_token(app_request)
```
angela-tran, line 211: A note here for @machikoyasuda and @lalver1, just because we looked at this briefly at the end of pairing session yesterday: 

I said at the time that this test shouldn't be responsible for asserting state about the Littlepay session. (Same idea with `test_pre_logout()` in `tests/pytest/oauth/test_hooks.py`.)

I feel like my reasoning was mostly influenced by wanting to try and finish out the commit and maybe hurrying through thinking about this. I looked at this closer and realized it's not hard to check that the Littlepay session attributes were updated correctly, so I went ahead and did that, i.e. maintained the assertion using the refactored code.

tests/pytest/oauth/test_hooks.py
```diff
@@ -50,7 +51,7 @@ def test_pre_logout(app_request, mocked_oauth_analytics_module):
 
     mocked_oauth_analytics_module.started_sign_out.assert_called_once_with(app_request)
     assert not session.logged_in(app_request)
-    assert session.enrollment_token(app_request) is False
```
angela-tran, line 53: Same note as above, I realized it's not hard to maintain the assertion :relaxed:

tests/pytest/core/test_session.py
angela-tran: (see https://github.com/cal-itp/benefits/pull/2910/files#diff-fcf5a55d9174de68ae6f8d1b72f014264bfd81d9647acf0d6a38b9aa58bd7adcR174)

tests/pytest/oauth/test_hooks.py
angela-tran: (see https://github.com/cal-itp/benefits/pull/2910/files#diff-c4e66e5bbdbde74f5062e2a19cb1b0374dc4180b1de7f0b932e169f60ef25e02R54)

angela-tran:
I ran through the app with this branch, and enrollment still works :white_check_mark: :tada:

benefits/enrollment_littlepay/session.py
```diff
@@ -0,0 +1,54 @@
...
+
+from django.http import HttpRequest
+
+logger = logging.getLogger(__name__)
+
+
+class Session:
+
+    _keys_access_token = "enrollment_littlepay_access_token"
+    _keys_access_token_expiry = "enrollment_littlepay_access_token_expiry"
```
thekaveman, line 12: Nit: we can probably leave `enrollment` out of the keys here. But this is a very small deal.

benefits/enrollment_littlepay/session.py
```diff
@@ -0,0 +1,54 @@
...
+    @access_token_expiry.setter
+    def access_token_expiry(self, value: str) -> None:
+        self.session[self._keys_access_token_expiry] = value
+
+    def access_token_valid(self):
+        """True if the request's session is configured with a valid card tokenization access token. False otherwise."""
+        if bool(self.access_token):
+            exp = self.access_token_expiry
+            # ensure token does not expire in the next 5 seconds
+            valid = exp is None or exp > (time.time() + 5)
```
thekaveman, line 51: Isn't `exp` a string here?

thekaveman: I guess this matches what the original code was doing, so this seems OK.

Maybe I'm tripped up on the type hints for `access_token_expiry`. Should that be `int`?

tests/pytest/core/test_session.py
thekaveman: I think this is fine / good to keep the asserts since they were relatively easy.

Another approach, since there is also good test coverage of `enrollment_littlepay.Session`, would be to mock/patch that class and just assert it was called with `reset=True` here.

thekaveman (approved):
Looks good :+1: 

I ran through enrollment locally and it worked.

Issue #2902, opened 2025-05-13
Create `enrollment_switchio` app


angela-tran:
Somewhat depends on #2907

The basic structure can be worked on without any hold up from #2907. To close out #2839, this ticket needs to also define the `index` view and a view equivalent to `token` (called `gateway_url`) for Switchio.

You can skim over #2884 to get an idea of the sort of branching that we need (keep in mind: that PR organized the implementations into classes whereas our new, current approach organizes them into proper Django apps).

## Acceptance Criteria

- [ ] `index` view returns the Switchio template
- [ ] `gateway_url` view returns a 200 status

PR #2911, opened 2025-05-13
Feat: `enrollment_switchio` app

lalver1:
Part of #2839 and its subtask #2902 

This PR introduces an `enrollment_switchio` app, which is where the Switchio-specific implementation of enrollment (models, views, templates, session, etc.) should go.


lalver1:
@angela-tran I'm not sure why Django originally had those extra steps in the migrations for the model changes in this branch 🤔 
Anyway, to see if I could get a different set of migrations, I ended up first changing the field `switchio_config` to FK `"enrollment_switchio.SwitchioConfig"` and moving `SwitchioConfig` to the new app and asking Django to make a migration. Then I deleted `SwitchioConfig` from `core` and asked for another migration. The [resulting migration steps](https://github.com/cal-itp/benefits/blob/58e2bf741f7a973cc38ac616029e56b9400196b3/benefits/core/migrations/0050_alter_transitagency_switchio_config.py) are the ones that are in this branch, with no more `RemoveField` operations in the migration.

angela-tran (approved):
Nice, that looks great :clap: :raised_hands: 

I ran the migration locally, and it looks good
- Checked out `main` and reset the database - loads the `SwitchioConfig` from sample fixtures
Screenshot from 2025-05-13 16-11-57
- Checked out this branch and ran `./bin/init.sh` script - runs migrations
Screenshot from 2025-05-13 16-12-23
- Migrations succeeded, no errors, and previous `SwitchioConfig`s are gone which is fine 
Screenshot from 2025-05-13 16-12-31

Issue #2870, opened 2025-04-30
Class-based views: Eligibility confirm


thekaveman:
View function: https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L137

## Acceptance Criteria

- [x] Mixin class enforces `AgencySessionRequired`
- [x] Mixin class enforces `FlowSessionRequired`
- [x] Mixin class enforces `RecaptchaEnabled`
- [x] `get()` method
- [x] `post()` method
- [x] Implement as [`FormView`](https://docs.djangoproject.com/en/5.2/ref/class-based-views/generic-editing/#formview)? Or maybe just `TemplateView` since the form changes per flow
- [x] Move any corresponding copy context into [`get_context_data()`](https://docs.djangoproject.com/en/5.1/topics/class-based-views/generic-display/#adding-extra-context)
- [x] Refactor to class-based tests for all view tests
- [x] Refactor view tests away from using the test Django `Client` to make requests, instead testing the individual attributes/methods of the view class

PR #2913, opened 2025-05-13
Refactor: Class-based views - Eligibility index

### LINKED ISSUE #2868: Class-based views: Eligibility index
View function: https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L22

## Acceptance Criteria

- [x] Mixin class enforces `AgencySessionRequired`
- [x] Mixin class enforces `RecaptchaEnabled`
- [x] `get()` and `post()` methods

machikoyasuda:
closes #2868 

This PR converts the Eligibility Index view method from a function-based view into a class-based view (CBV, reference here https://ccbv.co.uk/). This PR uses both mixins from #2875 and #2882.

## What this PR does
- uses a FormView https://ccbv.co.uk/projects/Django/5.2/django.views.generic.edit/FormView/ https://docs.djangoproject.com/en/5.2/ref/class-based-views/generic-editing/#django.views.generic.edit.FormView 
- uses both AgencySessionRequiredMixin and RecaptchaEnabledMixin in class-based index view (from https://github.com/cal-itp/benefits/pull/2887)
- updates url to reflect this change.

## Notes
- I kept the comments there for posterity. Do we still want/need them?
- I updated one test. One part of a spec was failing because it now was returning like this: `assert ['eligibility/index.html'] == 'eligibility/index.html'`, b/c `FormView` returns a list of template names, instead of a string. I changed the test to match what FormView now does, instead of changing FormView itself to return a string like the test expects.

### DIFF
--- a/benefits/eligibility/views.py
+++ b/benefits/eligibility/views.py
@@ -7,7 +7,7 @@ from django.shortcuts import redirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
 from django.utils.decorators import decorator_from_middleware
-from django.views import View
+from django.views.generic import FormView
 
 from benefits.routes import routes
 from benefits.core import recaptcha, session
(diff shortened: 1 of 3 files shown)


benefits/eligibility/views.py
```diff
@@ -7,30 +7,40 @@
...
-def index(request):
+class IndexView(AgencySessionRequiredMixin, RecaptchaEnabledMixin, View):
     """View handler for the enrollment flow selection form."""
-    agency = session.agency(request)
-    session.update(request, eligible=False, origin=agency.index_url)
 
-    # clear any prior OAuth token as the user is choosing their desired flow
-    # this may or may not require OAuth, with a different set of scope/claims than what is already stored
-    session.logout(request)
+    template_name = "eligibility/index.html"
```
machikoyasuda, line 25: Should this template name be defined on line 19, under `TEMPLATE_CONFIRM`, for consistency?

machikoyasuda:
After reading https://github.com/cal-itp/benefits/issues/2870, I started thinking this could be written as a FormView instead. Any thoughts on which way to go? @angela-tran @lalver1 @thekaveman

angela-tran:
> After reading #2870, I started thinking this could be written as a FormView instead. Any thoughts on which way to go? @angela-tran @lalver1 @thekaveman

I think I'm the  one with the least hands-on experience with class-based views, so take this with a grain of salt, but I think it's definitely worth looking into using a FormView and whether that makes the code simpler / more readable or not.

benefits/eligibility/views.py
angela-tran: I think these global variables for template names were an artifact of having view functions. With class-based views, one of the main benefits is that everything is more cohesive and together, so I don't think we need to keep using these global variables for template names.

benefits/eligibility/views.py
```diff
@@ -25,25 +25,27 @@ class IndexView(AgencySessionRequiredMixin, RecaptchaEnabledMixin, FormView):
...
         kwargs = super().get_form_kwargs()
         kwargs["agency"] = self.agency
         return kwargs
 
-    def get(self, request, *args, **kwargs):
-        """Handle GET requests: initialize session and clear any prior OAuth tokens."""
+    def dispatch(self, request, *args, **kwargs):
+        """Initialize session state before handling the request."""
+        if not self.agency:
+            return TemplateResponse(request, "200-user-error.html")
```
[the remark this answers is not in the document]
thekaveman: Hmm this is a little more nuanced than I previously understood. 

We have `AgencySessionRequiredMixin` which defines its logic in the `dispatch()` method. The original idea was that CBVs inheriting from this mixin would first have the mixin's `dispatch()` called (which does the check, sets the `self.agency` variable) and then if that was successful, fall through to the CBV's `dispatch()` (or return `user_error` if not successful).

This is sort of how it works.

### How MRO affects dispatch calls

When we have a view like class `IndexView(AgencySessionRequiredMixin, View)`, and `view_instance.dispatch()` is called, Python looks for the `dispatch` method in this order (simplified):

1. `IndexView`
2. `AgencySessionRequiredMixin`
3. `View` (e.g., `django.views.View`)
4. Other parent classes of `View` (if any)

### What this means for mixins

E.g. `AgencySessionRequiredMixin` relies on its own `dispatch` method being executed to perform the agency check.

**Scenario A:** The view using the mixin does NOT define `dispatch`

If `IndexView` does not have its own `dispatch` method, then `AgencySessionRequiredMixin.dispatch` is found and executed. The mixin runs its check, and if an agency is present, `super().dispatch()` in the mixin correctly passes control to `View.dispatch`.

**Scenario B**: The view using the mixin DOES define `dispatch`

If `IndexView` does define its own `dispatch` method (like in this PR), then `IndexView.dispatch` will be called instead of `AgencySessionRequiredMixin.dispatch`.

If `IndexView.dispatch` simply does its own thing and does not call `super().dispatch()`, then the mixin's `dispatch` logic is entirely bypassed.

However, if `IndexView.dispatch` does call `super().dispatch()` (like in this PR), then control will be passed to `AgencySessionRequiredMixin.dispatch` (as it's next in the MRO). The mixin can then perform its check, and its own `super().dispatch()` call will continue the chain to `View.dispatch`.

### TLDR;

If possible, I think we should avoid implementing `dispatch` in our CBVs. Can this logic move to the `get` method or similar? 

If not possible (e.g. we absolutely need to implement some logic in the CBVs `dispatch` method), it MUST call `super().dispatch()` _early_, before taking any action that would require the results of the mixin's `dispatch` (e.g. the session check).

thekaveman (changes requested):

angela-tran:
Is there still a change that's needed here?

machikoyasuda:
Will get this out today beofre EOD!

machikoyasuda:
Closing for the time being. Will re-open in next slush.

PR #2928, opened 2025-05-16
Feat: define Switchio Index and GatewayUrl views

### LINKED ISSUE #2902: Create `enrollment_switchio` app

lalver1:
Closes #2902

This PR sets up the basic structure of the `enrollment_switchio` app. It adds two class-based views, `IndexView` which handles the bare minimum to render the index page and `GatewayUrlView` which is a placeholder for the future implementation of the Switchio tokenization gateway registration.

## Reviewing
~You can preview the `enrollment_switchio` index page by setting up a transit agency with only a Switchio config and running through an enrollment flow.~ The new views are not reachable from the application yet, but running through the tests using the debugger can be used to verify the behavior of the new views.


benefits/enrollment_switchio/views.py
```diff
@@ -0,0 +1,30 @@
...
+    """CBV for the enrollment landing page."""
+
+    template_name = "enrollment_switchio/index--switchio.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        flow = session.flow(self.request)
+        context.update(
+            {
+                **flow.enrollment_index_context,
```
angela-tran, line 17: I'm envisioning that we will get rid of `enrollment_index_context` from `EnrollmentFlow`, so here it would be better to take what's in `enrollment_index_context` and simply inline it here.

 `enrollment_index_context` includes a context dict from `SwitchioConfig` so you'll need to inline that too. Let me know if any of this is unclear. It's a little hard to describe what I mean with just a PR comment.

benefits/enrollment_switchio/views.py
```diff
@@ -0,0 +1,30 @@
+from django.http import JsonResponse
+from django.views.generic import TemplateView, View
+
+from benefits.core import models, session
+
+
+class IndexView(TemplateView):
+    """CBV for the enrollment landing page."""
```
angela-tran, line 8: I don't think it's necessary for this docstring to be so specific to this view being a class-based view. We could just say "View handler" like the existing comments in `benefits.enrollment.views` or just "View".

benefits/enrollment_switchio/views.py
```diff
@@ -0,0 +1,30 @@
...
+                **flow.enrollment_index_context,
+                "cta_button": "tokenize_card",
+                "enrollment_method": models.EnrollmentMethods.DIGITAL,
+            }
+        )
+        return context
+
+
+class GatewayUrlView(View):
+    """CBV for the tokenization gateway registration"""
```
angela-tran, line 26: Same comment as above - no need to document that this is a class-based view. Suggestion is to say "View handler" or "View" in place of "CBV"

tests/pytest/enrollment_switchio/test_views.py
```diff
@@ -0,0 +1,31 @@
+import json
+import pytest
+
+from benefits.enrollment_switchio.views import GatewayUrlView, IndexView
+
+
+class TestIndexView:
```
angela-tran, line 7: Having these tests is especially helpful since the views are not actually being used by the application yet. :100:

angela-tran (changes requested):

benefits/enrollment_switchio/models.py
```diff
@@ -57,7 +57,7 @@ def transit_processor_context(self):
 
     @property
     def enrollment_index_template(self):
-        return "enrollment/index--switchio.html"
+        return "enrollment_switchio/index--switchio.html"
```
angela-tran, line 60: Now that this template is in the `enrollment_switchio` app, let's rename this template to just `index.html`. No need for the `--switchio` part any more.

benefits/enrollment_switchio/views.py
```diff
@@ -0,0 +1,30 @@
+from django.http import JsonResponse
+from django.views.generic import TemplateView, View
+
+from benefits.core import models, session
+
+
+class IndexView(TemplateView):
```
angela-tran, line 7: The analog to this view in the `enrollment` app is decorated to [require an eligible session](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/views.py#L55). Should we have a mixin on this class-based view to preserve this?

benefits/enrollment_switchio/views.py
```diff
@@ -0,0 +1,30 @@
...
+            {
+                **flow.enrollment_index_context,
+                "cta_button": "tokenize_card",
+                "enrollment_method": models.EnrollmentMethods.DIGITAL,
+            }
+        )
+        return context
+
+
+class GatewayUrlView(View):
```
angela-tran, line 25: Similar comment as https://github.com/cal-itp/benefits/pull/2928/files#r2093772867 - should we have a mixin on this class-based view?

benefits/enrollment_switchio/views.py
angela-tran: I need `EligibleSessionRequiredMixin` as well for #2910, and I realize now that it hasn't been written yet. I suggest we have a separate PR that just introduces that mixin for both of our PRs to use.

thekaveman: I created #2931 and will work on it now!

angela-tran: @lalver1 if you rebase onto `main`, you can add `EligibleSessionRequiredMixin` to this view now

benefits/enrollment_switchio/views.py
lalver1: Ah that's right, thanks for the reminder! I inlined the context in 6dda1ac788b290f9a582aa70fa6cd51739d781a8 (it still includes the context dict associated with the `SwitchioConfig`)

tests/pytest/enrollment_switchio/test_views.py
lalver1: Thanks, yep I used them to confirm the behavior of the new views. I did make a small change to `TestIndexView`. The view itself is now passed as a fixture to the test, anticipating that we'll have more tests that will use it and it's nicer to set it up this way.

benefits/enrollment_switchio/views.py
lalver1: Good catch, thanks! I added the `EligibleSessionRequiredMixin` mixin to both views.

tests/pytest/enrollment_switchio/test_views.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestGatewayUrlView:
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("mocked_session_eligible")
+    def test_get_gateway_url(self, app_request):
+
+        response = GatewayUrlView.as_view()(app_request)
+
+        assert response.status_code == 200
+        assert json.loads(response.content) == {"gateway_url": "https://server/gateway/uuid"}
```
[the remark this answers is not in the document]
lalver1: Yep, I was also thinking the same thing but I was getting

`AttributeError: 'JsonResponse' object has no attribute 'json'` 

when trying to use `response.json()`.

It turns out that if the response is generated by the Django test client, like `response = client.get(path)`, the `json` attribute is added to the response object, I guess for convenience, and we can use `response.json()`. But we are generating the response in the test using `response = GatewayUrlView.as_view()(app_request)` (since we don't have a `url`s definition yet), and the response object does not have a `json` attribute. It's kind of tricky because both ways of generating the response produce `JsonResponse` objects (as shown in the debugger), yet there is a difference in their attributes.

benefits/enrollment_switchio/views.py
angela-tran: Thanks for making that change! I see a problem here though, which is that the values here for `headline`, `next_step`, `partner_post_link`, and `alert_include` are not correct for Agency Card or CalFresh flows.

In #2930, @machikoyasuda and I inlined the context into the Littlepay `index` view by [copying over some implementation](https://github.com/cal-itp/benefits/pull/2930/files#diff-3e4beacead98dafe8a82198d46543eec79d30b841659e8e0523df3ca88c35f81R45-R47) from `flow.enrollment_index_context`.

Now that I'm looking at the two views though, I think it'll actually be best to do the following:
- keep `enrollment_index_context` on `EnrollmentFlow`
- have `EnrollmentFlow.enrollment_index_context` be responsible _only_ for returning the context dict with copy for the template
- have each of the `Index` views update their context with `flow.enrollment_index_context` _AND_ add their own values to the context

angela-tran: Maybe we could briefly talk through how to get this in cleanly whenever it makes sense for you and/or @machikoyasuda

lalver1: Thanks @angela-tran, I was just looking over this and I think I understand what you are proposing. I'll start on making the change and we can look at it together anytime today.

tests/pytest/enrollment_switchio/test_views.py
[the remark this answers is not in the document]
lalver1: I just tried it out but unfortunately we don't get the `json` attribute we need, the data is still just in raw json inside `content`. But it was worth trying!

tests/pytest/enrollment_switchio/test_views.py
```diff
@@ -0,0 +1,38 @@
...
+        assert "alert_include" in context
+        assert "transit_processor" in context
+
+
+class TestGatewayUrlView:
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("mocked_session_eligible")
+    def test_get_gateway_url(self, app_request):
+
+        response = GatewayUrlView.as_view()(app_request)
```
angela-tran, line 35: It might be better to call `dispatch` specifically here so that we don't have to worry about satisfying the `EligibleSessionRequiredMixin` `dispatch` logic (i.e. that's not the responsibility of this test)

angela-tran: Or maybe we do want a test that checks that we put `EligibleSessionRequiredMixin` on the view...

lalver1: It's interesting, if I replace

`response = GatewayUrlView.as_view()(app_request)` with 

`response = view.dispatch(app_request)` (with an appropriate `view` fixture) 

and remove `@pytest.mark.usefixtures("mocked_session_eligible")`, we get a `ContentNotRenderedError("The response content must be rendered before it can be accessed.")` error on the json assert. I thought it had something to do with the rendering mechanism maybe not being ready after `dispatch` is called. But, it turns out (and it makes sense in hindsight) that `dispatch` from `EligibleSessionRequiredMixin` ran and `response.content` got content related to the error and `rendered_content` got the error page's html.  I then added `mocked_session_eligible` back and the test passed, so it looks like using either `.as_view` or `.dispatch` still requires satisfying the `EligibleSessionRequiredMixin`.

angela-tran: Oh, interesting. That does make sense in hindsight. Thanks for trying it out and writing up this explanation!

lalver1: No problem! It was good to try this out as we are settling on how to test these class-based views. I ended up using `response = view.get(app_request)` for the test since it most accurately reflects the method of `GatewayUrlView` that we are testing.

tests/pytest/enrollment_switchio/test_views.py
```diff
@@ -0,0 +1,50 @@
...
+    @pytest.fixture
+    def view(self, app_request):
+        """Fixture to create an instance of IndexView."""
+        v = GatewayUrlView()
+        v.setup(app_request)
+
+        return v
+
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("mocked_session_eligible")
```
thekaveman, line 44: This fixture can be removed

thekaveman (changes requested):
Small test clean up

lalver1:
Putting this in Draft until it is rebased on #2930.

angela-tran (approved):
This looks great! Thanks @lalver1 :pray:

thekaveman (approved):
Sorry didn't realize I was blocking this!

lalver1:
No worries! I also overlooked following up on it 😅

PR #2930, opened 2025-05-17
Refactor: add enrollment_littlepay index view

angela-tran:
Part of #2907 

This PR moves Littlepay-specific code from `benefits.enrollment.views.index` into `benefits.enrollment_littlepay.views`. 

Note that with this PR, the Switchio template is no longer reachable, which is ok. We will add it back whenever we start needing to run through it for development.

## Reviewing
Enrollment with Littlepay should still work.


angela-tran:
Still a WIP - next step is to clean up `enrollment_littlepay`'s `index` function and convert it into a class-based view, if possible.

angela-tran:
Handing this off to @machikoyasuda so closing this PR so a new one can be opened

angela-tran:
Ran through enrollment locally, and it still works for me :pray:

benefits/enrollment/views.py
```diff
@@ -62,81 +62,44 @@ def index(request):
...
-            "form_retry": tokenize_retry_form.id,
-            "form_server_error": tokenize_server_error_form.id,
-            "form_success": tokenize_success_form.id,
-            "form_system_error": tokenize_system_error_form.id,
-            "overlay_language": overlay_language,
-        }
-        context.update(flow.enrollment_index_context)
-
-        return TemplateResponse(request, agency.enrollment_index_template, context)
+    return LittlepayIndexView.as_view(enrollment_result_handler=_enrollment_result_handler)(request)
```
lalver1, line 65: Just a thought, and I'm not sure if it would be of much benefit, but we could put `_enrollment_result_handler` in `enrollment/enrollment.py` and import it in `enrolment_littlepay` for `IndexView` to use, that way we don't pass it as a `kwarg`. Even though the current approach is perfectly fine, just thought about mentioning this in case there are benefits to this alternate approach that I am not seeing.

angela-tran: Interesting idea. I think that might cause a circular import though because regardless of where `_enrollment_result_handler` is, it needs to call the `success`, `retry`, etc. view functions which are in `benefits.enrollment.views`.

angela-tran:
I had to add one more commit to this PR to add `EligibleSessionRequiredMixin` onto the view. I don't know of anything else I need to add, so you should be good to keep reviewing this! :pray:

benefits/enrollment/views.py
lalver1: Ah yes I missed that part, definitely a circular import.

lalver1 (approved):
This looks good to me!
I also ran through an Older Adult flow and enrolled with Littlepay and everything still works 👍

tests/pytest/enrollment_littlepay/test_views.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def test_form_invalid(self, view):
+        with pytest.raises(Exception, match="Invalid card token form"):
+            form = view.form_class()
+            view.form_invalid(form)
+
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("mocked_session_eligible", "mocked_session_agency", "mocked_session_flow")
+    def test_index_view(self, mocked_enrollment_result_handler, app_request):
+        response = IndexView.as_view(enrollment_result_handler=mocked_enrollment_result_handler)(app_request)
```
angela-tran, line 70: Split out into two lines with a `view` variable assignment

tests/pytest/enrollment_littlepay/test_views.py
```diff
@@ -0,0 +1,73 @@
...
+        handler_spy.assert_called_once_with(view.request, Status.SUCCESS, None)
+
+    def test_form_invalid(self, view):
+        with pytest.raises(Exception, match="Invalid card token form"):
+            form = view.form_class()
+            view.form_invalid(form)
+
+    @pytest.mark.django_db
+    @pytest.mark.usefixtures("mocked_session_eligible", "mocked_session_agency", "mocked_session_flow")
+    def test_index_view(self, mocked_enrollment_result_handler, app_request):
```
angela-tran, line 69: Add a similar test where the session is not eligible (i.e. the negative scenario)

angela-tran: This would be really easy to do, but I don't want to block #2928 any longer, so going to push this off to a follow-up ticket :pray:

tests/pytest/enrollment_littlepay/test_views.py
angela-tran: This would be really easy to do, but I don't want to block #2928 any longer, so going to push this off to a follow-up ticket :pray:

angela-tran: Created #2937

tests/pytest/enrollment_littlepay/test_views.py
angela-tran: Created #2937

PR #2934, opened 2025-05-19
Refactor: add enrollment_littlepay token view

angela-tran:
Part of #2907 

This PR moves the entire `token` view from `benefits.enrollment.views` into `benefits.enrollment_littlepay.views`. It also refactors the view into a class-based view.

The URL for the view was moved under `enrollment_littlepay` with a namespace of `littlepay` so that the URL is something like `<base URL>/littlepay/token`. The route was updated as well so that all consumers of the route use the new URL.

## Reviewing
Enrollment with Littlepay should still work.

## Screenshot

image

### DIFF
--- a/tests/pytest/enrollment_littlepay/test_views.py
+++ b/tests/pytest/enrollment_littlepay/test_views.py
@@ -36,7 +36,7 @@ def mocked_enrollment_result_handler():
 
 @pytest.mark.django_db
 def test_token_ineligible(client):
-    path = reverse(routes.ENROLLMENT_TOKEN)
+    path = reverse(routes.ENROLLMENT_LITTLEPAY_TOKEN)
 
     response = client.get(path)
 
@@ -62,7 +62,7 @@ def test_token_refresh(mocker, client):
         ),
     )
 
-    path = reverse(routes.ENROLLMENT_TOKEN)
+    path = reverse(routes.ENROLLMENT_LITTLEPAY_TOKEN)
     response = client.get(path)
 
     assert response.status_code == 200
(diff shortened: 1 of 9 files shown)


angela-tran:
This PR will need to coordinate with the one for the `index` view. I can wait to rebase on top of that one

thekaveman:
Putting this back in Draft while (someone) resolves the conflicts.

thekaveman (approved):
OK I rebased on the latest `main` (I hope I didn't break anything :sweat_smile:)

* Tests are all passing :white_check_mark: 
* Was able to run through a full enrollment locally :white_check_mark: 

I did make one _tiny_ change during the rebase: it hasn't been our pattern to use `unittest.mock` directly in tests, we typically use the `mocker` fixture provided by `pytest-mock`. I adjusted one of the tests to follow our typical pattern for consistency.

tests/pytest/enrollment_littlepay/test_views.py
```diff
@@ -12,6 +34,170 @@ def handler(request, status, exception):
...
+    assert response.status_code == 200
+    data = response.json()
+    assert "token" in data
+    assert data["token"] == mock_token["access_token"]
+
+
+@pytest.mark.django_db
+@pytest.mark.usefixtures("mocked_session_agency", "mocked_session_eligible")
+def test_token_valid(mocker, client):
+    mocker.patch.object(Session, "access_token", "enrollment_token")
```
thekaveman, line 77: This was the adjustment:

Previously this test used the `@patch()` decorator from `unittest.mock` to override the `Session.access_token` attribute with a `PropertyMock` (also from `unittest.mock`).

The small change was to use `mocker.patch.object` on `Session` to achieve the same result.

angela-tran: Makes sense! Thanks for the adjustment

lalver1 (approved):
Looks good! I also made sure that I could still enroll with Littlepay and everything worked.

PR #2946, opened 2025-05-29
Feat: Switchio API module

### LINKED ISSUE #2903: Add module to `enrollment_switchio` app for working with Switchio registration API
Depends on #2902 

## Acceptance Criteria

- [x] function for `POST /api/v1/registration` - _The operation establishes a registration request for the tokenization gate._
- [x] function for `GET /api/v1/registration/{regId}` - _With this operation, the carrier finds out the current registration request status._
- [x] unit tests

## Additional context

- contains functions, classes, etc. for interacting with the Switchio Registration API (see Section 8 of Switchio documentation linked in [Slack](https://cal-itp.slack.com/archives/C089LUYDMSP/p1742811592950169))
- ~~how to test? pytest tests that hit the real API (integration tests)? write the functions in a way where you can call them from the command line?~~
  - a test driver was used in #2936 for developing / iterating. It ran against a real test API environment, which was helpful for iterating quickly (didn't have to spend time mocking API behavior or making sure that mocked API behavior was correct). However, for automated tests that run in CI, we should use mocked API behavior.
  - we may add a workflow similar to [check-api.yml](https://github.com/cal-itp/benefits/actions/workflows/check-api.yml) that would run against a real test API environment

angela-tran:
Closes #2903

This PR introduces a `Client` class that encapsulates interaction with the Switchio registration API. This API is responsible for managing the back-end state of a card tokenization session. The two API endpoints allow you to
- establish a session and get back an ID and corresponding URL (`Client.request_registration`)
- get the status of a session, by ID (`Client.get_registration_status`)

Note about their naming: "registration" here is not "enrollment". If it helps at all, think of "registration" here as "card tokenization" or a "card tokenization session".

### DIFF
--- a/benefits/enrollment_switchio/api.py
+++ b/benefits/enrollment_switchio/api.py
@@ -50,13 +50,13 @@ class Client:
         self.client_certificate = client_certificate
         self.ca_certificate = ca_certificate
 
-    def _signature_input_string(self, timestamp, method, request_path, body=None):
+    def _signature_input_string(self, timestamp: str, method: str, request_path: str, body: str = None):
         if body is None:
             body = ""
 
         return f"{timestamp}{method}{request_path}{body}"
 
-    def _stp_signature(self, api_secret, timestamp, method, request_path, body=None):
+    def _stp_signature(self, api_secret: str, timestamp: str, method: str, request_path, body: str = None):
         input_string = self._signature_input_string(timestamp, method, request_path, body)
 
         # must encode inputs for hashing, according to https://stackoverflow.com/a/66958131
--- a/benefits/enrollment_switchio/api.py
+++ b/benefits/enrollment_switchio/api.py
@@ -56,11 +56,11 @@ class Client:
 
         return f"{timestamp}{method}{request_path}{body}"
 
-    def _stp_signature(self, api_secret: str, timestamp: str, method: str, request_path, body: str = None):
+    def _stp_signature(self, timestamp: str, method: str, request_path, body: str = None):
         input_string = self._signature_input_string(timestamp, method, request_path, body)
 
         # must encode inputs for hashing, according to https://stackoverflow.com/a/66958131
-        byte_key = api_secret.encode("utf-8")
+        byte_key = self.api_secret.encode("utf-8")
         message = input_string.encode("utf-8")
         stp_signature = hmac.new(byte_key, message, hashlib.sha256).hexdigest()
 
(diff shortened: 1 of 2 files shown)


benefits/enrollment_switchio/api.py
```diff
@@ -0,0 +1,126 @@
...
+        api_secret,
+        private_key,
+        client_certificate_file,
+        ca_certificate,
+    ):
+        self.api_url = api_url
+        self.api_key = api_key
+        self.api_secret = api_secret
+        self.private_key = private_key
+        self.client_certificate_file = client_certificate_file
```
angela-tran, line 50: Ugh, I never settled on whether I like `_file` here or not... These parameters, `private_key`, `client_certificate`(`_file`), and `ca_certificate`, are expecting file paths.

I'm just gonna remove `_file` from this

thekaveman: I like consistency more than anything :grin:

benefits/enrollment_switchio/api.py
```diff
@@ -0,0 +1,126 @@
...
+from datetime import datetime
+from enum import Enum
+import hashlib
+import hmac
+import json
+import requests
+
+
+@dataclass
+class Registration:
```
thekaveman, line 11: :pray: :pray: Love this naming :grin:

tests/pytest/enrollment_switchio/test_api.py
```diff
@@ -0,0 +1,107 @@
...
+
+    stp_signature = client._stp_signature(timestamp=timestamp, method=method, request_path=request_path, body=body)
+
+    # calculate the expected value
+    byte_key = client.api_secret.encode("utf-8")
+    input_string = client._signature_input_string(timestamp=timestamp, method=method, request_path=request_path, body=body)
+    message = input_string.encode("utf-8")
+    expected = hmac.new(byte_key, message, hashlib.sha256).hexdigest()
+
+    assert stp_signature == expected
```
thekaveman, line 46: I think this is fine (to reimplement the algorithm here in the test), an alternative could be:

* Hardcode a single `method`
* Hardcode a single `body`
* Hardcode a `timestamp`
* Pre-calculate what the signature should be for these inputs, and hardcode that to `expected`
* Then just call the function with these known inputs, and assert the result against the expected output

Since the test above `test_client_signature_input_string_with_body` validates the format for many inputs, we can be confident that part will be correct regardless of inputs here (_maybe_ spy on that helper to ensure it was called with the known inputs).

I don't think all the parameterization is absolutely necessary here, and it would remove the need to dupe the signature algorithm and slightly simplify the test.

angela-tran: Those are great points, thanks for helping me see some alternatives. I can look into simplifying this down

angela-tran:
Ready for re-review

thekaveman (approved):
This looks great :+1:

Issue #2906, opened 2025-05-13
Implement Switchio `IndexView`


angela-tran:
This ticket can be worked on in parallel to #2904 and #2905.

When #2904 calls `POST /api/v1/registration`, it needs to provide an `eshopRedirectUrl` which is where Switchio will redirect the user when they either click the Back button or Submit button. We will use the URL of the Enrollment Index view for `eshopRedirectUrl`.

This means `IndexView` needs to do the following for Switchio:
- check the status of the registration (call `GET /api/v1/registration/{regId}`)
- if the status indicates the user did not start tokenization, then show the user the normal Enrollment Index page
- if the status indicates the user has completed tokenization, then the card token should be in the response, and this is where we would call the new Switchio API to whitelist that card token. For this ticket, we won't be able to do that, so you can just redirect the user to the success page as if whitelisting had succeeded.

## Acceptance Criteria

- [ ]

## Additional context

This is the parameter in the response that represents the registration status: `regState`

![Image](https://github.com/user-attachments/assets/1df908cd-09aa-4010-8742-a82b96c5ae04)

Issue #2905, opened 2025-05-13
Implement Switchio enrollment template


angela-tran:
- call the endpoint to trigger the view function, and then given the gateway URL, launch the modal with that gateway URL
   - don't worry about defining a handler or any callbacks
   - don't have to do anything after launching the modal b/c Switchio will handle telling the backend of Benefits that something happened

## Acceptance Criteria

- [ ]

Issue #2904, opened 2025-05-13
Implement `GatewayUrlView`


angela-tran:
Depends on #2903 

In `enrollment/views.py`, we currently have a `token` function where we initiate tokenization for Littlepay by requesting a card tokenization access token. 

We need a similar view function that will initiate tokenization for Switchio. For Switchio, it will initiate tokenization by calling `POST /api/v1/registration` which will establish a registration request and return 2 things: a registration ID and a tokenization gateway URL. The tokenization gateway is what the template needs to continue the flow in the front-end.

When POST-ing to that endpoint, the `eshopRedirectUrl` parameter should be set to the `/enrollment` URL. (see #2906 for more context)

## Acceptance Criteria

- [x] Save the registration ID in the session to be used later on by the backend - #2952 
- [x] Return the tokenization gateway URL as JSON for the template to use - #2952 
- [x] 400 and 500 errors from calling the POST endpoint are handled appropriately - #2960 
- [ ] (optimization) If registration request has not expired yet, do not request a new one

## Additional context

Re: "(optimization) If registration request has not expired yet, do not request it again", here is a truncated screenshot of Switchio documentation about the Lifecycle of a Registration request:

<img src="https://github.com/user-attachments/assets/df3a3830-420d-42b8-bf8f-66aec6b6ce64" width=450>

PR #2948, opened 2025-05-29
Refactor: redirect to enrollment views

angela-tran:
Related to / needed by #2904, #2905, and #2906

This PR refactors the base `enrollment` app's `index` function to redirect to the Littlepay index view rather than call it as a function. Some needed code also had to be refactored to use `redirect` as well.

## More details

The motivation for this was that we will at some point need to modify the `index` function to have an if-check and send the user to either the Littlepay index view or the Switchio index view. In yesterday's dev workshop, we had a discussion about favoring usage of `redirect` rather than directly calling views because it is more in line with how Django's views work, so I thought I'd go ahead with that refactor so that we don't add another direct view call (to the Switchio index view).

It turns out that using `redirect` allows us to avoid a circular import and simplify our code some. See https://github.com/cal-itp/benefits/pull/2930#discussion_r2100610190 for more context on that, but basically we were able to move the enrollment-result handling logic to the `benefits.enrollment.enrollment` module.


benefits/enrollment/enrollment.py
```diff
@@ -14,3 +21,41 @@ class Status(Enum):
...
+            sentry_sdk.capture_exception(exception)
+            return redirect(routes.ENROLLMENT_SYSTEM_ERROR)
+
+        case Status.EXCEPTION:
+            analytics.returned_error(request, str(exception))
+            raise exception
+
+        case Status.REENROLLMENT_ERROR:
+            analytics.returned_error(request, "Re-enrollment error.")
+            return redirect(routes.ENROLLMENT_REENROLLMENT_ERROR)
```
angela-tran, line 61: If you're asking yourself why each of these blocks isn't just in their respective view functions, I think it's because for `SYSTEM_ERROR`, `EXCEPTION`, and `REENROLLMENT_ERROR`, we're wanting to do things with the `exception` variable, but it wouldn't be in scope anymore once we're redirected to the view function.

tests/pytest/enrollment_littlepay/test_enrollment.py
angela-tran, line 1: This was really just a move from `tests/pytest/enrollment/test_enrollment.py` to here... I then added a new `tests/pytest/enrollment/tests_enrollment.py` with unit tests for the `handle_enrollment_results` function introduced in this PR.

angela-tran, line 1: This was really just a move from `tests/pytest/enrollment/test_enrollment.py` to here... I then added a new `tests/pytest/enrollment/tests_enrollment.py` with unit tests for the [`handle_enrollment_results`](https://github.com/cal-itp/benefits/pull/2948/files#diff-433a540990e4f97e79fc8295208b143b53155897205cba1c9c75340d24c14702R26) function introduced in this PR.

thekaveman:
> It turns out that using `redirect` allows us to avoid a circular import and simplify our code some.

:tada: :sparkles: :tada: :sparkles: :tada:

thekaveman (approved):
Code changes look good, I like the clean up! :broom: :sponge: 

I didn't review the tests _that_ closely but since they're all passing, seems good :+1: 

Successfully ran through a complete enrollment flow locally on this branch.

PR #2977, opened 2025-06-24
Feat: Switchio enrollment redirect

### LINKED ISSUE #2906: Implement Switchio `IndexView`

angela-tran:
Closes part of #2905 and closes #2906 

The part for #2905: the Switchio enrollment template now calls `tokenizeWithRedirect` from the st-rp-sdk library. When the user clicks the "Enroll" button, they are taken to an external page (the Switchio tokenization gateway for their registration ID) and then after submitting their card details, they will be redirected back to Benefits.

The part for #2906: the Switchio `IndexView` was updated to check the registration status, and if its `regState` is `tokenization_finished`, we redirect to the success page. When we get to implementing card registration, this is where we would take the card token and try to enroll it.

By using `tokenizeWithRedirect` instead of `tokenizeInIframe`, we don't have to worry about changing any Content-Security Policy settings or any cookie settings (see #2975 for comparison).

## Reviewing

Make sure you're configured to use Switchio (see #2952 for steps), and try enrolling a test card with Switchio. If you set a breakpoint in `IndexView` on [line 47](https://github.com/cal-itp/benefits/pull/2977/files#diff-d6e9f70aaccdb9bc5d303838b7a9481105acb8aca883c47580f46f25a489ca3aR47) or so, you can look at `response.registration_status.tokens` to see the card token.

Also, try the "Back" button on the external Switchio page.

## Other notes

I will add sending of analytic events for `started card tokenization` and `finished card tokenization` in a separate PR, which is why #2905 cannot be closed yet. It requires a bit more thought to figure out how and where to send the `finished card tokenization` event.


lalver1 (approved):
Looks great!
Tested locally (the back button too) and it worked as expected 👍

PR #2982, opened 2025-06-27
Feat: Switchio tokenization analytics

### LINKED ISSUE #2905: Implement Switchio enrollment template

angela-tran:
Closes #2905 

Sends the "started card tokenization" and "finished card tokenization" analytic events.

### DIFF
--- a/benefits/enrollment_switchio/views.py
+++ b/benefits/enrollment_switchio/views.py
@@ -77,6 +77,9 @@ class IndexView(AgencySessionRequiredMixin, EligibleSessionRequiredMixin, FormVi
                     return self.render_to_response(context=context_data)
                 elif reg_state == "verification_failed":
                     return redirect(routes.ENROLLMENT_RETRY)
+                elif reg_state == "tokenization_failed":
+                    sentry_sdk.capture_exception(Exception("Tokenization failed"))
+                    return redirect(routes.ENROLLMENT_SYSTEM_ERROR)
             else:
                 sentry_sdk.capture_exception(response.exception)
 
--- a/tests/pytest/enrollment_switchio/test_views.py
+++ b/tests/pytest/enrollment_switchio/test_views.py
@@ -147,6 +147,34 @@ class TestIndexView:
 
     @pytest.mark.django_db
     @pytest.mark.usefixtures("mocked_session_flow")
+    def test_get_with_session_registration_id_tokenization_failed(
+        self, view, app_request, mocker, model_TransitAgency, model_SwitchioConfig
+    ):
+        model_TransitAgency.switchio_config = model_SwitchioConfig
+        gateway_url = "https://example.com/cst/?regId=1234"
+        mocked_get_registration_status = mocker.patch(
+            "benefits.enrollment_switchio.views.get_registration_status",
+            return_value=RegistrationStatusResponse(
+                status=Status.SUCCESS,
+                registration_status=RegistrationStatus(
+                    regState="tokenization_failed",
+                    created=datetime.now(),
+                    mode="register",
+                    eshopResponseMode="query",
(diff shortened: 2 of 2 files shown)


benefits/enrollment_switchio/views.py
```diff
@@ -53,7 +53,11 @@ def get(self, request: HttpRequest, *args, **kwargs):
             response = get_registration_status(switchio_config=switchio_config, registration_id=session.registration_id)
             if response.status is Status.SUCCESS:
                 if response.registration_status.regState == "tokenization_finished":
-                    return redirect(routes.ENROLLMENT_SUCCESS)
+                    # this is where we will enroll the card token that's inside response.registration_status.tokens
+
+                    # render index template which will send "finished card tokenization" analytics event and then redirect
+                    # to success page
+                    return super().get(request=request, *args, **kwargs)
```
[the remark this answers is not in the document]
angela-tran: Yeah, that's a good point that enrollment and its analytic events would happen before the "finished tokenization" event gets fired, which is not what we want. I've been re-working this PR so will mark as draft to reflect that. Thanks for this note!

benefits/enrollment_switchio/views.py
```diff
@@ -52,8 +66,17 @@ def get(self, request: HttpRequest, *args, **kwargs):
...
+                reg_state = response.registration_status.regState
+                if reg_state == "tokenization_finished":
+                    # give card token to index template so it can send
+                    # "finished card tokenization" event and POST either the
+                    # CardTokenizeSuccessForm or CardTokenizeFailForm.
+                    context_data = self.get_context_data(**kwargs)
+
+                    context_data["card_token"] = get_latest_active_token_value(response.registration_status.tokens)
+                    return self.render_to_response(context=context_data)
+                elif reg_state == "verification_failed":
```
thekaveman, line 78: Is there a possibility of the `else` condition happening here? E.g.

`reg_state != "tokenization_finished" and reg_state != "verification_failed"`

benefits/enrollment_switchio/views.py
```diff
@@ -64,6 +87,15 @@ def get(self, request: HttpRequest, *args, **kwargs):
 
         return super().get(request=request, *args, **kwargs)
 
+    def form_valid(self, form):
+        card_token = form.cleaned_data.get("card_token")  # noqa
+
+        # (this is where we will enroll the card token)
```
thekaveman, line 93: Nice refactor 👍

benefits/enrollment_switchio/views.py
angela-tran: It is possible, and if so, the logic here is to fall through to line 88 and simply return the index template.

I've been working off the documented list of `regState` values in section 15.1 "Lifecycle" > "Registration request" of the PDF for the registration API.

From that list, we're handling `expired` and `deleted` in `GatewayUrlView` and handling `tokenization_finished` and `verification_failed` in `IndexView`. The other documented values don't seem like states we need to handle or that we would ever even see. They seem like values that would be set in Switchio's back-end as it is performing tokenization.

Maybe `tokenization_failed` is the only other state that we should handle here?

thekaveman: Hmm I see yeah, I was more worried about like a dead-end, but it does just fall down to line 88 👍 

> Maybe `tokenization_failed` is the only other state that we should handle here?

Seems like a good idea, e.g. sending a Sentry error so we get notified of this.

PR #2989, opened 2025-07-03
Refactor: SwitchioConfig model - support enrollment API

### LINKED ISSUE #2886: Switchio integration: refactor `TransitAgency.switchio_config` to be a OneToOneField
We learned that the API key and secret will be specific to each agency, so we should update the model definition to more correctly model the relationship between the two entities (`TransitAgency` and `SwitchioConfig`). This is important to protect against misconfiguration.

## Acceptance Criteria

- [x] `TransitAgency.switchio_config` is a OneToOneField
- [x] Validation logic and tests are updated to reflect that
- [x] `SwitchioConfig` has an `agency_slug` field since it represents the configuration for a specific agency

### LINKED ISSUE #2992: Switchio integration: update `SwitchioConfig` model to hold authorization header for new API
The new Switchio API does not require an API key or secret, but rather uses basic HTTP authentication via an `Authorization` header. The `SwitchioConfig` needs to store this value so we have it available when instantiating the client for calling the new API.

## Acceptance Criteria
 
- [ ] `SwitchioConfig` model has a new field for the authorization header

angela-tran:
Closes #2992 and closes #2886 

Model changes needed to support connecting to Switchio enrollment API.

Also went ahead and simplified `TransitAgency.switchio_config` field to be a one-to-one field and added `agency_slug` to `SwitchioConfig` since it'll be for a specific agency.


angela-tran:
Closing this in favor of new PRs that take a different approach to what's in this PR

Issue #2993, opened 2025-07-08
Switchio integration: update Switchio `IndexView` to enroll the card token


angela-tran:
## Acceptance Criteria
 
- [x] A card can be enrolled in a non-expiring discount

angela-tran:
The endpoints for adding and removing a group to/from a token are not working currently, so waiting for updates from the Switchio dev team

PR #3027, opened 2025-07-21
Fix: Switchio API datetime parsing

angela-tran:
Part of #2993

Adds a `__post_init__` method to `RegistrationStatus` and `GroupExpiry` so that their datetime fields are parsed into Python datetime objects. We need `GroupExpiry.expiresAt` to be a datetime object to implement enrollment of expiring discounts.

### DIFF
--- a/tests/pytest/enrollment_switchio/test_views.py
+++ b/tests/pytest/enrollment_switchio/test_views.py
@@ -92,7 +92,7 @@ class TestIndexView:
                 status=Status.SUCCESS,
                 registration_status=RegistrationStatus(
                     regState="tokenization_finished",
-                    created=datetime.now(),
+                    created=datetime.now().isoformat(),
                     mode="register",
                     eshopResponseMode="query",
                     tokens=[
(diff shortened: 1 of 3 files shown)


tests/pytest/enrollment_switchio/test_views.py
```diff
@@ -92,7 +92,7 @@ def test_get_with_session_registration_id_tokenization_finished(
                 status=Status.SUCCESS,
                 registration_status=RegistrationStatus(
                     regState="tokenization_finished",
-                    created=datetime.now(),
+                    created=datetime.now().isoformat(),
```
thekaveman, line 95: I'm assuming this is what the dates look like coming out of the API?

In `littlepay` we have to replace the `Z` ISO offset character in the input datetime string (that we get from the API) with `+00:00` for Python's parsing to handle it correctly: https://github.com/cal-itp/littlepay/blob/main/littlepay/api/funding_sources.py#L47

thekaveman (approved):
👍 we do the same thing in the [`littlepay` library](https://github.com/cal-itp/littlepay/blob/main/littlepay/api/funding_sources.py#L35)

PR #3033, opened 2025-07-22
Switchio: Help page

### LINKED ISSUE #2909: Switchio integration: update Help page copy
Update the Help page based on Figma designs, as clarified in #2864.

## Acceptance Criteria

- [ ] Help page copy is updated for the use case where the user has not yet selected an agency (coming from the Home page)
- [ ] Help page copy is updated for the use case where the user has selected an agency

machikoyasuda:
closes #2909 

## What this PR does
- Add `littlepay_config`, `switchio_config` to Help page context processor
- Add 3 variations of help copy: Copy before the user has selected an agency, copy for Littlepay and copy for Switchio

## How to test
- Test for help page w/o clicking an agency
- Test after clicking an agency with Littlepay
- Go to Admin to turn off Littlepay, and turn on Switchio for the agency config. Then select the agency and check help page


benefits/core/templates/core/help.html
```diff
@@ -116,19 +116,51 @@ <h2 class="h2-sm pt-4 pt-lg-8" id="{{ flow_help.id }}">{{ flow_help.headline }}<
...
+              transit benefit to the contactless card of your choice.
+            {% endblocktranslate %}
+          </p>
+          <p class="pt-3 pt-lg-4">
+            {% blocktranslate trimmed %}
+              For more information on Littlepay, please visit the
+              <a href="https://littlepay.com" target="_blank" rel="noopener noreferrer">Littlepay website</a>.
+            {% endblocktranslate %}
+          </p>
+        {% elif agency.switchio_config %}
```
thekaveman, line 134: This should also check `agency`, like the `if` condition, right? 

```
{% elif agency and agency.switchio_config %}
```

thekaveman (changes requested):
Small fix in the `elif`

benefits/core/templates/core/help.html
thekaveman: I think the `else` is fine. If no agency or no config is present, it makes sense to show the full string / all options.

thekaveman (approved):
✅ Confirm I see copy for both when I haven't selected an agency
✅ Confirm I see copy for both when neither Littlepay nor Switchio are configured for the selected agency
✅ Confirm I see copy for Littlepay only when that is the configured transit processor for the selected agency
✅ Confirm I see copy for Switchio only when that is the configured transit processor for the selected agency
:white_check_mark: Confirm I see Spanish copy for all scenarios as well

PR #3056, opened 2025-07-31
Fix: allow `ventura` slug to go through `TransitAgencyPathConverter`

### LINKED ISSUE #3055: Saving a `TransitAgency` with a slug of `ventura` results in `NoReverseMatch`
## To Reproduce

1. Log in to Benefits Admin as superuser
2. Add or change a TransitAgency so that its slug is `ventura`
3. Click Save
4. See `NoReverseMatch` error
5. Try going to the main app (`/`)
6. Also see `NoReverseMatch` error

From the stacktrace, we can see that a context processor is attempting to read the `eligibility_index_url` property from the session transit agency, which attempts to reverse the URL pattern `<agency: agency>/eligibility`. However, the regex on the `TransitAgencyPathConverter` matches only up to 5 characters.

As a quick fix to unblock production validation of Switchio enrollment for VCTC, we can update the regex to increase the maximum number of characters. 

I also think we should spend some time looking into what `eligibility_index_url` is being used for and if that can be simplified. Also, is it necessary to have this character limit on `TransitAgencyPathConverter`?

Stack trace:
```
  File "/calitp/app/benefits/core/context_processors.py", line 62, in active_agencies
    return {"active_agencies": [_agency_context(agency) for agency in agencies]}
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/calitp/app/benefits/core/context_processors.py", line 62, in <listcomp>
    return {"active_agencies": [_agency_context(agency) for agency in agencies]}
                                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/calitp/app/benefits/core/context_processors.py", line 26, in _agency_context
    "eligibility_index_url": agency.eligibility_index_url,
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/calitp/app/benefits/core/models/transit.py", line 178, in eligibility_index_url
    return reverse(routes.AGENCY_ELIGIBILITY_INDEX, args=[self.slug])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/calitp/.local/lib/python3.11/site-packages/django/urls/base.py", line 88, in reverse
    return resolver._reverse_with_prefix(view, prefix, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/calitp/.local/lib/python3.11/site-packages/django/urls/resolvers.py", line 831, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'agency_eligibility_index' with arguments '('ventura',)' not found. 1 pattern(s) tried: ['(?P<agency>[a-zA-Z]{3,5})/eligibility\\Z']
```

## Expected behavior
`ventura` should be supported as a slug on `TransitAgency`.

## Screenshots

Image

angela-tran:
Closes #3055

See the issue above for context. 

This PR removes the upper limit for what values can go through the `TransitAgencyPathConverter`. Increasing the limit just sets us up to run into it again in the future when / if we have an agency with a longer slug.

### DIFF
--- a/benefits/core/urls.py
+++ b/benefits/core/urls.py
@@ -17,7 +17,7 @@ class TransitAgencyPathConverter:
     """Path converter to parse valid TransitAgency objects from URL paths."""
 
     # used to test the url fragment, determines if this PathConverter is used
-    regex = "[a-zA-Z]{3,5}"
+    regex = "[a-zA-Z]{3,7}"
 
     def to_python(self, value):
         """Determine if the matched fragment corresponds to an active Agency."""
--- a/benefits/core/urls.py
+++ b/benefits/core/urls.py
@@ -17,7 +17,7 @@ class TransitAgencyPathConverter:
     """Path converter to parse valid TransitAgency objects from URL paths."""
 
     # used to test the url fragment, determines if this PathConverter is used
-    regex = "[a-zA-Z]{3,7}"
+    regex = "[a-zA-Z]{3,}"
 
     def to_python(self, value):
         """Determine if the matched fragment corresponds to an active Agency."""
--- a/benefits/core/urls.py
+++ b/benefits/core/urls.py
@@ -48,10 +48,10 @@ app_name = "core"
 urlpatterns = [
     path("", views.index, name=routes.name(routes.INDEX)),
     path("help", views.help, name=routes.name(routes.HELP)),
+    path("logged_out", views.logged_out, name=routes.name(routes.LOGGED_OUT)),
+    path("error", views.server_error, name=routes.name(routes.SERVER_ERROR)),
     path("<agency:agency>", views.agency_index, name=routes.name(routes.AGENCY_INDEX)),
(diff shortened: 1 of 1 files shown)


lalver1 (approved):
This looks great!

I also saw the same behavior described in #3055 when setting up `ventura` to be able to review #3058 and this PR fixed it 👍

I agree, it looks like we don't need an upper limit on the regex since [`models.SlugField`](https://docs.djangoproject.com/en/5.2/ref/models/fields/#slugfield) has a default max length of 50 anyway, so it seems that this is what would eventually validate the upper limit.

PR #3061, opened 2025-08-01
Fix: remove anonymous blob access to storage account

### LINKED ISSUE #3059: Disable anonymous blob access on storage account
We don't have any need for anonymous blob access on our storage account, and DevSecOps requested we disable it.

The property to set is [`allow_nested_items_to_be_public`](https://registry.terraform.io/providers/hashicorp/azurerm/3.117.0/docs/resources/storage_account#allow_nested_items_to_be_public-3)

You can view the setting in the Azure portal at the storage account's `Settings` > `Configuration` > `Allow Blob anonymous access`.

## Additional context

- https://stackoverflow.com/questions/73741756/disabling-allow-public-blob-access-using-terraform
- https://learn.microsoft.com/en-us/azure/storage/blobs/anonymous-read-access-prevent?tabs=portal

angela-tran:
Closes #3059 

Adds `allow_nested_items_to_be_public = false` to storage account definition.

### DIFF
--- a/terraform/storage.tf
+++ b/terraform/storage.tf
@@ -1,9 +1,10 @@
 resource "azurerm_storage_account" "main" {
-  name                     = "sacdtcalitp${lower(local.env_letter)}001"
-  location                 = data.azurerm_resource_group.main.location
-  resource_group_name      = data.azurerm_resource_group.main.name
-  account_tier             = "Standard"
-  account_replication_type = "RAGRS"
+  name                            = "sacdtcalitp${lower(local.env_letter)}001"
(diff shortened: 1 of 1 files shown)


angela-tran:
From the `terraform plan` in the [pipeline run](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=94858&view=logs&jobId=3f3994ef-4604-51e5-3e41-c7d80cf9338d&j=5af648c4-4294-55a7-6072-dd70820d3dba&t=42899ae2-6397-58a2-7641-cec3214b180b) for this PR:

image

Issue #2777, opened 2025-03-21
Refactor to class-based views


thekaveman:
We've made good progress getting a better handle on the templates via our new `context` approach. The next evolution is a more modern Django approach of utilizing [class-based views](https://docs.djangoproject.com/en/5.1/topics/class-based-views/intro/) rather than simple view functions. 

The main goal is to organize and consolidate the views, reducing the footprint by co-locating the context where appropriate and relying on more OOTB Django functionality to streamline the code.

## Acceptance criteria

- [ ] Refactor all view functions to class-based views
- [ ] Organize via sub package of view modules e.g. `benefits/core/views/__init__.py`
- [ ] One file per view class, named the same as the view function e.g. `benefits/core/views/index.py`
- [ ] Move all corresponding copy context into [`get_context_data()`](https://docs.djangoproject.com/en/5.1/topics/class-based-views/generic-display/#adding-extra-context) implementation per view class
- [ ] Refactor to class-based tests for all view tests

## Questions

- Can we create view classes that are pre-wrapped with middleware? And then create an inheritance hierarchy to leverage base classes that e.g require a transit agency or flow or eligibility? Are view class mixins an option here?

- What context can/should move from `context_processors` into `get_context_data()` or similar for specific views? Can we utilize view inheritance or mixins to provide common context to views that need it?

PR #3065, opened 2025-08-06
Refactor: class-based view for Eligibility Index

### LINKED ISSUE #2868: Class-based views: Eligibility index

angela-tran:
Closes #2868 

This PR converts the eligibility `index` view function into a class-based view.

The corresponding copy context for the Eligibility Index was moved into the CBV's `get_context_data` method.

## Open question

This PR is all working and technically ready to merge, but I wanted to call something out for discussion and to get ahead of this PR possibly having a "surprising" effect :persevere: ...

In our other CBV refactors, I haven't seen us move the copy context objects into the CBV, but I thought this was one of the main reasons for moving to CBVs (#2777). I'm trying my best to help us achieve what we set out to do but am feeling like this PR might cause controversy because it's deviating from the norm.


lalver1 (changes requested):
Thanks for highlighting how this PR differs from previous ones in that it moves the context into the CBV as we set out initially, and thanks for this reminder 🙂 

I tend to prefer this approach since the context and the view are now in the same place. If we need to update or add copy later, it seems that this setup makes it easier to find it since the context is closer to the view (and consequently the `html` via `template_name`). The code inside the views does get longer, but maybe that's ok? 

Just noticed a minor typo, unrelated to the open question, in the comment below

thekaveman:
RE: your open question @angela-tran 

I'm glad you bit the bullet and did this here. I was kind of avoiding it because it seemed like another layer of things to think about (OK OK, I was lazy :sweat_smile:).

But I agree with you and Luis, it makes a lot of sense to keep things self-contained in the CBVs. 

_Originally_ originally, we even had the idea to have a module (file) for each CBV. As we move context data into the views, we may want to go back to that idea to make for more reasonable and readable file sizes.

thekaveman:
@angela-tran you may want to rebase on the latest `main` to get the `pip install` fixes.

angela-tran:
Thanks @lalver1 and @thekaveman for adding your thoughts. It makes sense that there was hesitation around moving the context over since clearly I had hesitation too about what questions it would potentially open up. Glad to hear this is still the direction we want to go.

PR #3070, opened 2025-08-07
Fix: secret references

thekaveman:
~In Terraform, the `locals.secret_prefix` is like a function call to `@Microsoft.KeyVault(...` and so it needs the trailing `)` to complete the reference.~

~You can see the issue in the App Service environment variables list. E.g. compare `SSO_SHOW_FORM_ON_ADMIN_PAGE` with `SWITCHIO_PROD_API_BASE_URL`:~

image

~See https://learn.microsoft.com/en-us/azure/app-service/app-service-key-vault-references~

While all of the above is technically correct... we don't actually need these secrets as environment variables. They are referenced from model instances of `LittlepayConfig` and `SwitchioConfig`: https://github.com/cal-itp/benefits/pull/3070#discussion_r2261155054

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -57,27 +57,21 @@ resource "azurerm_linux_web_app" "main" {
     "REQUESTS_READ_TIMEOUT"    = "${local.secret_prefix}requests-read-timeout)",
 
     # Django settings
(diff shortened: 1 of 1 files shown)


angela-tran (approved):
:raised_hands:

PR #3079, opened 2025-08-11
Fix: set TransitAgency for in-person eligibility URL

### LINKED ISSUE #2938: AttributeError: 'NoneType' object has no attribute 'enrollment_flows'
Sentry Issue: [BENEFITS-CT](https://sentry.calitp.org/organizations/sentry/issues/81781/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'enrollment_flows'
(3 additional frame(s) were not displayed)
...
  File "benefits/in_person/views.py", line 28, in eligibility
    "form": forms.InPersonEligibilityForm(agency=agency),
  File "benefits/in_person/forms.py", line 29, in __init__
    flows = agency.enrollment_flows.filter(supported_enrollment_methods__contains=models.EnrollmentMethods.IN_PERSON)
```

lalver1:
Closes #2938

This PR sets the transit agency for the user in the case that the user directly goes to the `/in_person/eligibility` URL without signing in first at `/admin`. It also refactors the `eligibility` view to a class-based view.

### DIFF
--- a/benefits/in_person/views.py
+++ b/benefits/in_person/views.py
@@ -59,13 +59,14 @@ class EligibilityView(FormView):
         return kwargs
 
     def form_valid(self, form):
-        """If the form is valid, set enrollment flow and redirect."""
+        """If the form is valid, set enrollment flow, eligible session, and redirect."""
 
         flow_id = form.cleaned_data.get("flow")
         flow = models.EnrollmentFlow.objects.get(id=flow_id)
-        session.update(self.request, flow=flow)
+        session.update(self.request, flow=flow, eligible=True)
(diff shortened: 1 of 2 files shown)


thekaveman (changes requested):
I'm testing #3076 and we need a few slight changes in this view to make it so we can reuse the CBVs from Digital:

- We need to set session `eligibility = True` on `form_valid` in this view
  - This is because the `Enrollment` CBVs in Digital use the `EligibleSessionRequiredMixin`
- We could also change how the eligibility analytics are sent
  - `returned eligibility` is currently sent in the `enrollment` view, it can get moved to the `form_valid` here (since that means the agency user verified the rider's eligibility and submitted the form), so we're basically sending `selected`, `started`, and `returned` one right after the other.

PR #3082, opened 2025-08-12
Feat: base Enrollment view for in-person

### LINKED ISSUE #3075: In-person: define and register base EnrollmentIndex view
We need a base Enrollment view for In-person, that redirects to the appropriate transit processor specific Enrollment view (e.g. [Littlepay](#3067) or [Switchio](#3069)). 

## Acceptance Criteria

- [ ] New `RedirectView` CBV inheriting from the index view on the digital side (see #3073)
- [ ] URL registration is updated to point to this view
- [ ] Redirect to the appropriate In-person enrollment view based on transit agency

lalver1:
Closes #3075

This PR defines a base Enrollment view for in-person. The new base view inherits from `IndexView` on the Digital side and redirects to the appropriate in-person enrollment view based on the transit agency.

## Notes

- The in_person `urls.py` file has placeholders for `enrollment/littlepay` and `"enrollment/switchio` but both are associated with the current `enrollment` (Littlepay) view so that the app continues to work during this in-person refactor.
- ~The refactor in be4445ae1512eabadbe8970a7cb76e563a8df996 tried to make the least amount of changes to `TransitAgency` yet still introduce a way for it to support in-person too. We could replace this with a more significant refactor (such as changing `enrollment_index_route` from a property to a function and passing in the enrollment URLs, like the proposed `get_route_by_processor`, where the view is responsible for passing in the URLs). This wouldn't require too many changes since `enrollment_index_route` is only called in one place in the code, most changes would be related to updating tests.~

## Reviewing

1. Start the app with `F5`
2. In the browser, enter the URL `/admin`
3. Log in with `cst-user`
4. Run through an enrollment and verify that you reach the `in_person/enrollment/littlepay` page after selecting an eligibility type

Also ensure that the Digital side still works.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -28,11 +28,8 @@ class IndexView(AgencySessionRequiredMixin, EligibleSessionRequiredMixin, Redire
 
     route_origin = routes.ENROLLMENT_INDEX
 
-    def get_target_route_name(self):
-        return self.agency.enrollment_index_route
-
     def get_redirect_url(self, *args, **kwargs):
-        route_name = self.get_target_route_name()
+        route_name = self.agency.enrollment_index_route
         return reverse(route_name)
 
     def get(self, request, *args, **kwargs):
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -26,8 +26,12 @@ logger = logging.getLogger(__name__)
 class IndexView(AgencySessionRequiredMixin, EligibleSessionRequiredMixin, RedirectView):
     """CBV for the enrollment landing page."""
 
+    def get_target_route_name(self):
+        return self.agency.enrollment_index_route
+
     def get_redirect_url(self, *args, **kwargs):
-        return reverse(self.agency.enrollment_index_route)
+        route_name = self.get_target_route_name()
+        return reverse(route_name)
 
     def get(self, request, *args, **kwargs):
         session.update(request, origin=reverse(routes.ENROLLMENT_INDEX))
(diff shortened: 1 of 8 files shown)


thekaveman (changes requested):
So I think this all works (I tried it locally, I'm not fully configured but it seems to step through correctly).

I wonder if we can simplify just slightly though: introduce a new prop on the `TransitAgency` class like:

```python
@property
def in_person_enrollment_index_route(self):
     # mirror what enrollment_index_route() does
```

Essentially, don't make external callers figure out how to get what they need (having to call the `get_route_by_processor` in some cases but not others). Just make it explicit through a new prop that we need.

This will also be friendlier when we expand to additional transit processors -- no function signatures or call sites to change, just a prop implementation that will need updating.

benefits/enrollment/views.py
```diff
@@ -26,11 +26,17 @@
 class IndexView(AgencySessionRequiredMixin, EligibleSessionRequiredMixin, RedirectView):
     """CBV for the enrollment landing page."""
 
+    route_origin = routes.ENROLLMENT_INDEX
+
+    def get_target_route_name(self):
+        return self.agency.enrollment_index_route
```
thekaveman, line 32: Not the biggest deal, but it seems like `get_target_route_name` doesn't do much more / provide a lot of value over the existing `RedirectView.get_redirect_url`... could you just do this logic there? Then override `get_redirect_url` in the sub-class?

thekaveman (approved):
Not a blocker, 1 more suggestion if you want to take it or leave it!

benefits/enrollment/views.py
lalver1: Ah I agree, this is a nice simplification 👍 Thanks for the suggestion! Made the change in 35d16ca7d1970b995b395f6bfee5c8113e529d4f

Issue #3067, opened 2025-08-06
In-person: Refactor `enrollment` view to CBV inheriting from Littlepay `IndexView`


thekaveman:
A clear and concise description of the task.

## Acceptance Criteria

- [ ] Rewrite/rename existing `templates/enrollment/index.html` to be specific to Littlepay
- [ ] Update URL / paths
- [ ] Allow for different error handling / system errors

PR #3084, opened 2025-08-13
Refactor: convert in-person token view to CBV

lalver1:
attributes are extracted from TokenView so that the new in-person
class-based view (InPersonTokenView) can reuse the existing logic.
the in-person token/ URL is updated to use the new CBV.


benefits/enrollment_littlepay/views.py
```diff
@@ -32,12 +39,12 @@ def get(self, request, *args, **kwargs):
             elif response.status is Status.SYSTEM_ERROR or response.status is Status.EXCEPTION:
                 logger.debug("Error occurred while requesting access token", exc_info=response.exception)
                 sentry_sdk.capture_exception(response.exception)
-                analytics.failed_pretokenization_request(request, response.status_code)
+                self._handle_analytics(request, response.status_code)
```
thekaveman, line 42: Take a look at [#3085](https://github.com/cal-itp/benefits/pull/3085/files#diff-d6e9f70aaccdb9bc5d303838b7a9481105acb8aca883c47580f46f25a489ca3aR28) for how I did this... 

I think it is more straightforward to have an attribute for the `enrollment_method` and just pass that in vs. a new method to override. Similar to the `route_*` attributes, makes for a simpler inheritance.

benefits/in_person/urls.py
```diff
@@ -10,7 +10,9 @@
     path(
         "eligibility/", admin.site.admin_view(views.EligibilityView.as_view()), name=routes.name(routes.IN_PERSON_ELIGIBILITY)
     ),
-    path("token/", admin.site.admin_view(views.token), name=routes.name(routes.IN_PERSON_ENROLLMENT_TOKEN)),
+    path(
+        "token/", admin.site.admin_view(views.InPersonTokenView.as_view()), name=routes.name(routes.IN_PERSON_ENROLLMENT_TOKEN)
```
thekaveman, line 14: This endpoint is only relevant for `littlepay` enrollment, e.g. should be `/in_person/enrollment/littlepay/token`

benefits/in_person/views.py
```diff
@@ -71,34 +70,17 @@ def form_valid(self, form):
         return redirect(routes.IN_PERSON_ENROLLMENT)
 
 
-def token(request):
+class InPersonTokenView(TokenView):
```
thekaveman, line 73: We are in the `in_person` package, so we don't need/want to prefix classes with `InPerson`.

Further, this view is only relevent for Littlepay. Take a look at [#3085](https://github.com/cal-itp/benefits/pull/3085/files#diff-2ac584c1a64835c39f2c48a86098c554eaf2d49d9409c73b4b1abe4f104c0cacR271), rename to e.g. `LittlepayTokenView`.

thekaveman (changes requested):

benefits/in_person/views.py
lalver1: Yep, renamed it to `LittlepayTokenView` 👍

lalver1:
Thanks @thekaveman! I made the last small change in eb0ec91d5adbd8658c81d758a2bea1b07a0cb626 so you can review whenever you get a chance!

lalver1:
>I'm not sure exactly what is going on... I actually see this same behavior in main, so I don't think it is necessarily related to this specific PR/refactor.

🤔 that's interesting. Yep, that makes sense to address as part of #3067 👍

PR #3114, opened 2025-08-21
Fix: allow language cookie to travel with user

### LINKED ISSUE #3081: App switches to English after eligibility check
App does not retain preference for copy in Spanish.

## To Reproduce

1. Go to 'test-benefits.calitp.org/vctc'
2. Click on `Español` button to change language to Spanish.
3. Proceed to benefit options. Choose an enrollment pathway. I used Older Adults.
4. Complete ID verification using Login.gov (via the IdG).
5. When I return to the app, the copy on the eligibility confirmation view is in English.

## Expected behavior

Eligibility confirmation view displays in Spanish.

## Screenshots

(https://share.zight.com/9ZupBvmr)

- OS: MacOS Sequoia 15.6
- Browser: Safari
- Version: 18.6

## Additional context

- Occurs regardless of agency choice.
- May happen in other contexts in which the rider returns to the app from an external site, e.g. Switchio.

First noted in

thekaveman:
Similar to the session cookie, we need to have the language cookie's `SameSite` flag set to `Lax` so that the cookie stays with the user during redirects e.g. for IdG/Login.gov and Switchio.

Otherwise, the user gets the app's default language (`en`) upon redirection back, even if they originally left the app with a different language selection (e.g. `es`).

Fixes #3081 

See more: https://docs.djangoproject.com/en/5.2/ref/settings/#language-cookie-samesite

## Reviewing

1. Checkout this branch, start the app with `F5`
2. Click the `Español` button to change to `es`
3. Use the browser dev tools to confirm a cookie was set with name `django_language`, value `es`, and `SameSite = Lax`
4. Select a flow that takes you through Login.gov, e.g. Older Adult
5. Complete identity proofing with Login.gov
6. Confirm that the app is still displayed in `es` when you are redirected back
7. (The same behavior should be seen for Switchio enrollment through their gateway redirect)

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -211,7 +211,10 @@ AUTH_PASSWORD_VALIDATORS = [
 LANGUAGE_CODE = "en"
 
 LANGUAGE_COOKIE_HTTPONLY = True
-LANGUAGE_COOKIE_SAMESITE = "Strict"
+# `Lax` allows the cookie to travel with the user and be sent back to Benefits
+# during redirection e.g. through IdG/Login.gov or a Transit Processor portal
+# ensuring the app is displayed in the same language
+LANGUAGE_COOKIE_SAMESITE = "Lax"
 LANGUAGE_COOKIE_SECURE = True
 
 LANGUAGES = [("en", "English"), ("es", "Español")]


angela-tran (approved):
Makes sense! I tested it locally with Login.gov and Switchio, and it looks great :raised_hands:

Issue #2581, opened 2024-12-10
Agency onboarding: El Dorado Transit


mariearaneta:
Brian from El Dorado Transit has completed their onboarding form, which can be found [here](https://app.hubspot.com/submissions/5519226/form/2ec0c588-723e-4c98-9e16-b044819c83e2/submissions/0e3cc3c4-94d2-48a5-8275-7cc833259921)

They are currently set to launch contactless in March.

indexing:
@o-ram tagging you for awareness.

mariearaneta:
Hi @indexing we estimate starting configuration late May

mariearaneta:
FYI El Dorado is waiting on their hardware from Kuba. They are still estimating beginning configuration late May/early June.

mariearaneta:
Launch has been pushed back and is now expected to be September 1

o-ram:
@indexing thanks for the questions, it surfaced some form items that weren't behaving as we wanted them to, so I've updated things a bit! You should be able to see some updates either in the original link provided in the issue body, or at [this link](https://app.hubspot.com/contacts/5519226/record/0-3/35497212810/), the latter of which anyone with an account should be able to see, regardless of permissions (just make sure you are on the From Submissions tab).

Based on that:

- El Dorado (like Nevada County) does not have an agency short name, which is what the form answer `No shorter name` indicates. I noticed that the flag/explanation for that wasn't showing in the form view, though, so I updated it so it does. When @mariearaneta has a chance to chime in, she might have thoughts on what they might be ok with using in the slug, much like we went with NevCo for Nevada.
- Enrollment pathways - The property that confirms the Benefits pathways they want (vs. the overview of all their current discount groups) had restricted viewing permissions, which I've since fixed. Based on that, the form responses Marie recorded appear to show they want `older adults` and `Medicare cardholder `pathways. But, again, will let her confirm in case they have since mentioned to her any interest in other pathways, like Vets.

mariearaneta:
@indexing 

- El Dorado would like to go with **EDCTA** for their short name. 
- Enrollment Pathway - Yes, they initially requested `Medicare` and `Older Adult`, but I am checking in to see if they'd like to add `Veteran` and/or `CalFresh`. Will update once I hear back. 


Also, they are still set to begin installs the week of 8/25 with a tentative "go live" date of 10/1. Littlepay is transitioning them to a new portal, and once that is complete they'll provide us with group IDs.

indexing:
[...]

PR #3141, opened 2025-09-09
Feat: EDCTA copy

angela-tran:
Doing this as part of #2581

This PR raises the question: what is our source of truth for Benefits copy? When does that get updated?

I used the information from El Dorado's onboarding form, linked in the description of #2581, to fill in the copy the best I could.

Below are comments noting copy that needs to be checked.


benefits/enrollment/context/flow.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, transit_benefit, transportation_type):
     SystemName.AGENCY_CARD.value: AgencyCardEnrollmentSuccess(
         transit_benefit=_("a CST Agency Card transit benefit"), transportation_type=_("a CST bus")
     ),
+    AgencySlug.EDCTA: DefaultEnrollmentSuccess(transportation_type=_("EDCTA local fixed route services")),
```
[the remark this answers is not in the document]
angela-tran: Changed it to `an EDCTA bus` in https://github.com/cal-itp/benefits/pull/3141/commits/a483c7ecafd75c12c3db633db80dd85ddeba32ff

thekaveman (approved):
I think in the absence of "official" copy in Ditto / Figma, this looks like a great start. We can always adjust if needed, but getting the slug in there is a critical first step 👍

benefits/enrollment/context/flow.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, transit_benefit, transportation_type):
     SystemName.AGENCY_CARD.value: AgencyCardEnrollmentSuccess(
         transit_benefit=_("a CST Agency Card transit benefit"), transportation_type=_("a CST bus")
     ),
+    AgencySlug.EDCTA: DefaultEnrollmentSuccess(transportation_type=_("an EDCTA bus")),
```
Scotchester, line 96: Should this have had a `.value` like the other `AgencySlug`based keys in this dict?

angela-tran: That's a good catch! Yes, I think it should've had `.value` there, i.e. it was not intentional to omit it.

Can you set your transit agency to have a slug of `edcta` and see if the Enrollment Success page raises an error due to this? Let me know if you need any help with setting that up.

angela-tran: I'm curious to know whether we actually need `.value` or not. 

While looking into your question, I noticed that none of the keys in the [`enrollment_index` dict](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/context/flow.py#L45-L49) have `.value`, which is also not what I'd expect.

Scotchester: @angela-tran You're right! It works either way.

```
>>> enrollment_success['edcta']
DefaultEnrollmentSuccess(success_message='You were not charged anything today. When boarding an EDCTA bus, tap your contactless card and you will be charged a reduced fare. You will need to re-enroll if you choose to change the card you use to pay for transit service.', thank_you_message='Thank you for using Cal-ITP Benefits!')
```

Sticking with the `.value` might be slightly nicer when debugging, though. If you inspect `enrollment_success` as it is now, it shows the key as `AgencySlug.EDCTA` when the `.value` isn't used, but I think it's nicer to see the actual string slugs. What do you think?

Snippet of VS Code Debug Console showing the items of the enrollment_success dictionary

angela-tran: Yeah, I agree the keys should all be of the same type at least. I'll open a PR to add `.value` here. Thanks again for catching and for testing things out!

PR #3214, opened 2025-10-02
feat: logo strip for landing page

### LINKED ISSUE #3126: Implement logo strip for active agencies

## Acceptance Criteria

- [ ]

jgravois:
resolves #3126

* adds a logo-strip to display _existing_ transit logos on homepage
* omit logo-strip from agency specific homepage
* swap out white border box on mobile and replace with opaque blue overlay
* translated text still TK

desktop:
Screenshot 2025-10-09 at 3 05 12 PM

mobile:
Screenshot 2025-10-09 at 3 05 33 PM

i'm proposing that we swap in the new logos as soon as we land #3217 (deprecating the current modal-based picker)

### DIFF
--- a/benefits/core/templates/core/includes/form.html
+++ b/benefits/core/templates/core/includes/form.html
@@ -18,7 +18,9 @@
 
           {{ field }}
 
-          {% if field.help_text %}<small class="d-block mt-2 pt-1 form-text text-body">{{ field.help_text }}</small>{% endif %}
+          {% if field.help_text %}
+            <small class="d-block mt-2 pt-1 small-text form-text text-body">{{ field.help_text }}</small>
+          {% endif %}
         </div>
       {% endfor %}
     </div>
--- a/benefits/core/templates/core/includes/recaptcha-text.html
+++ b/benefits/core/templates/core/includes/recaptcha-text.html
@@ -1,6 +1,6 @@
 {% load i18n %}
 
-<p class="form-text text-body mt-0">
+<p class="small-text form-text text-body mt-0">
   {% blocktranslate trimmed %}
     This site is protected by reCAPTCHA and the Google <a href="https://policies.google.com/privacy" target="_blank" rel="noopener noreferrer">Privacy Policy</a> and <a href="https://policies.google.com/terms" target="_blank" rel="noopener noreferrer">Terms of Service</a> apply.
   {% endblocktranslate %}
--- a/benefits/core/templates/core/index-base.html
+++ b/benefits/core/templates/core/index-base.html
@@ -18,10 +18,10 @@
 
 {% block main-content %}
   <div class="container">
-    <div class="row align-items-end align-items-lg-center">
+    <div class="row align-items-center">
       <div class="col-lg-6">
         <div class="box py-3 px-3 py-lg-0 px-lg-0 mt-lg-5 pt-lg-4">
-          <h1 class="eyebrow-text mb-3">{% translate "Cal-ITP connects your transit benefit to your contactless card" %}</h1>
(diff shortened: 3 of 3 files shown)


benefits/static/css/styles.css
```diff
@@ -121,11 +121,12 @@ a:visited:not(.btn) {
 a[target="_blank"]::after {
   content: " ";
   background-color: currentColor;
-
+  /* prettier-ignore */
   mask-image:
     url("/static/img/external-link.svg"),
     linear-gradient(transparent, transparent);
   mask-position: center center;
+  /* prettier-ignore */
```
jgravois, line 129: i'm not sure exactly what's going on but i'd guess the jockeying between prettier's auto-formatting in my containerized dev environment and what is applied in https://github.com/cal-itp/benefits/pull/3214/commits/4608a4b261fc8dedb564d4c674157084833ac109 is a side-effect of a version mismatch in the latest/greatest VS Code extension and the (deprecated) pre-commit mirror.

https://github.com/prettier/prettier/discussions/16227

happy to explore a solution that's more than a band-aid in a followup PR, but i didn't want to get sidetracked here.

Scotchester:
Do we have an established way to track things that have to be done ASAP after deployment, like needing to update the agency's logo images as soon as possible after this and #3217 are deployed?

And now that I think about it, also tracking the fact that those two PRs need to go together so that we don't need a third image format? (Unlikely they wouldn't, I suppose.)

thekaveman:
Just some general feedback as I'm starting to review:

- This feels like a lot of different changes for a single commit: 6bc9533ae73fb688016c116dbc7718acf49ac850. E.g. I wondered why the `form` include was updated, and it took some close looking to see that a CSS rule was also updated that affected both. 
- Separate but related, that commit also changed the docs (fixing a typo). I'm glad you caught the typo! But please, let's have commits focused on a single area / "thing", it's fine to have many different commits!
- I will admit this is a personal pet-peeve of mine, and definitely not a big deal at all, but can we please try to avoid these backward merge commits? There's really no reason for them, you can just `git rebase -i main` to get the latest without creating a merge commit
image

- The commit `chore(pre-commit): autofix run` 126d6b417d211c969ea77c3583522b7ddb4672ce suggests that these commits weren't made in an environment where pre-commit was active? I would expect this to get caught before the commit is made? Or was this like, it worked fine in the local run and then the CI run complained?

jgravois:
okay, i just force pushed a set of commits that i think represent the changes more intelligibly.

thekaveman (changes requested):
Great work so far on this @jgravois!

A few observations from comparing with Figma...

## Desktop

The `text-align: left` could work when we have enough logos, but I think this calls for a more flexbox type approach:

* Items should be aligned left
* Items should have an 80px horizontal gap
* Items should not wrap

Image

## Mobile

Again, I think flexbox is probably going to work better here:

* Items should be justified and aligned center
* Items should have a 24px horizontal gap
* Items should wrap
* Items should have a 24px vertical gap

Image

Personally, I'm terrible at flexbox and always forget how it works. So I use this page to refresh myself often: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

thekaveman (changes requested):
I'm an idiot 😢 I failed to actually _read_ the HTML closely, my bad. You obviously saw all that in Figma... I guess the left-align on Desktop is really the only thing?

jgravois:
> I'm an idiot 😢 I failed to actually read the HTML closely, my bad. You obviously saw all that in Figma...

nah, maybe tomorrow you can show me how you exposed the discrete gaps in Figma. i was just eyeballing it 😆 

> I guess the left-align and increasing the gap up to 80px on Desktop are really the only (small) things?

addressed in 2812174dccd39f22c288119e3856e62452b9f09f

Issue #3125, opened 2025-08-28
Implement IndexView as FormView with agency select


thekaveman:
Current view: https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L25

Figma: https://www.figma.com/design/SeSd3LaLd6WkbEYhmtKpO3/Benefits--Full-Application-?node-id=9754-52797

## Acceptance Criteria

- [ ] Using HTML / browser-native `<select>` element
- [ ] Reuse existing focus-ring style e.g. on buttons, links for the select
- [ ] Agency selection is required to submit the form, browser native validation message indicates that a selection is required
- [ ] Active agencies are listed in alphabetical order in the select
- [ ] Selecting an agency updates the homepage copy to reflect that agency (in-place without a redirect/page change)
- [ ] Clicking Next takes you to the Eligibility Start view

PR #3217, opened 2025-10-06
feat: implement landing page transit agency picker as a <select/>

jgravois:
🚧 work in progress 🚧 . will (eventually) resolve https://github.com/cal-itp/benefits/issues/3125


tests/pytest/core/test_forms.py
```diff
@@ -0,0 +1,35 @@
...
+    form = ChooseAgencyForm({"select_transit_agency": model_TransitAgency.slug})
+    assert form.has_changed()
+    assert form.is_valid()
+
+
+@pytest.mark.django_db
+def test_ChooseAgencyForm_invalid():
+    form = ChooseAgencyForm({"select_transit_agency": "invalid"})
+    with pytest.raises(Exception, match=r"object has no attribute 'cleaned_data'"):
+        form.clean()
```
thekaveman: This is technically accurate, and testing what happens -- an exception is raised here on the call to `form.clean()`

Maybe the test (and code) could be a little more explicit about what is going on though, and graceful in the handling?

E.g. when I debug this test, in `form.clean()` I see:

Image

So `super().clean()` returned `{}` and the lookup in `cleaned_data["select_transit_agency"]` is raising an exception because the dict is empty and the `[]` lookup expects the key to exist.

Since we only want to set `self.selected_transit_agency` when the user _actually selected a valid agency_, we should represent this in the code with some more checks. E.g. use a safer method to access the dict:

```python
cleaned_data = super().clean()
# .get() returns None (or another default) if the key isn't found
agency_slug = cleaned_data.get("select_transit_agency")

if agency_slug:
    self.selected_transit_agency = models.TransitAgency.by_slug(agency_slug)
```

And then the test could be adapted for this case: `form.selected_transit_agency` should be `None`.

_Typing out the above, you may have to set a default in `__init__`: `self.selected_transit_agency = None`_

benefits/core/templates/core/includes/form--select-agency.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+
+<form class="{{ form.classes }}" id="{{ form.id }}" method="{{ form.method | default:"post" | upper }}">
+  {{ form.as_p }}
```
Scotchester: Doesn't feel right to wrap this field in a `<p>`. Removing this will revert it to just being a `<div>`:


```suggestion
  {{ form }}
```

This does require a little additional styling, which I'll suggest in a later comment.

benefits/static/css/styles.css
```diff
@@ -574,6 +575,21 @@ footer .footer-links li a.footer-link:visited {
   gap: 1.5rem;
 }
 
+/* Forms: Select Inputs */
+/* Styles specific to the agency chooser on the landing page */
+
```
Scotchester: Compensating for not being wrapped in a `<p>` anymore, and adding a bit of line-height:


```suggestion

#form-agency-selection label {
  font-size: var(--font-size-20px);
  line-height: var(--bs-body-line-height);
  letter-spacing: calc(var(--font-size-20px) * var(--letter-spacing-2));
}

```

benefits/static/css/styles.css
```diff
@@ -574,6 +575,21 @@ footer .footer-links li a.footer-link:visited {
...
+/* override a few bootstrap defaults */
+#form-agency-selection .form-select {
+  border-radius: var(--border-radius);
+  padding: 11px;
```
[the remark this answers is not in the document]
jgravois: > I'm not seeing existing variables for general spacing increments

since we can't extend bootstrap's [`$spacers`](https://getbootstrap.com/docs/5.3/utilities/spacing/#sass-maps) with Sass, something like this 👇 might be a good start....

```css
:root {
  --grid-spacer: calc(1rem / 2); /* 8px */
}

#form-agency-selection .form-select {
  padding: calc(var(--grid-spacer) * 1.5);
}
```

tests/pytest/core/test_forms.py
jgravois: thanks for pointing this out. i learned a bit more about the internals of the form (and the test) and i _think_ i implemented your suggestion appropriately in 5230a23.

benefits/core/templates/core/index-base.html
```diff
@@ -18,15 +18,17 @@
...
-        <div class="box py-3 px-3 py-lg-0 px-lg-0 mt-lg-5 pt-lg-4">
+        <div class="box py-3 px-3 py-lg-0 px-lg-0 pt-lg-4">
           <h1 class="small-text mb-3">{% translate "Cal-ITP connects your transit benefit to your contactless card" %}</h1>
           <h2 class="h1 text-start p-0 mb-4">
             {% block headline %}
             {% endblock headline %}
           </h2>
           <span role="img" aria-label="{% translate "A person holds a contactless debit card next to a card reader on a bus." %}"></span>
+          {% block select-agency %}
+          {% endblock select-agency %}
```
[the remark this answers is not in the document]
jgravois: i like this suggestion.

tbh, when i first did that a few weeks ago i didn't understand the inheritance chain of the two sub-templates yet.

tests/pytest/core/test_views.py
```diff
@@ -37,6 +37,13 @@ def test_get(self, view, app_request, mocked_session_reset):
         assert response.status_code == 200
         mocked_session_reset.assert_called_once()
 
+    def test_form_valid(self, view, model_TransitAgency):
+        form = view.form_class(data=dict(select_transit_agency=model_TransitAgency.slug))
+
+        assert form.is_valid()
+        view.form_valid(form)
+        assert view.success_url
```
thekaveman: We have the `model_TransitAgency` fixture in this test already, may as well assert that this `success_url` equals the expected value (`eligibility_index_url`).

thekaveman (changes requested):
One last small update to the test.

tests/pytest/core/test_views.py
jgravois: good call. i implemented this suggestion and cleaned up the commit history.

Scotchester (approved):

PR #3222, opened 2025-10-07
chore: Update fixtures and sample .env for local Switchio testing

Scotchester:
- Modifies Littlepay/Switchio secret name values in the local fixtures to standardize on our desired conventions
- Adds placeholders for all needed Switchio env vars to `.env.sample`
  - Removes vars named `prod` because I don't think we'll ever be testing those locally, but correct me if I'm wrong.


thekaveman (approved):
👍

Issue #3101, opened 2025-08-19
El Dorado: Littlepay configuration complete; ready for configuration in Cal-ITP Benefits.


indexing:
- [x] Group(s) are created in Littlepay QA for each enrollment flow
- [x] Product(s) are linked to Group(s) in Littlepay QA for each enrollment flow
- [x] API access granted to Littlepay QA

thekaveman:
We received API access for El Dorado for Littlepay's QA and Production environments. We'll continue with the next steps for this configuration.

thekaveman:
Confirmed group and product configuration in QA using the [littlepay CLI](https://github.com/cal-itp/benefits/issues/3103)

```shell
$ littlepay groups products
👥 Matching groups (2): qa, edcta
GroupResponse(id='b9b0a617-66d7-48b9-aa1a-301952055443', label='Discounted', participant_id='qa-eldorado-transit')
  🛒 Linked products (4)
  ProductResponse(id='50cb829e-a144-4099-8694-5316ab8ae8d6', code='Daily Max - Discounted', status='ACTIVE', type='CAPPING', description='Daily Max - Discounted', participant_id='qa-eldorado-transit')
  ProductResponse(id='71990c94-4a84-4f5c-a980-13dcf51a9ab9', code='Discount - Local - One-Way', status='ACTIVE', type='DISCOUNT', description='Discount - Local - One-Way', participant_id='qa-eldorado-transit')
  ProductResponse(id='9e38f25b-ac90-4aef-a6f3-dde438e93c45', code='Paratransit', status='ACTIVE', type='DISCOUNT', description='Paratransit', participant_id='qa-eldorado-transit')
  ProductResponse(id='fe5683c6-6cc8-42ca-b336-27e7d042dc78', code='Monthly Pass - July - Discounted', status='INACTIVE', type='CAPPING', description='Monthly Pass - July - Discounted', participant_id='qa-eldorado-transit')
GroupResponse(id='ff30abe9-f52f-4d99-b355-d2126e472212', label='General Public', participant_id='qa-eldorado-transit')
  🛒 Linked products (2)
  ProductResponse(id='63c35bea-2eea-4076-9a89-83de52e79ce9', code='General - Commuter - One-Way', status='ACTIVE', type='DISCOUNT', description='General - Commuter - One-Way', participant_id='qa-eldorado-transit')
[...]

PR #3233, opened 2025-10-08
docs: Overhaul guide to adding a new agency

Scotchester:
This takes the existing "Configuring a new transit agency" page and makes some major updates to it, fixing the outdated parts and bringing in additional steps and details that were not previously captured.

Opening as a draft for now, as I think I might want to clarify a few things about what can be done in parallel, and also come back to the Mermaid diagram of the process I've been working on, but I think it would be good for anyone who is interested to take a look at what I've done so far.

### DIFF
--- a/docs/getting-started/README.md
+++ b/docs/getting-started/README.md
@@ -28,7 +28,7 @@ DJANGO_LOCAL_PORT=9000
 
     You will need to change `DJANGO_LOCAL_PORT` to a specific value in order to test locally with the CDT Identity Gateway's dev environment. Compiler developers, this value can be found in our shared notes in LastPass.
 
-See [Configuration](../configuration) for more details on supported environment variables and their settings.
+See [Configuration](../configuration/) for more details on supported environment variables and their settings.
 
 ## Run the build script
 
@@ -46,7 +46,7 @@ docker compose build --no-cache client
 
 ## Start the client in a VS Code dev container
 
-From this point forward, the recommended local development setup is to run the app with the [VS Code Dev Containers extension](devcontainers).
+From this point forward, the recommended local development setup is to run the app with the [VS Code Dev Containers extension][devcontainers].
 
 See [Development](../development/) for more details on setting up a dev container and developing with it.
 
@@ -54,7 +54,7 @@ Alternatively, read on to run the app traditionally with Docker Compose.
 
 ## Start the client traditionally
 
-The optional `-d` flag will start in _detatched_ mode and allow you to continue using the terminal session.
+The optional `-d` flag will start in _detached_ mode and allow you to continue using the terminal session.
 
 ```bash
 docker compose up -d client
@@ -78,7 +78,7 @@ docker compose down
 
 ## Minimum configuration needed for your first manual end-to-end test
 
-The following updates must be made to run a full end-to-end test using the sample agenct (CST), the Older Americans flow via Login.gov and the CDT Identity Gateway(IdG) dev environment, and the Littlepay QA environment.
+The following updates must be made to run a full end-to-end test using the sample agency (CST), the Older Americans flow via Login.gov and the CDT Identity Gateway(IdG) dev environment, and the Littlepay QA environment.
 
 !!! note
 
@@ -92,7 +92,7 @@ The following updates must be made to run a full end-to-end test using the sampl
 
 !!! warning
 
-    Be sure to rebuild your devcontainer beefore proceeding.
+    Be sure to rebuild your devcontainer before proceeding.
 
 ### Django Admin updates
 
@@ -118,7 +118,7 @@ For details on creating an identity-proofed account for testing in the Login.gov
 
 ---
 
-You should now be ready to perform a complete end-to-end test in your local environment! When arriving at the Littlepay form, use any of the acceptable forms of [test data for Visa or MasterCard accounts](test-cards).
+You should now be ready to perform a complete end-to-end test in your local environment! When arriving at the Littlepay form, use any of the acceptable forms of [test data for Visa or MasterCard accounts][test-cards].
 
 ## Managing test data going forward
 
@@ -141,7 +141,4 @@ Or, if you made a temporary change to one of the objects created by the fixtures
 
 [docker]: https://www.docker.com/products/docker-desktop
 [devcontainers]: https://code.visualstudio.com/docs/devcontainers/containers
-[data-migration]: https://github.com/cal-itp/benefits/tree/main/benefits/core/migrations
-[logingov-internal]: https://docs.google.com/document/d/1bFynuiLy9POXYEYnLVEQJpAm1TVqlappN9lU9ev3Bvg/edit?tab=t.0#heading=h.bvwe1dyv15c5
-[logingov-external]: https://developers.login.gov/testing/#testing-identity-proofing
 [test-cards]: https://docs.stripe.com/testing?testing-method=card-numbers#cards


Scotchester:
Rebased to bundle all of the significant changes (with feedback addressed) into [one commit](https://github.com/cal-itp/benefits/pull/3233/commits/efa2ae0b272f83707a246d8586f573933f2949f4) that shows them as modifications to the original file's content (as opposed to the total changeset showing an add/delete).

docs/configuration/transit-provider.md
```diff
@@ -0,0 +1,203 @@
...
+   - Choose one of the new `EnrollmentFlow`s.
+   - Set the group value to the corresponding **testing** group ID (from production Littlepay) for production validation.
+     - This will be set back to the QA group value after final production configuration is complete.
+1. Cal-ITP creates a new `LittlepayConfig` in the Benefits test environment:
+   - Set Environment to **Production** for production validation.
+     - This will be set back to QA after final production configuration is complete.
+   - Choose the new `TransitAgency`.
+   - Retrieve Audience and Client ID values for the **production** config from shared LastPass note.
+   - [Create the client secret in the Azure Key Vault](../../deployment/secrets/) for the test environment, then paste its name in the Client Secret Name field.
+     - Be sure to refresh the secrets for this to take effect! _**ADD MORE DETAILED INSTRUCTIONS**_
```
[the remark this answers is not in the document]
lalver1: No problem! I agree, the text instructions are pretty clear! 👍

thekaveman (approved):
Awesome work Scott! And great review and comments everyone.

docs/configuration/transit-provider.md
```diff
@@ -0,0 +1,209 @@
...
+_Typically performed by an engineer._
+
+Certain parts of the codebase must be updated to support a new transit provider:
+
+- [`benefits/core/context/agency.py`](https://github.com/cal-itp/benefits/blob/main/benefits/core/context/agency.py) (requires a [migration](../../development/models-migrations/))
+- [`benefits/eligibility/views.py`](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py)
+- [`benefits/enrollment/context/flow.py`](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/context/flow.py)
+- [Message files](../../development/i18n/)
+
+[Here is an example PR that makes these updates for a new transit provider.](https://github.com/cal-itp/benefits/pull/3141)
```
jgravois: given that #3141 is missing a translation, you're also free to link to https://github.com/cal-itp/benefits/pull/3246 now 😇

docs/configuration/transit-provider.md
```diff
@@ -1,49 +1,118 @@
...
+
+### Initial Littlepay configuration
+
 For development and testing, only a Littlepay customer group is needed since there is no need to interact with any discount product. (We don't have a way to tap a card against the QA system to trigger a discount and therefore have no reason to associate the group with any product.)
 
-### Steps
+This work can begin once the transit provider has a contract in place with Littlepay.
+
+- Cal-ITP uses the transit provider's Littlepay merchant ID to create a customer group in the Littlepay QA environment for each type of eligibility (e.g. senior).
+  - _Typically performed by transit provider's Account Manager_
```
angela-tran: Calling out this step as one that is in a limbo state.

This step as written doesn't really make sense because it essentially contradicts itself and reflects an SOP that seems to have been (intentionally? unintentionally?) abandoned.

It contradicts itself because "using a merchant ID to create customer groups" is something an engineer would do using the (old) Littlepay API. But then we say that this step is "_Typically performed by transit provider's Account Manager_", which doesn't make sense because they're not expected to have to use any APIs. (Also, minor aside, if an engineer were doing this, they would not be using a merchant ID which was part of the Littlepay Customer API; we've been using the newer Backoffice API since [early 2024](https://github.com/cal-itp/benefits/pull/1905).)

At this point, you might think, "ok, well maybe this step is really just saying _someone_ should use _some way_ to create a customer group in the Littlepay QA environment for the transit provider."  Two thoughts on that:

1. Is that actually something someone should do at this point? This is what I mean by an SOP that seems to have been abandoned. It seems these days that when we add new agencies to test-benefits, we go straight into production validation, as opposed to the much, much older SOP where the agency would spend some pointed at their groups in Littlepay QA. So it's not actually at this point that we absolutely need the groups to exist in the QA environment, right? 

For example, El Dorado is in production-validation mode right now. Someone did create [groups for them in the QA environment](https://github.com/cal-itp/benefits/issues/3101#issuecomment-3304701865). But those groups haven't been used anywhere yet.

2. It is true that the account manager can create customer groups. They usually work with a staff member from the transit agency to create customer groups using Littlepay's Control Portal... for _production_. Does the staff member have access to a Control Portal for QA? I have no idea.

docs/configuration/transit-provider.md
```diff
@@ -1,49 +1,118 @@
...
 ## Configuration for production validation
 
-For production validation, both a customer group and discount product are needed. The customer group used here is a temporary one for testing only. Production validation is done against the Benefits **test environment** to avoid disruption of the production environment.
+**Production validation** is the process of doing a end-to-end test of enrolling a real person's card through the Benefits app and using it to ride with a discounted fare. The word "production" here refers to **Littlepay's production environment** (which must be used to take a ride in real life), but the Benefits application's test environment is used for the enrollment process to avoid disruption of the Benefits production environment.
 
-### Steps
+For production validation, both a customer group and discount product are needed. The customer group used here is a temporary one for testing only.
 
 1. Transit provider staff creates the discount product in production Littlepay (if it does not already exist).
 1. Transit provider staff takes a screenshot of the discount product in the Merchant Portal, making sure the browser URL is visible, and sends that to Cal-ITP.
```
angela-tran: This is outdated in a number of ways.

- This was written back when the transit provider staff used something called a "Merchant Portal" to create the products, and the browser URL was the only place that the product ID was exposed. I'm not sure why we decided to ask them to take a screenshot rather than just copy-pasting the URL for us :joy: but nonetheless, these days, they're using something called the Control Portal, which exposes the IDs more clearly.

- When this was written, Cal-ITP engineers needed the product IDs to link the products to the groups. These days, account managers are working with agency staff directly to link products and groups, so engineers are no longer involved in this step nor do we need the product IDs.

docs/configuration/transit-provider.md
```diff
@@ -52,22 +121,89 @@ At this point, Cal-ITP and transit provider staff can coordinate to do on-the-gr
...
+
+### Remove test group in production Littlepay
+
+The customer group that was created in production Littlepay for testing purposes can be deleted.
 
-### Cleanup
+### Update transit provider configuration in test environment
 
-At this point, the customer group that was created in production Littlepay for testing purposes can be deleted. The temporary production validation objects in the Benefits database can also be deleted.
+The transit provider's configuration in the test environment should be updated to change the production values back to the QA values for its steady state going forward.
```
angela-tran: >... configuration in the test environment should be updated to change the production values **back** to the QA values ...

Just want to highlight this to re-iterate my earlier point that our SOP seems to have changed such that QA values are never actually used / needed until this point. So here we're not changing it back to QA, but rather finally needing QA values for the first time!

docs/configuration/inc/add-enrollmentflow.md
```diff
@@ -0,0 +1,9 @@
+1. Cal-ITP creates new `EnrollmentFlow`s in the Admin:
+   - For each type of discount the new transit provider will be using, add an enrollment flow with the following:
+     - System name
+     - Label (retype the system name)
```
angela-tran: It's not entirely correct to say "retype the system name", right? Like for a system name of `medicare`, we would want a label of `Medicare Cardholder`. The label is used as the radio button label in in-person enrollment, so it's a user-facing string.

I don't need to block this PR for this, but just wanted to check our understanding.

docs/configuration/transit-provider.md
angela-tran: Caveat: engineers would need the product IDs if we were to go with the approach of engineers managing test groups in production for production validation, as mentioned in https://github.com/cal-itp/benefits/pull/3233#discussion_r2457111176

docs/configuration/transit-provider.md
```diff
@@ -1,49 +1,118 @@
...
+_Typically performed by an engineer._
+
+Certain parts of the codebase must be updated to support a new transit provider:
+
+- [`benefits/core/context/agency.py`](https://github.com/cal-itp/benefits/blob/main/benefits/core/context/agency.py) (requires a [migration](../../development/models-migrations/))
+- [`benefits/eligibility/views.py`](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py)
+- [`benefits/enrollment/context/flow.py`](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/context/flow.py)
+- [Message files](../../development/i18n/)
+
+[Here is an example PR that makes these updates for a new transit provider.](https://github.com/cal-itp/benefits/pull/3246)
```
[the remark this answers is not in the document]
angela-tran: Another way to solve this is if all CA transit agencies really did discount transit fares by 50% on all transportation types. But until then...

docs/configuration/inc/add-enrollmentflow.md
Scotchester: Yeah, John and I noticed this a couple days ago when trying out this doc with RABA. I'll put off updating it at this moment so it doesn't reset the PR approval, but let's come back to it with the other updates you're raising.

docs/configuration/transit-provider.md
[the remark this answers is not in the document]
Scotchester: In addition to nodding along with everything you've said here, I have some questions/comments about the form itself that I think would be good to discuss with the Product and Customer Success teams. It seems to be set up for more than just onboarding a transit provider to Benefits, which is its stated goal.

Issue #3127, opened 2025-08-28
Update logos for active agencies


thekaveman:
Export from Figma: https://www.figma.com/design/SeSd3LaLd6WkbEYhmtKpO3/Benefits--Full-Application-?node-id=20426-7647&t=0mlkwfikxTG43q7S-4

## Acceptance Criteria

- [x] CST is configured with the new logo in `dev`
- [x] CST + active agencies are configured with new logos in `test`
- [x] Active agencies are configured with the new logos in `prod`

cmajel:
@jgravois yes I can support that. I've blocked off some time to look into this today and may schedule a quick pair.

cmajel:
I agree that these aren't high priority fixes, but more nice to haves. 

I went back to the source logos and attached a realigned MST and a higher resolution 3x VCTC. This will help with 1 and some with 3 (though there will be some detail loss at this size)

I see a few underlying issues with this logo pattern:
-  We get a variety of logos of varying detail, and they're not readable at this size (not just MST, but also Nevada County Connects and others). 
- We can probably only add 1-2 more logos here before we'll outgrow this pattern on desktop.

So, long way to say if we have bandwidth let's fix those two logos, but I'd recommend revisiting how we show logos for better readability and partner growth in the future. cc @indexing for thoughts.

[logo-updates.zip](https://github.com/user-attachments/files/23396879/logo-updates.zip)

PR #3237, opened 2025-10-10
chore: extract shareable form handling code

jgravois:
## What this PR does

this is a sliver of the work from #3127 that i thought might be helpful to review and make a decision on in isolation.

our new `<select/>` handler is only going to make use of the custom form validation, but it seemed sensible enough to leave the existing callback intact as the recaptcha logic is very much opt-in.

## Testing the PR

no behavior changes should be encountered. clicking <kbd>Choose this benefit</kbd> in [/eligibility](https://test-benefits.calitp.org/eligibility) without selecting a radio should continue to display a custom error message.
Screenshot 2025-10-21 at 10 27 32 AM

i wasn't able to trigger the [recaptcha](https://docs.calitp.org/benefits/configuration/recaptcha/) by spoofing the "User agent string" as suggested [here](https://stackoverflow.com/questions/48224799/test-invisible-recaptcha). if y'all have a technique for accomplishing that, i'd be interested in testing that too. 😇


thekaveman (commented):
@jgravois can you please say a little more about the motivation for this refactor?

It seems straightforward enough (and we've done this plenty of times): extract some code into an include, then re-include it where it was originally.

I'm just not following the _why_ for what this include gets us over `form.html`? In my understanding, all forms that we build on this app should use `form.html` already?

thekaveman:
> my thinking was that it would be most expedient to create another -form.html variant

I gotcha I gotcha, thanks for pointing me to that other PR. I definitely see what you mean and this does seem straightforward for the homepage update while retaining the common logic in one place. Looks like there's some other stuff going on in the `form.html` around label generation with the required indicator that we probably wouldn't want here either, so this makes sense to me :+1: 

It does make me think about maintainability and potential digressions. Small example: the original `form.html` sets the `<form action="...">` attribute but that is missing in `select-agency-form.html` -- was that intentional? Or an omission?

I don't need to quibble too much about specific ways of implementing, because at the end of the day it just has to work. I will suggest/point out some of our naming patterns (at least that we've tried to move towards lately) for templates though:

* When we have a single "component" we just name it what it is, e.g. `analytics.html`, `logo-strip.html`
* When we have a "component" with variations, we try to name them all with the "component" as a prefix, e.g. `button--index.html`, `button--origin.html`
* The same goes for pages with a template hierarchy, e.g. `index-base.html` (base template), `index.html` (variant for the homepage), `index--agency.html` (variant for the agency homepage)

thekaveman (approved):

PR #3247, opened 2025-10-22
fix: make EDCTA headline translatable

jgravois:
while we were working on onboarding RABA, @Scotchester and i noticed that El Dorado's headline is currently untranslated.

https://test-benefits.calitp.org/edcta
Screenshot 2025-10-22 at 3 55 18 PM


thekaveman (approved):
🔎

Issue #2833, opened 2025-04-08
Update version of Terraform, `azurerm` provider


thekaveman:
From recent pipeline runs:

```log
Terraform v1.8.5
on linux_amd64

Your version of Terraform is out of date! The latest version
is 1.11.3. You can update ...
```

`1.8.5` is pinned in the [`terraform/pipeline/deploy.yml` file](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/deploy.yml#L12) file.

The [`azurerm` provider](https://github.com/cal-itp/benefits/blob/main/terraform/main.tf#L7) is also using a fairly outdated reference:

```hcl
required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = "~> 3.0"
    }
  }
```

Resulting in:

```log
Initializing provider plugins...
- Reusing previous version of hashicorp/azurerm from the dependency lock file
- Installing hashicorp/azurerm v3.117.0...
- Installed hashicorp/azurerm v3.117.0 (signed by HashiCorp)
```

[`hashicorp/azurerm v3.117.0`](https://registry.terraform.io/providers/hashicorp/azurerm/3.117.0/docs) is from... "five months ago" according to their docs.

The [latest version is `v4.26.0`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs).

## Acceptance criteria

- [ ] The version of `terraform` used in the Azure DevOps pipeline is updated to the latest (currently `1.11.3`)
- [ ] The version of the `azurerm` provider is updated to `"~> 4.0"`
- [ ] Minimal (or no changes) and no errors in e.g. `plan` output

## Additional context

- Corresponding issue in DDRC: Office-of-Digital-Services/cdt-ods-disaster-recovery#47
- Its associated PR Office-of-Digital-Services/cdt-ods-disaster-recovery#50

jgravois:
i don't have permission to assign myself, but i'm interested in taking a look at this one.

thekaveman:
Just want to make sure you see this too, I don't know if they are totally related but maybe can be knocked out together/similarly? #2520

Issue #2520, opened 2024-11-12
Address Terraform warnings in Azure DevOps Pipeline


lalver1:
During the past two releases ([2024.11.1](https://github.com/cal-itp/benefits/releases/tag/2024.11.1) and [2024.11.2](https://github.com/cal-itp/benefits/releases/tag/2024.11.2)) we noticed two warnings related to the Terraform tasks in the infra pipeline:

- `Downloading task: TerraformInstaller (0.203.0)
##[warning]Task 'Terraform tool installer' version 0 (TerraformInstaller@0) is dependent on a Node version (10) that is end-of-life. Contact the extension owner for an updated version of the task. Task maintainers should review Node upgrade guidance: https://aka.ms/node-runner-guidance`
- `Downloading task: TerraformTaskV3 (3.209.33)
##[warning]Task 'Terraform' version 3 (TerraformTaskV3@3) is dependent on a Node version (10) that is end-of-life. Contact the extension owner for an updated version of the task. Task maintainers should review Node upgrade guidance: https://aka.ms/node-runner-guidance`

## Acceptance Criteria

- [ ] The warnings are resolved.

## Additional context

Initial suggestions are to use `TerraformInstaller@1` and `TerraformTaskV4`. We can [check on the `task.json` file of the extension to see which version of Node it uses](https://learn.microsoft.com/en-us/azure/devops/extend/develop/add-build-task?view=azure-devops#q-how-can-i-upgrade-my-custom-task-to-the-latest-node).

Issues related to these warnings:

- https://github.com/microsoft/azure-pipelines-terraform/issues/257
- https://github.com/microsoft/azure-pipelines-terraform/issues/253

jgravois:
DevSecOps is looking into adding federated credentials to our existing service connections. if that doesn't unblock the upgrade path all on its own we can come back to this ticket at our leisure.

jgravois:
related https://github.com/cal-itp/benefits/issues/2832

PR #3250, opened 2025-10-24
chore: upgrade terraform

jgravois:
resolves https://github.com/cal-itp/benefits/issues/2833 (and half of #2520)

locally i'm seeing the same plan results as are appearing in the azure pipeline. i don't think there's anything amiss, but its possible i'm misunderstanding something about the subscription architecture.

pipeline run: https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=110749&view=logs&j=5af648c4-4294-55a7-6072-dd70820d3dba&t=42899ae2-6397-58a2-7641-cec3214b180b

prior art: https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/pull/50

ref: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/4.0-upgrade-guide


terraform/app_service.tf
```diff
@@ -45,7 +49,6 @@ resource "azurerm_linux_web_app" "main" {
   }
 
   app_settings = {
-    "DOCKER_REGISTRY_SERVER_URL"          = "https://ghcr.io/"
```
jgravois: initially i assumed that this (now invalid declaration) would need to be replaced with the block below.

```tf
site_config {
  application_stack {
    docker_registry_url = "https://ghcr.io/"
  }
}
```

but that's error inducing (unless you also specify `docker_image_name`_. to verify that the appropriate default is preserved, i ran a plan locally with this:

```tf
site_config {
  application_stack {
    python_version = "3.7"
  }
}
```
the plan output made me more confident that "https://ghcr.io" is still being set:

```tf
~ application_stack {
    - docker_image_name        = "cal-itp/benefits:c6af905c75df45d6d81a8f20fc3dc3bb2f472aae" -> null
    - docker_registry_url      = "https://ghcr.io/" -> null
    + python_version           = "3.7"
    # (10 unchanged attributes hidden)
}
```

terraform/environment.tf
```diff
@@ -4,7 +4,6 @@ locals {
   is_dev              = !(local.is_prod || local.is_test)
   env_name            = local.is_prod ? "prod" : terraform.workspace
   env_letter          = upper(substr(local.env_name, 0, 1))
-  subscription_letter = local.is_prod ? "P" : "D"
```
jgravois: this is the same unused variable that was yanked in https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/pull/50

terraform/pipeline/deploy.yml
```diff
@@ -5,11 +5,11 @@ parameters:
     type: string
 
 steps:
-  # https://github.com/microsoft/azure-pipelines-terraform/tree/main/Tasks/TerraformInstaller#readme
-  - task: TerraformInstaller@0
+  # https://github.com/microsoft/azure-pipelines-terraform/blob/main/Tasks/TerraformInstaller/TerraformInstallerV1/README.md
+  - task: TerraformInstaller@1
```
jgravois: this one was a clean swap. the TerraformTask upgrade is going to require a little bit of spelunking.

terraform/storage.tf
```diff
@@ -1,10 +1,11 @@
...
-  account_tier                    = "Standard"
-  account_replication_type        = "RAGRS"
-  allow_nested_items_to_be_public = false
+  name                             = "sacdtcalitp${lower(local.env_letter)}001"
+  location                         = data.azurerm_resource_group.main.location
+  resource_group_name              = data.azurerm_resource_group.main.name
+  account_tier                     = "Standard"
+  account_replication_type         = "RAGRS"
+  allow_nested_items_to_be_public  = false
+  cross_tenant_replication_enabled = true
```
thekaveman: Curious why this was set to `true` here? I don't think we need this enabled. In https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/pull/50 we made it `false`.

terraform/app_service.tf
```diff
@@ -23,8 +23,12 @@ resource "azurerm_linux_web_app" "main" {
...
 
   site_config {
-    ftps_state             = "Disabled"
-    vnet_route_all_enabled = true
+    ftps_state                                    = "Disabled"
+    vnet_route_all_enabled                        = true
+
+    # https://github.com/hashicorp/terraform-provider-azurerm/issues/25244
+    ip_restriction_default_action                 = "Allow"
+    scm_ip_restriction_default_action             = "Allow"
```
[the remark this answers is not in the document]
thekaveman: Seems fine. `Allow` is the default under `4.x`: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#ip_restriction_default_action-2

And yeah, this service is meant to be publicly available.

thekaveman (commented):
Looks good 👍 My local `plan` also matches the pipeline run you linked to.

Just one question below (and some other comments).

terraform/storage.tf
jgravois: terraform reports the same change whether the prop is omitted or set to `false` but i'll readily admit i didn't think things through any further than that. 😇 

```tf
~ resource "azurerm_storage_account" "main" {
      ~ cross_tenant_replication_enabled   = true -> false
```

jgravois:
> Let's go for it! 🚀

looks like the [`terraform apply`](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=110774&view=logs&j=5af648c4-4294-55a7-6072-dd70820d3dba&t=39708da4-e251-5130-285a-f2fe36c9a5aa) failed. i'll dig in and figure out the fix tomorrow.

thekaveman:
Any word here @jgravois? Our deployments are blocked for now...

jgravois:
the gist of it is that `storage_account_name` > `storage_account_id` wasn't the clean swap i assumed it would be in https://github.com/cal-itp/benefits/pull/3250#discussion_r2496455224.

the value we need to provide must match the pattern here:
```
/subscriptions/{subscription guid}/resourceGroups/{resource group}/providers/Microsoft.Storage/storageAccounts/{storage_account_name}"
```
but locally, even when i hardcode the appropriate values i'm encountering errors planning. happy to pair on it if its time-sensitive or keep plugging away at it on my own. just let me what makes more sense.

PR #3251, opened 2025-10-24
docs: revisions to agencies still onboarding

indexing:
- Removed the green checkmarks in enrollment pathway columns for all agencies that are not yet live in Cal-ITP Benefits.
- Added an asterisk in place of an en dash in enrollment pathway columns for all agencies that are not yet live in Cal-ITP Benefits  to clarify an en dash (–) denotes the agency doesn't offer that enrollment option.
- changed the target launch date for El Dorado to November


jgravois (approved):
this looks more than reasonable to me

thekaveman (approved):

Issue #2263, opened 2024-08-01
Admin interface: add support for Microsoft SSO


angela-tran:
Microsoft is a common identity provider. Let's add usage of [`django-microsoft-sso`](https://megalus.github.io/django-microsoft-sso/#) to the admin interface.

Microsoft authentication seems to differ from Google in that it has the concept of [tenants](https://learn.microsoft.com/en-us/entra/external-id/tenant-configurations) which narrows down the "universe" of users to the users that are a part of the tenant.

When configuring the library, we give it an [application ID and client secret](https://megalus.github.io/django-microsoft-sso/quick_setup/#the-application-client-secret), both of which are associated with an application that is registered with a tenant.

For this task, we can test using an instance of Microsoft Entra ID using our Compiler Azure subscription. A separate task will cover setting up the production Microsoft Entra ID instance.

## Acceptance Criteria

- [ ] Users in the test Microsoft Entra ID instance can log in to the admin interface

angela-tran:
We've assigned this and #2266 to sprint `2025 - Sprint 11` as items to research / prototype / do as a spike so that we have a better idea of the amount of effort it will require to implement them for real.

thekaveman:
I think so, we can use this epic for when we do the actual config with agencies?

PR #3270, opened 2025-11-05
chore: Eliminate unnecessary CSP directives

### LINKED ISSUE #3257: CSP: Eliminate style-src unsafe-inline directive
Follow up to #2968 

When doing that work, I noted that we were still allowing unsafe inline styles and wondered how easy it would be to get rid of it. It's only needed for some of the styling built into the django-google-sso, but it will require some overriding of those styles to keep things looking as expected.

## Acceptance Criteria

- [ ] `unsafe-inline` directive is removed
- [ ] Google SSO login button looks the same as before

## Additional context

- django-google-sso also has some inline JavaScript for the purpose of updating the post-login destination path in the querystring, but we are not currently allowing `unsafe-line` on `script-src`, so that's already broken for us. [I think [this existing Sentry issue](https://sentry.calitp.org/organizations/sentry/issues/70178/) captures that, but I'm not sure why it triggers on other front-end URLs, too.] We could consider adding our own external .js file to enable that.
- I see a couple other CSP issues in Sentry that are probably worth investigating, as well:
  - https://sentry.calitp.org/organizations/sentry/issues/88635/
  - https://sentry.calitp.org/organizations/sentry/issues/85947/

Scotchester:
script-src: code.jquery.com and style-src: unsafe-inline were both required only by the django-google-sso UI. This PR overrides the package's template and styling to eliminate the need for those.

Closes #3257 

~~To do before finalizing: Review CSP errors in Sentry.~~

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -317,7 +317,7 @@ CONTENT_SECURITY_POLICY = {
             SELF,
             "https://api.amplitude.com/",
             "https://cdn.jsdelivr.net/npm/@switchio",
-            "https://cdn.jsdelivr.net/npm/bootstrap",
+            "https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/",
             "https://cdn.jsdelivr.net/npm/jquery",
         ],
         "default-src": [SELF],
@@ -330,7 +330,7 @@ CONTENT_SECURITY_POLICY = {
             SELF,
             "https://cdn.amplitude.com/libs/",
             "https://cdn.jsdelivr.net/npm/@switchio",
-            "https://cdn.jsdelivr.net/npm/bootstrap",
+            "https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/",
             "https://cdn.jsdelivr.net/npm/jquery",
             "*.littlepay.com",
             NONCE,  # https://django-csp.readthedocs.io/en/latest/nonce.html


jgravois:
alright, `main` is unborked now so the next time you push a commit to this branch i expect you'll see ✅ again. sorry for the inconvenience!

benefits/static/css/admin/styles.css
```diff
@@ -128,6 +128,11 @@ iframe.card-collection {
 
 /* Login Page */
 
+#content-main {
+  /* Overriding Django Admin base styles */
+  float: none;
+}
```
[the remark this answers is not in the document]
Scotchester: You're totally right, Angela. I wondered about this as I was writing it and then forgot to actually log in and see how things looked 🤦 

It _was_ needed because otherwise the buttons at the bottom have to be cleared in order to get proper margin, but I think a later change I made seems to have negated that need. Removed!

benefits/templates/google_sso/login_sso.html
```diff
@@ -0,0 +1,37 @@
...
+<script nonce="{{ request.csp_nonce }}">
+    ready(function() {
+        const queryString = window.location.search;
+        const urlParams = new URLSearchParams(queryString);
+        document.querySelectorAll('.sso-login-btn').forEach((el) => {
+            console.log(el);
+            el.href = el.href + '?next=' + urlParams.get('next');
+        });
+    });
+</script>
```
angela-tran: I was able to verify that on dev-benefits, the `href` of the SSO login button does not get updated with this `?next=/admin/` stuff, which is to say the inline Javascript coming from 

https://github.com/megalus/django-google-sso/blob/518ca8e73f31552f99593e452e7972fde7a8eda0/django_google_sso/templates/google_sso/login_sso.html#L40-L46

```html
<script>
    $(".btn-content").click(function (event) {
        const queryString = window.location.search;
        const urlParams = new URLSearchParams(queryString);
        event.originalEvent.currentTarget.href = event.originalEvent.currentTarget.href + "?next=" + urlParams.get('next');
    });
</script>
```

was blocked from running.

Now that you overrode the libary's template with this one and used `nonce=...` here, I see the `href` gets set as intended. Nice!

Benefits - CSP change admin login setting href

benefits/static/css/admin/styles.css
```diff
@@ -217,43 +217,28 @@ iframe.card-collection {
...
+  transition: all 0.15s; /* Matching standard Django Admin button style seen on main log in button */
 }
 
-.login .google-btn-label {
-  color: var(--bs-primary);
-  width: auto;
-  font-weight: var(--medium-font-weight);
-  font-size: var(--font-size-16px);
-  line-height: 1;
+.sso-login-btn-logo {
```
angela-tran: I've been going back and forth on whether I agree or not with the decision to implement a cleaned up version of 

https://github.com/megalus/django-google-sso/blob/518ca8e73f31552f99593e452e7972fde7a8eda0/django_google_sso/templates/google_sso/login_sso.html#L24-L38

```html
<div class="login-box">
    <div class="login-btn-area">
        {% for provider in sso_providers %}
            <div class="{{ provider.name }}-login-btn">
                <a class="btn-content" href="{{ provider.login_url }}" role="button">
                    <div class="{{ provider.name }}-btn-logo">
                        <img width="20px" alt="{{ provider.name }} sign-in"
                             src="{{ provider.logo_url }}"/>
                    </div>
                    <div class="{{ provider.name }}-btn-label">{{ provider.text }}</div>
                </a>
            </div>
        {% endfor %}
    </div>
</div>
```

and then rename the CSS styles based on that re-implementation (lines [21-24](https://github.com/cal-itp/benefits/pull/3270/files#diff-32d10354312b4f7ac5466f4faf06f37eb59b3e9ddd9f20afa41bb9d1e5191f2bR21-R24) of `benefits/templates/google-sso/login_sso.html`).

The alternative would've been to leave it as-is since what we're mainly wanting to override is the Javascript part.

However, given that we're completely overriding the template, which doesn't extend from a parent or anything, I guess we aren't in any danger of styles breaking from a version update, dependencies updating, etc.

Another pro for this decision is that we'll be adding `django-microsoft-sso` at some point hopefully (#2263), and we'll have to do this overriding again for [its template](https://github.com/megalus/django-microsoft-sso/blob/v6.0.2/django_microsoft_sso/templates/microsoft_sso/login_sso.html), so with these provider-agnostic CSS class names, we'll be able to re-use the template you've written and not have to change any CSS class names.

benefits/settings.py
```diff
@@ -324,12 +324,10 @@ def RUNTIME_ENVIRONMENT():
             "https://cdn.amplitude.com/libs/",
             "https://cdn.jsdelivr.net/",
             "*.littlepay.com",
-            "https://code.jquery.com/jquery-3.6.0.min.js",
```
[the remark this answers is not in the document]
angela-tran: I'll add in my two cents too that this comment probably was referring to how `django-google-sso` loads in jQuery 3.6.0 in its [template](https://github.com/megalus/django-google-sso/blob/v8.0.0/django_google_sso/templates/google_sso/login_sso.html#L39).

benefits/templates/google_sso/login_sso.html
```diff
@@ -0,0 +1,37 @@
...
+    </a>
+  {% endfor %}
+{% endwith %}
+
+<script nonce="{{ request.csp_nonce }}">
+    ready(function() {
+        const queryString = window.location.search;
+        const urlParams = new URLSearchParams(queryString);
+        document.querySelectorAll('.sso-login-btn').forEach((el) => {
+            console.log(el);
```
angela-tran: Did you mean to leave this in?

benefits/templates/google_sso/login_sso.html
```diff
@@ -0,0 +1,37 @@
+{% comment %}
+    This file overrides the django-google-sso pacakge's login_sso.html template
```
jgravois: 😇 
```suggestion
    This file overrides the django-google-sso package's login_sso.html template
```

benefits/templates/google_sso/login_sso.html
Scotchester: No, thanks for the catch!

benefits/templates/google_sso/login_sso.html
```diff
@@ -0,0 +1,37 @@
...
+    You can pass a custom list of SSO providers to this
+    template using the `sso_providers` context variable.
+    If not provided, the template will use the default
+    `define_sso_providers` defined in the sso tags.
+{% endcomment %}
+{% with sso_providers|default:None as sso_providers %}
+  {% if sso_providers is None %}
+    {% define_sso_providers as sso_providers %}
+  {% endif %}
+  {% for provider in sso_providers %}
```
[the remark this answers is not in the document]
angela-tran: Eh, I think it's ok. It just took me a bit to figure out where this came from while reviewing because I had the [8.0.0 version](https://github.com/megalus/django-google-sso/blob/v8.0.0/django_google_sso/templates/google_sso/login_sso.html) of the template pulled up for comparison since it's the [version we're on](https://github.com/cal-itp/benefits/blob/main/pyproject.toml#L20) and didn't see all this logic. I'll leave it up to you, but I don't think it's necessary

benefits/settings.py
```diff
@@ -312,7 +312,7 @@ def RUNTIME_ENVIRONMENT():
 CONTENT_SECURITY_POLICY = {
     "DIRECTIVES": {
         "base-uri": [NONE],
-        "connect-src": [SELF, "https://api.amplitude.com/"],
+        "connect-src": [SELF, "https://api.amplitude.com/", "https://cdn.jsdelivr.net/"],
```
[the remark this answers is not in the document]
jgravois: ah, good sleuthin 🕵️

Scotchester:
Thanks for the reviews! I agree that I could have made the commits a bit more atomic for easier review.

I also should have generally commented more about my thought process in the PR description. I think that's partly still getting used to the process of putting up a draft PR and then later making it ready for review. I'll try to do better :)

To quickly summarize how it went:

- In our codebase, the only inline style is in this Google logo SVG
- But removing that breaks the button styling a bit
- Looking at the existing markup and style to figure out best how to un-break it, I saw that there was also an inline style block and inline script block in the package's template, which we can't add a nonce to
- Therefore we needed to start from a place of overriding the package's template to get rid of those unsafe blocks
- We also had a somewhat large number of styles, overriding a lot of what the package provided
- And wow, the button markup that the package provides is way more complicated than it needs to be
- Therefore I concluded that the best path forward was rewrite the markup and simplify the styles accordingly, leaving in the ability to add buttons for additional SSO providers

I'll update this on Monday for a final review! (Or pause until after the sprint?)

benefits/templates/google_sso/login_sso.html
jgravois: once you land something like https://github.com/cal-itp/calitp.org/issues/513 here i'll be out of a job. 🙃

benefits/templates/google_sso/login_sso.html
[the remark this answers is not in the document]
Scotchester: Ohhh, I see. I missed that we weren't on the latest.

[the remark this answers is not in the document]
angela-tran: Yeah, I think it's fine

benefits/templates/google_sso/login_sso.html
angela-tran: Removed in https://github.com/cal-itp/benefits/pull/3270/commits/4e7664f535e03a9938fc5e698cad73b224511ada

benefits/templates/google_sso/login_sso.html
angela-tran: Fixed in https://github.com/cal-itp/benefits/pull/3270/commits/ec112a172a405f99cd0710e31a0f1e81c1beacb3

angela-tran:
LP + Switchio enrollment still works with updated `connect-src`

image

thekaveman:
I note the `bootstrap.bundle.min.js` request from `jsdelivr.net` failed due to the CSP in your screenshots above, and see the same behavior locally. Is that something to be worried about?

benefits/settings.py
```diff
@@ -323,14 +329,14 @@ def RUNTIME_ENVIRONMENT():
         "script-src": [
             SELF,
             "https://cdn.amplitude.com/libs/",
-            "https://cdn.jsdelivr.net/",
+            "https://cdn.jsdelivr.net/npm/@switchio",
+            "https://cdn.jsdelivr.net/npm/bootstrap",
```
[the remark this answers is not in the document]
angela-tran: Fixed in https://github.com/cal-itp/benefits/pull/3270/commits/9c57c5d9936fff5772d78752875bfff1f8848712

thekaveman (approved):
LGTM!

jgravois (commented):
🔏❤️🙏

PR #3281, opened 2025-11-05
fix: make admin courtesy card confirmation agency agnostic

jgravois:
just something small @Scotchester and i caught out testing the release today. 

Screenshot 2025-11-05 at 10 55 06 AM

i'm under the impression that even though we don't plan to offer additional digital courtesy card enrollments going forward, they will still be able to be configured in-person.


thekaveman:
This is not correct. Courtesy Card is the name of the MST agency card program.

Issue #3272, opened 2025-11-05
Password reset: Wire up PasswordResetConfirmView


thekaveman:
Figma: https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/Cal-ITP-Benefits-Admin?node-id=2513-4573&t=mxopuhqD8NBf99L5-4

This view is responsible for verifying the one-time code from the link the user clicks in their email, and then actually performing the password reset.

We can (maybe?) just use the existing view without subclassing. We need to implement the template ourselves.

## Acceptance Criteria

- [ ] The URLConf has a route for PasswordResetConfirmView
- [ ] When the form has errors, they are shown inline and submission is blocked
- [ ] After successfully resetting their password, the user is redirected to the Login page, which shows a success message
  - Contrary to the standard Django flow, the user is NOT redirected to `PasswordResetCompleteView`.

## Additional context

[`PasswordResetConfirmView` Django docs](https://docs.djangoproject.com/en/5.2/topics/auth/default/#django.contrib.auth.views.PasswordResetConfirmView)

[`PasswordResetConfirmView` on CCBV](https://ccbv.co.uk/projects/Django/5.2/django.contrib.auth.views/PasswordResetConfirmView/)

Scotchester:
Note to self for addressing later:

<img width="524" height="136" alt="Image" src="https://github.com/user-attachments/assets/2c88e7f5-c0b6-46cc-bdb2-a13d1b1b54cf" />

This log in link points to `/accounts/login/`, which is a 404. It should be `/admin/login/`.

Scotchester:
Two questions I've come across:

1. Do we need reCAPTCHA on this view, or do we think that Django's built in confirmation code is enough protection?
2. We don't have designs in Figma for the standard errors on this form. Without changing or adding to what we already have in place, they look like this:
   <img width="468" height="548" alt="Password reset form with two errors shown" src="https://github.com/user-attachments/assets/b6ee655e-43c7-4eb3-8195-72487addb4df" />
   Does that seem acceptable for now?

Issue #3271, opened 2025-11-05
Password reset: Subclass and implement PasswordResetView


thekaveman:
Figma: https://www.figma.com/design/uqb32dSI8sWqV0XDOeD6z2/-2940---passwords?node-id=2514-5634&m=dev

We probably don't need much in the way of _logic_ in the subclass. But in order to provide some overrides for e.g. templates, it could be easier and more testable to do this as a subclass vs. in the route registration.

## Acceptance Criteria

- [ ] When the user clicks the link from #3269, they are directed to a form that collects their username
- [ ] When the form has errors, they are shown inline and submission is blocked
- [ ] When the form is submitted successfully, a confirmation message is shown to the user
- [ ] When the form is submitted successfully, if the user account exists, a one-time link is generated and emailed to the user

## Additional context

[`PasswordResetView` Django docs](https://docs.djangoproject.com/en/5.2/topics/auth/default/#django.contrib.auth.views.PasswordResetView)

[`PasswordResetView` in CCBV](https://ccbv.co.uk/projects/Django/5.2/django.contrib.auth.views/PasswordResetView/)

Issue #2940, opened 2025-05-27
Existing admin users can initiate account password reset


indexing:
Review #3037 for complete context.

## Goals

- Provide transit provider staff with a workflow to reset their account password using the app's functionality, eliminating the need to contact their Cal-ITP account manager.

## User stories

> As a transit agency staff person, I want to reset my administrator account password so I can authenticate and use the app without needing to contact my Cal-ITP Customer Success account manager for support.

## Basic flow

1. User (with an existing account) goes to the Admin login page, and selects "Reset password"
2. User enters their existing account username
3. Benefits sends an email to the associated account with a one-time link
4. User clicks the link and is directed to the reset page, user can reset their password
5. The link expires after it is used once

See https://dev.to/earthcomfy/django-reset-password-3k0l for the inspiration for this flow / implementation ideas.

ohmegasquared:
Question of https://github.com/cal-itp/benefits/issues/2384 - do we want logged in users to be able to change passwords for non SSO accounts?

ohmegasquared:
This is ready to chat with dev about decision points: https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/branch/uqb32dSI8sWqV0XDOeD6z2/Cal-ITP-Benefits-Admin?node-id=3-4&p=f&t=iga3gSpMQvNEWQqg-0

Andy suggested we hold off until the sprint/schedule is calmer.

PR #3285, opened 2025-11-10
feature: Enable password reset with custom request form and confirmation

### LINKED ISSUE #3269: Password reset: Login form presents a "Forgot password?" link
Figma designs: https://www.figma.com/design/uqb32dSI8sWqV0XDOeD6z2/-2940---passwords?node-id=28-1644&m=dev

## Acceptance Criteria

- [ ] Users can initiate self-service password reset from the Admin Login page by clicking the link

Scotchester:
The first parts of epic #2940!

Current changes on this branch would close #3269 and #3271 – the stock Django password reset flow is enabled, which automatically adds the link to the login form given the code we already have in place, the initial reset form is customized to match Figma designs and support reCAPTCHA. The confirmation of the initial form completion is also included here.

### Notes

- One part of the Figma design I have not matched is the styling of the reCAPTCHA disclaimer text. I left it as currently styled on the main login form.
- I'm not sure if this should be merged until we have a proper email sending solution in place? Or perhaps it could be merged before that if we hide the reset link on the login form. Open to suggestions.
- ~~Also open to opinions on whether I should add the template customization for the other three views (reset form completion, new password form, and new password form completion) in this PR or in (an)other(s). The new password form (`PasswordresetConfirmView`) is the only one with its own subtask currently (#3272).~~
  - Looking closer at acceptance criteria, I realized that #3271 implicitly included the reset form completion view (`PasswordResetDoneView`), so that is included here. And in #3272, the acceptance criteria indicates that we don't need the password form completion view (`PasswordResetCompleteView`).
  - As it stands going into final review, this PR implements a completely working password reset flow, but is only customized for the requirements in #3269 and #3271. Will tackle #3272 in a separate PR.
- ~~I probably also want to pull out the reCAPTCHA template code into an include, since I currently just duplicated what the login template has into the new reset template here.~~

### DIFF
--- a/benefits/urls.py
+++ b/benefits/urls.py
@@ -15,7 +15,7 @@ from django.http import HttpResponse
 from django.urls import include, path, re_path
 from django.views.static import serve
 
-from .core.admin.views import BenefitsPasswordResetView
+from .core.admin.views import BenefitsPasswordResetView, BenefitsPasswordResetDoneView
 
 logger = logging.getLogger(__name__)
 
@@ -75,7 +75,7 @@ password_reset_patterns = [
     ),
     path(
         "admin/password_reset/done/",
-        auth_views.PasswordResetDoneView.as_view(extra_context={"site_header": admin.site.site_header}),
+        BenefitsPasswordResetDoneView.as_view(extra_context={"site_header": admin.site.site_header}),
         name="password_reset_done",
     ),
     path(
--- a/tests/pytest/core/admin/test_views.py
+++ b/tests/pytest/core/admin/test_views.py
@@ -1,0 +1,38 @@
+import pytest
+
+from django.urls import reverse
+
+from benefits.core.admin import views
+
+
+@pytest.mark.django_db
+class TestBenefitsPasswordResetView:
+
+    @pytest.fixture
+    def view(self, app_request):
+        """Fixture to create an instance of BenefitsPasswordResetView."""
+        v = views.BenefitsPasswordResetView()
+        v.setup(app_request)
+        return v
+
+    def test_form_valid(self, view):
+        email = "mail@example.com"
+        form = view.form_class(data={"email": email})
+        assert form.is_valid()
+
+        view.form_valid(form)
+        assert view.success_url == f"{reverse("password_reset_done")}?email={email}"
+
+
+@pytest.mark.django_db
+class TestBenefitsPasswordResetDoneView:
+    @pytest.fixture
+    def view(self, app_request):
+        """Fixture to create an instance of BenefitsPasswordResetDoneView."""
+        v = views.BenefitsPasswordResetDoneView()
+        v.setup(app_request)
+        return v
+
+    def test_get_context_data(self, view):
+        context_data = view.get_context_data()
+        assert "email" in context_data
--- a/benefits/core/admin/views.py
+++ b/benefits/core/admin/views.py
@@ -1,8 +1,24 @@
-from django.contrib.auth.views import PasswordResetView
+from django.contrib.auth.views import PasswordResetView, PasswordResetDoneView
+from django.urls import reverse_lazy
 
 from benefits.core.admin.forms import BenefitsPasswordResetForm
 from benefits.core.mixins import RecaptchaEnabledMixin
 
 
 class BenefitsPasswordResetView(RecaptchaEnabledMixin, PasswordResetView):
+    """Subclass of stock PasswordResetView to enable reCAPTCHA and pass email to done view"""
+
     form_class = BenefitsPasswordResetForm
+
+    def form_valid(self, form):
+        self.success_url = f"{reverse_lazy("password_reset_done")}?email={form.cleaned_data["email"]}"
+        return super().form_valid(form)
+
+
+class BenefitsPasswordResetDoneView(PasswordResetDoneView):
+    """Subclass of stock PasswordResetDoneView to receive email param and add to template context"""
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["email"] = self.request.GET.get("email")
+        return context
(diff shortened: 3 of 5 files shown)


Scotchester:
Ah, the check failure reminds me that I wanted to confirm whether or not we are translating copy in the Admin at this time. Looking at other admin templates we already have, seems like no? I can remove those translate tags that were introduced by using Django's stock `password_reset_form.html` as a starting point.

jgravois:
> I wanted to confirm whether or not we are translating copy in the Admin at this time. Looking at other admin templates we already have, seems like no?

do you not see this when Spanish is selected?
Screenshot 2025-11-10 at 1 05 09 PM

> perhaps it could be merged before that if we hide the reset link on the login form.

fwiw, it seems perfectly reasonable to me to merge bite-sized pieces with the new UI hidden.

benefits/core/admin/forms.py
```diff
@@ -0,0 +1,12 @@
+from django.contrib.auth.forms import PasswordResetForm
+from django.forms import ValidationError
+
+from benefits.core import recaptcha
+
+
+class BenefitsPasswordResetForm(PasswordResetForm):
```
jgravois: this is subjective as hell so consider this piece of feedback so file it under 'take it or leave it', but i think 'Admin' is more descriptive/specific in this context than `Benefits`.


```suggestion
class AdminPasswordResetForm(PasswordResetForm):
```

benefits/core/admin/forms.py
```diff
@@ -0,0 +1,12 @@
...
+
+    def clean(self):
+        if not recaptcha.verify(self.data):
+            raise ValidationError("reCAPTCHA failed, please try again.")
+        return super().clean()
```
jgravois: thanks for resurfacing the tip. i think the UI here looks quite nice.

not sure how hard it would be, but if we could avoid displaying the 'default' form validation that complains about the forms sole field, that would be ideal yeah?

Image

Scotchester: You mean, not bother showing the error about it being a bad email address if the reCAPTCHA failed?

thekaveman (changes requested):

benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -1001,3 +1001,15 @@ msgid ""
...
+msgid "Error:"
+msgstr ""
+
+msgid "Enter your email and we’ll send you a link to reset your password."
+msgstr ""
+
+msgid "Email"
+msgstr ""
+
+msgid "Send password reset link"
```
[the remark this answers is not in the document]
thekaveman: cc @indexing to confirm my understanding :point_up:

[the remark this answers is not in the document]
Scotchester: Sounds good, thanks for confirming.

thekaveman:
> I'm not sure if this should be merged until we have a proper email sending solution in place? Or perhaps it could be merged before that if we hide the reset link on the login form. Open to suggestions.

I think I'm agnostic, whatever is the least amount of fiddling we have to do. I don't anticipate we'll want to _release_ any of this until all of it is in place. If it gets to `dev` (or even `test`) in the meantime though, that is probably fine.

benefits/urls.py
```diff
@@ -64,5 +67,28 @@ def test_secret(request):
...
+        BenefitsPasswordResetDoneView.as_view(extra_context={"site_header": admin.site.site_header}),
+        name="password_reset_done",
+    ),
+    path(
+        "reset/<uidb64>/<token>/",
+        auth_views.PasswordResetConfirmView.as_view(extra_context={"site_header": admin.site.site_header}),
+        name="password_reset_confirm",
+    ),
+    path(
+        "reset/done/",
```
thekaveman: Is there a reason to have these paths at the top-level vs. under `admin/`?

benefits/templates/admin/login.html
```diff
@@ -43,63 +43,20 @@
...
-      {% if password_reset_url %}
-        <div class="password-reset-link">
-          <a href="{{ password_reset_url }}">Forgotten your login credentials?</a>
-        </div>
-      {% endif %}
       <div class="login-submit-row">
         <input type="submit" value="Log in">
       </div>
 
       {% if request.recaptcha %}
```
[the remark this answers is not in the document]
Scotchester: Oops, good catch! I think it makes more sense to keep it in the main template.

tests/pytest/core/admin/test_views.py
```diff
@@ -0,0 +1,23 @@
...
+    assert response.url == f"{done_path}?email={email}"
+
+
+@pytest.mark.django_db
+def test_benefits_password_reset_done_view_context(client):
+    form_path = reverse("admin_password_reset")
+    email = "mail@example.com"
+    response = client.post(form_path, {"email": email}, follow=True)
+    # breakpoint()
+    assert response.context["email"] == email
```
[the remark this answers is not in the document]
thekaveman: (We also like to write these CBV tests _as test classes_ vs. functions)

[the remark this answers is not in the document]
thekaveman: Yeah, and I fully admit that our view tests are sort of all over the place right now! There are some good examples of the new pattern in https://github.com/cal-itp/benefits/blob/main/tests/pytest/eligibility/test_views.py

benefits/core/admin/forms.py
Scotchester: How did you even get that to submit, though? The browser considers `john@compilerl.a` to be a valid email address, but Django doesn't?

jgravois: hah. now that you mention it... my screenshot is an example of a simulated reCAPTCHA failure **and** an invalid email address. 🤦‍♂️ 

when i posted originally i (wrongly) assumed that the reCAPTCHA failure was triggering the field to be considered invalid all on its own.

carry on!

Issue #3268, opened 2025-11-04
Password reset: Implement and configure Azure Communication + Email Services


thekaveman:
For password reset, we'll need a service to send email to users. 

## Acceptance Criteria

- [ ] Terraform defines the necessary resources
- [ ] Use the default sending address in `dev` and `test`
- [ ] Configure a custom sending address for `prod` -- will need help from DevSecOps (ideally this would be something like `DoNotReply@calitp.org` or `DoNotReply@benefits.calitp.org`)
- [ ] Django settings are configured to use this email sending service

## Additional context

We implemented this in CDRC: https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/pull/220

PR #3293, opened 2025-11-13
Feat: Azure Communication and Email services

lalver1:
Part of #3268

This PR sets up the [`django-azure-communication-email`](https://github.com/retech-us/django-azure-communication-email) package for Django email sending with Azure and defines the necessary services in Terraform. It closely follows the CDRC implementation in https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/pull/220. The [`django-azure-communication-email`](https://github.com/retech-us/django-azure-communication-email) package is third-party (open source / Apache), with a fairly straightforward implementation of a [custom Django email backend](https://docs.djangoproject.com/en/5.2/topics/email/#defining-a-custom-email-backend) using the official [`azure-communication-email`](https://pypi.org/project/azure-communication-email/) package from Microsoft.

## Followups

- [ ] Configure the following DNS records for `benefits.calitp.org`
  - [x] TXT for domain status verification
  - [x] TXT for SPF Status
  - [x] CNAME for DKIM1 Status
  - [x] CNAME for DKIM2 Status
- [ ] Before making a `Benefits` release, import the email-related resources that were created by DevSecOps in the `prod` Azure environment

## Reviewing

`Rebuild and reopen` your devcontainer and verify that `django-azure-communication-email` is installed in the container. You can do this by running `pip freeze` in the dev container's terminal and checking that `django-azure-communication-email` shows up in the list of installed packages.

Ensure that the `terraform plan` command in the DevOps infra pipeline shows ~`6 to add, 1 to change`~ `5 to add, 1 to change`. These correspond to the following resources that will be created/modified in the Azure `dev` environment:

1. Communication Service: `azurerm_communication_service.main`
2. Email Communication Services: `azurerm_email_communication_service.main`
3. Email Communication Services Domain (free `azurecomm.net` address): `azurerm_email_communication_service_domain.azure_managed`
4. Associate domain with Email Service:`azurerm_communication_service_email_domain_association.azure_managed`
5. Secret for the Azure Email Backend setup environment variable: `azurerm_key_vault_secret.azure_communication_connection_string`
6. ~Secret for the "from email" environment variable: `azurerm_key_vault_secret.azure_communication_from_email`~
7. Changes to the web app's settings: `azurerm_linux_web_app.main`

### DIFF
--- a/terraform/app_service.tf
+++ b/terraform/app_service.tf
@@ -61,7 +61,7 @@ resource "azurerm_linux_web_app" "main" {
 
     # Django Azure Email Backend
     "AZURE_COMMUNICATION_CONNECTION_STRING" = "${local.secret_prefix}${local.azure_communication_connection_string_name})",
-    "DEFAULT_FROM_EMAIL"                    = "${local.secret_prefix}${local.azure_communication_from_email_name})",
+    "DEFAULT_FROM_EMAIL"                    = local.sender_email,
 
     # Django settings
     "DJANGO_ALLOWED_HOSTS"        = "${local.secret_prefix}django-allowed-hosts)",
--- a/terraform/email.tf
+++ b/terraform/email.tf
@@ -3,7 +3,6 @@ locals {
   data_location                              = "United States"
   custom_domain_name                         = "benefits.calitp.org"
   azure_communication_connection_string_name = "azure-communication-connection-string"
-  azure_communication_from_email_name        = "azure-communication-from-email"
   # Determine the correct sender domain based on the environment.
   # If is_prod is true, use the custom domain; otherwise, use the Azure-managed domain.
   sender_domain = local.is_prod ? azurerm_email_communication_service_domain.custom[0].mail_from_sender_domain : azurerm_email_communication_service_domain.azure_managed[0].mail_from_sender_domain
@@ -15,18 +14,6 @@ resource "azurerm_key_vault_secret" "azure_communication_connection_string" {
   value        = azurerm_communication_service.main.primary_connection_string
   key_vault_id = azurerm_key_vault.main.id
   content_type = "password"
-}
-
-resource "azurerm_key_vault_secret" "azure_communication_from_email" {
-  name         = local.azure_communication_from_email_name
-  value        = local.sender_email
-  key_vault_id = azurerm_key_vault.main.id
-
-  # This resource depends on the creation of one of the two possible domains.
-  depends_on = [
-    azurerm_email_communication_service_domain.azure_managed[0],
-    azurerm_email_communication_service_domain.custom[0]
-  ]
 }
 
 resource "azurerm_communication_service" "main" {


terraform/app_service.tf
```diff
@@ -59,6 +59,10 @@ resource "azurerm_linux_web_app" "main" {
     "REQUESTS_CONNECT_TIMEOUT" = "${local.secret_prefix}requests-connect-timeout)",
     "REQUESTS_READ_TIMEOUT"    = "${local.secret_prefix}requests-read-timeout)",
 
+    # Django Azure Email Backend
+    "AZURE_COMMUNICATION_CONNECTION_STRING" = "${local.secret_prefix}${local.azure_communication_connection_string_name})",
+    "DEFAULT_FROM_EMAIL"                    = "${local.secret_prefix}${local.azure_communication_from_email_name})",
```
thekaveman: Does this need to be a secret? We are basically hardcoding it in the `email.tf` anyway. I think we can just pass this in as an env var (it isn't sensitive).

terraform/email.tf
```diff
@@ -0,0 +1,94 @@
+locals {
+  communication_service_name                 = "ACS-CDT-PUB-VIP-CALITP-${local.env_letter}-001"
+  data_location                              = "United States"
+  custom_domain_name                         = "benefits.calitp.org"
+  azure_communication_connection_string_name = "azure-communication-connection-string"
+  azure_communication_from_email_name        = "azure-communication-from-email"
+  # Determine the correct sender domain based on the environment.
+  # If is_prod is true, use the custom domain; otherwise, use the Azure-managed domain.
+  sender_domain = local.is_prod ? azurerm_email_communication_service_domain.custom[0].mail_from_sender_domain : azurerm_email_communication_service_domain.azure_managed[0].mail_from_sender_domain
```
jgravois: this line got me thinking...

do y'all get an indication in VS Code for formatting errors in .tf files? i do not. outside the container i confirmed that the [`hashicorp.terraform`](https://github.com/cal-itp/benefits/blob/97132f78ef47f3ef68611f405c058f9c08674f5e/.devcontainer/devcontainer.json#L29) extension needs `terraform` to be installed too.

seems like the juice would be worth the squeeze to do that inside the container and to enable auto-formatting, but i won't rat-hole on it unless y'all agree its worth creating a ticket for slush time.

thekaveman: Yeah I think we've discussed installing `terraform` in the devcontainer before, I can't remember where those discussions are. I also can't remember why we haven't done it yet... Slush ticket sounds good :+1:

terraform/app_service.tf
thekaveman: It may be a different story during `apply`... but let's try it! (You could try the `apply` locally too, against `dev`, to see)

benefits/settings.py
```diff
@@ -382,3 +382,18 @@ def RUNTIME_ENVIRONMENT():
     REQUESTS_READ_TIMEOUT = 20
 
 REQUESTS_TIMEOUT = (REQUESTS_CONNECT_TIMEOUT, REQUESTS_READ_TIMEOUT)
+
+# Email
+# https://docs.djangoproject.com/en/5.1/ref/settings/#email-backend
```
thekaveman: We can update this link to point to the `5.2` Django docs

thekaveman (approved):
Minor nit, but nonblocking. Looks ready to try in `dev`!

Issue #2266, opened 2024-08-01
Admin: Set up production Microsoft Entra ID instance


angela-tran:
Depends on #2263, where we will learn more about what needs to be set up

We will probably need help from CDT DevSecOps on this.

The idea is to set up a tenant in the CDT Azure environment and create a registered application to represent the Benefits app. This registered application will have a client secret. The application ID and client secret are then used to configure Microsoft SSO for the Benefits app. 

External users are then invited to have access to the tenant which gives them access to the registered application. Because of that relationship, they should be able to successfully authenticate into the Benefits admin with Microsoft SSO.

## Acceptance criteria
- [ ] There is a tenant set up in the production CDT Azure environment which allows us to invite external users to access the registered application for Benefits

## Additional context
[Adding Microsoft credentials](https://megalus.github.io/django-microsoft-sso/credentials/)

PR #3295, opened 2025-11-14
Spike: log in to local Benefits Admin with Microsoft SSO

angela-tran:
Research for #2266 and #2263

### DIFF
--- a/benefits/core/admin/users.py
+++ b/benefits/core/admin/users.py
@@ -99,7 +99,9 @@ def add_transit_agency_staff_user_to_group(user, request):
 
 
 def pre_login_user(user, request):
+    # this needs to be updated to work for Microsoft too
+
     logger.debug(f"Running pre-login callback for user: {user.username}")
     add_google_sso_userinfo(user, request)
     add_staff_user_to_group(user, request)
-    add_transit_agency_staff_user_to_group(user, request)
+    # add_transit_agency_staff_user_to_group(user, request)


angela-tran:
Side note: this "Copy response" button in Gemini is awesome -- it copied the content in Markdown format. I didn't have to do anything to it :pray:

angela-tran:
@thekaveman and I followed the instructions from the Gemini output (which is basically the same as what's in the [Microsoft docs](https://learn.microsoft.com/en-us/entra/identity/multi-tenant-organizations/cross-tenant-synchronization-configure?pivots=same-cloud-synchronization)) to test this in tenants that we set up in Compiler's Azure environment.

We were able to verify that this setup will achieve what we need. We used a "Security group" to hold users in the source tenant and saw that the scheduled provisioning job was able to sync those users into the target tenant. We also checked that removing a user from the group in the source tenant will cause that user to be removed from the target tenant.

Note that the P1 license needed to be assigned to the group rather than the users for it to work.

angela-tran:
For this research, we set everything up via web UIs:
- [entra.microsoft.com](https://entra.microsoft.com) (Microsoft Entra admin center)
- [admin.cloud.microsoft](https://admin.cloud.microsoft) (Microsoft 365 admin center)

We didn't draft any Terraform code because:
1. it was easier that way, and this was a spike after all
2. we're not going to be the ones doing the configuration for either source or target tenant. (We don't have the sufficient roles nor do we really need to be the ones doing it, I think.) 

However, we could consider coming up with the needed Terraform if that would help the external teams that we'll be working with (DevSecOps and the transit agency IT teams).

Issue #2389, opened 2024-09-20
App container is calculating an incorrect version number from tag


thekaveman:
## Background

[In Release `2024.07.2`](https://github.com/cal-itp/benefits/releases/tag/2024.07.2) we went to a tag-based deployment model. See #1527 / #2213 for more context.

Our most recent [Release was for `2024.09.2`](https://github.com/cal-itp/benefits/releases/tag/2024.09.2).

I was looking at enrollment data, and noticed that ever since that `2024.07.2` release, our version numbers have been calculating incorrectly at runtime:

Event Type | Client Event Time | Version | Version should be
-- | -- | -- | --
returned enrollment | June 29, 2024, 4:09 PM | `2024.6.1` |  :white_check_mark: 
returned enrollment | July 1, 2024, 6:03 PM | `2024.6.1` | :white_check_mark: 
returned enrollment | July 4, 2024, 6:15 PM | `2024.7.1` | :white_check_mark: 
returned enrollment | July 16, 2024, 11:06 AM | `2024.7.1` | :white_check_mark: 
returned enrollment | July 19, 2024, 3:54 PM | `2024.7.1` | :white_check_mark: 
returned enrollment | July 29, 2024, 10:33 AM | `2024.7.1` | :white_check_mark: 
returned enrollment | August 1, 2024, 4:11 PM | `2024.7.3.dev0+gcd3b083.d20240731` | `2024.7.2`
returned enrollment | August 3, 2024, 11:05 AM | `2024.7.3.dev0+gcd3b083.d20240731` | `2024.7.2`
returned enrollment | August 9, 2024, 2:36 PM | `2024.7.3.dev0+gcd3b083.d20240731` | `2024.7.2`
returned enrollment | August 9, 2024, 7:17 PM | `2024.7.3.dev0+gcd3b083.d20240731` | `2024.7.2`
returned enrollment | August 13, 2024, 1:37 PM | `2024.7.3.dev0+gcd3b083.d20240731` | `2024.7.2`
returned enrollment | August 27, 2024, 1:03 PM | `2024.8.2.dev0+g7664917.d20240821` | `2024.8.1`
returned enrollment | September 2, 2024, 1:15 PM | `2024.8.2.dev0+g7664917.d20240821` | `2024.8.1`
returned enrollment | September 3, 2024, 11:43 AM | `2024.8.2.dev0+g7664917.d20240821` | `2024.8.1`
[...]

PR #3314, opened 2025-11-18
Fix: Benefits version number

### LINKED ISSUE #3311: Version number looks wrong in `prod` metrics
There's something funky going on with our version number in prod. If we check the deployed [`requirements.txt`](https://benefits.calitp.org/static/requirements.txt), the version looks correct (`2025.11.1`) but in looking at Amplitude events, we are seeing the "dirty" version string variant. E.g. enrollments over the last 90 days grouped by Version:

Image

From that chart, it seems like something broke here after the `2025.07.1` release.

## Expected behavior

We should see our normal "clean" version strings in Amplitude events.

## Additional context

We had a similar problem before, see #2389 (which, coincidentally, was right after the `2024.07.1` release... something about July?? 😅)

angela-tran:
Closes #3311

## Background 

Notice in `pyproject.toml` that our version is [dynamic](https://github.com/cal-itp/benefits/blob/main/pyproject.toml#L3). Python packaging build tools can set the version - see https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata.

We use [`setuptools-scm`](https://setuptools-scm.readthedocs.io/en/latest/) during our [Python-package build process](https://github.com/cal-itp/benefits/blob/4f9c3ac134e57157d0a3d1c5b3757e470270ad87/pyproject.toml#L55) so that it will calculate our version number for us by looking at the Git repo's metadata. The [default versioning scheme](https://setuptools-scm.readthedocs.io/en/latest/usage/#default-versioning-scheme), which we use, uses the name of the latest annotated tag, whose name follows a [version tag format](https://setuptools-scm.readthedocs.io/en/latest/usage/#version-tag-formats), as the version number if there are no pending changes.

## The problem

Starting with `2025.08.1`, the calculated version number is not what we expect; it is formatted with `{next_version}.dev{distance}+{scm letter}{revision hash}.dYYYYMMDD` rather than `{tag}`. This tells us that during the build process, something is causing the build tool to see that there are changes to the repo's files.

In the diff of `2025.07.1` to `2025.08.1`, we see that some directories were added to `.dockerignore` - https://github.com/cal-itp/benefits/compare/2025.07.1...2025.08.1#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5.

These directories are tracked by Git, so that explains why the build process running from within the Docker build context is seeing changes.

## The fix

Remove those directories from `.dockerignore` so that all git-tracked files will be seen during the build.

## Testing

I followed the steps listed in #2389 to first reproduce the issue and then confirm that with the fix, we get the correct version number.

To see that the fix works,
- (Outside of the devcontainer) Check out this branch - `git checkout fix/version-number`
- Create an annotated tag - `git tag -a 2025.11.2`, use whatever test message you want for the tag annotation
- Run `docker compose build --no-cache client; docker compose up -d client; docker compose exec -ti client /bin/bash`
- Run `python manage.py shell`
- Run `import benefits; print(benefits.VERSION);`
  - You should see `2025.11.2` :white_check_mark: 
- Now, clean-up :broom: :
  - Run `exit()`
  - Run `exit`
  - Run `git tag -d 2025.11.2`

## Screenshots

| Version | Output |
| --- | --- |
| `2025.07.1` :white_check_mark: | Screenshot from 2025-11-17 18-37-25 |
|`2025.08.1` :x: | Screenshot from 2025-11-17 18-37-40 |
| `2025.11.1` :x:  | Screenshot from 2025-11-17 18-38-15 |
| `2025.11.2` (my tag for testing) :white_check_mark: | Screenshot from 2025-11-17 18-38-29 |

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 **/__pycache__/
 node_modules/
+.pytest_cache/
 static/
 .coverage
 *.db
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ __pycache__/
 .DS_Store
 .sent_emails/
 .venv
+
+# including this just to be explicit. Pytest itself already ensures its cache directory is ignored by Git.
+.pytest_cache/


thekaveman:
Damn it, pip caching changes strike again! Sorry :sob: Thanks for finding and fixing!

.dockerignore
```diff
@@ -1,16 +1,9 @@
...
 .coverage
 *.db
 *.egg-info
-.flake8
-.pre-commit-config.yaml
-compose.yml
 *.log
 *.mo
 .DS_Store
-.env
```
jgravois: this one isn't tracked in version control, correct?

thekaveman: If we want to be sure, we can also add `.pytest_cache/` to our `.gitignore`

thekaveman:
> The commit message of https://github.com/cal-itp/benefits/commit/31ec22467f24f1b0fbc63972a2da56cb595d4149 also sounds like it was just some clean-up we did while we were there rather than a necessary change for caching.

Yep this is correct

thekaveman:
Thank you @angela-tran!! :pray: :flushed:

jgravois (approved):
> TIL of the acronym CIL... "check-in later"?

haha. don't ask me. i've always assumed it was 'see inline' but that doesn't make any sense. apparently its actually '[comments inline](https://www.acronymfinder.com/Comments-In_Line-(CIL).html)'

PR #3286, opened 2025-11-10
Password reset: implement email templates

jgravois:
resolves #3273

### design
https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/Cal-ITP-Benefits-Admin?node-id=2515-5606&m=dev
(note that in the comment thread [here](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP?node-id=2515-5606&m=dev#1513595711) Product/Design landed in favor of using the term 'Administrator' instead of 'Admin', to match the existing login UI)

### dev
this work attempts to keep the email styles as primitive and portable as possible (ie: no flexbox, bootstrap, etc.)

in anticipation of the actual email pipeline being stood up in #3293, i just riffed in Codepen: https://codepen.io/jgravois/pen/PwNoNOW?editors=1100

### testing
i cherry-picked this commit on top of @Scotchester's work in #3285 and confirmed that the same email is written to the console on submit with no additional code changes. 😎

PR #3321, opened 2025-11-19
chore: advertise a looser reset password link expiration

jgravois:
## TL;DR

* advertise a 24 hour expiration for password reset links instead of 15 minutes. (relevant [Figma discussion](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP?node-id=2515-5606&m=dev#1520258848))
* use a `<hr/>` instead of a plain old `<div/>` where appropriate
* add wrapping tags to ensure compatibility with legacy email clients

## deets

after i merged #3286 i got to thinking that i probably should have wrapped the template in the wrapper tags we all know and love. 

```html
<html>
  <body> 
    { message }
  </body>
</html>
```

i checked a few raw emails and found some that omit them, but the conventional wisdom is definitely still that there's a benefit to ☝️ (for example: [mailchimp templates](https://github.com/mailchimp/email-blueprints/blob/4d4daa85a8107baf128b88dd3baaf16137befb8e/templates/simple-letterhead-centerlogo.html) include them)

the resulting HTML is a bit uglier, but i'm more confident that it will render correctly if a transit agency is using an outdated email client. i suspect the odds of that are higher than 0%.

while i was rearranging the closet, i swapped in a more semantic [`<hr/>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/hr) tag too.

new: https://codepen.io/jgravois/pen/PwNoNOW?editors=1100
old: https://codepen.io/jgravois/pen/PwNoNOW?editors=1100

regardless, sorry for the noisy diff.

### DIFF
--- a/benefits/templates/registration/password_reset_email.html
+++ b/benefits/templates/registration/password_reset_email.html
@@ -1,63 +1,71 @@
-<link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&display=swap" rel="stylesheet">
-<style>
-  body {
-    color: #535456;
-    font-family: 'Noto Sans', sans-serif;
-    font-size: 14px;
-  }
-
-  h1 {
-    color: #212121;
-    font-size: 24px;
-  }
-
-  a {
-    color: unset;
-  }
-
-  .message-body {
-    max-width: 576px;
-  }
-
-  .link-wrapper {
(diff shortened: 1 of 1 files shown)


jgravois:
i'll revert this one to draft until the entire team agrees regarding what our password link expiration should be 😇

PR #3326, opened 2025-11-19
fix: Handle multiple forms when including reCAPTCHA on one of them

### LINKED ISSUE #3325: reCAPTCHA failing on front-end forms on Dev
On front-end forms like https://dev-benefits.calitp.org/eligibility/, reCAPTCHA is always failing due to a regression introduced in #3285.

Specifically, when pulling out the reCAPTCHA template code into an include, the selector used to attach the event listener to the submit element was simplified, and it now incorrectly selects the language chooser button, which is the first `[type="submit"]` that appears on the page.

Investigation and discussion of how to fix it in Slack: https://cal-itp.slack.com/archives/C037Y3UE71P/p1763576790694369

Scotchester:
Fixes #3325

Adds a unique identifier to the hidden reCAPTCHA field and uses that to ensure we attach the event listener to the correct form submit element.

### Testing instructions

1. To reproduce the issue, with your local repo still on `main`:
   1. Set up your local env to enable reCAPTCHA by adding the env vars found in the "Benefits reCAPTCHA config" note in LastPass to your `env` object in `launch.json`
   1. Go to your locally running application
   1. Select CST
   1. Try to choose an eligibility flow
   1. Observe error
1. To test the fix:
   1. Checkout this branch
   1. Follow reproduction steps again
   1. You should not encounter the error this time

### DIFF
--- a/benefits/core/templates/core/includes/recaptcha.html
+++ b/benefits/core/templates/core/includes/recaptcha.html
@@ -4,9 +4,14 @@ Adapted from https://stackoverflow.com/a/63290578/453168
 {% endcomment %}
 
 {% comment %}
+Create a pseudorandom identifier from the current microseconds to ensure we attach the event to the right button.
+{% endcomment %}
+{% now "u" as unique_id %}
+
+{% comment %}
 hidden input field will later send g-recaptcha token back to server
 {% endcomment %}
-<input type="hidden" name="{{ request.recaptcha.data_field }}" value="">
+<input type="hidden" name="{{ request.recaptcha.data_field }}" data-recaptcha-id="{{ unique_id }}" value="">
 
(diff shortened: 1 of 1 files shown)


benefits/core/templates/core/includes/recaptcha.html
```diff
@@ -3,10 +3,15 @@
 Adapted from https://stackoverflow.com/a/63290578/453168
 {% endcomment %}
 
+{% comment %}
+Create a pseudorandom identifier from the current microseconds to ensure we attach the event to the right button.
+{% endcomment %}
+{% now "u" as unique_id %}
```
thekaveman: Cool, didn't know about the `{% now %}` tag. Nice way to get a unique ID!

thekaveman (approved):
It works!

PR #3333, opened 2025-11-20
refactor(core): Convert agency_index to class-based view

### LINKED ISSUE #2997: Class-based views: Core agency index
View function: https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L37

## Acceptance Criteria

- [ ] Updates agency in session

Scotchester:
Closes #2997 

### Testing instructions

1. Pull branch
2. Visit `/cst` on localhost and see that the main heading is still customized as expected


tests/pytest/core/test_views.py
```diff
@@ -46,37 +47,29 @@ def test_form_valid(self, view, model_TransitAgency):
...
+        v = views.AgencyIndexView()
+        converter = TransitAgencyPathConverter()
+        v.setup(app_request, agency=converter.to_python(model_TransitAgency.slug))
+        return v
 
+    def test_view(self, view):
+        assert view.template_name == "core/index--agency.html"
 
-@pytest.mark.django_db
-def test_agency_index_multiple_flow(
```
Scotchester: If my spelunking is correct, it looks like this test was left over from a time where an agency with only one enrollment flow would redirect from its agency index to `/eligibility/start/`, but let me know if I got that wrong and this is still a scenario that needs to be tested.

tests/pytest/core/test_views.py
```diff
@@ -46,37 +47,29 @@ def test_form_valid(self, view, model_TransitAgency):
...
-    mocked_session_update.assert_called_once()
-
-    assert response.status_code == 200
-    assert response.template_name == "core/index--agency.html"
+class TestAgencyIndexView:
+    @pytest.fixture
+    def view(self, app_request, model_TransitAgency):
+        v = views.AgencyIndexView()
+        converter = TransitAgencyPathConverter()
+        v.setup(app_request, agency=converter.to_python(model_TransitAgency.slug))
```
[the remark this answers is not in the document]
thekaveman: > so it is probably safe to assume for this test that it's doing its job correctly and just set `agency=model_TransitAgency`.

Yeah I think this is probably fine, reduces some noise slightly in the test.

tests/pytest/core/test_views.py
thekaveman: Hmm I honestly can't remember. I know we have some shortcut redirect tricks like that for e.g. `/<slug>/eligibility` and `/<slug>/agency-card`, but I think you are right that this was an outdated test that doesn't need to be represented in the new structure.

thekaveman (dismissed):
Yup, looks good 👍

PR #3347, opened 2025-11-24
Fix: process mixins before dispatching in Enrollment retry

thekaveman:
Follow-up to #3346, testing is hard :sob:


jgravois (approved):
🌈🌈

PR #3353, opened 2025-12-01
Refactor: convert In-person Enrollment retry to CBV

### LINKED ISSUE #3110: Class-based views: In-person Enrollment reenrollment retry
View function: https://github.com/cal-itp/benefits/blob/main/benefits/in_person/views.py#L134

## Acceptance Criteria

- [ ] Inherit from the CBV created for #3011
- [ ] Apply the mixin from #3108

lalver1:
Closes #3110 

This PR refactors `benefits/in_person/views.retry` to a class-based view. The in-person retry page at `http://localhost/in_person/enrollment/retry` still looks like:

image

Note the small change in d01786715d34da9a307d6bd6b56a8fc9638d36dd, where the type of enrollment method was added to the Digital `RetryView` so that we can take advantage of inheritance for the In-person `RetryView`.

## Reviewing

A quick way to review this PR is to set up your Littlepay `dev` fixtures and go through an in-person flow. When you get to the enrollment page, in the browser's console, enter `document.getElementById("form-card-tokenize-fail-retry").submit();`. This will force calling the `RetryView` and then you'll see the desired template. You can do the same on the app's regular Digital path just to make sure that the Digital path still works. Also, the Switchio enrollment page does not call the `RetryView` so there is no need to test this PR for the Switchio processor.

### DIFF
--- a/benefits/in_person/templates/in_person/enrollment/retry.html
+++ b/benefits/in_person/templates/in_person/enrollment/retry.html
@@ -13,6 +13,6 @@
   </div>
   <div class="col-6">
     {% url routes.IN_PERSON_ENROLLMENT as url_try_again %}
-    <a href="{{ url_try_again }}" class="btn btn-lg btn-primary d-block">Try again</a>
+    <a href="{{ url_try_again }}" class="btn btn-lg text-white btn-primary d-block">Try again</a>
   </div>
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
LGTM!

PR #3355, opened 2025-12-02
Refactor: convert Enrollment system error to CBV

### LINKED ISSUE #3012: Class-based views: Enrollment system error
View function: https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/views.py#L59

## Acceptance Criteria

lalver1:
Closes #3012

This PR refactors `benefits/enrollment/views.system_error` to a class-based view. The system error page at `http://localhost/enrollment/error` still looks like:

image

`SystemErrorView` in this PR is very similar to [`RetryView`](https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/views.py#L84) since both are used to submit the result of the enrollment (either from Littlepay or Switchio) back to the server. We could have implemented `SystemErrorView` just like `RetryView` but the approach in this PR maybe seems more in line with Django? Both `get` and `post` are defined (in case at some point the view is called by a GET) and `post` reuses the `get` logic. I was a little concerned about using `dispatch` and overlooking something in the view's lifecycle that would produce any weird behavior.

Also, note the small fix ~in 8715f88032a251db66c47cb43e005b58644e60df~ to get rid of the following warning that we've been seeing (for example in this [workflow run](https://github.com/cal-itp/benefits/actions/runs/19641881581/job/56247233922#step:8:91)) when running the test suite:

```
tests/pytest/in_person/test_mixins.py:8
  /calitp/app/tests/pytest/in_person/test_mixins.py:8: PytestCollectionWarning: cannot collect test class 'TestView' because it has a __init__ constructor (from: tests/pytest/in_person/test_mixins.py)
```

The fix relies on ~[telling Pytest that the class isn't a test even though it starts with `Test`](https://docs.pytest.org/en/stable/example/pythoncollection.html#customizing-test-collection), as documented towards the end of that section.~ simply renaming the class so that it isn't interpreted as a test by pytest.

## Reviewing

Set up your Littlepay (or Switchio) `dev` fixtures and go through any flow. When you get to the enrollment page, in the browser's console, enter `document.getElementById("form-card-tokenize-fail-system-error").submit();`. This will force calling the `SystemErrorView` and then you'll see the desired template.

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -120,6 +120,10 @@ class SystemErrorView(AgencySessionRequiredMixin, FlowSessionRequiredMixin, Elig
         return super().get(request, *args, **kwargs)
 
     def post(self, request, *args, **kwargs):
+        # the Javascript in enrollment_littlepay/index.html and enrollment_switchio/index.html sends a form POST to this view
+        # rather than implementing this view as a FormView, which requires instantiating the
+        # enrollment.forms.CardTokenizeFailForm, we implement post() to simply return the template via get()
+        # we thus avoid interfering with the view's lifecycle and dispatch() method
         return self.get(request, *args, **kwargs)
 
 


tests/pytest/in_person/test_mixins.py
```diff
@@ -6,6 +6,7 @@
 
 
 class TestView(mixins.CommonContextMixin, View):
```
thekaveman: The purpose of this view is to test the `CommonContentMixin`? Maybe we just rename it to something like `SampleView` or something so we can avoid the funky dunder attribute?

benefits/enrollment/views.py
```diff
@@ -113,15 +110,19 @@ def dispatch(self, request, *args, **kwargs):
...
 
-    return TemplateResponse(request, TEMPLATE_SYSTEM_ERROR)
+    def get(self, request, *args, **kwargs):
+        # overwrite origin so that CTA takes user to agency index
+        agency = self.agency
+        session.update(request, origin=agency.index_url)
+        return super().get(request, *args, **kwargs)
+
+    def post(self, request, *args, **kwargs):
+        return self.get(request, *args, **kwargs)
```
thekaveman: Can you help me understand why we need a POST handler here? Is it the same story we dealt with the `RetryView` where JS POSTs a form back to this view?

thekaveman: Oh sorry, I see you described this in the PR. Maybe an inline comment here would help us in the future.

PR #3361, opened 2025-12-04
Fix: de-duplicate the `display_order` values on sample EnrollmentFlows

angela-tran:
The need for this fix was discovered while testing #3348. You can read https://github.com/cal-itp/benefits/pull/3348#issuecomment-3614268062 for more details and context, but the gist is that we need the `display_order`s to be unique, sequential values in order for `django-admin-sortable2` to work correctly. (I think loading fixtures doesn't run any constraint validation, so that'd be how this invalid config went undetected.)

The two flows that had the same `display_order` value were Older Adult and Calfresh Cardholder:

https://github.com/cal-itp/benefits/blob/d7ce8e8ba8159956a868ba231afb0688d29c46f5/benefits/core/migrations/local_fixtures.json#L235-L240

https://github.com/cal-itp/benefits/blob/d7ce8e8ba8159956a868ba231afb0688d29c46f5/benefits/core/migrations/local_fixtures.json#L277-L285

Fixtures to fix:
- [x] `local_fixtures.json`
- [ ] ~~Fixtures in LastPass~~ -- I looked over these, and they were fine as-is. Nice!

### DIFF
--- a/benefits/core/migrations/local_fixtures.json
+++ b/benefits/core/migrations/local_fixtures.json
@@ -252,7 +252,7 @@
     "fields": {
       "system_name": "veteran",
       "label": "U.S. Veteran",
-      "display_order": 3,
+      "display_order": 4,
       "oauth_config": "00e5cbdb-242a-4df5-a147-45d0c1e4dceb",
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
🔍 Nice find!

PR #3367, opened 2025-12-04
Chore: add isort for devcontainer, pre-commit

### LINKED ISSUE #3245: Integrate isort for enforcement of import order
> Nit: I would expect the new `import benefits.core.analytics` to come here (based on our informal rules of everything `benefits` last)
>
> Maybe it's time to finally get `isort` working properly in the devcontainer?

_Originally posted by @thekaveman in https://github.com/cal-itp/benefits/pull/3243#discussion_r2449909140_

thekaveman:
Closes #3245 

## For discussion / decision

Should we just install the tooling (the current state of the PR), and let files get modified over time as they are touched?

Or should we also run the tooling on all existing files and commit that in this PR as well?

I don't really have a preference, `isort` makes it easy to run against all current files if we wanted, but that could be a big commit. Maybe [@Scotchester's suggestion on the Cal-ITP website project](https://github.com/cal-itp/calitp.org/pull/537#pullrequestreview-3540780000) applies here as well?

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# sort imports with isort
+6e1d32e50a54d1436de91383c3f003fb373ce49c


jgravois:
> let files get modified over time as they are touched?

pre-commit says, 'no'. 😆

thekaveman:
I'm thinking that maybe the `pre-commit` aspect of this is a little overblown. Do we really need a check for `isort`?

Scotchester:
Oh hey :) https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

thekaveman:
OK this is now ready for review.

Thanks to @Scotchester for pointing out the `black` profile for `isort`, which makes those tools play nicely together. Also thanks for the idea of using a `.git-blame-ignore-revs` file.

I went ahead and ran `isort .` and committed the results in 6e1d32e50a54d1436de91383c3f003fb373ce49c to:

1. Avoid `pre-commit.ci` doing the same thing
2. Add the resulting SHA to `.git-blame-ignore-revs`

Read more about [`black`'s recommendations on `.git-blame-ignore-revs`](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame) and the the functionality [within `git` itself](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-filefile).

[GitHub also respects the `.git-blame-ignore-revs` file](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view).

PR #3370, opened 2025-12-05
Refactor: convert In-person Enrollment server error to CBV

### LINKED ISSUE #3112: Class-based views: In-person Enrollment server error
View function: https://github.com/cal-itp/benefits/blob/main/benefits/in_person/views.py#L160

## Acceptance Criteria

- [ ] Apply the mixin from #3108

lalver1:
Closes #3112 

This PR refactors `benefits/in_person/views.server_error` to a class-based view. The server error page at `http://localhost/in_person/error` still looks like:

image

Since this In-person view has no Digital counterpart, it is implemented as a `TemplateView`. The Javascript in `in_person/index_littlepay.html` sends a form POST to this view so we follow the same strategy of other error views that implement `post()` to simply return the template via `get()`.

## Reviewing

Set up your Littlepay `dev` fixtures and go through an In-person flow. When you get to the enrollment page, in the browser's console, enter `document.getElementById("form-card-tokenize-fail-server-error").submit();`. This will force calling the `ServerErrorView` and then you'll see the desired template. Also make sure that clicking on the button takes you back to the Admin Dashboard. There's no need to test against Switchio since this view is only used for Littlepay.

### DIFF
--- a/benefits/in_person/views.py
+++ b/benefits/in_person/views.py
@@ -131,9 +131,6 @@ class ServerErrorView(mixins.CommonContextMixin, AgencySessionRequiredMixin, Tem
 
     template_name = "in_person/enrollment/server_error.html"
 
-    def get(self, request, *args, **kwargs):
-        return super().get(request, *args, **kwargs)
-
     def post(self, request, *args, **kwargs):
         # the Javascript in in_person/index_littlepay.html sends a form POST to this view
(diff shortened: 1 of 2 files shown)


angela-tran (approved):
Sorry for my delayed response - this looks great, thank you @lalver1!

PR #3373, opened 2025-12-08
Refactor: convert agency eligibility index to class-based view

### LINKED ISSUE #2998: Class-based views: Core agency eligibility
View function: https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L46

## Acceptance Criteria

- [ ] Implement as `RedirectView`
- [ ] Resets session
- [ ] Updates agency in session

thekaveman:
Closes #2998

This view redirects to the `/eligibility` route with a session pre-configured with the agency parsed from the path via [`TransitAgencyPathConverter`](https://github.com/cal-itp/benefits/blob/main/benefits/core/urls.py#L16)

## Small implementation detail

We have to remove the `agency` key from kwargs incoming to the `get()` call before passing along to `super().get()`. This is because `super().get()` eventually calls `reverse(self.pattern_name, *args, **kwargs)`, where `pattern_name` is specified in this new CBV. The `reverse()` call would fail because this pattern doesn't expect any extra kwargs.

This is fine, as the CBV itself still has a reference to the original `self.kwargs`, if needed.

Alternatively, we could just call `super().get(request)` without passing along additional incoming args/kwargs, as we know they are unused. Happy to adjust the PR to go that direction if it seems cleaner.

## Reviewing

1. Checkout the branch and start the app locally
2. In the browser, enter the URL path `/cst/eligibility`
3. See the Eligibility Index page with CST's flows; confirm in the `DEBUG` bar that CST is stored as the agency in session

## Note

#3377 and #3378 are stacked on top of this one as separate PRs, but I could easily put the commits into this single PR and close all 3 issues with one merge.

### DIFF
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -7,7 +7,7 @@ from django.shortcuts import redirect
 from django.template import loader
 from django.template.response import TemplateResponse
 from django.utils.decorators import method_decorator
-from django.views.generic import TemplateView
+from django.views.generic import RedirectView, TemplateView
 from django.views.generic.edit import FormView
 
 from benefits.core.forms import ChooseAgencyForm
@@ -60,13 +60,20 @@ class AgencyIndexView(TemplateView):
         return context
 
 
-@pageview_decorator
-def agency_eligibility_index(request, agency: models.TransitAgency):
+class AgencyEligibilityIndexView(RedirectView):
     """View handler forwards the request to the agency's Eligibility Index (e.g. flow selection) page."""
-    session.reset(request)
-    session.update(request, agency=agency, origin=agency.index_url)
 
-    return redirect(routes.ELIGIBILITY_INDEX)
+    pattern_name = routes.ELIGIBILITY_INDEX
+
+    @method_decorator(pageview_decorator)
+    def get(self, request, *args, **kwargs):
+        # keep a reference to the agency before removing from kwargs
+        # since the eventual reverse() lookup doesn't expect this key in the kwargs for routes.ELIGIBILITY_INDEX
+        # self.kwargs still contains the agency if needed
+        agency = kwargs.pop("agency")
+        session.reset(request)
+        session.update(request, agency=agency, origin=agency.index_url)
+        return super().get(request, *args, **kwargs)
 
 
 @pageview_decorator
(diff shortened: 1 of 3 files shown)


lalver1 (approved):
This looks good! I tested locally and saw no issues 👍

About the alternative approach, using `kwargs.get("agency")` and returning `return super().get(request)`. I agree that it may be cleaner looking but it seems like we would still want to include a comment that would explain why we aren't using `return super().get(request, *args, **kwargs)` which is the pattern we usually see. It seems like the root issue is having familiarity with `RedirectView` and its `get_redirect_url()`, so in my mind both approaches have the same level of "clarity". Because I find them equivalent I'll go ahead and approve the current approach 😄

benefits/core/views.py
```diff
@@ -60,13 +60,20 @@ def get_context_data(self, **kwargs):
...
+
+    @method_decorator(pageview_decorator)
+    def get(self, request, *args, **kwargs):
+        # keep a reference to the agency before removing from kwargs
+        # since the eventual reverse() lookup doesn't expect this key in the kwargs for routes.ELIGIBILITY_INDEX
+        # self.kwargs still contains the agency if needed
+        agency = kwargs.pop("agency")
+        session.reset(request)
+        session.update(request, agency=agency, origin=agency.index_url)
+        return super().get(request, *args, **kwargs)
```
angela-tran: > ## Small implementation detail
> 
> We have to remove the `agency` key from kwargs incoming to the `get()` call before passing along to `super().get()`. This is because `super().get()` eventually calls `reverse(self.pattern_name, *args, **kwargs)`, where `pattern_name` is specified in this new CBV. The `reverse()` call would fail because this pattern doesn't expect any extra kwargs.
> 
> This is fine, as the CBV itself still has a reference to the original `self.kwargs`, if needed.

So just to check that I understand, the alternative would be something like:
```python
   agency = kwargs.get("agency")
   ...
   return super().get(request)
```
?

angela-tran: (This is not necessarily an actual suggestion, just an attempt to convey what I think the alternative would be)

```suggestion
        agency = kwargs.get("agency")
        session.reset(request)
        session.update(request, agency=agency, origin=agency.index_url)
        return super().get(request)
```

angela-tran (dismissed):
I agree with @lalver1's comments in https://github.com/cal-itp/benefits/pull/3373#pullrequestreview-3568691300

angela-tran:
(ahhh sorry, please ignore the dismissed review)

benefits/core/views.py
thekaveman: Yep, exactly

PR #3400, opened 2025-12-19
docs: updates to agency adoption table

indexing:
Today we activated SLORTA in `prod` so they can begin rider pre-enrollment.


thekaveman (approved):
Looks great!

PR #3416, opened 2026-01-09
Fix: admin `<title>` element

### LINKED ISSUE #3288: Admin: `<title>` element empty in most views
Most views in the Admin (excluding in-person enrollment) are not currently populating the `<title>` element. I believe this was an accidental regression in #2773 when some of the template inheritance was changed, but I haven't traced the template logic enough to determine exactly why.

Based on manual testing and a quick look at our codebase, I think the only ones currently showing a title are in-person enrollment views and the logout confirmation.

## To Reproduce

1. Go to the Admin in any environment
2. See that the login page doesn't have a title
3. Log in as a superuser and browser around

## Expected behavior

All admin views would have a populated `<title>` element.

lalver1:
Closes #3288

This PR slightly simplifies the overridden admin templates' inheritance chain to 

`benefits/admin/agency-dashboard-index.html -> benefits/admin/agency-base.html -> benefits/admin/base_site.html -> admin/base.html`

to more closely match Django's admin app inheritance chain:

`index.html -> base_site.html -> base.html`

and adds a `title` block to `benefits/admin/base_site.html` so that admin pages don't have an empty title.

Compare this to the original inheritance chain for the in-person app:

`benefits/admin/agency-dashboard-index.html -> benefits/admin/agency-base.html -> benefits/admin/base.html -> admin/base.html`

While this worked, the relationship between `benefits/admin/base.html` and `admin/base.html` seemed brittle, it may have worked because of how the Django apps were ordered in the settings, and I guess it could break with a different ordering.

Similarly, for the admin app the original inheritance chain was

`admin/index.html -> benefits/admin/base_site.html -> benefits/admin/base.html -> admin/base.html`

with `benefits/admin/base.html` not really being needed.

## Reviewing

Start the debugger and navigate to the admin page. For different pages on the site, verify that the title on the browser displays the correct title instead of the page's url:

image

image

For pages on the in-person app, verify that the title on the browser still displays the correct title:

image

image


benefits/locale/en/LC_MESSAGES/django.po
```diff
@@ -6,7 +6,7 @@
 msgid ""
 msgstr ""
 "Report-Msgid-Bugs-To: https://github.com/cal-itp/benefits/issues \n"
-"POT-Creation-Date: 2025-11-18 09:00-0800\n"
+"POT-Creation-Date: 2026-01-12 11:21-0800\n"
```
thekaveman: This is a nit, but do we want to include a bump to these timestamps when the file wasn't actually modified?

thekaveman (approved):
Looks great, tested locally, saw the expected titles throughout the Admin.

Thanks for all the notes and explanation @angela-tran and @lalver1, was really helpful understanding the nuance of this PR/change!

PR #3429, opened 2026-01-12
Update and consolidate environment variables

### LINKED ISSUE #3371: Password reset: add documentation of new environment variables
From https://github.com/cal-itp/benefits/pull/3293/files#diff-088d9f35d23a4347d221d71dd49b02b95001dff4abe637a40fe0bc04d502049cR13-R15, we have 2 new environment variables related to using an Azure email service.

```
# Email sending
AZURE_COMMUNICATION_CONNECTION_STRING=
DEFAULT_FROM_EMAIL=noreply@example.calitp.org
```

Let's update the ~~environment variables~~ Admin interface reference page (https://docs.calitp.org/benefits/reference/admin-interface/) to include these and brief explanation on what they're for.

## Acceptance Criteria

- [ ] Admin interface reference page contains the above variables and brief explanation on what they're for.

## Additional context

- #3293 
- #2940 (epic)

angela-tran:
Closes #3371

This PR adds documentation of the environment variables introduced by #3293 and also does some minor clean-up to the "Admin interface" reference documentation (update line numbers in link, re-order some call-outs).

### DIFF
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -210,7 +210,7 @@ Comma-separated list of User-Agent strings which, when present as an HTTP header
 
     [The Django admin site](https://docs.djangoproject.com/en/5.0/ref/contrib/admin/)
 
-The configuration values that can be stored in the application database are managed directly in the [Django Admin interface](./admin-interface.md) at the `/admin` endpoint.
+The configuration values that can be stored in the application database are managed directly in the Django Admin interface at the `/admin` endpoint.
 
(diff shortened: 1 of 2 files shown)


thekaveman:
> apologies for the delay

No worries at all!!

> Should we then also move the environment variables from...

Sure why not put them all on one page? I don't see a good reason not to have that be the case.

> I could do it in this PR

I think that would be fine. Just rename the PR to "Consolidate environment variables reference" or something like that and it makes sense to me!

thekaveman (approved):
This looks great Angela! Nice consolidation work, I think the overall env vars reference reads a lot cleaner being on a single page. And I like reducing the number of sections/pages in the sidebar nav!

PR #3440, opened 2026-01-16
docs: marks El Dorado Transit Authority as launched

indexing:
We activated EL Dorado in the production environment this morning, so this provider is now accepting enrollments for reduced fares.


thekaveman (approved):
LGTM

PR #3452, opened 2026-01-28
feat: Update environment list to be dev, test, and prod

### LINKED ISSUE #3386: Introduce `Acceptance` environment
More details in this comment/ticket: https://github.com/cal-itp/benefits/issues/3384#issue-3724807685

> Maybe another follow up here is to add an option to the Environments drop-down for `Acceptance` to make it really clear that we aren't using `Prod` in `test` for Switchio. This would require code changes in the models for logic that determines which URL to use, so maybe not worth it at this time.

This is mostly a ["labeling thing"](https://github.com/cal-itp/benefits/issues/3384#issuecomment-3648172201) where we want the labels in `test` to clearly indicate the values are pointed at Switchio ACC (acceptance).

## Acceptance Criteria

- [ ] The drop downs for `LittlepayConfig` and `SwitchioConfig` contain an entry for `Acceptance`
- [ ] The logic for Switchio is updated to consider the `Acceptance` environment and use appropriate secrets for `tokenization-base-url` and `enrollment-base-url`
- [ ] Switchio secrets in `test` are updated to reflect the new environment
- [ ] The logic for Littlepay is updated to reject `Acceptance` as a selection

## Additional context

We may want to consider _further_ model / logic changes to take into consideration the [`RUNTIME_ENVIRONMENT`](https://github.com/cal-itp/benefits/blob/main/benefits/settings.py#L37) such that:

- In `dev`, only `QA` can be selected
- In `test`
  - For Switchio, only `Acceptance` can be selected
  - For Littlepay, only `QA` or `Production` can be selected
- In `prod`, only `Production` can be selected

Scotchester:
Closes #3386 

---

- Adds `dev` to the `Environment` choices, supports use of it in Switchio configs, and prevents use of it in Littlepay configs
- Renames choice `qa` to `test`

### Testing instructions

1. Pull branch
2. Update your local Switchio fixtures file:
   - `core.transitprocessorconfig` – `"environment": "dev"`
   - Replace remaining instances of `qa` with `int`
4. Ensure the Switchio fixtures are set to load on rebuild
5. Update your `.env` to replace `qa` with `int` in the names of your Switchio secrets
6. Rebuild container
7. Start debug session
8. Run through a full enrollment and confirm that it works

### TODO after merge:

- [x] Update Littlepay fixtures with secrets in LastPass
- [x] Update Switchio fixtures with secrets in LastPass
- [x] Switchio secrets in `dev` are updated to reflect the new environment
  - [x] Duplicate existing `switchio-qa-*` secrets in Azure Key Vault with new `switchio-int-*` names
  - [x] Update `SwitchioConfig`s in Admin to change to the `acc` secrets
  - [x] Delete the `qa` secrets in Azure Key Vault for `dev`

### TODO after deployment to `test`:

- [x] Switchio secrets in `test` are updated to reflect the new environment
  - [x] Duplicate existing `switchio-prod-*` secrets in Azure Key Vault with new `switchio-acc-*` names
  - [x] Update `SwitchioConfig`s in Admin to change to the `acc` secrets
  - [x] Change VCTC and RABA's `SwitchioConfig.environment` values to `test` (which will be used for both production validation and steady-state, going forward)
  - [x] Delete the `prod` secrets in Azure Key Vault for `test`

(There should be no need to make and config changes in `prod`.)

### DIFF
--- a/benefits/core/migrations/0074_update_env_choices.py
+++ b/benefits/core/migrations/0074_update_env_choices.py
@@ -3,6 +3,15 @@
 from django.db import migrations, models
 
 
+def migrate_data(apps, schema_editor):
+    TransitProcessorConfig = apps.get_model("core", "TransitProcessorConfig")
+
+    for config in TransitProcessorConfig.objects.all():
+        if config.environment == "qa":
+            config.environment = "test"
+        config.save()
+
+
 class Migration(migrations.Migration):
 
     dependencies = [
@@ -18,4 +27,5 @@ class Migration(migrations.Migration):
                 help_text="A label to indicate which environment this configuration is for.",
             ),
         ),
+        migrations.RunPython(migrate_data),
     ]


benefits/core/migrations/local_fixtures.json
```diff
@@ -134,7 +134,7 @@
     "pk": 2,
     "fields": {
       "tokenization_api_key": "123",
-      "tokenization_api_secret_name": "cst-switchio-qa-api-secret",
+      "tokenization_api_secret_name": "cst-switchio-acc-api-secret",
```
thekaveman: For all of these fixtures changes, I think we want to leave the default of `qa` since that is what we should be using locally. 

[`acc` is for use in our `test` environment](https://github.com/cal-itp/benefits/issues/3386#issuecomment-3807284725), where these local fixtures don't come into play.

thekaveman (changes requested):

thekaveman:
> I would like to propose that we not use a label of `qa` in our code to refer to both Littlepay QA and Switchio INT

I think this is a good suggestion. We started with QA because that is what Littlepay calls it,  but I agree it might be better to do something else. I would propose we match our own naming convention of `dev`, `test`, `prod` -- so we aren't relying on whatever each transit processor comes up with (likely would be different for others as we look to onboard more in the future). And then we can be consistent / use a single `TextChoices` for all.

Scotchester:
@thekaveman I had a similar thought to use our env naming convention, but if we only use a single `TextChoices` list, how should we handle it for processors that only have two envs?

Options off the top of my head:
- Show all three options in the admin list and just throw a validation error if they select the irrelevant option (basically what the original issue proposed)
- Show all three options in the admin list and silently map both `dev` and `test` to the processor's lower environment
- Figure out a way to inject some custom CSS to hide the irrelevant option

benefits/core/migrations/0074_update_env_choices.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name="transitprocessorconfig",
+            name="environment",
+            field=models.TextField(
+                choices=[("dev", "Development"), ("test", "Testing"), ("prod", "Production")],
+                help_text="A label to indicate which environment this configuration is for.",
+            ),
+        ),
+    ]
```
thekaveman: I think we need a data migration here as well? I'm seeing that all the existing records that listed `qa` still do after running this migration.

Issue #3357, opened 2025-12-03
Evaluate packages to enable database-backed copy translation


thekaveman:
As another phase of the work in #3356 to simplify agency onboarding, we can evaluate some Django packages that move copy translation management into the models/database.

## Acceptance Criteria

- [ ] Find a good candidate area of context to zero in on for this package evaluation
- [ ] Decide on one or more packages to evaluate
- [ ] Implement the package for the context, and write up notes about the features and UX
- [ ] Present to engineering, product, and design the results of this research

## Additional context

Some package options we identified already (non-exhaustive)

- https://github.com/feincms/django-translated-fields
- https://github.com/deschler/django-modeltranslation

Issue #3356, opened 2025-12-03
Evaluate and normalize agency-specific copy


thekaveman:
We want to make more progress on simplifying agency onboarding, reducing the places in code that need to be updated. 

Three areas to focus on:

- Agency-specific landing page heading/blurb, e.g in this context: https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L63
- Eligibility start page blurb, e.g. in this context: https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L47
- Enrollment success blurb, e.g. in this context: https://github.com/cal-itp/benefits/blob/main/benefits/enrollment/views.py#L178

## Acceptance Criteria

- [x] Evaluate above context for duplicate/snowflakey copy
- [ ] Propose update(s) that reduces need for duplicate hard-coded copy
- [ ] Present to product and design for evaluation and potential implementation

Issue #3057, opened 2025-07-31
Use agency short-name or long-name where appropriate in copy context objects


angela-tran:
https://github.com/cal-itp/benefits/blob/99a76c6f2f1521603affbd3a59edb10b7ca227e1/benefits/core/context/agency.py#L25-L35 is an example of copy context objects in which we have some form of the transit agency's name hard-coded in the string.

This issue is to suggest that we make use of `agency.short_name` or `agency.long_name` in these (and possibly other copy context objects) so that we don't have to do a deployment / release if that value needs to be changed.

jgravois:
the bold values in the table below are updates I'd recommend we make to our list of active agencies to support this effort ☝️

| Agency (long_name) | short_name |
| -- | -- |
| California State Transit | CST |
| El Dorado Transit | **El Dorado Transit*** |
| Monterey-Salinas Transit | MST |
| Nevada County Connects | **Nevada County Connects*** |
| Redding Area Bus Authority | RABA |
| **Roseville Transit*** | **Roseville Transit*** |
| Sacramento Regional Transit | SacRT |
| Santa Barbara MTD | SBMTD |
| San Luis Obispo RTA | ~RTA~ **SLO RTA*** |
| Ventura County Transportation Commission | VCTC |

The guiding principles being:
* The long name needs to be unique among all agencies, the short name does not
* the short name should be recognizable to end-users
* the long name and short name can match
* if the agency is just a place-name, we should tack the word 'Transit' (or 'transit') onto it

With those changes in place, @cmajel and I did some brainstorming and came up with alternative standardized options for the three locations where we currently maintain bespoke copy for each agency:

## Agency landing page

current copy:
> Get your reduced fare on [CST] public transit when you tap to ride

proposed:
> Get your reduced fare when you tap to ride on [short_name]

## Eligibility

current copy:
> All [RABA] transit benefits reduce fares by 50% for bus service on fixed routes

proposed:
> All transit benefits at [short_name] reduce fares by 50% or more for bus service on fixed routes.

or (more generically)

> All transit benefits at [short_name] reduce fares by 50% or more.

## Success

current copy:
> When boarding a [MST] bus, tap this card and you will be charged a reduced fare.

remains unchanged:
> When boarding a [short_name] bus, tap this card and you will be charged a reduced fare.

or (to avoid calling out agency's supported modes)

> When boarding public transit provided by [short_name], tap this card and you will be charged a reduced fare.

[...]

PR #3460, opened 2026-02-03
Refactor: consolidate context data - Agency Index

angela-tran:
This is a follow-up to #2777.

The `context` modules were originally introduced so that we could move away from having to manually copy and edit a set of templates every time we add a new agency.

Building upon that work, we set out to move towards class-based views so that we could then consolidate those context dictionaries into the views' `get_context_data`. Due to the varying complexity of those class-based view refactors, that consolidation was not always done, so we currently have a mix of views that get their context from a `@property` on a model vs. views that define their context in `get_context_data`.

As a way of gaining momentum on simplifying our copy, we should get our code in a more consistent and consolidated state.

The goal is to get rid of the `context` modules entirely since their purpose was to be an incremental step in figuring out a sustainable approach to copy management.

This PR starts with the `AgencyIndexView`.


angela-tran:
We've been thinking and talking about cleaning up our copy a lot, with tickets like #3356 and #3357. And https://github.com/cal-itp/benefits/issues/3057 is a ticket that's been lingering.

I wanted to just try this out and see if folks feel this could help remove a distracting inconsistency and give us some momentum.

PR #3470, opened 2026-02-05
chore: Replace logging in sentry module with basic print statements

### LINKED ISSUE #1379: Fix logging during application startup
## Acceptance Criteria
- [ ] During application startup, log messages are recorded as expected

## Additional context

In #1360, we made a small change to the `sentry` module to use a `logger` instead of `print`. However, we're not seeing the log statements. It may be due to the way our logging is [configured](https://github.com/cal-itp/benefits/blob/dev/benefits/settings.py#L210) and/or because `sentry` is at the root-level of the Django project.

Scotchester:
Fixes #1379

---

I made them look mostly like the standard `logging` output format, but without digging into how to output the current line number dynamically. Seems good enough for this use case.

In this sample, the first line is a standard `logging` statement, and the second two lines are the custom `print`s implemented here:

```
[05/Feb/2026 15:37:51] INFO django.utils.autoreload:265 /calitp/app/benefits/sentry.py changed, reloading.
[05/Feb/2026 15:37:52] INFO benefits.sentry Enabling Sentry for environment 'local', release '6bff0df4ec0458ef4073bfb1b2d65e99bf8ed980'...
[05/Feb/2026 15:37:52] INFO benefits.sentry SENTRY_TRACES_SAMPLE_RATE set to: 0.1
```

### DIFF
--- a/benefits/sentry.py
+++ b/benefits/sentry.py
@@ -1,4 +1,4 @@
-import logging
+import datetime
 import os
 import shutil
 import subprocess
@@ -9,8 +9,6 @@ from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.scrubber import DEFAULT_DENYLIST, EventScrubber
 
 from benefits import VERSION
-
-logger = logging.getLogger(__name__)
 
 SENTRY_CSP_REPORT_URI = None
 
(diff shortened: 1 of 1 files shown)


thekaveman (approved):
👍

PR #3479, opened 2026-02-11
Fix: replace 'qa' with 'test' for the sample LittlepayConfig

angela-tran:
Small fix to something we missed when we reviewed #3452

The `environment` on the sample `LittlepayConfig` needed to be updated from `qa` to `test`.

Using an invalid value for `environment` causes this error when running through the app:
Screenshot from 2026-02-11 10-44-56

### DIFF
--- a/benefits/core/migrations/local_fixtures.json
+++ b/benefits/core/migrations/local_fixtures.json
@@ -107,7 +107,7 @@
     "model": "core.transitprocessorconfig",
     "pk": 1,
     "fields": {
-      "environment": "qa",
+      "environment": "test",
       "transit_agency": 1,
       "portal_url": "https://www.transit-processor-portal.com"
     }


angela-tran:
> This PR updates `local_fixtures.json`. After merging, please be sure to make a corresponding update to the fixtures with secrets in our shared LastPass folder and let other developers know that they should update their local fixtures accordingly.

I checked over the fixtures in LastPass, and the `environment` labels there are all good.

jgravois (commented):
> Using an invalid value for environment causes this error when running through the app:

i'm nearly certain this is user-error, but after getting in sync with github and checking out `main` locally, i didn't have any trouble running through an e2e littlepay enrollment for CST.

in my .env i confirmed i am pointing at 'local_fixtures.json' and ran bin/init.sh before starting the local server. what else could i be missing?

jgravois (approved):
kudos to @angela-tran not only for the fix but for dragging the code reviewer along every step of the way. 🤗

Issue #3444, opened 2026-01-21
Transition `dev` infrastructure to a single Azure subscription


lalver1:
When we originally set up the Benefits Azure infrastructure with the help of DevSecOps, we utilized two different Azure subscriptions:

* One for the `dev` and `test` environments (`CDT/ODI Development`)
* One for the `prod` environment (`CDT/ODI Production`)

Read more in the Benefits docs: https://docs.calitp.org/benefits/deployment/infrastructure/#environments

The impacts of these dual subscriptions manifest in e.g. the [`terraform/init.sh`](https://github.com/cal-itp/benefits/blob/main/terraform/init.sh) helper script, which configures the subscription based on which environment you are working on. This is also why there are two different service connections in [the infra pipeline](https://github.com/cal-itp/benefits/blob/main/terraform/azure-pipelines.yml) and (part of) why we need [environment setup helpers](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/workspace.py) for that pipeline as well.

For DDRC (which started as a clone of Benefits infra) we only [use a single subscription](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/commit/7333151e7e1bea613a55e31ce5613db1935fc5f0), with a [single service connection](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/commit/8ebf66dee2fe8e22f316ea6f78ddf33e24633cf5). This is DevSecOps preferred approach going forward.

## Checklist

Referring to the table below, the general approach is to move each resource tagged as `MOVE? Yes` using `az resource move` to `CDT/ODI Production`, remove the references to the moved resources in the Terraform state, import the moved resources into the Terraform state, update the Terraform helper scripts that select the Azure subscription, and run a `plan` and `apply` using the `CDT/ODI Production` subscription. The moved `dev` resources will already be in `CDT/ODI Production` and the other resources will be created by the Terraform configuration.

| NAME | TYPE | IN TERRAFORM? |  MOVE?   |
|------|------|----------|----------|
RG-CDT-PUB-VIP-CALITP-D-001 | Resource group | Only reference to it (`data`), not managed by Terraform |  |
| benefits-notify Slack channel email | Action group | Yes | |
AS-CDT-PUB-VIP-CALITP-D-001 | App Service | Yes | Yes |
ASP-CDT-PUB-VIP-CALITP-D-001 | App Service plan | Yes | Yes |
[...]

Issue #2831, opened 2025-04-08
Transition `test` infrastructure to a single Azure subscription


thekaveman:
When we originally set up the Benefits Azure infrastructure with the help of DevSecOps, we utilized two different Azure subscriptions:

* One for the `dev` and `test` environments (`CDT/ODI Development`)
* One for the `prod` environment (`CDT/ODI Production`)

Read more in the Benefits docs: https://docs.calitp.org/benefits/deployment/infrastructure/#environments

The impacts of these dual subscriptions manifest in e.g. the [`terraform/init.sh`](https://github.com/cal-itp/benefits/blob/main/terraform/init.sh) helper script, which configures the subscription based on which environment you are working on. This is also why there are two different service connections in [the infra pipeline](https://github.com/cal-itp/benefits/blob/main/terraform/azure-pipelines.yml) and (part of) why we need [environment setup helpers](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/workspace.py) for that pipeline as well.

For DDRC (which started as a clone of Benefits infra) we only [use a single subscription](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/commit/7333151e7e1bea613a55e31ce5613db1935fc5f0), with a [single service connection](https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/commit/8ebf66dee2fe8e22f316ea6f78ddf33e24633cf5). This is DevSecOps preferred approach going forward.

## Checklist

Referring to the table below, the general approach is to move each resource tagged as `MOVE? Yes` using `az resource move` to `CDT/ODI Production`, remove the references to the moved resources in the Terraform state, import the moved resources into the Terraform state, update the Terraform helper scripts that select the Azure subscription, and run a `plan` and `apply` using the `CDT/ODI Production` subscription. The moved `test` resources will already be in `CDT/ODI Production` and the other resources will be created by the Terraform configuration.

| NAME | TYPE | IN TERRAFORM? |  MOVE?   |
|------|------|----------|----------|
RG-CDT-PUB-VIP-CALITP-T-001 | Resource group | Only reference to it (`data`), not managed by Terraform |  |
| benefits-notify Slack channel email | Action group | Yes | |
AS-CDT-PUB-VIP-CALITP-T-001 | App Service | Yes | Yes |
ASP-CDT-PUB-VIP-CALITP-T-001 | App Service plan | Yes | Yes |
[...]

PR #1886, opened 2024-02-12
Feat: configure Azure file share backup

thekaveman:
Closes #1816 

## What this PR does

1. Defines an Azure Recovery Services Vault; this is where snapshots are stored. The default Geo-redundant storage is configured, so that backups are stored across various Azure geographic regions for maximum availability in case of e.g. a data center outage. The cost difference for our level of transactions and storage is mostly irrelevant at this point.
2. Links the existing Storage Account, containing the file(s) to be backed up, to this new Recovery Services Vault.
3. Creates an Azure File Share backup and retention policy: a daily snapshot taken at 14:00 UTC (roughly 6-7am Pacific), keeping 1 backup per day and 5 per week (1 for each weekday). The thinking here is we likely make changes throughout the day, so the snapshot will be at the beginning of a new day, capturing the end-state of any previous day's changes.

I also added a quick update to our Terraform docs for getting setup with a local `terraform.tfvars` file, which is required to run e.g. `terraform plan` locally.

## Relevant Azure docs

* [About Azure file share backup](https://learn.microsoft.com/en-us/azure/backup/azure-file-share-backup-overview)
* [Overview of share snapshots for Azure Files](https://learn.microsoft.com/en-us/azure/storage/files/storage-snapshots-files)
* [Manage Azure file share backups](https://learn.microsoft.com/en-us/azure/backup/manage-afs-backup)
* [Azure Storage redundancy](https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy)
* [Azure Files pricing](https://azure.microsoft.com/en-us/pricing/details/storage/files/)

## Relevant Terraform docs

* [`azurerm_recovery_services_vault`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/recovery_services_vault)
* [`azurerm_backup_container_storage_account`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/backup_container_storage_account)
* [`azurerm_backup_policy_file_share`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/backup_policy_file_share)

PR #3492, opened 2026-02-13
Chore: remove Azure file share and mount

lalver1:
Part of #3444

This PR will help with #3444 and #2831 by having less resources to move to the new subscription and by freeing up the App Service from the storage account, simplifying the move.

## Azure file share

Since for each environment we are storing the Sqlite database file in the Azure managed `/home` directory, and not in an Azure file share mount, we can remove this. The file share was originally created to persist the Sqlite database in https://github.com/cal-itp/benefits/pull/1775 but after noticing some issues with the database being locked, the database was moved to the Azure managed `/home` directory in https://github.com/cal-itp/benefits/pull/2166.

As an extra verification of the behavior described above, we confirmed that the database file in the file share of each environment (`dev`, `test`, and `prod`) was last modified in June 2024, matching the date of https://github.com/cal-itp/benefits/pull/2166.

## Recovery services vault

A recovery services vault was initially configured in #1886, but since we are not using the Azure file share anymore, we can remove this.

As an extra verification that it is not being used, we confirmed in the Azure Portal that for each environment, `Protected items -> Backup items` shows 0 backed up items for each category of backup type.

## Terraform plan

Terraform plan shows

> Plan: 0 to add, 1 to change, 4 to destroy.

as expected.


lalver1:
There is one more resource that can be removed, moving this back to draft to add it...

lalver1:
Thanks for the review everyone! I'll keep an eye out on this as it gets deployed.

And that's right @jgravois, we won't move `sacalitpd001` and since it isn't being managed by Terraform, no need to clean up anything in Terraform. So it'll get deleted once we delete the `RG-CDT-PUB-VIP-CALITP-D-001` resource group in `CDT/ODI Development`.

PR #3511, opened 2026-02-20
Refactor: invert `TransitAgency` to `TransitProcessorConfig` relationship

### LINKED ISSUE #3507: Invert the relationship between `TransitAgency` and `TransitProcessorConfig`
Currently `LittlepayConfig` and `SwitchioConfig` have a foreign key to the `TransitAgency` (via the base `TransitProcessorConfig` model).

Going the other way would allow us to handle regional use-cases like SLO/SLORTA, where we can still have 2 different `TransitAgency` instances, referencing the same `LittlepayConfig`. This will also help us with the Ventura County use-case, where all 6 of those agencies can reference the same `SwitchioConfig`.

This should also make it more straightforward to continue improvements to the agency onboarding and maintenance experience within Benefits Administrator, regardless of the exact implementation details there (a wizard-like flow, or a consolidated form/view).

## Acceptance Criteria

- [x] The `TransitProcessorConfig.transit_agency` field is removed
- [x] A new `TransitProcessorConfig.label` field is added as a `TextField` that will be used for internal (Benefits Administrator) display only, to help distinguish between different instances
- [x] The `TransitProcessorConfig.__str__` method is updated to combine the `environment` and new `label` fields
- [x] A new `TransitAgency.transit_processor_config` field is added as a `ForeignKey` to the `TransitProcessor` model
- [x] The existing calculated props `TransitAgency.littlepay_config` and `TransitAgency.switchio_config` are modified to read from the new field added above
- [x] A data migration exists that:
  - [x] Sets the `TransitProcessorConfig.label` value based on the existing selection of `TransitAgency.slug`
  - [x] Sets the `TransitAgency.transit_processor_config` value based on the existing (inverted, to be removed) relationship for each `TransitProcessorConfig`

## Additional context

[Related Slack thread](https://cal-itp.slack.com/archives/C037Y3UE71P/p1771542859659989)

angela-tran:
Closes #3507 

This PR makes the changes described in #3507; no need for me to repeat those details here :sweat_smile:  

I also needed to update the `clean` methods for `LittlepayConfig` and `SwitchioConfig`. I ran through enrollment for both those transit processors to make sure it's all still working. :white_check_mark:

### DIFF
--- a/benefits/core/migrations/local_fixtures.json
+++ b/benefits/core/migrations/local_fixtures.json
@@ -108,7 +108,7 @@
     "pk": 1,
     "fields": {
       "environment": "test",
-      "transit_agency": 1,
+      "label": "cst",
       "portal_url": "https://www.transit-processor-portal.com"
     }
   },
@@ -117,7 +117,8 @@
     "pk": 2,
     "fields": {
       "environment": "dev",
-      "transit_agency": null
+      "label": "",
+      "portal_url": ""
     }
   },
   {
@@ -153,7 +154,8 @@
       "info_url": "https://www.agency-website.com",
       "phone": "1-800-555-5555",
       "customer_service_group": 2,
-      "logo": "agencies/cst.png"
+      "logo": "agencies/cst.png",
+      "transit_processor_config": 1
     }
   },
   {


angela-tran:
> This PR adds a Django migration without updating any fixtures. Is a corresponding fixture update needed?

Done :white_check_mark: 

> This PR updates `local_fixtures.json`. After merging, please be sure to make a corresponding update to the fixtures with secrets in our shared LastPass folder and let other developers know that they should update their local fixtures accordingly.

Will do, and will update this comment when it's done

thekaveman:
> there not being a Switchio group configured for the flow

Doh, this has to be it. I totally spaced that part, assumed they were in my fixtures, but alas, they are not!

angela-tran:
>     1. I'm using the VCTC `INT` credentials for the Switchio config, that should all work as far as we know right? I am using CST as the agency (locally) but otherwise configured with VCTC `INT` credentials.

Yep, those are the ones I'm using too, and they are working.


>     2. Has anyone seen this 400 blows up into unhandled exception before?

We have ticket https://github.com/cal-itp/benefits/issues/3468 open to address the unhandled exception. I can spend a little time seeing if it's an easy fix since we're already here looking at it.

Issue #3180, opened 2025-09-18
Automate creation of parent and child issues required to onboard an agency to Cal-ITP Benefits


indexing:
We are making strides to standardize and refine the process of launching a new agency in Cal-ITP Benefits. We now have an `epic` to track progress on the onboarding of each agency. We also have a specific set of child issues required to complete the launch of an agency. See #3102 as an example.

The process to create the set of issues in manual and rather time intensive. The goal of this effort is to automate the creation of this set of issues when we kick off onboarding.

## Acceptance Criteria

- [ ] Automated process generates an `epic` issue with a title starting with `Agency onboarding:`. This is the parent for a set a sub issues to track the progress of onboarding.
- [ ] Automated process generates a set of `story` issues with a title starting with `<agency shortname>:`. All of these issue are associated with the `epic` created in the step above as the parent.
- [ ] User can specify agency full name
- [ ] User can specify agency short name (for prefix: in child issues)
- [ ] The content output is easily modified if the process requires changes.

PR #3538, opened 2026-02-27
docs: marks RABA live, adds City of SLO, and revises table layout

indexing:
- Redding is now live
- moved the `agency card` column to the end
- updated the column title for `Low income` to `CalFresh cardholders` to align with the name of the enrollment pathway in the app copy
- FIRST AGENCY TO OFFER ELIGIBILITY FOR RIDERS THAT RECEIVE CALFRESH BENEFITS 🎉


thekaveman:
Just noting for your awareness @indexing, altering the structure of this table ~will~ may have to be reconciled with the automated issue scaffolding from #3180. 

Since that workflow attempts to add a row to that table for the agency.

~@jgravois already has an in-progress PR where we can handle this change.~

It turns out this specific change shouldn't matter for the issue scaffolding workflow. But the advice still stands.

indexing:
@thekaveman Thanks for the heads-up. This speaks to why we have PR reviews.

thekaveman (approved):
LGTM 👍

PR #3546, opened 2026-03-02
chore: Reset migrations

### LINKED ISSUE #3445: Squash/reset database migrations
We'd like to reduce the number of [migration files](https://github.com/cal-itp/benefits/tree/main/benefits/core/migrations) which is currently at around 72 migrations. The migrations can either be 

- [Squashed](https://docs.djangoproject.com/en/6.0/topics/migrations/#squashing-migrations), which involves reducing an existing set of many migrations down to one (or sometimes a few) migrations which still represent the same changes
- Reset, which involves deleting all existing migration files, deleting the table that keeps track of the migrations, and creating a new `0001_initial.py` file

Scotchester:
Closes #3445 

---

Tested with `local_fixtures.json`.

To-do before merging:

- [x] Verify it works with secret fixtures
- [x] Verify it works after loading a copy of the prod DB, wiping the migration table, and faking the initial migrations
- [x] [Document our data migration workflow](https://github.com/cal-itp/benefits/issues/3445#issuecomment-3929088298)
- ~~[ ] Update new migration filenames with the date of when this is merged~~

## Testing instructions

1. Pull branch
2. Purge the current migration table in the database
   - There are several ways to accomplish this, but the direct SQLite command is: `DELETE FROM django_migrations`
3. Fake running the new migrations with `python manage.py migrate --fake`
4. Check out the admin and front-end and verify that things look/work the same as before

[New guide page](https://benefits-3546--cal-itp-previews.netlify.app/guides/migrations-and-fixtures/)


Scotchester:
> I take it that the reason for naming the initial migration files as `0001_reset_<date>.py` rather than just `0001_initial.py` is to signal to anyone who _may_ be running Benefits with a version before this that there was a reset of the migrations. I think it'd be fine to not worry about that and have them named as `0001_initial.py`, but I don't have a strong opinion for either way so this is fine.

I thought it might be nice to have an prominent reminder of when we reset the migrations, but I also don't feel strongly about it :) Maybe we can get another vote or two before merging?

> Just checking since I don't see this written anywhere: we need some post-merge steps for uploading new SQLite `.db` files with faked migrations for dev, test, and prod, correct?

That's right, each time this code gets deployed to an env, we'll need to do that. Kegan laid out the steps in this comment on the original issue: https://github.com/cal-itp/benefits/issues/3445#issuecomment-3917511854

benefits/core/migrations/0002_create_staff_group.py
```diff
@@ -8,7 +15,7 @@ def create_staff_group(apps, schema_editor):
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("core", "0012_remove_transitagency_merchant_id"),
+        ("core", "0001_reset_20260226"),
```
thekaveman: Can this migration not be merged into the `0001` migration?

PR #3548, opened 2026-03-03
Chore: local fixtures cleanup

angela-tran:
Last bit of follow-up for #3523 

This PR updates `local_fixtures.json` so that the ordering and included fields match what is exported by running the `./bin/dumpdata.sh` helper script. I used `HEAD` as the commit for both "pre" and "post" model changes.

This makes it so when making model changes, we can also use the helper script to update the local fixtures.


lalver1 (approved):
Thanks @angela-tran! Verified this locally and it looks good! 👍

Issue #3527, opened 2026-02-26
Migrate database from SQLite to PostgreSQL for `dev` environment (Cutover)


lalver1:
Load data and cutover to use PostgreSQL.

lalver1:
## Manual Data Loading

- [x] Data
- [x] Agency logos

The Benefits application running on Container Apps now shows agencies, flows, etc. as well as agency logos.

## Making database backups via `dumpdata`

I verified that a backup of the `django` database in Postgres can be made using `dumpdata`. Note that the ordering of some of the JSON objects does not follow the same order as the backup made from the SQLite database, but the content is the same.

Issue #3411, opened 2026-01-08
⛰️ Migrate from Azure App Service to Container Apps


lalver1:
## Motivation

Currently the Benefits app runs on the Azure App Service platform. While App Service has been reliable and meets the current requirements, we'd like to migrate to Azure Container Apps to take advantage of the following features:

- Native support for running background containerized worker services
- More consistency across other web app projects, both in terms of the type of Cloud resources used and CI/CD setup
- A more natural event-driven horizontal scaling

## High-level approach

The migration will be handled in three phases:

- [**Prerequisites:**](https://github.com/cal-itp/benefits/issues/3443) This will require moving the `dev` and `test` infrastructure from the CDT/ODI Development subscription to the CDT/ODI Production subscription and setting up the Terraform pipeline to have the capability to use a Federated Identity.
- [**Migrate the database from SQLite to PostgreSQL:**](https://github.com/cal-itp/benefits/issues/3412) This will require resetting migrations, exporting the data, setting up the PostgreSQL database, and loading the data.
- [**Migrate from App Service to Container Apps:**](https://github.com/cal-itp/benefits/issues/3413) This will require migrating the infrastructure to use the new services, refactoring the Terraform code to a more modular style, and deciding how to handle user uploaded data.

PR #3555, opened 2026-03-05
Feat: define Azure infra for `postgres`

### LINKED ISSUE #3447: Migrate database from SQLite to PostgreSQL for `dev` environment (Infra)
This focuses on writing Terraform for changes relevant to CI/CD and Azure infrastructure.

This can mostly be done in parallel with #3446. Depending on how the initial deploy scenario is handled for `local`, there may be a loose dependency on #3446 since we may want to do some of the initial deploy steps in Terraform for the Azure environments.

## Acceptance Criteria

The `dev` Benefits instance will have all the required resources to use PostgreSQL instead of SQLite.

lalver1:
Closes #3447

This PR sets up the required infrastructure for a new PostgreSQL Flexible Server that will replace the current SQLite database and moves the App Service to the virtual network shown below. Django will connect to the PostgreSQL server as part of the cutover in #3527.

### App Service

- Moves the App Service from the DevSecOps managed VNet/Subnet (`VNET-CDT-PUB-SHRD-W-P-001`/`SN-CDT-PUB-CALITP-D-001`) to the new `VNET-CDT-PUB-VIP-CALITP-D-001`/`SNET-CDT-PUB-VIP-CALITP-D-APP` VNet/Subnet defined below
- [Sends all outbound traffic through the VNet](https://learn.microsoft.com/en-us/azure/app-service/configure-vnet-integration-routing#configure-application-routing) (`vnet_route_all_enabled = true`). Although it was already set up like this before this PR, the only difference is that it was using the DevSecOps managed VNet.

### Database

- Uses PostgreSQL version 17 (18 not available yet in Azure as reported by the result of `terraform plan`: `expected version to be one of ["15" "14" "11" "16" "13" "12" "17"], got 18`)
- Uses the smallest `storage_mb` (max storage) allowed since it can only be scaled up, scaling down `storage_mb` forces a new PostgreSQL Flexible Server to be created
- Sets up the admin credentials in Terraform code

### Networking

- Sets up the `VNET-CDT-PUB-VIP-CALITP-D-001` virtual network (with no subnet overlap) as `10.0.0.0/24` (256 addresses - `10.0.0.0` to `10.0.0.255`)
  - App (`APP`) subnet `10.0.0.0/26` (64 addresses - `10.0.0.0` to `10.0.0.63`)
    - For a subnet, [Azure uses 5 addresses for networking and each app instance temporarily doubles its number of addresses when scaling.](https://learn.microsoft.com/en-us/azure/app-service/overview-vnet-integration#subnet-requirements). So with 64 addresses we can have (64-5)/2=29 app instances. Possibly a little less given that some addresses are used for platform upgrades. But in any case, 64 addresses cover the maximum scale of a single multitenant App Service plan.
    - [Includes delegation `Microsoft.Web/serverFarms`](https://learn.microsoft.com/en-us/azure/app-service/configure-vnet-integration-enable?tabs=portal#prerequisites)
    - [Default outbound access disabled](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access?tabs=portal#why-is-disabling-default-outbound-access-recommended) (`default_outbound_access_enabled = false`)
    - [NAT Gateway](https://learn.microsoft.com/en-us/azure/app-service/overview-nat-gateway-integration) for the Service App's outbound connectivity
  - Database (`DB`) subnet `10.0.0.64/26` (64 addresses - `10.0.0.64` to `10.0.0.127`)
    - [The smallest CIDR range is `/28` (16 addresses). For a subnet, Azure uses 5 addresses for networking and each flexible server instance uses 4 addresses.](https://learn.microsoft.com/en-us/azure/postgresql/network/concepts-networking-private#virtual-network-concepts). So with 64 addresses we can have (64-5)/4=14 database servers.
    - [Includes delegation `Microsoft.DBforPostgreSQL/flexibleServers`](https://learn.microsoft.com/en-us/azure/postgresql/network/concepts-networking-private#virtual-network-concepts)
    - [Default outbound access disabled](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access?tabs=portal#why-is-disabling-default-outbound-access-recommended) (`default_outbound_access_enabled = false`)
  - Available `10.0.0.128/25` (128 addresses - `10.0.0.128` to `10.0.0.255`)
- Sets up a [Private DNS zone](https://learn.microsoft.com/en-us/azure/postgresql/network/concepts-networking-private#use-a-private-dns-zone) and a [virtual private link](https://learn.microsoft.com/en-us/azure/dns/private-dns-getstarted-portal#virtual-private-links) so that Django running on the App Service can resolve the database server's name (see this [Terraform sample](https://learn.microsoft.com/en-us/azure/dns/dns-private-zone-terraform?tabs=azure-cli#implement-the-terraform-code))
- Sets up the [Network Security Groups (NSG)](https://learn.microsoft.com/en-us/azure/virtual-network/network-security-groups-overview) listed below. Note that NSGs cannot filter traffic based on URLs, only on IP addresses and Ports.
  - (`APP` Subnet) App -> Internet: Allowed on port 443 only (`AllowHTTPSOutbound`)
  - (`DB` Subnet) App -> Database: Allowed (`AllowPostgresInboundFromApp`).
  - (`DB` Subnet) Internet -> Database: Blocked (`DenyAllInbound`).
- Initial testing in the Azure Portal suggests that we can create all of these resources with our existing `Contributor` role
## Reviewing

While we can't test the database connection in this PR (we'll have to wait until we do the cutover to PostgreSQL in #3527), we should test the behavior of the App Service in the new `VNET-CDT-PUB-VIP-CALITP-D-001` virtual network:

- [x] Ensure we can still reach `https://dev-benefits.calitp.org/`
- [x] Go through a full enrollment with Littlepay and Switchio to ensure we can still reach the IdG and transit processors
- [x] If Sentry is active, ensure the app can send Sentry notifications
- [x] Log in to the admin using Google SSO

## Follow-ups

- [x] Once testing has confirmed that the App Service works in the new VNet, remove the [reference to the old VNet in Terraform](https://github.com/cal-itp/benefits/blob/main/terraform/network.tf#L3-L8)
- [ ] Although the infrastructure is already secure, adding the following security-related items could be an Epic (part of #3411) to make some further security improvements:
  - [ ] Set up a firewall to complement the configured NSGs since firewalls allow filtering traffic based on URLs. This would enable filtering traffic on a per-site basis, something NSGs cannot do because they can only filter based on IP addresses and these regularly change for a particular service.
  - [ ] To make the firewall usage practical, we would need to define a single source of truth list containing the allowed sites. The list would have to be accessible from the application and from Terraform, this would make it less error-prone to leave out a site when adding/removing sites to/from the list.

### DIFF
--- a/terraform/network.tf
+++ b/terraform/network.tf
@@ -2,12 +2,16 @@ locals {
   subnet_prefix = "SNET-CDT-PUB-VIP-CALITP-${local.env_letter}"
   network_subnets = {
     "APP" = {
-      prefix     = ["10.0.0.0/26"] # 64 addresses - 10.0.4.0 to 10.0.4.63
-      delegation = "Microsoft.Web/serverFarms"
+      prefix            = ["10.0.0.0/26"] # 64 addresses - 10.0.4.0 to 10.0.4.63
+      delegation        = "Microsoft.Web/serverFarms"
+      actions           = ["Microsoft.Network/virtualNetworks/subnets/action"]
+      service_endpoints = null
     }
     "DB" = {
-      prefix     = ["10.0.0.64/26"] # 64 addresses - 10.0.0.64 to 10.0.0.127
-      delegation = "Microsoft.DBforPostgreSQL/flexibleServers"
+      prefix            = ["10.0.0.64/26"] # 64 addresses - 10.0.0.64 to 10.0.0.127
+      delegation        = "Microsoft.DBforPostgreSQL/flexibleServers"
+      actions           = ["Microsoft.Network/virtualNetworks/subnets/join/action"]
+      service_endpoints = ["Microsoft.Storage"]
     }
   }
 }
@@ -33,11 +37,13 @@ resource "azurerm_subnet" "main" {
   virtual_network_name = azurerm_virtual_network.main.name
   resource_group_name  = data.azurerm_resource_group.main.name
   address_prefixes     = each.value.prefix
+  service_endpoints    = each.value.service_endpoints
 
   delegation {
     name = "delegation-${lower(each.key)}"
     service_delegation {
-      name = each.value.delegation
+      name    = each.value.delegation
+      actions = each.value.actions
     }
   }
 


thekaveman (approved):
I'm good with it, let's goooooooo 🚀

PR #3556, opened 2026-03-05
docs: fix a few broken links to migrations and fixtures page

jgravois:
just something small i noticed when i tested #3546 locally.

one thing that surprised me was that the ids in the `django_migrations` table didn't reset when i faked the migrations after blowing away all the previous rows.

Screenshot 2026-03-05 at 8 25 11 AM

i assume its harmless. i was just temporarily at a loss thinking that somehow @Scotchester's PR was able to preserve the original migration history without the files 😆


Scotchester:
Huh, I didn't notice that before, but I see it now, too. IIRC, databases keep track of auto-incrementing PKs and won't restart them without some additional effort.

Scotchester (approved):
Thank you, and sorry I neglected to test these links.

Issue #3544, opened 2026-03-02
Update agency onboarding scaffolding workflow to also add Issues to the Customer Success project board


thekaveman:
CS uses their own [GitHub Project Board](https://github.com/orgs/cal-itp/projects/23) to track tasks relevant to their work streams. Lately, they have been manually adding some of the agency onboarding tickets to this board as well, to increase visibility on the CS team, especially for those folks that don't spend a lot of time on the USB.

With the new [Agency Onboarding Issue Scaffold](https://github.com/cal-itp/benefits/blob/main/.github/workflows/agency-onboarding.yml) workflow, we have an opportunity to save everyone a few clicks, and automate adding these issues to this additional board.

## Acceptance Criteria

- [ ] All issues created by the Agency Onboarding Issue Scaffold workflow get added to the Customer Success project board automatically

PR #3561, opened 2026-03-06
chore: for all issues created with an 'agency-onboarding' label, set a 'Todo' project status immediately

### LINKED ISSUE #3543: Update agency onboarding scaffolding workflow to mark Issues as `Todo`
Whenever a new issue is added to this repo, we have [a workflow](https://github.com/cal-itp/benefits/blob/main/.github/workflows/add-to-project-issues.yml) that automatically adds the issue to our USB. By default, whenever Issues are added to the USB, they go into the status `Needs Shaping`. We use this status as the default status so that we ensure a proper product/design/dev alignment is achieved for each task before it is groomed from the Backlog and into a Sprint (or Stretch time) for work.

This treatment is the same for all Issues created by our new [Agency Onboarding Issue Scaffold workflow](https://github.com/cal-itp/benefits/blob/main/.github/workflows/agency-onboarding.yml).

Since the idea of the agency onboarding issue scaffold is to create a known checklist set of tasks to complete for onboarding, the `Needs Shaping` status doesn't align as well with how we view these tasks. In other words, this set of tasks is "shaped" and ready to be worked on, as soon as they are created.

## Acceptance Criteria

- [ ] New issues created by the Agency Onboarding Issue Scaffold workflow are in the `Todo` status

jgravois:
resolves #3543

initially i thought we were going to need a fancier approach for creating the initial issue, but the snippet @thekaveman shared helped me realize that the easier option is to just wire up a follow-up action that looks for new (and moved) issues that are labeled as <kbd>agency-onboarding</kbd> and assign the appropriate project status after the fact.

### DIFF
--- a/.github/workflows/add-to-project-dependabot.yml
+++ b/.github/workflows/add-to-project-dependabot.yml
@@ -10,7 +10,7 @@ jobs:
     # see https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events
     if: github.actor == 'dependabot[bot]' || github.actor == 'pre-commit-ci[bot]'
     steps:
-      - uses: actions/add-to-project@main
+      - uses: actions/add-to-project@1
         with:
           project-url: ${{ secrets.USB_PROJECT_URL }}
           github-token: ${{ secrets.USB_PROJECT_TOKEN }}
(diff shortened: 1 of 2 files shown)


thekaveman:
In fact, I think the same approach could be taken for https://github.com/cal-itp/benefits/issues/3544. Up to you @jgravois if you want to handle in the same PR or a different one.

PR #3570, opened 2026-03-10
chore: retire agency card specific enrollment success copy variant

### LINKED ISSUE #3565: chore: remove agency card success variant in favor of one generic message
As a part of the Ditto refactor and reducing content variations, we want to revise the enrollment success message to us a common pattern. 

Currently, agency cards have a distinct message, and we want to move towards a general message template.

**Current copy for agency cards:**

> Your contactless card is now enrolled in a CST Agency Card transit benefit. When boarding a CST bus, tap this card and you will be charged a reduced fare. You will need to re-enroll if you choose to change the card you use to pay for transit service. 
> 
> You were not charged anything today. Thank you for using Cal-ITP Benefits! 

**Updated copy (same copy for agency cards as other enrollment paths):**

> You were not charged anything today. When boarding public transit provided by {{transitProviderShortName}}, tap this card and you will be charged a reduced fare. 
>
> You will need to re-enroll if you choose to change the card you use to pay for transit service.
>
> Thank you for using Cal-ITP Benefits!
> 

In the near future, there will also be a regional variant of this language:
> You were not charged anything today. When boarding public transit at the following providers, tap this card and you will be charged a reduced fare:
>
> [ list of transit providers]
> 
> You will need to re-enroll if you choose to change the card you use to pay for transit service. 
> 
> Thank you for using Cal-ITP Benefits! 

## Acceptance Criteria

- [ ] Translations requested 

## Additional context

* [Ditto instance with the updated copy](https://app.dittowords.com/projects-beta/69966fa1712cf0458cca9293?viewState=text&detailsPanel=EDIT&selectedBlockIds=69a87ff2b93ac72d48f2eb4b)

jgravois:
closes #3565

This PR unifies the success copy for agency cards and all the other existing flows.

Screenshot 2026-03-11 at 1 25 45 PM

español:
Screenshot 2026-03-11 at 1 26 44 PM

### DIFF
--- a/benefits/enrollment/views.py
+++ b/benefits/enrollment/views.py
@@ -170,36 +170,30 @@ class SystemErrorView(AgencySessionRequiredMixin, FlowSessionRequiredMixin, Elig
         return self.get(request, *args, **kwargs)
 
 
-@dataclass
-class EnrollmentSuccess:
-    short_name: str
-
-    re_enrollment_message: str = _(
-        "You will need to re-enroll if you choose to change the card you use to pay for transit service."
-    )
-    success_message: str = ""
-    thank_you_message: str = _("Thank you for using Cal-ITP Benefits!")
-
-    def __post_init__(self):
-        self.success_message = _(
-            "You were not charged anything today. When boarding public transit provided by {short_name}, tap this "
-            "card to receive a reduced fare.",
-            short_name=self.short_name,
-        )
-
-    def dict(self):
-        return asdict(self)
-
-
 class SuccessView(PageViewMixin, FlowSessionRequiredMixin, EligibleSessionRequiredMixin, TemplateView):
     """View handler for the final success page."""
 
     template_name = "enrollment/success.html"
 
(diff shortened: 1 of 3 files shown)


benefits/enrollment/views.py
```diff
@@ -172,69 +172,33 @@ def post(self, request, *args, **kwargs):
 
 @dataclass
 class EnrollmentSuccess:
```
thekaveman: IMHO we can dump this `@dataclass` entirely now. It was really only useful for a consistent structure for the variants. But now without the variants... we can just inject context directly?

jgravois: makes sense to me. 🤙

in JS I would have probably reached for a function to isolate the logic for composing the strings but there's so little going on now I'm fine with just inlining it too.

thekaveman (approved):
Looks great, nice cleanup!

Tested locally with CST / Agency card, CST / Older Adult, English and Spanish.

PR #3573, opened 2026-03-11
docs: Point Django docs links to stable version

Scotchester:
I propose that when we link to the Django docs in our own docs, we link to the Stable version. Yes, technically we are behind that currently, but in general we move fairly quickly to update Django, and I'm not sure it's worth the maintenance burden of updating these URLs every time we update.

If others disagree, though, that's OK!


jgravois (dismissed):
if there's anywhere we can afford to be loose about versions, its here IMO. 🤗

thekaveman (dismissed):

Issue #3305, opened 2025-11-14
SLO Transit: Transit provider configured in Cal-ITP Benefits `test` environment for production validation


indexing:

indexing:
City of SLO will onboard to Cal-ITP Benefits as part of SLO RTA. Thus, they will not onboard as a distinct agency.

lalver1:
@indexing just wanted to see what you thought about using `slotransit` as the slug? We mainly use it for the direct agency URL. I noticed that the [city's transit website](https://www.slocity.org/government/department-directory/public-works/slo-transit-bus) uses `slotransit@slocity.org` as a contact email, so we could also do `slocity` but I guess `slotransit` feels more appropriate? In case it helps us decide on a slug, San Luis Obispo RTA uses `slorta` as a slug.

laneymangan:
@lalver1 no problem at all--I find this to be a particularly confusing situation, so thanks for doing your due diligence to make sure it's all set up correctly!

laneymangan:
> [@lalver1](https://github.com/lalver1) `slotransit` seems like the best slug to me. [@laneymangan](https://github.com/laneymangan) please confirm with Alex.

Just noting here that we confirmed via slack to move forward with slotransit in order to get them online--the agency is fine with this for now, but due to potential changes in the city's branding direction, they _may_ request a change once Alex confirms with a colleague

PR #3587, opened 2026-03-13
Fix: avoid accessing a nonexistent `TransitProcessorConfig`

lalver1:
This PR addresses two small issues we saw when trying to set up the [City of SLO for production validation](https://github.com/cal-itp/benefits/issues/3305):

1. Fixes an issue we saw when trying to set up a new `TransitProcessorConfig`
  When we tried saving a new `TransitProcessorConfig` in the Django Admin, we got 

    >ValueError at /admin/enrollment_littlepay/littlepayconfig/add/
    >
    >'LittlepayConfig' instance needs to have a primary key value before this relationship can be used.

    The problem was that the [`clean()` method of `LittlepayConfig`](https://github.com/cal-itp/benefits/blob/122a8b9cddcafa445dfcfb92dfd16d8072f75ba4/benefits/enrollment_littlepay/models.py#L44) was trying to call `self.transitagency_set` which in turn tried to use the `LittlepayConfig` instance before it was saved to the database, resulting in the error. 

    Since the check in the `clean()` method is only relevant for `LittlepayConfig`s that have been linked to a `TransitAgency`, checking first for `self.pk` ensures that the logic in the `clean()` method is only evaluated for these cases and solves the problem. Similarly for `SwitchioConfig`.

2. Makes `transit_processor_config` optional on `TransitAgency`
  It is convenient to make setting a `transit_processor_config` optional if an agency is not active yet. This way the creation of a `TransitAgency` and a `TransitProcessorConfig` can be done in any order. Note that for an active agency, when using the Admin, the field is still required (see the model's [clean() method](https://github.com/cal-itp/benefits/blob/122a8b9cddcafa445dfcfb92dfd16d8072f75ba4/benefits/core/models/transit.py#L208), "Must fill out configuration for either Littlepay or Switchio.").

### 
With either your local Littlepay or Switchio fixtures and resetting your database (`bin/reset_db.sh`) so the database migrations run:

For 1.
- In the Admin, set up a new `LittlepayConfig` (`SwitchioConfig`) instance and save it
- Confirm that the instance was saved
- In the Admin, create a new inactive `TransitAgency`, select the `LittlepayConfig` (`SwitchioConfig`) instance you just created for it
- Confirm that the `TransitAgency` was saved
- Set `TransitAgency` to active
- Go to the `LittlepayConfig` (`SwitchioConfig`) you created and try saving it with an incomplete configuration
- Confirm that you can't save it until you complete the configuration

For 2.
- Set `TransitAgency` to inactive
- In the Admin, leave `Transit processor config:` blank and save
- Confirm that the `TransitAgency` was saved
- Set `TransitAgency` to active
- Confirm that you can't save it until you complete the configuration

### DIFF
--- a/benefits/enrollment_littlepay/models.py
+++ b/benefits/enrollment_littlepay/models.py
@@ -41,7 +41,7 @@ class LittlepayConfig(TransitProcessorConfig):
     def clean(self):
         field_errors = {}
 
-        if self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
+        if self.pk and self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
             message = "This field is required when this configuration is referenced by an active transit agency."
             needed = dict(audience=self.audience, client_id=self.client_id, client_secret_name=self.client_secret_name)
             field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
--- a/benefits/enrollment_switchio/models.py
+++ b/benefits/enrollment_switchio/models.py
@@ -103,7 +103,7 @@ class SwitchioConfig(TransitProcessorConfig):
     def clean(self):
         field_errors = {}
 
-        if self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
+        if self.pk and self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
             message = "This field is required when this configuration is referenced by an active transit agency."
             needed = dict(
                 tokenization_api_key=self.tokenization_api_key,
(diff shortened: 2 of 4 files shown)


benefits/core/models/transit.py
```diff
@@ -119,6 +119,7 @@ class Meta:
         TransitProcessorConfig,
         on_delete=models.PROTECT,
         null=True,
+        blank=True,
```
jgravois: just to make sure i'm understanding correctly:

1. when a transit agency is created initially, the value of this column is `null`, which is acceptable
1. attempting to remove an associated Transit processor config results in the value being a blank string, which this PR ensures is acceptable too.

jgravois (approved):
this code looks reasonable enough to me and i was definitely able to run through the repro steps locally and confirm that:

1. agencies and transit configs can now be created in either order
1. it is now possible to reset the transit config for an existing (inactive) agency. 🤙

PR #3592, opened 2026-03-16
feat: Add TransitAgencyGroup model

### LINKED ISSUE #3530: Define a new model to group `TransitAgency` instances
Based on research in #3437, we need a way to group `TransitAgency` instances in a lightweight model accessible from the Benefits Administrator.

## Acceptance Criteria

- [x] `TransitAgencyGroup` model exists
- [x] `TransitAgencyGroup` is represented in the Benefits Administrator
- [x] `TransitAgencyGroup` can be managed by superusers and members of the `Cal-ITP` group
- [x] `TransitAgencyGroup` has at least 1 (required) field: `label` (an internal display label for within the Benefits Administrator)
- [x] A direction decision is made and implemented to establish a relationship between `TransitAgency` and `TransitAgencyGroup`
- [x] A new helper is added to the `TransitAgency` model to calculate the set of all distinct `TransitAgency` instances associated with any `TransitAgencyGroup` instances associated with the current `TransitAgency`

### Model definition

From https://github.com/cal-itp/benefits/issues/3437#issuecomment-3813236444:

> The model that relates agencies is basically just a grouping of agencies, so we can call it something like ... `TransitAgencyGroup`.
>
> The model definition would be something like
>
> ```python
> class TransitAgencyGroup(models.Model):
>   id = ...
>   name = models.TextField(...)
>   transit_agencies = models.ManyToManyField(TransitAgency,...)
> ```
>
> Alternatively, you could put the `ManyToMany` field on the `TransitAgency` model.
>
> ```python
> class TransitAgencyGroup(models.Model):
>   id = ...
>   name = models.TextField(...)
>
> 
> class TransitAgency(models.Model):
>   ...
>   transit_agency_group = models.ManyToManyField(TransitAgencyGroup, ...)
> ```
>

### Helper function

Pseudocode:

```python
def group_agencies(self: TransitAgency):
    """The set of all agencies in all groups associated with this agency."""
    agencies = set()
    # update for all groups
    for group in self.transit_agency_group.all():
        agencies.update(group.transit_agencies)
    return agencies
```

Scotchester:
Closes #3530 

---

### Testing instructions

Using the Littlepay fixtures:

1. Create two additional agencies in the admin:
   1. One that shares the same Littlepay config (leave it inactive)
   1. One that uses a Switchio config (will also need to be created)
1. Try creating a group that includes the Switchio agency and see that it fails due to mixing processors
1. Create a group that includes the original CST agency and the new Littlepay agency you created
1. In a devcontainer terminal:
   1. ```sh
      python manage.py shell
      ```
   1. ```py
      cst = TransitAgency.objects.first()
      ```
   1. ```py
      cst.group_agencies()
      ```
      should return an empty list
1. Return to the admin and make the new Littlepay agency active
1. Back in the Django shell:
   1. ```py
      cst.group_agencies()
      ```
      should now return a list with one member, the new Littlepay agency

### TODO:
- [x] Add unit tests
- [ ] Update fixtures?

### DIFF
--- a/tests/pytest/core/models/test_transit.py
+++ b/tests/pytest/core/models/test_transit.py
@@ -44,12 +44,13 @@ class TestCardSchemes:
 
 
 @pytest.mark.django_db
-def test_TransitProcessorConfig_str():
-    label = "cst"
-    transit_processor_config = TransitProcessorConfig.objects.create(environment="dev", label=label)
-    environment_label = Environment(transit_processor_config.environment).label
+class TestTransitProcessorConfig:
+    def test_str(self):
+        label = "cst"
+        transit_processor_config = TransitProcessorConfig.objects.create(environment="dev", label=label)
+        environment_label = Environment(transit_processor_config.environment).label
 
-    assert str(transit_processor_config) == f"({environment_label}) {label}"
+        assert str(transit_processor_config) == f"({environment_label}) {label}"
 
 
 @pytest.mark.django_db
@@ -87,9 +88,7 @@ class TestTransitAgency:
 
         assert result == model_TransitAgency
 
-    def test_by_id_nonmatching(
-        self,
-    ):
+    def test_by_id_nonmatching(self):
         with pytest.raises(TransitAgency.DoesNotExist):
             TransitAgency.by_id(99999)
 
@@ -98,9 +97,7 @@ class TestTransitAgency:
 
         assert result == model_TransitAgency
 
-    def test_by_slug_nonmatching(
-        self,
-    ):
+    def test_by_slug_nonmatching(self):
         result = TransitAgency.by_slug("nope")
 
         assert not result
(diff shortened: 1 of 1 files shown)


thekaveman (changes requested):

benefits/core/models/transit.py
```diff
@@ -125,7 +125,9 @@ class Meta:
     )
 
     def __str__(self):
-        return self.long_name
+        if self.long_name:
+            return self.long_name
+        return self.short_name
```
Scotchester: This was just something I noticed when entering test data by hand for local testing. `long_name` isn't required, so I thought it'd be nice to fall back to `short_name`.

tests/pytest/core/models/test_transit.py
```diff
@@ -266,3 +275,86 @@ def test_TransitAgency_in_person_enrollment_index_route_no_config(model_TransitA
...
+def test_TransitAgency_group_agencies(model_TransitAgency, model_TransitAgencyGroup):
+    agency2 = TransitAgency.objects.create(
+        slug="cst2",
+        short_name="TEST",
+        long_name="Test Transit Agency",
+        info_url="https://example.com/test-agency",
+        phone="800-555-5555",
+        active=True,
+        logo="agencies/cst.png",
+    )
```
Scotchester: This `agency2` definition is repeated in each of my tests. Would it preferable to go ahead with another global fixture, or make a class and add a local fixture?

thekaveman: That's fine if you want to go for the whole file!

Maybe separate commits though? :smile:

PR #3642, opened 2026-03-31
Fix: show in-person transit processor portal button

lalver1:
At some point, the [attribute](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/transit.py#L118) that refers to `TransitProcessorConfig` on `TransitAgency` changed from `transitprocessorconfig` to [`transit_processor_config`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/transit.py#L118). This PR updates the [attribute name in the in-person part of the code that sends the Transit Processor's URL to the template via its context](https://github.com/cal-itp/benefits/blob/main/benefits/admin.py#L60).

We also update the `if` condition to avoid running into a `AttributeError at /admin/ 'NoneType' object has no attribute 'portal_url'` for the case when we load the in-person dashboard for a transit agency that has been set up without a `TransitProcessorConfig` (this setup is only possible for an inactive agency, but it's still possible to reach this state when setting up an agency).

## Reviewing

Using the combined fixtures, log in to in-person using `ast-user` and confirm that you see the control portal button:

image

### DIFF
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -57,8 +57,8 @@ class BenefitsAdminSite(AdminSite):
             if agency is not None:
                 has_permission_for_in_person = agency.customer_service_group in request.user.groups.all()
 
-                if hasattr(agency, "transitprocessorconfig"):
-                    transit_processor_portal_url = agency.transitprocessorconfig.portal_url
+                if agency.transit_processor_config is not None:
+                    transit_processor_portal_url = agency.transit_processor_config.portal_url
                 else:
                     transit_processor_portal_url = ""
 


benefits/admin.py
```diff
@@ -57,8 +57,8 @@ def index(self, request, extra_context=None):
             if agency is not None:
                 has_permission_for_in_person = agency.customer_service_group in request.user.groups.all()
 
-                if hasattr(agency, "transitprocessorconfig"):
-                    transit_processor_portal_url = agency.transitprocessorconfig.portal_url
+                if agency.transit_processor_config is not None:
+                    transit_processor_portal_url = agency.transit_processor_config.portal_url
                 else:
                     transit_processor_portal_url = ""
```
Scotchester, line 63: This works fine, but the template already checks for a truthy value of the context variable `transit_processor_portal_url`, so we could just remove this whole thing and directly set the context var to `agency.transitprocessorconfig.portal_url` below.

Scotchester (approved):
Tested and approved, with one suggestion if you'd like to simplify slightly.

benefits/admin.py
lalver1: Thanks, Scott! Yep, the template checks for a truthy value but if we create a transit agency with an empty `transit_processor_config` and save it (it has to be set as an inactive agency, but this scenario is allowed), we still need this check in the Python code to avoid getting a `'NoneType' object has no attribute 'portal_url'` before the template can check the truthiness of `transit_processor_portal_url`. I tested implementing the suggested change but ran into that error.

Scotchester: Ahhh, right, I was conflating the config and its `portal_url` field in my head.

lalver1: Thanks for looking at it though and the suggestion, it would've been nice to simplify it some more. I'll go ahead and merge this!

Issue #2153, opened 2024-06-10
Refactor how data is persisted in Azure


thekaveman:
Recently, we have been seeing errors in `dev` related to attempting sign in to the Admin and the database being in a locked state:

```log
OperationalError: database is locked
```

* https://sentry.calitp.org/organizations/sentry/issues/75387
* https://sentry.calitp.org/organizations/sentry/issues/75388

While we have only seen this in `dev` so far, it seems inevitable to come up in `test` and/or `prod`, where it will really be a problem if we can't access the Admin or if the app is locked for user requests.

The root of the issue is due to the changes from #1773 to persist the (Sqlite) database file in a mounted Azure file share. This _mostly_ works for reads; but now that we have the potential for more writes via the Admin, it is starting to break down (as expected, see e.g. [this SO post](https://stackoverflow.com/questions/788517/sqlite-over-a-network-share)).

The idea is to reuse as much of the existing infrastructure and process as possible, with two changes:

* Azure Web App storage is turned on, to enable Azure to persist the `/home` directory
* The database is moved under `/home` to a non-file share directory

This should solve the Sqlite-over-a-network issue, while allowing us to maintain data configured via the Admin across restarts/deployments.

## Acceptance Criteria

- [x] The setting `WEBSITES_ENABLE_APP_SERVICE_STORAGE=true`
- [x] The database file sits within the app container in the `/home` directory, not on a network file share

## Plan

- [x] Make a backup of the `django.db`
- [x] https://github.com/cal-itp/docker-python-web/issues/42
- [x] Move the code out of the `/home` directory to work around the issue above #2156 
- [x] https://github.com/cal-itp/docker-python-web/issues/43
- [x] Update `DJANGO_DB_DIR` secret to a location under `/home`
- [x] Copy `django.db` into the location
- [x] Turn on `WEBSITES_ENABLE_APP_SERVICE_STORAGE=true` and restart the app
- [x] Verify everything starts up and the flows work as expected

## Additional context

This FAQ article about Azure Web App Containers has been very informative: https://techcommunity.microsoft.com/t5/apps-on-azure-blog/things-you-should-know-web-apps-and-linux/ba-p/392472

We can also think about dumping fixture data on some frequency into the network file share for a backup, following the commands from the notes in https://github.com/cal-itp/benefits/issues/1856#issue-2086716269

[...]

PR #3646, opened 2026-04-01
Feat: Azure Container Apps infrastructure

### LINKED ISSUE #3636: Migrate app from App Service to Container Apps for `dev` environment (Infra)
Terraform configuration for migrating Benefits from an App Service to a Container App. For this ticket, `dev` Benefits is running side by side as an App Service and Container App to ensure that the Container App works as expected. The actual cut over to running `dev` Benefits exclusively as a Container App is covered in #3637.

## Acceptance Criteria

- [x] App Service & Container App side-by-side execution
- [x] New Postgres management container deployed in the same Container Apps Environment as the Benefits application

lalver1:
Closes #3636

This PR sets up the infrastructure for a side-by-side migration of the Benefits app from an App Service to a Container App. This approach will allow us to validate network integration, database connectivity, Key Vault access, and outbound traffic routing using the Container App's default URL before routing live traffic to it. It also adds a Database Management Container to allow us to test database connectivity more conveniently and to get us ready for the migration to PostgreSQL in #3527. 

### Container Apps

- Uses the [Consumption plan](https://learn.microsoft.com/en-us/azure/container-apps/plans#consumption) type
- Environment
  - Uses the [Workload profile](https://learn.microsoft.com/en-us/azure/container-apps/environment#types)
  - Since the App Service can be reached from the Internet ([`public_network_access_enabled = false` is not explicitly set](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#public_network_access_enabled-1) and [`ip_restriction_default_action = "Allow"`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#ip_restriction_default_action-1)), to match it, [set `internal_load_balancer_enabled = false`](https://learn.microsoft.com/en-us/azure/container-apps/networking#accessibility-level).
  - Used by both the Benefits app and the pgAdmin container
  - Uses the same Log Analytics Workspace as the App Service (since the App Service will be retired anyway)
  - Ignores changes to `infrastructure_resource_group_name` via the `lifecycle` argument. This is necessary because by default, [Azure manages this and may change it. If we don't ignore it, changes that Azure makes will make Terraform create the environment again and therefore replace the existing one.](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/container_app_environment#infrastructure_resource_group_name-1)
- Benefits App
  - [Single revision mode](https://learn.microsoft.com/en-us/azure/container-apps/revisions#revision-modes) since it covers our use case (only one version at a time running from the latest image in GHCR)
  - Since the App Service can be reached from the Internet ([`public_network_access_enabled = false` is not explicitly set](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#public_network_access_enabled-1) and [`ip_restriction_default_action = "Allow"`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/linux_web_app#ip_restriction_default_action-1)), to match it, [set `external_enabled = true`](https://learn.microsoft.com/en-us/azure/container-apps/networking#public-network-access).
  - [Ingress](https://learn.microsoft.com/en-us/azure/container-apps/ingress-how-to?pivots=azure-cli#ingress-settings) set to allow external connections, but only `HTTPS` (`allow_insecure_connections = false`), like the App Service setup.
  - [Azure Files volume](https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts?tabs=smb&pivots=azure-cli#azure-files) for storing agency logos
- Database Management Container
  - Based on the [pgAdmin Docker image](https://www.pgadmin.org/docs/pgadmin4/9.14/container_deployment.html) which includes [`pg_dump`, `pg_dumpall`, `pg_restore` and `psql`](https://www.pgadmin.org/docs/pgadmin4/9.14/container_deployment.html#postgresql-utilities) and a web GUI for quickly browsing the database.
  - [Scaling down to zero during periods of inactivity, or cool down period](https://learn.microsoft.com/en-us/azure/container-apps/scale-app?pivots=azure-cli#scale-behavior). This creates a [delay known as cold-start time](https://learn.microsoft.com/en-us/azure/container-apps/cold-start) but it is not an issue since this container will seldomly be used. Set the cool down period to 3600 seconds (1 hour) so that if we are using the `psql` CLI (the CLI generates no HTTP requests to keep the container alive) we'll have a window of 1hr before we need to send an HTTP request to the container.
  - Set up [IP ingress restrictions](https://learn.microsoft.com/en-us/azure/container-apps/ip-restrictions?pivots=azure-portal) to block traffic from all IPs to the pgAdmin Container App until the Azure Portal's user adds their IP address to the allowed list. To implement denying all IPs, but still allowing us to add our local IP in the Azure Portal when we are ready to interact with the container, we set a restriction to Allow only the dummy IP `127.0.0.1/32`, also known as localhost and which is not a routable address on the Internet. If we were to use Deny, we would only be able to set a range of IPs to deny, which is not what we need.
  - For convenience, sets environment variables for [pgAdmin](https://www.pgadmin.org/docs/pgadmin4/9.14/container_deployment.html#environment-variables) (`PGADMIN_DEFAULT_EMAIL`, `PGADMIN_DEFAULT_PASSWORD`, `PGADMIN_CONFIG_SERVER_MODE`) and [`psql`](https://www.postgresql.org/docs/current/libpq-envars.html) (`PGHOST`, `PGUSER`, `PGDATABASE`, `PGSSLMODE`, `PGPASSWORD`).

### Networking

- Uses existing `VNET-CDT-PUB-VIP-CALITP-D-001` virtual network (256 addresses - `10.0.0.0` to `10.0.0.255`)
  - `ACAPP` subnet `10.0.0.128/26` (64 addresses - `10.0.0.128` to `10.0.0.191`)
    - New [subnet](https://learn.microsoft.com/en-us/azure/container-apps/custom-virtual-networks?tabs=workload-profiles-env#subnet) used for Benefits Container App. Minimum subnet size is `/27`. 50 (64-14=50) available addresses given that Container Apps reserves 14 IP addresses. Since we use the Consumption workload profile, we plan for 1 IP address per 10 replicas, which would give us 500 replicas. But since the required address space is doubled for a short period of time in order to support zero downtime deployments, the maximum is 250 replicas.
    - [Delegated to `Microsoft.App/environments`](https://learn.microsoft.com/en-us/azure/container-apps/custom-virtual-networks?tabs=workload-profiles-env#subnet)
    - [Default outbound access disabled](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access?tabs=portal#why-is-disabling-default-outbound-access-recommended) (`default_outbound_access_enabled = false`)
    - [NAT Gateway](https://learn.microsoft.com/en-us/azure/container-apps/custom-virtual-networks?tabs=workload-profiles-env#nat-gateway-integration) association for the Container App's outbound connectivity. Note that the [single existing NAT Gateway currently used by the App Service can also be used by the Container App's subnet](https://learn.microsoft.com/en-us/azure/nat-gateway/nat-gateway-resource#nat-gateway-architecture).
- Sets up / updates the [Network Security Goups (NSG)](https://learn.microsoft.com/en-us/azure/virtual-network/network-security-groups-overview) listed below
  - (`ACAPP` Subnet) Container App -> Internet: Allowed on port 443 only (`AllowHTTPSOutbound`). Uses the existing `NSG-CDT-PUB-VIP-CALITP-D-APP` network security group created originally only for the App Service.
  - (`DB` Subnet) Container App -> Database: Allowed (`AllowPostgresInboundFromContainerApp`) via new security rule. Regarding `priority`, the security group first checks if the traffic is from the `APP` and then if it's from the `ACAPP` subnet. It denies all inbound if the traffic is coming from anywhere else.
  
### Terraform

- Upgraded Terraform's `azurerm` provider to version 4.67 since the previous version (4.51.0) in the lock file did not support the `cooldown_period_in_seconds` argument for `azurerm_container_app`. This argument is essential for the Database Management Container to work properly and not time-out prematurely.

### Azure DevOps Pipeline

- Passes the image tag (registry URL and commit SHA) to Terraform deploy so that it can be used by the `container`'s `image` property in the Terraform code.

### Local environment

- This PR adds a `pgdmin` service to test out commands and the GUI locally before using it for management tasks on the Azure Postgres server. ~If we decide this is useful, we may want to remove `pgweb` since it would become redundant. We would revert to using the more standard port `8081` for pgAdmin also.~
  
## Reviewing

For the local component of this PR we can check the following:

- Start the pgAdmin container (from outside of the dev container, if you have it running) with `docker compose up -d pgadmin`
- Enter the pgAdmin GUI by navigating to `http://localhost:8081/browser/`
- Fill in "Add New Server" to connect to with:
  - General
    - Name: `postgres`
  - Connection
    - Host name/address: `postgres`
    - Port: `5432`
    - Maintenance database: `postgres`
    - Username: `postgres`
    - Password: `postgres`
    - Save Password? `Yes`
- Confirm that you can see the `postgres` and `django` databases and tables in the pgAdmin GUI
- Confirm that you have a new Docker volume at `benefits_pgadmin_data`, stop the pgAdmin container, restart it and confirm that your connection settings were saved so that you don't need to "Add New Server" again. This should also hold if you delete and recreate the container.
- Inside the running pgAdmin container, run `export PATH=$PATH:/usr/local/pgsql-17`. This step is necessary because the [`pgAdmin` image has different versions of the Postgres utilities installed](https://www.pgadmin.org/docs/pgadmin4/9.14/container_deployment.html#postgresql-utilities).
- Ensure that you can run `psql --version` and that you see `psql (PostgreSQL) 17.9`
- Run `psql -d django` to ensure you can connect to the `django` database. Run `\dt;` inside `psql` and confirm that you can see all the tables used by Benefits.


lalver1:
Putting this back to draft to make a few changes:

- Allow the App Service to continue using SQLite and the Container App to switch to using Postgres. This will require manually setting `USE_POSTGRES=true` only for the Container App, independently of the value of `use-postgres` in the Key Vault.
- Taking a closer look at the values for `django-trusted-origins`, `django-allowed-hosts`, and `google-sso-allowable-domains`.

Scotchester:
Just doing a bit of pre-reading on this before the PRPR and wanted to ask about this:

> [Azure Files volume](https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts?tabs=smb&pivots=azure-cli#azure-files) for storing agency logos

I had been assuming we'd use Azure Blob Storage, but if I understand how you've set this up, it seems like Azure Files might work with no Django settings updates simply by mounting the volume in the same location as our previous data?? That's awesome!

lalver1:
That's right, @Scotchester! Mounting the volume as a File share makes it easier to work with since we just mount it in the same location where we've been storing our data, there's no need to access a Blob through HTTP.

Also, for images, text, etc. a File share works well, but we found out that for storing a SQLite database, it isn't recommended. Since a File share is like a network drive, [writing to a SQLite file over a network produces file lock problems and other issues](https://github.com/cal-itp/benefits/issues/2153). This is one of the reasons why we are sort of forced to deploy Container Apps + Postgres and can't do an intermediate setup of Container Apps + SQLite.

terraform/modules/application/app_benefits.tf
```diff
@@ -0,0 +1,86 @@
+# The Benefits Container App
+resource "azurerm_container_app" "benefits" {
+  name                         = "ca-cdt-pub-vip-calitp-${lower(var.env_letter)}-001"
```
thekaveman: Maybe `aca-cdt-pub...` for consistency

terraform/modules/application/app_pgadmin.tf
```diff
@@ -0,0 +1,109 @@
+# The pgAdmin Container App
+resource "azurerm_container_app" "pgadmin" {
+  name                         = "ca-cdt-pub-vip-calitp-${lower(var.env_letter)}-pgadmin"
```
thekaveman, line 3: `aca-`

terraform/network.tf
```diff
@@ -13,6 +13,12 @@ locals {
       actions           = ["Microsoft.Network/virtualNetworks/subnets/join/action"]
       service_endpoints = ["Microsoft.Storage"]
     }
+    "ACAPP" = {
```
thekaveman: What do you think about just `ACA` here?

terraform/network.tf
```diff
@@ -108,6 +114,12 @@ resource "azurerm_subnet_nat_gateway_association" "app" {
   nat_gateway_id = azurerm_nat_gateway.main.id
 }
 
+# Associate NAT Gateway with the ACAPP Subnet
+resource "azurerm_subnet_nat_gateway_association" "acapp" {
```
thekaveman: `aca`?

terraform/network.tf
```diff
@@ -138,13 +150,18 @@ resource "azurerm_subnet_network_security_group_association" "app" {
   network_security_group_id = azurerm_network_security_group.app.id
 }
 
+resource "azurerm_subnet_network_security_group_association" "acapp" {
```
thekaveman: `aca`?

thekaveman (commented):
A bunch of naming nits incoming:

terraform/network.tf
lalver1: Thanks Kegan, I agree with these naming changes. I also included two more changes in 3b2cb4dd976ab9636cbf4e755015388cf8cac9c8 that hopefully will make things more clear:

- Changed the Benefits Container App's name from `ca-cdt-pub-vip-calitp-${lower(var.env_letter)}-001` to `aca-cdt-pub-vip-calitp-${lower(var.env_letter)}-benefits`
- Changed the `ACA` subnet id variable from `infrastructure_subnet_id` to `subnet_aca_id`

The Benefits Container App change more closely follows [Azure's naming conventions](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming#choose-naming-components) since it shouldn't use instance-based naming but instead use application-based naming. I had forgotten about these docs and I'm glad Angela brought them up yesterday.

Scotchester: If we want to stick to the conventions in that doc (or rather, the abbreviations reference it links to), shouldn't we be using `ca` instead of `aca`?

lalver1: Ah, turns out `aca-cdt-pub-vip-calitp-${lower(var.env_letter)}-benefits` is one character too long (the limit for a Container App name is 32 characters) so the pipeline failed.

~What do you think about removing `vip` from the naming convention for both Container Apps? This corresponds to `Project Category` but I'm wondering if `cdt-pub-calitp` is enough? Or I could go back to using `ca-cdt-pub-vip-calitp-${lower(var.env_letter)}-001` too (or even using `aca-cdt-pub-vip-calitp-${lower(var.env_letter)}-bnfts`) but I wanted to avoid that if possible.~

thekaveman: I like having `-benefits` in the name, it is helpful when you have multiple container apps running. 

I'm fine with removing `vip`, that seems extra, was never sure why they added that.

lalver1: Thanks Scott, sorry, I didn't see your comment until after I refreshed the page just now. 

Right, we could go back to using [`ca`](https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations#containers) since that actually is Azure's recommended abbreviation. This would also solve the extra character issue in the name of the Benefits Container App I mentioned above. And for consistency, I could change the name of the subnet to `CA`. Actually, this might be nice since we'll end up with `DB` and `CA` subnets, both being 2 letter abbreviations.

thekaveman: Or we could take the tactic from CDRC and use `-web` or `-app`.

thekaveman: I'm good with whatever you guys land on. I'll just advocate for consistency, and ease of identification. E.g. in logs and elsewhere, this name shows up, so it is helpful to be able to easily filter by some pre or postfix.

lalver1: Yep, this makes sense, I'll use `ca` and `web` (like CDRC). I like this better, since Benefits is the whole application, and this particular container is only running the `web` component of it.

lalver1: Ah, I overlooked to change the name of the Terraform resource from `benefits` to `web` and the file name too for consistency. Will do this shortly.

lalver1:
## Notes for Terraform `apply` from local

### `web` Container App
- On our first local `apply` we saw an issue that the Container App did not associate any of the "secret" environment variables with the key vault secret they corresponded to. This may have been a case of the Key Vault Policy / Container App Identity not being ready when it was deployed. Running `apply` a second time fixed this issue, the environment variables now point to the correct Key Vault secret.
- I forgot to set `django-db-name`, `django-db-user` in the Key Vault before deploying so starting the Container App failed. So I set them and the Container App [automatically re-read them and set the corresponding environment variables (within 30 minutes of a new secret version, Container Apps will fetch the new value)](https://learn.microsoft.com/en-us/azure/container-apps/manage-secrets?tabs=azure-portal#reference-secret-from-key-vault).
- I couldn't reach the Container App form its URL. Per the error message, I added the Container App's `Application Url` to `django-allowed-hosts`, and I was able to reach the app. I also updated `django-trusted-origins` since we'll need it updated anyway. Note that `django-trusted-origins` requires `https://` in the URL, while `django-allowed-hosts` doesn't. Restarted the app manually so I didn't have to wait for the automatic secret value update.
- Added a [custom HTTP liveness probe](https://learn.microsoft.com/en-us/azure/container-apps/health-probes?tabs=arm-template) using [Benefits' `Healthcheck` middleware](https://github.com/cal-itp/benefits/blob/main/benefits/core/middleware.py#L60) to check if the application is still running and responsive.

### `pgAdmin` Container App

- Similar to the first item above, on our first local `apply` we saw an issue that the Container App did not associate any of the "secret" environment variables with the key vault secret they corresponded to. This may have been a case of the Key Vault Policy / Container App Identity not being ready when it was deployed. ~Restarting the Container App fixed the issue, the environment variables now point to the correct Key Vault secret.~ <- Not sure this is correct, I probably meant "Running `apply` a second time fixed this issue" since this is what I am seeing in `test` while doing release `2026.04.1-rc1`.
- Remember that the revision scales to zero after no activity, so we need to add our IP to the allowed IPs ingress list, and make a request to it to "wake it up". If we don't add our IP we'll get an `RBAC: access denied` error when we try to reach its URL from the browser.
- The container failed when starting up with the following showing up in the Application Logs which suggested changing the [target port settings of the Container App](https://learn.microsoft.com/en-us/azure/container-apps/troubleshoot-target-port-settings) to `5050` and [setting the `PGADMIN_LISTEN_PORT` environment variable to `5050`](https://www.pgadmin.org/docs/pgadmin4/latest/container_deployment.html#mapped-files-and-directories).
  > sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
  > sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
  > [INFO] Starting gunicorn 23.0.0
  > [ERROR] connection to ('::', 80) failed: [Errno 13] Permission denied
  > [ERROR] Can't connect to ('::', 80)
- pgAdmin stores its configuration settings in a SQLite database by default. The SQLite database was being saved in the pgAdmin Container App's fileshare. Since SQLite is not supported over fileshares, we changed the pgAdmin configuration to use the Azure Postgres server itself to store the configuration settings in a database called `pgadmin_config`. Just noting that [the URL needs to be enclosed in single quotes since it gets read by a Python script at some point in pgAdmin's startup](https://www.pgadmin.org/docs/pgadmin4/latest/config_py.html#the-config-py-file).
- The pgAdmin image we are using does not have a certificate at `/home/pgadmin/.postgresql/root.crt` but sqlalchemy (used during pgAdmin's startup) expects it to be there by default. We set the [environment variable `PGSSLROOTCERT` to `system`](https://www.postgresql.org/docs/current/libpq-envars.html) to tell sqlaclhemy to use the system-provided certs instead.
- Finally, one of pgAdmin's [startup scripts also expects the environment variables `PGADMIN_SETUP_EMAIL` and `PGADMIN_SETUP_PASSWORD`.](https://github.com/pgadmin-org/pgadmin4/blob/master/web/pgadmin/setup/user_info.py#L34)

jgravois (dismissed):
i verified that the new container app is accessible in the browser (with no agencies configured) and that the pgAdmin container is not.

nice work @lalver1. i learned a lot riding along. 🚒 

makes sense to me to merge and confirm that the new SHA deployment succeeds all on its own if no one else has any concerns.

Scotchester:
I added my IP to the allowlist for pgAdmin and the RBAC error went away, but now I'm getting "stream timeout" after a long wait.  Any idea what else I might need to do?

Should we update any docs before merging this? Might be fine to wait until we are fully deployed in all envs and have any more wrinkles ironed out, though.

lalver1:
Thanks for the reviews!

@jgravois, I still need to add one more commit to change how pgAdmin's External Database Settings are configured, and once that is finalized I'll also need to make some updates to the docs. So I'll request another review soon 🙏 

@Scotchester that's good news that you can at least reach pgAdmin. The stream timeout is because pgAdmin is trying to use the SQLite database to read its settings and accessing it over a file share is problematic. I should have a fix for this soon though. I was also feeling the same way about the docs, I'll try to get some updates in this PR and hopefully that will cover most of the other things we may see in the other environments.

thekaveman:
FWIW I think the docs updates can wait until we have this more ironed out. Up to y'all.

.devcontainer/devcontainer.json
```diff
@@ -3,7 +3,7 @@
   "name": "cal-itp/benefits",
   "dockerComposeFile": ["../compose.yml"],
   "service": "dev",
-  "runServices": ["dev", "docs", "server", "postgres", "pgweb"],
+  "runServices": ["dev", "docs", "server", "postgres", "pgweb", "pgadmin"],
```
thekaveman: Maybe we can dump `pgweb` since we've got `pgadmin` going now?

lalver1: Yep! I was also thinking the same thing 👍

lalver1:
Hi everyone, I think this one is ready for a re-review. The major changes I made late last week were doing some cleanups, removing `pgweb` from our local setup, and adding a custom HTTP liveness probe to the Container App using our `Healthcheck` middleware. I've also deployed these changes from `local` so once this PR is merged, the only Terraform change we'll see will be related to the `web` container image, thanks!

terraform/variables.tf
```diff
@@ -14,3 +14,20 @@ variable "ENGINEERING_GROUP_OBJECT_ID" {
...
+
+variable "CONTAINER_REPOSITORY" {
+  description = "The repository path within the registry."
+  type        = string
+  default     = "cal-itp/benefits"
+}
+
+variable "CONTAINER_TAG" {
+  type        = string
+  description = "The specific tag of the image to deploy (e.g., 'main')."
```
jgravois: would a SHA or a calendar version be more appropriate to use as an example here?

compose.yml
```diff
@@ -67,23 +67,30 @@ services:
...
       postgres:
         condition: service_healthy
-    healthcheck:
-      test: ["CMD", "nc", "-vz", "127.0.0.1", "8081"]
-      interval: 5s
     environment:
-      - PGWEB_DATABASE_URL=postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOSTNAME}:${POSTGRES_PORT}/${DJANGO_DB_NAME}?sslmode=disable
+      - PGADMIN_DEFAULT_EMAIL=benefits-admin@calitp.org
+      - PGADMIN_DEFAULT_PASSWORD=admin
+      - PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED=False
```
jgravois: just an fyi that @angela-tran did a little refactoring recently to standardize on `false` instead of the pythonic variant in #3598

compose.yml
```diff
@@ -67,23 +67,30 @@ services:
...
       interval: 5s
 
-  pgweb:
-    container_name: pgweb
-    image: sosedoff/pgweb
+  pgadmin:
+    image: dpage/pgadmin4:latest
     ports:
-      - "${PGWEB_PORT:-8081}:8081"
+      - "${PGADMIN_PORT:-8081}:80"
```
jgravois, line 76: after rebuilding the container i'm not seeing anything running on http://localhost:8081 or http://localhost

am i missing something simple?

docs/reference/environment-variables.md
```diff
@@ -351,17 +351,13 @@ Comma-separated list of URIs. Configures the [`style-src`][mdn-csp-style-src] Co
 
     The official PostgreSQL Docker image has more on available [`postgres` environment variables][postgres]
 
-### `PGWEB_PORT`
-
-The port to serve a (local only) [`pgweb`](https://sosedoff.github.io/pgweb/) service on.
```
jgravois, line 356: should we be adding `PGADMIN_PORT` in place of this variable?

Scotchester: There are several other `PGADMIN_` vars, as well, though we discussed waiting on _adding_ to the docs until later.

jgravois: ah, thanks for the reminder. i'm all for it. 🤙

Scotchester (dismissed):
All the recent changes look good to me!

compose.yml
Scotchester: I got there (on http://localhost:8081/), though it took a couple minutes for it to finish starting up before the page would load.

However, it doesn't seem to be connecting to anything out of the box. Revisiting your testing instructions, seems that's expected, but surely we can configure it to auto-connect? (In the future – let's not hold up this PR for that.)

lalver1: Oh that's weird, this is what I see on my end using Podman

image

We can pair if you want to share your screen just to make sure I'm not missing anything either.

jgravois: thanks y'all. i was having some trouble initially with a conflict with the stale pgweb container. now that i've shut that down and RTFM, i was able to get connected locally too. 🚀

jgravois (approved):
🌈🌈

Scotchester:
> * Inside the running pgAdmin container, run `export PATH=$PATH:/usr/local/pgsql-16`. This step is necessary because the [`pgAdmin` image has different versions of the Postgres utilities installed](https://www.pgadmin.org/docs/pgadmin4/9.14/container_deployment.html#postgresql-utilities).
> * Ensure that you can run `psql --version` and that you see `psql (PostgreSQL) 16.13`
> * Run `psql -d django` to ensure you can connect to the `django` database. Run `\dt;` inside `psql` and confirm that you can see all the tables used by Benefits.

Not sure how important this is, but I did run into some issues with the above testing steps. Here's my output:

```
root@4e1a0fcdaab7:/# export PATH=$PATH:/usr/local/pgsql-16
root@4e1a0fcdaab7:/# psql --version
psql (PostgreSQL) 17.9 (Debian 17.9-1.pgdg13+1)
root@4e1a0fcdaab7:/# psql -d django
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: FATAL:  role "root" does not exist
```

jgravois:
> Not sure how important this is, but I did run into some issues with the above testing steps.

strange, that bit worked a treat on my box.

```
a5cb1a4f68c8:/pgadmin4$ export PATH=$PATH:/usr/local/pgsql-16
a5cb1a4f68c8:/pgadmin4$ psql --version
psql (PostgreSQL) 16.13
a5cb1a4f68c8:/pgadmin4$ psql -d django
psql (16.13, server 17.9 (Debian 17.9-1.pgdg13+1))
```

don't we want 17 like @Scotchester is seeing and not 16 though??

lalver1:
> don't we want 17 like @Scotchester is seeing and not 16 though??

Thanks for catching this John! I updated the original comment to reflect this, we do want version 17 😅

Scott, I think that you're seeing that message because maybe your Postgres database hasn't been set up yet? I'll confirm but my hunch is that the `ensure_db` command hasn't ran on the current Postgres container you have running now.

Scotchester:
> Scott, I think that you're seeing that message because maybe your Postgres database hasn't been set up yet? I'll confirm but my hunch is that the `ensure_db` command hasn't ran on the current Postgres container you have running now.

No, it's set up, I see the data in the web interface 🤔 

@jgravois Your prompt looks different than mine. How did you get into the pgadmin container? I was using the Terminal tab on the container in Podman Desktop.

lalver1:
Huh 🤔, let me do a little more digging then. I'm also using the `Terminal` tab in Podman Desktop.

EDIT: Oh I see you are logged in as `root` in your terminal. In my Terminal I am logged in as `pgadmin`, I think that's the reason for that error. I wonder why on your end it didn't log you in as `pgadmin` 🤔

Scotchester:
NEVERMIND. I was in the `postgres` container, not the `pgadmin` container 🤦

Scotchester (approved):

lalver1:
Ah, well I'm glad it wasn't something else! I have to say it's been a little more difficult keeping up with database users/pgAdmin users conceptually and with env variable naming but as we keep using the new database setup I think we'll get more comfortable with it.

lalver1:
Our [CI worked without any issues](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build/results?buildId=138115&view=logs&j=beac6bc9-9996-5711-f557-a14ba57d62ca&t=0649047d-26fc-5cb1-fb55-fbcdbe455e11) and we now have our Container App using e8a5fe07563aac07ad0b75de7926c6cd3db4841a 🎉 Thanks for all the reviews!

PR #3652, opened 2026-04-08
chore: Run fixture reminder workflow when any app has migration changes

### LINKED ISSUE #3651: Update fixture update reminder workflow to consider apps besides core
For some reason, when [creating the fixture update reminder workflow](https://github.com/cal-itp/benefits/pull/3473) I didn't have it run on updates to the models in apps other than `core`.

We should update this to run when models are updated in `enrollment_littlepay` and `enrollment_switchio`, as well.

## Acceptance Criteria

- [ ] PRs that only update models in `enrollment_littlepay` or `enrollment_switchio` also trigger the fixture update reminder workflow

## Additional context

Noticed because the workflow didn't run on #3650.

Scotchester:
Fixes #3651

### DIFF
--- a/.github/workflows/fixture-update-reminder.yml
+++ b/.github/workflows/fixture-update-reminder.yml
@@ -3,7 +3,7 @@ name: Fixture update reminder
 on:
   pull_request:
     paths:
-      - 'benefits/core/migrations/**'
+      - 'benefits/**/migrations/**'
 
 jobs:
   # Check which files have changed
--- a/.github/workflows/fixture-update-reminder.yml
+++ b/.github/workflows/fixture-update-reminder.yml
@@ -3,7 +3,7 @@ name: Fixture update reminder
 on:
   pull_request:
     paths:
-      - 'benefits/**/migrations/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/fixture-update-reminder.yml
```diff
@@ -3,7 +3,7 @@ name: Fixture update reminder
 on:
   pull_request:
     paths:
-      - 'benefits/core/migrations/**'
+      - 'benefits/**/migrations/**'
```
jgravois: i could be missing something obvious about the way Django apps are nested, but i wouldn't expect to have to traverse into _subfolders_ of core's siblings looking for migrations.

```suggestion
      - 'benefits/*/migrations/**'
```

jgravois (commented):
CIL for a question disguised as a suggestion.

.github/workflows/fixture-update-reminder.yml
Scotchester: You're right! Nice catch.

jgravois (approved):
thanks for humoring me. 🤗

PR #3653, opened 2026-04-09
fix: Update subfilters in fixture reminder workflow

Scotchester:
Follow up to https://github.com/cal-itp/benefits/issues/3651#issuecomment-4216482017


jgravois (approved):
if at first you don't succeed... 😅

PR #3571, opened 2026-03-10
Refactor: rename DIGITAL to SELF_SERVICE

thekaveman:
Django/Benefits part of #3542 

**Open question:** a733939fa5bc15bdbf99c620b478e45bb2087838 adds a data migration in the same transaction for updating the formerly `DIGITAL` values to `SELF_SERVICE`. Is this the right place to accomplish this goal? Ref: https://github.com/cal-itp/benefits/pull/3546#discussion_r2884896948

## Reviewing

_In the devcontainer_

1. Checkout `main` and reset your DB with the built-in `local_fixtures.json`
2. Observe the references to `digital` in the SQLite viewer for `EnrollmentEvent` and `EnrollmentFlow` tables
3. Checkout this branch and run `bin/init.sh`
4. Observe the references have updated to `self_service` in the SQLite viewer for `EnrollmentEvent` and `EnrollmentFlow` tables

PR #3655, opened 2026-04-13
Refactor: rename DIGITAL to SELF_SERVICE

### LINKED ISSUE #3542: Rename "digital" enrollment method to "self-service"
We recently had the epiphany that "self-service" is a much better descriptor than "digital" for the enrollment method where users independently go through the Benefits front end to enroll their card (because in-person enrollment led by agency staff is also digital).

Let's make that change in the back-end labeling.

## Acceptance Criteria

- [ ] All references to the "digital" enrollment method now call it "self-service"

jgravois:
Django/Benefits part of #3542 (supercedes #3571)

79842864cab3c27367a07a8efe4889b9998cd81d adds a data migration in the same transaction for updating the formerly `DIGITAL` values to `SELF_SERVICE`.

## Reviewing

_In the devcontainer_

1. Checkout `main` and reset your DB with the built-in `local_fixtures.json`
2. Observe the references to `digital` in your Postgres viewer of choice for `EnrollmentEvent` and `EnrollmentFlow` tables
3. Checkout this branch and run `bin/setup.sh`
4. Observe the references have updated to `self_service` in the db viewer for `EnrollmentEvent` and `EnrollmentFlow` tables

locally i tested both an self_service and in_person flow and ensured that updating the active enrollment methods for a specific EnrollmentFlow were reflected appropriately in the app.


lalver1:
This looks great @jgravois! I only have a comment related to the database migration, given https://github.com/cal-itp/benefits/pull/3571#issue-4054345157 and https://github.com/cal-itp/benefits/pull/3546#discussion_r2884896948 and how we have a schema and data migration in the same file.

Since we have a schema and a data migration, I tested locally using Postgres and the dev container by following these steps

- On `main`, start with a clean database using the the combo fixtures from LastPass and go through an enrollment
- Switch to this PR's branch, run the `setup.sh` script and force it to not reset the database, but to run the database migrations (we may need to look at this some more to make local testing easier going forward, but definitely out of scope for this PR)
- Verify that the data got migrated
- Go through another enrollment to ensure the new schema is being used

to make sure I didn't get any errors. My guess is that I didn't see any `OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.` errors because my local tables are still very small (even for prod our database tables are very small), so the DDL part (schema changes) runs fast enough that it doesn't create an issue for the DML part (data migration). I guess I could also test with SQLite, since that is what we are still using in our infra, but my guess is that we will see the same behavior, no `database locked` messages due to the small size of the database.

The recommendation in [Django's docs](https://docs.djangoproject.com/en/6.0/ref/migration-operations/#runpython) would be to have 2 files, one with the schema migration and another with the data migration. But for this PR maybe this isn't necessary given our current database size 🤔.

thekaveman:
Thanks @lalver1, just noting we also discussed this in the original PR: https://github.com/cal-itp/benefits/pull/3571#issuecomment-4077656018 and concluded we were fine with a single migration file.

lalver1 (approved):
Oh I missed that comment, thanks for noting it @thekaveman! 

I had no other comments @jgravois, this looks great! I guess the only other thing will be to coordinate this PR with https://github.com/cal-itp/data-infra/pull/4904 but I'll approve this so we are ready to go.

PR #3659, opened 2026-04-16
Centralize SwitchioConfig fields that are not unique to each agency

### LINKED ISSUE #3560: consider centralizing redundant `SwitchioConfig` properties
Currently each agency's `SwitchioConfig` declares several pieces of information that are common across agencies:

* Enrollment api authorization header
* Client certificate
* Ca certificate
* Private key

It would be ideal to centralize this information. ~~Perhaps in a new model called `SwitchioEnvConfig` associated with each `Environment` rather than agencies themselves?~~ _see discussion below_

If that makes sense, we might also consider surfacing the two props below in the very same model.

* switchio-{env}-enrollment-api-base-url
* switchio-{env}-tokenization-api-base-url

~~based on my cursory reading of the code, it looks more like they ended up secrets in the interest of making it convenient to switch between environments, not because they actually store something sensitive.~~ _see discussion below_

## Acceptance Criteria

- [ ] The `SwitchioConfig` model no longer has these fields that must be entered into each instance in the Admin:
  - `enrollment_api_authorization_header`
  - `client_certificate`
  - `ca_certificate`
  - `private_key`
- [ ] Values that were used in the above fields are stored in environment variables (secrets in Azure Key Vault for the cloud environments)
- [ ] The existing `client_certificate_data`, `ca_certificate_data`, and `private_key_data` properties are updated to get the env var directly, rather than looking for the removed fields
- [ ] A new `enrollment_api_authorization_header` property is added to access that env var
- [ ] By reusing those PEM data `_data` fields and the old `enrollment_api_authorization_header` field name for the new property name, no code changes outside of the model are required.

Scotchester:
Closes #3560 

---

Removes SwitchioConfig fields that aren't unique, converting them to properties that directly access env vars.

Because the Switchio environments line up 1:1 with our application environments and no switching between them is needed (even for production validation in the test environment), there is no longer a need to check the environment and use secrets with the env in the name.

### Testing instructions

1. Update your combo fixtures with the changes seen here in `local_fixtures.json`
2. Update your `.env` file to remove the `_int_` from the Switchio vars (except for `cst_switchio_int_api_secret`… but maybe we should do that, too?) and add the `switchio_enrollment_api_authorization_header` var.
3. Rebuild the container
4. Run through a CST enrollment flow

### To do before merge/deploy

- [x] Add `switchio-enrollment-api-authorization-header` secret to Azure Key Vault for each environment
  - [x] dev
  - [x] test
  - [x] prod
- [x] Add `switchio-client-cert`, `switchio-ca-cert`, and `switchio-private-key` secrets to Azure Key Vault for each environment
  - [x] dev
  - [x] test
  - [x] prod

### To do after merge

- [x] Update fixtures with secrets with the changes seen here in `local_fixtures.json`
- [x] Remind devs to update their local fixtures and `.env` files

### To do after deploys

- [x] Remove `switchio-<env>-*` secrets from the Azure Key Vault for the deployed env
  - [x] dev
  - [x] test
  - [x] prod


jgravois:
> Update your `.env` file to remove the `_int_` from the Switchio vars (except for `cst_switchio_int_api_secret`… but maybe we should do that, too?)

fwiw, i think its worth the trouble to omit the env signifier from these secrets too.

jgravois:
i was able to run through the testing steps successfully! 🚀  

imo it'd be helpful to make it explicit in the "Testing instructions" that one new secret needs to be added to the local .env file using information that is being scrubbed from the fixtures:

```bash
switchio_enrollment_api_authorization_header=shhh
```

Scotchester:
> > Update your `.env` file to remove the `_int_` from the Switchio vars (except for `cst_switchio_int_api_secret`… but maybe we should do that, too?)
> 
> fwiw, i think its worth the trouble to omit the env signifier from these secrets too.

Yeah, why not? Done in [feb9d02](https://github.com/cal-itp/benefits/pull/3659/commits/fbe9d02d4982d8fac0db4852d7a5f6c2155618f5)!

jgravois:
a few [onboarding docs](https://docs.calitp.org/benefits/guides/transit-provider-onboarding/configure-production-validation/#__tabbed_1_2) tweaks and simplification are in order too:

https://github.com/cal-itp/benefits/blob/c19939679e108a3c446042e3cb10f5638b27bfef/docs/guides/transit-provider-onboarding/configure-production.md?plain=1#L31-L37

https://github.com/cal-itp/benefits/blob/c19939679e108a3c446042e3cb10f5638b27bfef/docs/guides/transit-provider-onboarding/configure-production-validation.md#L53-L57

Scotchester:
> a few [onboarding docs](https://docs.calitp.org/benefits/guides/transit-provider-onboarding/configure-production-validation/#__tabbed_1_2) tweaks and simplification are in order too

Done, thanks for the reminder!

lalver1 (dismissed):
This looks great! I tested locally with the current combo fixtures (requiring me to run the migration which worked well) and with the updated environment variable naming and I was able to go through a complete enrollment.

Just one note, because in this PR we are removing the "per environment" secret naming, I wanted to refresh my memory about the origin of this naming convention and make sure I wasn't missing something. We had implemented it this way because parts of the Switchio API were being developed around the same time that we added Switchio support for Benefits, and at that time the 1:1 correspondence between environments was not clear / had not been implemented. Like this PR mentions in its description, I think that this refactor is safe to do now.

.gitignore
```diff
@@ -2,6 +2,7 @@
 *.env
 *fixtures.json
 !benefits/core/migrations/local_fixtures.json
+*.pem
```
jgravois, line 5: so the idea is that you'd retrieve a secret from azure, store it in a .pem file and reference the file in the .env instead of inline text?

jgravois (dismissed):
I retested after you pushed the last couple commits too and all was still well.

CIL for one question. overall, I'm loving this simplification.

.gitignore
Scotchester: Not quite; sorta the other way around.

The easiest way to add/update secrets in Azure Key Vault is via [these helper scripts in `terraform/scripts`](https://github.com/cal-itp/benefits/tree/main/terraform/secrets), but when running `file.sh` to set a multiline secret, the file has to be available in the devcontainer. So, I downloaded the `.pem` files from LastPass and put them in `terraform/secrets` for easiest completion when typing out the `./file.sh ...` command, but wanted to keep them out of version control.

Scotchester:
Thanks for the approvals, but putting back into draft until we can resolve an issue with a conflicting secret name in dev. [See discussion in Slack.](https://cal-itp.slack.com/archives/C037Y3UE71P/p1776806316220499)

Scotchester:
Updated to do the special casing discussed in the Slack thread in https://github.com/cal-itp/benefits/pull/3659/commits/82541f4a42b558afcdd1bce9ae9889274294b21c

Note that it affects local, as well, so to test the change locally you have to update your `.env` file.

Ready for one last look.

benefits/enrollment_switchio/models.py
```diff
@@ -98,17 +56,24 @@ def tokenization_api_secret(self):
     @property
     def client_certificate_data(self):
         """This SwitchioConfig's client certificate as a string."""
-        return self.client_certificate.data
+        if self.environment == Environment.DEV.value:
+            # Special case to handle un-purgeable test cert in Key Vault with the desired `switchio-client-cert` name
+            # See: https://cal-itp.slack.com/archives/C037Y3UE71P/p1776806316220499
+            # Also affects local setup using standard fixtures with secrets
+            # TODO: Remove this special case when the deleted test cert is automatically purged on July 20, 2026
```
jgravois: unless i've lost the plot completely, it'd be good to make it clear that this issue isn't in the `test` env.

```suggestion
            # Special case to handle un-purgeable cert in the azure dev environment Key Vault with the desired `switchio-client-cert` name
            # See: https://cal-itp.slack.com/archives/C037Y3UE71P/p1776806316220499
            # Also affects local setup using standard fixtures with secrets
            # TODO: Remove this special case when the deleted cert is automatically purged on July 20, 2026
```

Scotchester: Good point!

Scotchester: Tweaked slightly to fit under the line length limit and pushed in https://github.com/cal-itp/benefits/pull/3659/commits/2a4e895981f8d6b416c10018f2b419c19d6e7ea7

.env.sample
```diff
@@ -42,12 +42,13 @@ client_public_key='-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AM
 littlepay_qa_api_base_url=
 cst_littlepay_client_secret=
 
+switchio_enrollment_api_authorization_header=
+switchio_tokenization_api_base_url=
+switchio_enrollment_api_base_url=
 switchio_int_client_cert=
```
jgravois, line 48: if you don't mind humoring me, i think one more breadcrumb here would be extremely helpful.
```suggestion
# after 7/20/2026 this can safely be renamed 'switchio_client_cert'
switchio_int_client_cert=
```

jgravois (approved):
i tested e2e locally. still works a treat.

CIL for one more (non-blocking) suggestion.

.env.sample
Scotchester: Hmm, it'll _have_ to be along with the code change to remove the special case. Maybe it would be better to mention it in the comment on that special case code so it doesn't get overlooked?

jgravois: meh, forget i asked. the place the breadcrumb would be the most helpful (to me anyway) is the untracked .env itself. 😇

thekaveman:
@Scotchester I had requested @angela-tran to review this PR, and her and I spoke about her taking a look at this since she did all the original work on Switchio. Let's try to make sure we're using the "request a review" feature effectively for next time - just because the PR is unblocked for merging, take a double-check on the requests to ensure anyone else that wanted to / needed to look has a chance. I often self-request when I want to be sure to have a chance.

jgravois:
> I had requested @angela-tran to review this PR

ah, good catch @thekaveman. i noticed that last week, but lost track of it by today too. 

personally i wish github didn't conflate genuine requests for individual review like yours with other artifacts of its own. for example, it would be a lot easier to remember that a review from @angela-tran was still needed if we didn't see @lalver1 💬 immediately above with a comment (because his prior approval is now stale). 

Screenshot 2026-04-27 at 1 39 41 PM

that's just a tangent though. i think the technique you are suggesting for targeted reviews is a good one.

Scotchester:
My apologies! I totally missed that.

Angela, feel free to follow up with a post-hoc review, and if we'd like to wait to release this until such time as you can, I can revert the changes for now.

angela-tran:
Thanks all for giving / taking in the feedback. I know I can do better too with being more vocal myself with where I'm at and with not letting the weight of context-switching block me from starting. I started a post-hoc review earlier and will let y'all know when I'm done.

angela-tran:
I've finished going through all the discussion spread across #3560, this PR, and Slack as well as the code changes. No concerns or requests from me, and this clean-up / centralization makes a lot of sense. :white_check_mark:

PR #3664, opened 2026-04-20
docs: update transit agency statuses and target dates

indexing:
- marks `SLO Transit` LIVE
- updates target date for `SCMTD`
- removes target date for `SBCAG`; marks as `Planned`


jgravois (approved):
🎳

PR #3667, opened 2026-04-20
chore: pin docker-python-web and track using Dependabot

### LINKED ISSUE #3577: Pin upstream `ghcr.io/cal-itp/docker-python-web`, track with Dependabot
With https://github.com/cal-itp/docker-python-web/pull/72, I thought we could make a small improvement here to ensure we're getting the environment we expect.

Rather than pointing at `ghcr.io/cal-itp/docker-python-web:main` e.g.

- https://github.com/cal-itp/benefits/blob/be4e7b37acfbd40dabd08ecf3dd2b5b227d8fd3d/appcontainer/Dockerfile#L12
- https://github.com/cal-itp/benefits/blob/be4e7b37acfbd40dabd08ecf3dd2b5b227d8fd3d/appcontainer/Dockerfile#L58

We could instead (publish) and point at a specific version tag. [Dependabot](https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#docker) could be configured (with some additional work on the upstream) to auto-update via PR as new versions are published.

This will ensure we have the exact version we expect, and make it easier to track and manage updates through our standard Dependabot flow.

## Acceptance Criteria

- [ ] This Dockerfile pins the version of `ghcr.io/cal-itp/docker-python-web` to a numbered published version
- [ ] This repo configures Dependabot to watch for updates for the Docker image

## Additional context

Related: #3568

jgravois:
resolves #3577. while i was in here i did a little unrelated cleanup of our dependabot.yml` config file. 😇 

* start using [`directories`](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#defining-multiple-locations-for-manifest-files) (instead of maintaining two unique pip configurations)
* ~use a YAML anchor (to centralize/reuse common input arguments)~

i _think_ this is all that we'll need to ensure Dependabot opens a PR to update to docker-python-web@1.0.x when a new version is tagged.

after the PR is merged, i'll create a 1.0.1 tag to test and to make sure that we don't get bitten by https://github.com/dependabot/dependabot-core/issues/13383 given our multi-arch image.

## Steps to test

```bash
# outside the devcontainer
```

afterwards the DevContainer should build and the app should boot up as normal. i tested an e2e self-service medicare flow just to make sure.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,7 +28,9 @@ updates:
     labels:
       - "dependencies"
   - package-ecosystem: "docker"
-    directory: "/tests/playwright" # Playwright Docker image
+    directories:
+      - "/appcontainer" # appcontainer Docker image
+      - "/tests/playwright" # Playwright Docker image
     schedule:
       interval: "daily"
     commit-message:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,9 @@
 version: 2
 updates:
   - package-ecosystem: "pip"
-    directory: "/" # pyproject.toml
+    directories:
+      - "/" # pyproject.toml
+      - "/tests/playwright" # Playwright pytest plugin
     schedule:
       interval: "daily"
     commit-message:
@@ -27,15 +29,6 @@ updates:
       - "dependencies"
   - package-ecosystem: "docker"
     directory: "/tests/playwright" # Playwright Docker image
-    schedule:
-      interval: "daily"
-    commit-message:
-      prefix: "chore"
-      include: "scope"
-    labels:
-      - "dependencies"
-  - package-ecosystem: "pip"
-    directory: "/tests/playwright" # Playwright pytest plugin
     schedule:
       interval: "daily"
     commit-message:
--- a/appcontainer/Dockerfile
+++ b/appcontainer/Dockerfile
@@ -9,7 +9,7 @@ ARG PYTHONDONTWRITEBYTECODE=1 \
 #
 # stage 1: builds the benefits package from source
 #          using the git metadata for version info
-FROM ghcr.io/cal-itp/docker-python-web:main AS build_wheel
+FROM ghcr.io/cal-itp/docker-python-web:1.0.0 AS build_wheel
 
 # renew top-level args in this stage
 ARG PYTHONDONTWRITEBYTECODE \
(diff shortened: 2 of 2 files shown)


.github/dependabot.yml
```diff
@@ -3,43 +3,33 @@
...
-    directory: "/tests/playwright" # Playwright pytest plugin
-    schedule:
-      interval: "daily"
-    commit-message:
-      prefix: "chore"
-      include: "scope"
-    labels:
-      - "dependencies"
+    directories:
+      - "/appcontainer" # appcontainer Docker image
```
jgravois, line 32: this is the only line that intends to change the behavior of the action. it asks dependabot to start scanning the appcontainer Dockerfile for changes too.

lalver1:
Sorry about the delay @jgravois, this is looking good!
I tested locally building the Benefits image using the new `ghcr.io/cal-itp/docker-python-web:1.0.0` base image and had no issues, everything ran fine 👍. I'm just looking at the dependabot stuff one more time to make sure I'm not missing anything.

I'm wondering if `dependabot.yml` supports anchors? I would have thought so, but looking at https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd and how it only mentioned Compose files, it made me look around for dependabot specific yaml. I only found a few issues ([here](https://github.com/orgs/community/discussions/37496) and [here](https://github.com/dependabot/dependabot-core/issues/1582)) that suggested anchors aren't supported 🤔. I guess we can only know for sure by trying it out, it would be nice if they were supported!

jgravois:
> it would be nice if they were supported!

ah. bummer ☹️ i had found documentation indicating they were supported in more than just compose files, but your research appears conclusive that dependabot can't parse them. i'll push up another commit to revert. 🙏

bin/setup.sh
```diff
@@ -1,6 +1,13 @@
 #!/usr/bin/env bash
 set -ex
 
+# ensure fresh .env values are loaded
+if [ -f .env ]; then
+  set -a            # Automatically export all variables
+  source .env
+  set +a            # Disable automatic exporting
+fi
+
```
thekaveman: I'm curious about these lines, can you please say a little more about the use-case here?

jgravois: i didn't mean to include that in my commit. i've reverted the change but i'd like to open another PR to suggest it.

the gist of it is that i've been bitten a few times by the fact that bin/setup.sh loads stale .env values from memory. this tweak ensures it would take a fresh look at the file on disk each time its run. 😇

thekaveman: Is the use case running `bin/setup.sh` manually, when you're already inside the devcontainer? Closing / reopening the devcontainer should always reload env vars.

jgravois: exactly. typically its when i've needed to reset the db after updating `DJANGO_DB_FIXTURES`.

lalver1 (approved):
This looks great @jgravois! It's a shame that dependabot doesn't support `yaml` anchors but the consolidation you made did a good job of DRYing things up.

PR #3670, opened 2026-04-23
chore: Pin transitive dependencies and add script to update them

### LINKED ISSUE #3568: chore: consider pinning direct and transitive python dependencies
it would behoove us to get more explicit control over the exact versions of dependency code we deploy. initially we'd like to explore doing this without introducing new dedicated tooling like [`poetry`](https://python-poetry.org/) and [`uv`](https://docs.astral.sh/uv/).

## Acceptance Criteria

- [ ] for a particular branch, a fresh install lays down the same dependency versions on _any_ machine.
- [ ] the technique we use needs to play nicely with automated dependabot PRs.
- [ ] a convenient mechanism to update dependencies both surgically and in bulk will be needed.

## Additional context

* @Scotchester described a potentially viable path to maintain a homegrown lockfile in https://github.com/cal-itp/benefits/issues/3541#issuecomment-3986967474

Scotchester:
Closes #3568 

---

I want to revisit this next week and do some fresh testing of eligibility checks to make sure that the upgrades to some of the packages here are not affecting anything, but if anyone wants to take a look at the approach I'm proposing, please go ahead!

TODO:
- [ ] Update docs


jgravois (commented):
this seems sensible to me. my first question is, how do you expect the change to interact with Dependabot?

currently, it opens PRs that touch the `pyproject.toml` file directly. the equivalent would be for Dependabot to track `requirements.in`. is that possible?

even if it is, it would be better for it to track the `requirements.txt` file and notify us about updates to transitive deps too, right?

my instinct is that it would be. that would give us the flexibility to run an automated script to update stale dependencies in bulk and test them in aggregate or review/approve/merge individual Dependabot PRs.

example of the status quo: https://github.com/cal-itp/benefits/pull/3602

Scotchester:
Great questions, @jgravois. The Dependabot documentation leaves a bit to be desired, but based on that and a few conversations I looked through, I _think_ that the config I just pushed may work to:

- Check transitive dependencies (`dependency-type: "all"`)
- Only update `requirements.in`, not `requirements.txt` (`exclude-paths: "requirements.txt"`)

> even if it is, it would be better for it to track the `requirements.txt` file and notify us about updates to transitive deps too, right?

I don't think we want it looking directly at `requirements.txt`, because I believe it might then suggest updates to transitive deps that are not compatible with the ranges that our direct deps support.

I guess we'll have to merge it and wait up to a day to find out if it works, but does that approach make sense to you?

Scotchester:
Oops, _now_ I've pushed that Dependabot config update.

thekaveman:
Maybe following up to John's questions: is the `bin/dependencies.sh` script meant to be run by hand by us? In other words, how do we get a new `requirements.txt` when Dependabot updates versions in `requirements.in`?

Scotchester:
> Maybe following up to John's questions: is the `bin/dependencies.sh` script meant to be run by hand by us? In other words, how do we get a new `requirements.txt` when Dependabot updates versions in `requirements.in`?

That's also a good question. I had meant for it to be run manually, yes, but hadn't thought about it in context of Dependabot updates.

We could either make that be a part of the human review process for Dependabot PRs, or we could have another GHA workflow to run it automatically when Dependabot opens a PR that edits `requirements.in`, and update the PR to add in the changes to `requirements.txt`. Either way, we should add a workflow to block merging if `requirements.in` has been updated and `requirements.txt` has not.

jgravois:
> Check transitive dependencies (dependency-type: "all") ... Only update requirements.in, 

what would you expect it to look like for Dependabot to flag that an update to a transitive dep is needed if its only editing `requirements.ini`?

> I guess we'll have to merge it and wait up to a day to find out if it works, but does that approach make sense to you?

makes sense to me. i'm happy to start simple and rely on manual intervention from reviewers to get a feel for the mechanics. i have a (slight) worry that we're succumbing to [NIH](https://en.wikipedia.org/wiki/Not_invented_here) trying to juggle this manually and that it'd be more practical to use an existing solution for managing a lockfile, but i don't know the python ecosystem well enough to weigh the pros and cons well myself.

thekaveman:
> or we could have another GHA workflow to run it automatically when Dependabot opens a PR that edits `requirements.in`, and update the PR to add in the changes to `requirements.txt`

Yes I think this would be preferable to having to run this manually. Seems like something that would be easy to forget / lose track of, and moves us away from easier, faster PR review and release.

> Either way, we should add a workflow to block merging if `requirements.in` has been updated and `requirements.txt` has not.

:+1:

Scotchester:
> what would you expect it to look like for Dependabot to flag that an update to a transitive dep is needed if its only editing `requirements.ini`?

Hmm, another good question :)

[Here are the docs on that `dependency-type` option.](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#dependency-type-allow) The thing is, it doesn't explain how exactly it would handle an indirect dependency needing an update but not the direct dependency. (Also think about the scenario where there's no lockfile at all.) I also searched for some real user testimony about what would happen, but couldn't turn up any. Maybe it would open an issue instead of a PR? Total speculation, though.

I guess that brings me back to it being one of those things we just have to try and see how it works out.

Scotchester:
Updated to use the new `pylock.toml` support in pip 26.1 as discussed in #3568.

Tested locally:
- Building app container image from scratch
- Spinning up a plain app container, shelling in, and checking that dependencies were installed as expected
- Rebuilding dev container from scratch
- Running tests using VS Code's test runner and directly in the terminal
- Running the application
- A complete enrollment flow on the front end
- Clicking around through various models in the Admin
- Running the `dependencies.sh` updater script

Still to do:
- Add GHA check to catch the lockfile not being updated when it should
- See why a bunch of the checks are failing now

PR #3700, opened 2026-04-29
docs: updates target launch date for Gold Coast Transit District

indexing:


jgravois:
we'll need to close https://github.com/cal-itp/benefits/pull/3692 in favor of this one, correct? if so, does it still make sense timing-wise to open up an automated PR for agency onboarding to add to the adoption table?

if we onboard a few more agencies and find that its only helpful sometimes, we could make it an argument when running the workflow.

> Add the agency to the adoption table?
TRUE/FALSE

indexing:
Ah. I didn't realize/remember that the workflow autogenerates a PR for updating the adoption table. I'm in favor of keeping this step part of the automation, as I'll want to confirm the agency is in the table and that the target date is accurate. Can we auto assign the PR to me so that I receive a notification?

jgravois:
> Can we auto assign the PR to me so that I receive a notification?

sounds reasonable to me. go ahead and create a ticket to track it.

jgravois (approved):
lgtm!

indexing:
@jgravois 

> go ahead and create a ticket to track it.

I see my confusion. I believe this issue should be created as part of the autogeneration, but it wasn't. See SCMTD (manually created) compared to GCTD (auto generated). So I'll manually the issue for GCTD, but we'll update the script to include this issue going forward?

Screenshot 2026-04-30 at 11 46 14 AM
Screenshot 2026-04-30 at 11 47 07 AM

thekaveman:
@indexing I believe @jgravois was talking about "creating an issue" RE: your assignment on the PR that is created automatically by the workflow.

The automation doesn't create an _issue_ for updating the docs, it creates a (draft) PR with the update directly, with the idea that it can be a simple review/edit/merge from you.

Issue #3677, opened 2026-04-28
Cutover Benefits config to new hosted Sentry (`dev`)


thekaveman:
Once we have the new instance setup in #3669 we need to update the config in each Benefits environment to complete the cutover. This ticket is for the `dev` environment.

## Acceptance Criteria

- [x] `SENTRY_DSN` is updated to the new instance
- [x] `SENTRY_REPORT_URI` is updated to the new instance
- [ ] The new instance receives errors from the `dev` environment

jgravois:
i believe this one is safe to close out. i updated the secrets mentioned above in the test environment and afterward confirmed that 'dev' became available as a filterable environment.

<img width="340" height="236" alt="Image" src="https://github.com/user-attachments/assets/ac2eaa2c-2b2d-4c16-8b99-20888c617f56" />

routing to a non-existent agency path like /fake does _not_ trigger a sentry error because debug=True in that environment. when i set the same env variables and run the app locally with debug=False, i was able to see the issue come through and the 'local' environment appear for filtering in Sentry.

<img width="509" height="246" alt="Image" src="https://github.com/user-attachments/assets/1c738a3b-e2b5-4c54-8049-f7724c12f649" />

if there's another tried and true technique to trigger an error in dev that won't be swallowed, just say the word.

thekaveman:
> does _not_ trigger a sentry error because debug=True in that environment.

Ah right, I guess Django swallowing the exception for debug purposes prevents Sentry from seeing it? 

> confirmed that 'dev' became available as a filterable environment.

Anyway that seems good 👍

jgravois:
> I guess Django swallowing the exception for debug purposes prevents Sentry from seeing it?

my testing definitely seems to indicate that it does. and this makes sense imo. if you're actively debugging an app its probably premature for sentry to be logging.

Issue #1948, opened 2024-03-13
Delete refresh config app environment variable


thekaveman:
Azure has pulled out environment variable configuration from regular app configuration, into its own menu item:

![image](https://github.com/cal-itp/benefits/assets/1783439/726ad38b-046c-4653-b652-f33f451b1e94)

This includes a new item to re-fetch values from KeyVault, which also restarts the app:

![image](https://github.com/cal-itp/benefits/assets/1783439/6aa06145-dd2e-4e1f-ab7d-46efa47bc6b2)

![image](https://github.com/cal-itp/benefits/assets/1783439/551e67a3-a21a-4727-8733-51db6ccc493f)

Once this new menu item is available for @angela-tran, we can remove the app configuration environment variable we were using for this same purpose: `change_me_to_refresh_secrets`

## Acceptance Criteria

- [ ] `change_me_to_refresh_secrets` is removed from Terraform in [`app_service.tf`](https://github.com/cal-itp/benefits/blob/dev/terraform/app_service.tf#L53)

PR #2138, opened 2024-06-05
Refactor: delete app setting meant for refreshing secrets

angela-tran:
Closes #1948 

This is what I see in the Azure portal. Looks like they might've renamed the button

Screenshot 2024-06-05 113516   

<br>
<br>
Message when you click the button:
<br><br>
Screenshot 2024-06-05 113526

PR #3717, opened 2026-05-04
fix: ensure restart-app.sh helper script can also run on MacOS

jgravois:
## summary
this PR makes a small tweak to restart-app.sh to ensure it can be run successfully from both Linux and MacOS. i verified this by confirming that `date --utc` and `date -u` are analagous in a bash shell in the linux-based dev container.

it also updates the secrets tutorial to mention the current best practice for pulling fresh reference values. (see #1948 for more info)

## details

~while working on #3677 it wasn't immediately apparent to me how i could update the value of `change_me_to_refresh_secrets` manually to trigger an app restart. after making the tweak below, _then_ i was able to see the corresponding environment variable listed under "app service > settings > environment variables" in the Azure portal.~

~i assume this is 'by design', but if there's somewhere else i could have found the setting in the portal initially, i'd be curious to learn about it.~

### DIFF
--- a/docs/tutorials/secrets.md
+++ b/docs/tutorials/secrets.md
@@ -38,8 +38,13 @@ To make sure the Benefits application uses the latest secret values in Key Vault
 
 The steps are:
 
-1. After setting new secret values, go to the App Service configuration in Azure Portal, and change the value of the setting named `change_me_to_refresh_secrets`.
-1. Save your changes.
+1. After setting new secret values, cd into the `/terraform` directory
+2. run the command below, which updates a dummy environment variable called `change_me_to_refresh_secrets` in the relevant App Service configuration in Azure Portal.
+
+```bash
+cd terraform
+./restart-app.sh <environment_letter>
+```
 
 The effects of following those steps should be:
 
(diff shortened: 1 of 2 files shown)


thekaveman:
Glad you're looking at this. In #1948 / #2138 we actually removed this setting because Azure introduced UI to do it from the Portal. Obviously the docs and that helper script were not updated at the same time, sorry for the confusion.

Since we're about ready to dump App Service altogether, and the Portal has UI to do this, I say we just ignore this for now and/or blow away the script and docs since they are no longer relevant.

Scotchester:
Good point, @thekaveman, but we will need some kind of way to refresh updated secrets in the Container App world, too, right? I'd vote for the route of ignoring until we figure that process out.

thekaveman:
@Scotchester possibly. I think @lalver1 had done some research on that and Container Apps might be a little better at pulling new secret versions. In the Portal, you can always just deploy a new revision of the Container App (i.e. a new running instance of the already deployed infra/SHA/etc.), which _should_ pull new values.

jgravois:
ah, thanks for the extra context. i just pushed 267df66ee to mention the current best practice for App services. 

I agree we can/should revisit when the ground solidifies to update the doc again for container apps and leave the helper script alone in the interim.

if y'all would rather just close this one out, that's a-okay too.

jgravois:
> if y'all would rather just close this one out, that's a-okay too.

nm, just reread @thekaveman's original review and realized that's what he meant. 🤙

Issue #3639, opened 2026-03-30
Migrate app from App Service to Container Apps for `prod` environment (Cutover)


lalver1:
This ticket splits the migration from App Service to Container Apps into two parts: QA the side-by-side running of the App Service and Container Apps and the final Cutover to exclusively use Container Apps.

## QA the side-by-side running of the App Service and Container Apps

- [x] [Add the Container App's `Application Url` to `django-allowed-hosts`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] [Add the Container App's `Application Url` (requires `https://` in the URL) to `django-trusted-origins`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] Access Benefits running on the Container App via its Application URL
- [x] Postgres Database connection
- [ ] Access to external services
  - [x] reCAPTCHA
    Requires (temporarily) adding the Application URL's domain to the `Benefits` key's domain list in the Admin console (`benefits-recaptcha` project)
  - [x] MST Eligibility Server
  - [x] SBMTD Eligibility Server
  - [x] IdG
    Requires (temporarily) adding the Application URL to the IdG's allowed domains list
  - [x] Littlepay
  - [ ] Switchio
  - [ ] Amplitude
  - [x] Sentry
  - [x] Google SSO
    Requires (temporarily) adding the Application URL to the Authorized redirect URIs for `prod` in the Admin console (`benefits-admin` project)
  - [x] Benefits user password reset
- [x] In-person (ensure test user is in Django's database)
- [ ] Log in to Django admin using a Django user (not SSO) to ensure password integrity

## Cutover to exclusively use Container Apps

- [ ] Ensure the Postgres database has the latest data
- [ ] Coordinate with DevSecOps to update Front Door
  - [ ] Keep the existing endpoint
  - [ ] Add an origin with the Container App's Application URL in the existing Origin Group
  - [ ] Remove the origin that corresponded to the App Service
- [ ] Access Benefits running on the Container App via [`benefits`](https://benefits.calitp.org/)

lalver1:
## Container Apps QA

List any issues observed in the `prod` Container App here. These issues will need to be addressed before the cutover to exclusively use Container Apps.

If you do any testing, remember to add the `User-Agent: cal-itp/benefits-smoke-test` header to your browser for the `prod` Container App's `azurecontainerapps.io` domain (the application's URL listed in the Azure portal).

lalver1:
[...]

Issue #3638, opened 2026-03-30
Migrate app from App Service to Container Apps for `test` environment (Cutover)


lalver1:
This ticket splits the migration from App Service to Container Apps into two parts: QA the side-by-side running of the App Service and Container Apps and the final Cutover to exclusively use Container Apps.

## QA the side-by-side running of the App Service and Container Apps

- [x] [Add the Container App's `Application Url` to `django-allowed-hosts`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] [Add the Container App's `Application Url` (requires `https://` in the URL) to `django-trusted-origins`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] Access Benefits running on the Container App via its Application URL
- [x] Postgres Database connection
- [ ] Access to external services
  - [x] reCAPTCHA
    Requires (temporarily) adding the Application URL's domain to the `testing` key's domain list in the Admin console (`benefits-recaptcha` project)
  - [x] MST Eligibility Server
  - [x] SBMTD Eligibility Server
  - [x] IdG
    Requires (temporarily) adding the Application URL to the IdG's allowed domains list
  - [x] Littlepay
  - [x] Switchio
  - [ ] Amplitude
  - [x] Sentry
  - [x] Google SSO
    Requires (temporarily) adding the Application URL to the Authorized redirect URIs for `test` in the Admin console (`benefits-admin` project)
  - [x] Benefits user password reset
- [x] In-person (ensure test user is in Django's database)
- [ ] Log in to Django admin using a Django user (not SSO) to ensure password integrity

## Cutover to exclusively use Container Apps

- [ ] Ensure the Postgres database has the latest data
- [ ] Coordinate with DevSecOps to update Front Door
  - [ ] Keep the existing endpoint
  - [ ] Add an origin with the Container App's Application URL in the existing Origin Group
  - [ ] Remove the origin that corresponded to the App Service
- [ ] Access Benefits running on the Container App via [`test-benefits`](https://test-benefits.calitp.org/)
- [ ] Remove the App Service's URL from the following 
  - [ ] `django-trusted-origins`
  - [ ] `django-allowed-hosts`
- [ ] Remove the Container App URL from the 2 Google services
  - [ ] reCAPTCHA (`benefits-recaptcha` project)
  - [ ] SSO (`benefits-admin` project)

lalver1:
## Container Apps QA

[...]

Issue #3637, opened 2026-03-30
Migrate app from App Service to Container Apps for `dev` environment (Cutover)


lalver1:
This ticket splits the migration from App Service to Container Apps into two parts: QA the side-by-side running of the App Service and Container Apps and the final Cutover to exclusively use Container Apps.

## QA the side-by-side running of the App Service and Container Apps

- [x] [Add the Container App's `Application Url` to `django-allowed-hosts`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] [Add the Container App's `Application Url` (requires `https://` in the URL) to `django-trusted-origins`](https://github.com/cal-itp/benefits/pull/3646#issuecomment-4255937412)
- [x] Access Benefits running on the Container App via its Application URL
- [x] Postgres Database connection
- [ ] Access to external services
  - [x] reCAPTCHA
    Requires (temporarily) adding the Application URL's domain to the `testing` key's domain list in the Admin console (`benefits-recaptcha` project)
  - [x] MST Eligibility Server
  - [ ] ~SBMTD Eligibility Server~ N/A
  - [x] IdG
    Requires (temporarily) adding the Application URL to the IdG's allowed domains list
  - [x] Littlepay
  - [ ] ~Switchio~ N/A
  - [ ] ~Amplitude~ N/A
  - [ ] Sentry
  - [x] Google SSO
    Requires (temporarily) adding the Application URL to the Authorized redirect URIs for `dev` in the Admin console (`benefits-admin` project)
  - [x] Benefits user password reset
- [x] In-person (ensure test user is in Django's database)
- [ ] Log in to Django admin using a Django user (not SSO) to ensure password integrity

## Cutover to exclusively use Container Apps

- [ ] Ensure the Postgres database has the latest data
- [ ] Coordinate with DevSecOps to update Front Door
  - [ ] Keep the existing endpoint
  - [ ] Add an origin with the Container App's Application URL in the existing Origin Group
  - [ ] Remove the origin that corresponded to the App Service
- [ ] Update CI/CD
[...]

PR #3725, opened 2026-05-12
Refactor: trigger Container Apps deploy from GH Actions

lalver1:
Part of #3637, #3638, #3639

This PR refactors our CI/CD to "manually" trigger the DevOps Pipeline (which deploys the Container Apps) from the GH Actions `Deploy` workflow. We are still keeping the `Deploy to Azure Web App` step in the `Deploy` workflow to ensure that this PR is entirely an additive change so that we can continue deploying the App Service until we are completely cut over to Container Apps.

~We won't merge this PR until we make the manual changes to Front Door and confirm that the full cutover is complete.~ We decided to push forward with this change instead of waiting so that we are in a better state to continue the Postgres Stabilization and Optimization work.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -110,6 +110,64 @@ jobs:
           images: ghcr.io/${{ github.repository }}:${{ github.sha }}
           publish-profile: ${{ secrets.AZURE_WEBAPP_PUBLISH_PROFILE }}
 
+      - name: Trigger Azure Pipeline
+        id: trigger_pipeline
+        run: |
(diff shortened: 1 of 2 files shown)


jgravois (approved):
we tried running this workflow manually and a preceding step fails, presumably because it can't reference the appropriate azure devops pipeline context.

i'm comfortable merging this to `main` and reviewing the result of the subsequent run to verify that the container app deployed successfully.

jgravois:
after making a correction to one of the relevant secrets, the pipeline run was indeed successful.

PR #3729, opened 2026-05-12
Fix: branch name formatting for adoption table

thekaveman:
Replaces space characters with dash during branch name generation

See failed run here: https://github.com/cal-itp/benefits/actions/runs/25761548411/job/75663717809

Where the short name was `Thousand Oaks` and we saw a failure to create the branch:

> RequestError [HttpError]: refs/heads/docs/update-adoption-table-thousand oaks is not a valid ref name.

### DIFF
--- a/.github/workflows/agency-onboarding/helpers.js
+++ b/.github/workflows/agency-onboarding/helpers.js
@@ -180,7 +180,7 @@ export const updateAdoptionTable = async ({ github, context, parentIssue }) => {
   }
 
   // Create a new branch name
-  const branchName = `docs/update-adoption-table-${short_name.toLowerCase()}`;
+  const branchName = `docs/update-adoption-table-${short_name.toLowerCase().split(/\s+/).join("-")}`;
 
   // Get the main branch reference
   const mainRef = await github.rest.git.getRef({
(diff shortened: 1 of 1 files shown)


.github/workflows/agency-onboarding/helpers.js
```diff
@@ -180,7 +180,7 @@ export const updateAdoptionTable = async ({ github, context, parentIssue }) => {
   }
 
   // Create a new branch name
-  const branchName = `docs/update-adoption-table-${short_name.toLowerCase()}`;
+  const branchName = `docs/update-adoption-table-${short_name.toLowerCase().split(/\s+/).join("-")}`;
```
jgravois: i'm fine with this, but [`String.replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) would side-step the String > Array > String.

```suggestion
  const branchName = `docs/update-adoption-table-${short_name.toLowerCase().replaceAll(/\s+/g, "-")}`;
```

Image

ref: https://caniuse.com/wf-string-replaceall

jgravois (dismissed):
CIL for an optional suggestion

.github/workflows/agency-onboarding/helpers.js
thekaveman: Some light StackOverflow searching indicated `split/join` was faster, I don't really care either way.

jgravois (approved):
🌈🌈

Issue #3703, opened 2026-04-30
Add `GCTD Reduced Fare ID` to as an option for in-person enrollment


indexing:
### Summary
Add `GCTD Reduced Fare ID` as an in-person enrollment option so agency staff can enroll eligible riders for reduced fares when paying by open-loop payment. While this option serves Gold Coast Transit District, it will be available to all Ventura County agencies that comprise a region in Cal-ITP Benefits. 

### Problem
Ventura County agencies are preparing for a regional open-loop payments rollout, but riders who currently qualify for reduced fares through GCTD's Reduced Fare ID have no way to enroll through the self-service pathways Cal-ITP Benefits currently offers. That creates a gap: these riders need access to reduced fares via open-loop payments and, in particular, fare capping — which has a lower monthly cap for open-loop versus legacy fares. Gold Coast estimates roughly 25% of riders qualify for reduced fare.

### Why this matters
This closes an immediate launch gap for Ventura County’s reduced fare riders while broader self-service pathways are still in development. Gold Coast emphasized that they do not want vulnerable riders excluded from open-loop payments or regional fare capping during launch.

### In scope
- Add `GCTD Reduced Fare ID` as a new option in the `in-person enrollment` flow in Cal-ITP Benefits Administrator.
- This in-person enrollment option can be enabled in the configuration for all agencies in Ventura County that adopt Cal-ITP Benefits
- Set the staff-facing label for this benefit option to `GCTD Reduced Fare ID`.
- Set the staff attestation language as:  
  `I confirmed this rider’s identity using a government-issued photo ID and verified they possess a valid GCTD Reduced Fare ID.`
- Ensure successful card registration with Switchio in the `DISABLED` white-list.
- Add Amplitude tracking for enrollments completed through **GCTD Reduced Fare ID** by including a `GCTC_card` parameter on the `enrollment_flows` event

### Out of scope
- Self-service enrollment for GCTD Reduced Fare ID
- Expiration-date management during enrollment

## Key product notes / constraints
- Gold Coast’s existing reduced fare cards can expire either on a doctor-specified date or, if permanent, at a 5-year maximum. Gold Coast is comfortable launching without the ability to set expiration during enrollment. 
- Manual expiration date entry is in the product backlog, so we may want to consider prioritizing this feature. 

### Phase 1
`GCTD Reduced Fare ID` in-person enrollment is available for Gold Coast (#3690)  in the `test` environment by the end of May.

### Phase 2
[...]

PR #3771, opened 2026-05-15
feat: add support for GCTD reduced fare id enrollment flow (in person only)

### LINKED ISSUE #3763: Create `SystemName` for `gctd_card`

## Acceptance Criteria

- [ ]

### LINKED ISSUE #3764: Add `gctd_card` to allow list for In-Person flows

## Acceptance Criteria

- [ ]

### LINKED ISSUE #3765: Add policy language for `gctd_card` In-Person flow

## Acceptance Criteria

- [ ]

### LINKED ISSUE #3766: Add mapping for `gctd_card` to `AGENCY_CARD` Switchio group

## Acceptance Criteria

- [ ]

jgravois:
this branch tackles #3763, #3764, #3765 and #3766.

## `TODO:`
- [x] settle on implementation details
- [x] update shared fixtures and draft testing instructions
- [x] ~more test coverage?~
- [x] confirm https://github.com/cal-itp/benefits/issues/3703#issuecomment-4483032858
- [x] cleanup commit history once the ground stabilizes

## Testing instructions

In a new "[PROPOSED] GCTD Benefits fixtures with secrets" LastPass note, CST is configured for in-person GCTD Reduced Fare ID enrollment flow testing.

the note lists individual line item changes to "[COMBINED] Benefits fixtures...". if everything looks good i can update it in place after this PR is merged or we can toss it.

- [ ] in person, AST, Courtesy Card enrollment should work
- [ ] in person, BST, Reduced Fare Mobility ID enrollment should work
- [ ] in person, CST, GCTD Reduced Fare ID enrollment should work

If you attempt to configure GCTD Reduced Fare ID for self-service enrollment as-is, you'll notice that either a claims or api request are needed, along with a corresponding html template.


Scotchester:
> this is an experimental branch that tackles #3763, #3764, #3765 and #3766 holistically. it may prove more sensible to land the changes incrementally.

I think I recall Kegan saying in planning that one PR might be fine to cover all those issues, and seeing the total scope of the changes, I agree.

benefits/core/migrations/0005_alter_enrollmentflow_system_name.py
```diff
@@ -0,0 +1,30 @@
...
+            name="system_name",
+            field=models.SlugField(
+                choices=[
+                    ("calfresh", "Calfresh"),
+                    ("courtesy_card", "Courtesy Card"),
+                    ("medicare", "Medicare"),
+                    ("senior", "Older Adult"),
+                    ("mobility_pass", "Reduced Fare Mobility Id"),
+                    ("veteran", "Veteran"),
+                    ("agency_card", "Gctd Card"),
```
jgravois: the casing of `Gctd Card` is a little unfortunate, but since its purely internal it didn't seem worth the bother to try and one-off a fix.

happy to discuss if others disagree though.

benefits/core/models/enrollment.py
```diff
@@ -107,16 +108,18 @@ def api_public_key_data(self):
...
+
+
 class EnrollmentFlow(models.Model):
     """Represents a user journey through the Benefits app for a single eligibility type."""
 
-    supported_in_person_flows = (
-        SystemName.COURTESY_CARD.value,
-        SystemName.MEDICARE.value,
-        SystemName.OLDER_ADULT.value,
-        SystemName.REDUCED_FARE_MOBILITY_ID,
```
jgravois, line 117: i pulled this constant out of the class to avoid redeclaring it more than once.

the shuffle obscures the fact that i also swapped in `SystemName.REDUCED_FARE_MOBILITY_ID.value` for consistency after retesting to ensure that it doesn't induce an error.

benefits/enrollment_switchio/models.py
```diff
@@ -15,8 +15,7 @@ class SwitchioGroupIDs:
     CALFRESH = "LOW_INCOME"
     OLDER_ADULT = "OLDER_ADULT"
     VETERAN = "VETERAN"
-    # We have no Switchio agencies with Agency Card flows, but this is needed for testing.
-    COURTESY_CARD = "AGENCY_CARD"
+    GCTD_CARD = "AGENCY_CARD"
```
thekaveman, line 18: Right, this is where we map our internal name `GCTD_CARD` to the external Switchio group 👍

benefits/core/models/enrollment.py
```diff
@@ -30,6 +30,7 @@ class SystemName(models.TextChoices):
     OLDER_ADULT = "senior"
     REDUCED_FARE_MOBILITY_ID = "mobility_pass"
     VETERAN = "veteran"
+    GCTD_CARD = "agency_card"
```
thekaveman: I don't think this is correct. The system name should be our internal name `gctd_card` -- we do the mapping to the `AGENCY_CARD` Switchio group elsewhere. 

I think this is the value that is used for e.g. analytics events, so we definitely want the specific version.

benefits/core/migrations/0005_alter_enrollmentflow_system_name.py
thekaveman: Since this is a user-facing (internal users) label, I believe we want to make the casing correct here.

E.g. this is the value that shows up in the drop-down when selecting the system name right? And potentially the radio list for In-Person selection of the flow.

thekaveman (dismissed):

benefits/core/migrations/0005_alter_enrollmentflow_system_name.py
jgravois: > this is a user-facing (internal users) label

that's correct. but these days we only display the system name in the global enrollment flow.

when configuring what flows an individual agency will support and within in-person enrollment, we display the label.

Screenshot 2026-05-18 at 12 43 47 PM
Screenshot 2026-05-18 at 1 23 22 PM

for what its worth, "Reduced Fare Mobility ID" is also cased inconsistently, albeit less noticeably.

Screenshot 2026-05-18 at 11 53 21 AM

benefits/core/models/enrollment.py
jgravois: ah, thank you for pointing this out. 🙏

benefits/core/models/enrollment.py
Scotchester: Since the previous inconsistency taught us that we didn't need the `.value`, I wouldn't mind dropping those instead, but not a strong preference at all.

jgravois: no objections here. i'll do that unless anyone else disagrees.

edit: done in 22cb49c

benefits/core/models/enrollment.py
```diff
@@ -242,12 +245,15 @@ def clean(self):
         # we can't add a field-level validation error
         # so just create directly for a missing template
         for t in templates:
-            if not template_path(t):
+            # so far we only have templates for self-service flows
+            # if/when we templatize in-person, we'll need to revisit and ensure that
+            # all templates associated with the currently enabled paths are present
```
angela-tran: The history behind this code is not obvious at all, so understandably, this comment is not correct. I have to request that we not add this. It'll take me a second to write up why, please hold 😅

angela-tran (changes requested):

benefits/core/models/enrollment.py
angela-tran: We've gone through many approaches to handling agency-specific/flow-specific copy through the years.

At one point, the approach was to have fields on `TransitAgency` and `EnrollmentFlow` like
- `index_template`
- `eligibility_index_template`
- [`eligibility_start_template`](https://github.com/cal-itp/benefits/blob/2024.10.3/benefits/core/models.py#L335)
- [`selection_label_template`](https://github.com/cal-itp/benefits/blob/2024.10.3/benefits/core/models.py#L332)
- `unverified_template`

We would have a base template that defined the generic page (e.g. [`start.html`](https://github.com/cal-itp/benefits/tree/2024.10.3/benefits/eligibility/templates/eligibility) or [`selection-label.html`](https://github.com/cal-itp/benefits/tree/2024.10.3/benefits/eligibility/templates/eligibility/includes)) and then "concrete" child versions of them as the way to "fill in" the agency-specific/flow-specific copy (e.g. `start--calfresh.html`, `start--medicare.html`, etc. and `selection-label--calfresh.html`, `selection-label--medicare.html`).

Then, since each `EnrollmentFlow` was specific to a transit-agency and flow, the name of the concrete child template was put in to the model field. So for example, the MST Courtesy Card `EnrollmentFlow` would have a `selection_label_template` with a value of `eligibility/includes/selection-label--mst-courtesy.card.html`

At some point, we realized, if we're already naming the templates following a predictable pattern, we should just have the [code calculate the template name for us](https://github.com/cal-itp/benefits/blob/82b4b67b98501c06c0e18a08fb22f3df1816487e/benefits/core/models/enrollment.py#L186-L200) instead of putting it in ourselves. To protect against a situation where somehow we didn't name the template file correctly, we added in [validation logic](https://github.com/cal-itp/benefits/commit/19dc65660638b9a27894ddcfa21fa67c5ef0f386) to check that the calculated template name does in fact reference a file that exists. That's what this `template_path` code is doing.

To finish the story, at some point, we realized we could just move the agency-specific/flow-specific copy so that it's passed in as context dictionaries to a single, generic template. That refactor removed all the concrete child templates, _except_ for `selection_label_template`. That's where we are today.

(And of course as you know @jgravois, you helped further simplify so we no longer maintain agency-specific context dictionaries, solving the problem of needing a code deployment to add new agencies. 😃 )

So all of that is to say, the direction we want to go with this is to one day remove it, rather than achieve parity.

---

`selection_label_template` is special in that it has more variation than just copy. It's the HTML for the label of the transit-benefit radio buttons, so there are things like embedded buttons and links for opening modals and whatnot. There's nothing technically stopping us from refactoring it out, just time / other much-higher priorities.

benefits/core/models/enrollment.py
```diff
@@ -242,12 +245,15 @@ def clean(self):
...
+            if EnrollmentMethods.SELF_SERVICE in self.supported_enrollment_methods and not template_path(t):
```
angela-tran: I see now what prompted your comment. Definitely makes sense that we need to add in this check 💯

angela-tran: So I guess if the point of the comment is to explain why `EnrollmentMethods.SELF_SERVICE in self.supported_enrollment_methods` is needed, it could be something like 
```
# selection_label_template only applies to SELF_SERVICE
```

Now that I'm looking at it more, it's feeling like we should just clean this code up so it reads more clearly as validating the existence of `selection_label_template` if the flow supports `SELF_SERVICE`.

jgravois: thanks for taking the time to share all the context above.

> we should just clean this code up so it reads more clearly as validating the existence of `selection_label_template` if the flow supports `SELF_SERVICE`.

i took a stab at this in ~d171b0a34~4ef77ad. curious to hear what you think.

angela-tran (dismissed):
These code changes look good to me. Nice work @jgravois 

For testing this locally, I took the approach of going through it like I would if I were configuring it in dev / test / prod:
- I added a new, in-person-only `EnrollmentFlow` for the GCTD Reduced Fare ID and linked it to one of my transit agencies. 
- I logged in as the user for that transit agency and tried out an in-person enrollment.
- This threw an error and revealed to me that I needed to go create an `SwitchioGroup` that links up the transit agency and the new `EnrollmentFlow`.
- Then, I tried again, and the enrollment worked 🎉 

I have some thoughts about our fixtures that have been on my mind even before this PR. Your notes on how you had to update the fixtures are prompting me to start that discussion, but I don't think this PR is the right place, so I'll ask about it in Slack.

benefits/core/models/enrollment.py
angela-tran: That looks great! :heart:

jgravois:
> These code changes look good to me. Nice work @jgravois

thanks for the thorough review @angela-tran. i just:

1. squashed a few messy/intermediate commits
1. confirmed the diff remained the same after `-f` pushing
1.  tacked on 2c472c8 to address @thekaveman's suggestion in slack.

angela-tran (approved):

PR #3789, opened 2026-05-27
fix: avoid inappropriate redirects after successful switchio card tokenization during in-person enrollment

### LINKED ISSUE #3788: in person: duplicative redirects after successful card tokenization
i've seen incorrect redirects back and forth between our own app and switchio's full page takeover intermittently after successful card tokenization during in-person switchio enrollment locally, on `test` and on `production` as well.

## To Reproduce

1. login to admin as staff user for switchio agency
  (i've been able to repro with `cst-user` locally, `vctc-user` on `test` and `vctc-test-user` on `prod`)
1. start an in-person enrollment
1. enter card details
1. app redirects back and forth between our own loading screen and the switchio full page takeover endlessly

## Expected behavior

after successful card tokenization, the app should immediately redirect to the success page 

## Screenshots

https://cal-itp.slack.com/archives/C037Y3UE71P/p1779914658652049?thread_ts=1779912820.969619&cid=C037Y3UE71P

jgravois:
closes #3788

### DIFF
--- a/benefits/enrollment_switchio/templates/enrollment_switchio/script.html
+++ b/benefits/enrollment_switchio/templates/enrollment_switchio/script.html
@@ -16,60 +16,60 @@
   const startedEvent = "started card tokenization";
   const closedEvent = "finished card tokenization";
 
-  let returned_from_tokenization = window.location.href.search('state=tokenize') > -1;
+  const returned_from_tokenization = window.location.href.search('state=tokenize') > -1;
 
   if (returned_from_tokenization) {
     StRpClient.handleTokenizeWithRedirectCallback().then((response) => {
       if (response.type === 'success') {
         amplitude.getInstance().logEvent(closedEvent, {status: "success", enrollment_method: "{{ enrollment_method }}"});
 
-        let form = document.querySelector("form#{{ form_success }}");
+        const form = document.querySelector("form#{{ form_success }}");
         form.querySelector("#card_token").value = "{{ card_token }}";
         form.submit();
       } else if (response.error != "canceled") {
         amplitude.getInstance().logEvent(closedEvent, {status: "error", error: response.error, enrollment_method: "{{ enrollment_method }}"});
 
-        let form = document.querySelector("form#{{ form_system_error }}");
+        const form = document.querySelector("form#{{ form_system_error }}");
         form.submit();
       } else {
         amplitude.getInstance().logEvent(closedEvent, {status: "cancel", enrollment_method: "{{ enrollment_method }}"});
       }
     });
-  }
+  } else {
+    fetch("{% url gateway_url_route %}")
+      .then(response => response.json()
+      .then(data => {
+        if (data.redirect) {
+          // https://stackoverflow.com/a/42469170
+          // use 'assign' because 'replace' was giving strange Back button behavior
+          window.location.assign(data.redirect);
+          return; // exit early so the rest of this function doesn't execute
+        }
 
-  fetch("{% url gateway_url_route %}")
-    .then(response => response.json()
-    .then(data => {
-      if (data.redirect) {
-        // https://stackoverflow.com/a/42469170
-        // use 'assign' because 'replace' was giving strange Back button behavior
-        window.location.assign(data.redirect);
-        return; // exit early so the rest of this function doesn't execute
-      }
+        {% block gateway-url-handler %}
+        document.querySelector(".loading").remove();
 
-      {% block gateway-url-handler %}
-      document.querySelector(".loading").remove();
+        // remove invisible and add back visible, so we aren't left with
+        // a div with an empty class attribute
+        // (this is purely for clarity when reviewing the HTML, no UX impact)
+        const invisibleElement = document.querySelector(".invisible");
+        invisibleElement.classList.remove("invisible");
+        invisibleElement.classList.add("visible");
 
-      // remove invisible and add back visible, so we aren't left with
-      // a div with an empty class attribute
-      // (this is purely for clarity when reviewing the HTML, no UX impact)
-      let invisibleElement = document.querySelector(".invisible");
-      invisibleElement.classList.remove("invisible");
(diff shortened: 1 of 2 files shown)


benefits/enrollment_switchio/templates/enrollment_switchio/script.html
```diff
@@ -16,60 +16,61 @@
...
-    .then(response => response.json()
-    .then(data => {
-      if (data.redirect) {
-        // https://stackoverflow.com/a/42469170
-        // use 'assign' because 'replace' was giving strange Back button behavior
-        window.location.assign(data.redirect);
-        return; // exit early so the rest of this function doesn't execute
-      }
+  } else {
+    fetch("{% url gateway_url_route %}")
```
jgravois, line 39: i could be over-simplifying, but my reading of the chart below is that we should only attempt to identify the appropriate switchio gateway url _prior_ to an attempt to tokenize payment card details.

https://docs.calitp.org/benefits/explanation/development/application-logic/#switchio

benefits/enrollment_switchio/templates/enrollment_switchio/script.html
```diff
@@ -16,60 +16,60 @@
...
+      .then(response => response.json()
+      .then(data => {
+        if (data.redirect) {
+          // https://stackoverflow.com/a/42469170
```
jgravois, line 43: in all honesty, i'm not sure why this `fetch` only _sometimes_ interferes with the call to `form.submit();` on L28. 

my best guess is that it has to do with how quickly we are able to determine the appropriate gateway url, but a breakpoint inside this callback isn't hit even when the redirect prevents the form from being submitted.

benefits/settings.py
```diff
@@ -343,6 +343,7 @@ def RUNTIME_ENVIRONMENT():
         "connect-src": [
             SELF,
             "https://api.amplitude.com/",
+            "https://cdn.jsdelivr.net/sm/",  # centralized sourcemap hosting
```
jgravois, line 346: this is just a fix for a tangentially related error i saw in sentry.

i wouldn't expect this one to surface until/unless someone happens to have developer tools open 😇 

Image

jgravois (commented):
i'm still researching this, but i figured i'd open and annotate early for visibility.

benefits/settings.py
thekaveman: Is this like a canonical URL for _all_ source maps via `jsdelivr`? Or is this specific to Switchio?

jgravois: i've been assuming that "sm" stands for sourcemaps. i have no idea why jsdelivr would host them in a dedicated directory for the switchio package (and likely others).

they're not included in the listed package contents below, but since the package itself doesn't have an `index.min.js` file either i just chalked that up to jsdelivr adding some sugar of its own to optimize and annotate. 🤷 

https://www.npmjs.com/package/@switchio/st-rp-sdk?activeTab=code

in other words, if a package was published with sourcemaps of its own, i'd expect them to be hosted from the same directory as the rest of the code and have `/sm/` reserved for sourcemaps that jsdelivr generates itself, but that's entirely speculative.

benefits/enrollment_switchio/templates/enrollment_switchio/script.html
Scotchester: That's how I understand it, as well, and the addition of the `else` here seems to me like a logical way to fix it.

angela-tran: Yeah, I agree that the `else` here makes sense. 

I think the previous assumption was that it would never be that Switchio returns back to us (i.e. returns from tokenization) in a way that doesn't match the `returned_from_tokenization` condition. But I guess something changed or that was always a bad assumption. Either way, this seems more correct.

jgravois: thank you both for the sanity check. 🙏

angela-tran (approved):

benefits/enrollment_switchio/templates/enrollment_switchio/script.html
Scotchester: Strange that the breakpoint wouldn't be hit.

If I've traced things through the codebase correctly, it seems to me that the redirect here only happens if the request to get the gateway URL is not successful:

https://github.com/cal-itp/benefits/blob/0b8bd1e4e35a7db79d42c2c23e76a530adb03317/benefits/enrollment_switchio/views.py#L153-L159

However, we should also be seeing multiple forms of logging on this, but we seem not to be, so maybe I am off-base?

https://github.com/cal-itp/benefits/blob/0b8bd1e4e35a7db79d42c2c23e76a530adb03317/benefits/enrollment_switchio/views.py#L149-L151

jgravois: > it seems to me that the redirect here only happens if the request to get the gateway URL is not successful:

thats correct. i was adding a `debugger` statement directly above to try and catch out _any_ response.

```js
debugger
if (data.redirect) {
```

> we should also be seeing multiple forms of logging on this

i agree! you'd think if the request was still pending it wouldn't hijack and that if it did hijack it'd be in the logs.

Scotchester: Oh! It's not this redirect.

The _in-person_ form extends this template, but it overrides the `gateway-url-handler` block below to _immediately_ trigger the redirect, rather than requiring the user to click the button in the original block here that's used for self-service.

https://github.com/cal-itp/benefits/blob/0b8bd1e4e35a7db79d42c2c23e76a530adb03317/benefits/in_person/templates/in_person/enrollment/switchio_script.html#L3-L8

Scotchester: (which also explains why your breakpoint wasn't hit)

PR #3791, opened 2026-05-28
chore: chain promises when determining switchio gateway url

### LINKED ISSUE #3790: chore: chain promises instead of nesting when determining switchio gateway_url
it's so subtle i didn't even notice when i opened #3789 yesterday, but the two Promises below aren't actually chained, they are nested:

https://github.com/cal-itp/benefits/blob/159c43fbc30c22fce5f4bdab41c711738d47968f/benefits/enrollment_switchio/templates/enrollment_switchio/script.html#L40-L41

~based on my testing this appears to be a benign error that the browser fixes up for us behind the scenes somehow, but its still worth fixing.~ (see https://github.com/cal-itp/benefits/pull/3791#discussion_r3320418567 for more info)

going forward i think it'd be worth exploring improving the JS linting and auto-formatting in this repo holistically.

jgravois:
closes #3790

### DIFF
--- a/benefits/enrollment_switchio/templates/enrollment_switchio/script.html
+++ b/benefits/enrollment_switchio/templates/enrollment_switchio/script.html
@@ -37,7 +37,7 @@
     });
   } else {
     fetch("{% url gateway_url_route %}")
-      .then(response => response.json()
+      .then(response => response.json())
       .then(data => {
         if (data.redirect) {
           // https://stackoverflow.com/a/42469170
@@ -70,6 +70,6 @@
           tokenizeWithRedirect(data.gateway_url)
         });
         {% endblock gateway-url-handler %}
-      }));
+      });
   }
 </script>


benefits/enrollment_switchio/templates/enrollment_switchio/script.html
jgravois, line 1: edit: @Scotchester and i took a poke at this together and he reminded me that this works as-is because [`Response.json()`](https://developer.mozilla.org/en-US/docs/Web/API/Response/json) itself returns a [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).

ie: what we have currently is interpreted as:

```js
fetch(url)
  .then(response => { 
    return response.json()
      .then(data => {
        if (data.redirect) // ...
      })
  })
```
so to revise what i wrote earlier, this PR isn't so much fixing a problem as it is avoiding unnecessary nesting and ensuring that the expectation/intent and actual code flow are aligned.

Scotchester (approved):

benefits/enrollment_switchio/templates/enrollment_switchio/script.html
angela-tran: That all makes sense to me, and I agree this clean-up is good even if nothing is necessarily broken.

PR #3804, opened 2026-06-15
Chore: enable in-person for inactive agencies

### LINKED ISSUE #3802: Enable the In-person feature for an inactive transit agency
_edit:_
The long story short is that in-person enrollment doesn't work for inactive agencies. 
- For inactive, non-regional `TransitAgency`s, an error will be raised by `AgencySessionRequiredMixin` when you get to [`/in_person/enrollment`](https://github.com/cal-itp/benefits/issues/3757#issuecomment-4684116842)
- For inactive, regional `TransitAgency`s, an error will be raised immediately
  > Currently, if an inactive transit agency is part of a region, the in-person feature fails and returns a `200-user-error.html` self-service template when the regional view, [`AdditionalAgenciesView`](https://github.com/cal-itp/benefits/blob/main/benefits/in_person/views.py#L29), is called due to how it inherits from the self-service [`AdditionalAgenciesView`](https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L238) and how this class assumes an active agency in the [`AgencySessionRequiredMixin`](https://github.com/cal-itp/benefits/blob/main/benefits/core/mixins.py#L23) mixin. 

---
_original description:_

Currently, if an inactive transit agency is part of a region, the in-person feature fails and returns a `200-user-error.html` self-service template when the regional view, [`AdditionalAgenciesView`](https://github.com/cal-itp/benefits/blob/main/benefits/in_person/views.py#L29), is called due to how it inherits from the self-service [`AdditionalAgenciesView`](https://github.com/cal-itp/benefits/blob/main/benefits/core/views.py#L238) and how this class assumes an active agency in the [`AgencySessionRequiredMixin`](https://github.com/cal-itp/benefits/blob/main/benefits/core/mixins.py#L23) mixin. It is also possible that in-person will fail later on for inactive agencies since the `AgencySessionRequiredMixin` is used in subsequent in-person views.

It is also interesting to note that the failure in `AdditionalAgenciesView` is not visible if all the transit agencies in a region are inactive due to how [the logic in the in-person index site](https://github.com/cal-itp/benefits/blob/main/benefits/admin.py#L71) interacts with the return value of [TransitAgency.group_agencies()](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/transit.py#L206). 

For more details see the conversation in the [Simi Valley production validation ticket](https://github.com/cal-itp/benefits/issues/3757).

lalver1:
Closes #3802 

This PR enables the in-person flow for transit agencies that are marked as inactive. It does so by making the following minimal changes until we decide to implement a `status` field on the `TransitAgency` model:

- Refactoring `AgencySessionRequiredMixin` to allow in-person to work for inactive agencies ONLY if the user is an authenticated staff member, yet still fulfilling its original purpose of only allowing self-service for active transit agencies.
- Generalizing the `TransitAgency.group_agencies()` and `TransitAgency.group_agency_short_names()` methods to allow the caller to specify if all agencies or only active ones within a transit agency group should be returned. By using a default parameter value we keep the original behavior of the function unchanged.
- Keeping our original CBV pattern of using class attributes to set behavior for in-person or self-service views when the in-person view inherits from self-service.

## Reviewing

- Set your fixtures to use the combined fixtures in LastPass
- Verify the following cases:
  - [x] Aalifornia active, Balifornia active
    - [x] See the additional regional agencies view, can enroll a card, and see the regional agencies success view using in-person (using `ast-user`)
    - [x] Aalifornia and Balifornia in drop-down, see the additional agencies view, can enroll a card, see the regional agencies success view
  - [x] Aalifornia active, Balifornia inactive
    - [x] See the additional regional agencies view, can enroll a card, and see the regional agencies success view using in-person (using `ast-user`)
    - [x] Aalifornia in drop-down, no additional agencies view, can enroll a card, no regional agencies success view
  - [x] Aalifornia inactive, Balifornia active
    - [x] See the additional regional agencies view, can enroll a card, and see the regional agencies success view using in-person (using `ast-user`)
    - [x] Balifornia in drop-down, no additional agencies view, can enroll a card, no regional agencies success view
  - [x] Aalifornia inactive, Balifornia inactive
    - [x] See the additional regional agencies view, can enroll a card, and see the regional agencies success view using in-person (using `ast-user`)
    - [x] Only CST in drop-down so you can only proceed using CST

### DIFF
--- a/benefits/core/mixins.py
+++ b/benefits/core/mixins.py
@@ -16,15 +16,16 @@ class AgencySessionRequiredMixin:
 
     Gets the active `TransitAgency` out of session and sets an attribute on `self`.
 
-    If the session is not configured with an agency, return a user error.
+    Allows inactive agencies ONLY if the user is an authenticated staff member, otherwise returns a user error.
     """
 
     def dispatch(self, request, *args, **kwargs):
-        if session.active_agency(request):
-            self.agency = session.agency(request)
+        agency = session.agency(request)
+        if agency and (agency.active or getattr(request.user, "is_staff", False)):
+            self.agency = agency
             return super().dispatch(request, *args, **kwargs)
         else:
-            logger.warning("Session not configured with an active agency")
+            logger.warning("Session missing an agency, or user lacks permission to access an inactive agency")
             return user_error(request)
 
 
--- a/benefits/core/models/transit.py
+++ b/benefits/core/models/transit.py
@@ -215,14 +215,16 @@ class TransitAgency(models.Model):
 
         return list(agencies_in_group.exclude(pk=self.pk).order_by("short_name"))
 
-    def group_agency_short_names(self):
+    def group_agency_short_names(self, only_active=True):
         """A list of agency short names for this agency and any agencies it shares a group with.
+        If only_active is True, only active short names are returned. If only_active is False,
+        all short names are returned.
 
         The list begins with the current agency and the rest follow in alphabetical order.
         If an agency is not associated with any other agencies via TransitAgencyGroup,
         this returns an empty list.
         """
-        agencies = [self] + self.group_agencies()
+        agencies = [self] + self.group_agencies(only_active=only_active)
 
         if len(agencies) > 1:
             return [agency.short_name for agency in agencies]
--- a/benefits/core/models/transit.py
+++ b/benefits/core/models/transit.py
@@ -206,14 +206,16 @@ class TransitAgency(models.Model):
         this returns an empty list.
         """
 
-        agencies_in_group = TransitAgency.objects.filter(
-            transitagencygroup__in=list(self.transitagencygroup_set.all())
-        ).distinct()
+        agencies_in_group = (
+            TransitAgency.objects.filter(transitagencygroup__in=list(self.transitagencygroup_set.all()))
+            .distinct()
+            .exclude(pk=self.pk)
+        )
 
         if only_active:
             agencies_in_group = agencies_in_group.exclude(active=False)
 
-        return list(agencies_in_group.exclude(pk=self.pk).order_by("short_name"))
+        return list(agencies_in_group.order_by("short_name"))
 
     def group_agency_short_names(self, only_active=True):
         """A list of agency short names for this agency and any agencies it shares a group with.
--- a/benefits/core/models/transit.py
+++ b/benefits/core/models/transit.py
@@ -198,7 +198,7 @@ class TransitAgency(models.Model):
         return f"{self.short_name} Customer Service"
 
     def group_agencies(self):
-        """The set of all agencies in all groups associated with this agency, excluding itself.
+        """The set of all active agencies in all groups associated with this agency, excluding itself.
 
         If an agency is not associated with any other agencies via TransitAgencyGroup,
         this returns an empty list.
--- a/benefits/core/models/transit.py
+++ b/benefits/core/models/transit.py
@@ -197,19 +197,23 @@ class TransitAgency(models.Model):
         """Returns the standardized name for this Agency's customer service group."""
         return f"{self.short_name} Customer Service"
 
-    def group_agencies(self):
-        """The set of all active agencies in all groups associated with this agency, excluding itself.
+    def group_agencies(self, only_active=True):
+        """The set of agencies in all groups associated with this agency, excluding itself.
+        If only_active is True, only active agencies are returned. If only_active is False,
+        all agencies are returned.
 
         If an agency is not associated with any other agencies via TransitAgencyGroup,
         this returns an empty list.
         """
-        return list(
-            TransitAgency.objects.filter(transitagencygroup__in=list(self.transitagencygroup_set.all()))
-            .distinct()
-            .exclude(active=False)
-            .exclude(pk=self.pk)
-            .order_by("short_name")
-        )
+
+        agencies_in_group = TransitAgency.objects.filter(
+            transitagencygroup__in=list(self.transitagencygroup_set.all())
+        ).distinct()
+
+        if only_active:
+            agencies_in_group = agencies_in_group.exclude(active=False)
+
+        return list(agencies_in_group.exclude(pk=self.pk).order_by("short_name"))
 
     def group_agency_short_names(self):
         """A list of agency short names for this agency and any agencies it shares a group with.
--- a/tests/pytest/core/test_mixins.py
+++ b/tests/pytest/core/test_mixins.py
@@ -25,9 +25,9 @@ class TestAgencySessionRequiredMixin:
         v.setup(app_request)
         return v
 
-    def test_dispatch_without_active_agency(self, view, app_request, mocker):
+    def test_dispatch_without_agency(self, view, app_request, mocker):
         mock_session = mocker.patch("benefits.core.mixins.session")
-        mock_session.active_agency.return_value = False
+        mock_session.agency.return_value = None
 
         response = view.dispatch(app_request)
 
@@ -35,14 +35,25 @@ class TestAgencySessionRequiredMixin:
         assert response.status_code == 200
         assert response.template_name == TEMPLATE_USER_ERROR
 
-    def test_dispatch_with_active_agency(self, view, app_request, mocker):
+    @pytest.mark.django_db
+    def test_dispatch_with_active_agency(self, view, app_request, mocker, model_TransitAgency):
         mock_session = mocker.patch("benefits.core.mixins.session")
-        mock_session.active_agency.return_value = True
-        mock_session.agency.return_value = {"agency": "123"}
+        mock_session.agency.return_value = model_TransitAgency
 
         view.dispatch(app_request)
 
-        assert view.agency == {"agency": "123"}
+        assert view.agency == model_TransitAgency
+
+    @pytest.mark.django_db
+    def test_dispatch_with_inactive_agency_staff_user(self, view, app_request, mocker, model_TransitAgency):
+        mock_session = mocker.patch("benefits.core.mixins.session")
+        model_TransitAgency.active = False
+        mock_session.agency.return_value = model_TransitAgency
+        app_request.user = mocker.Mock(is_staff=True)
+
+        view.dispatch(app_request)
+
+        assert view.agency == model_TransitAgency
 
 
 class TestEligibleSessionRequiredMixin:
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -68,7 +68,9 @@ class BenefitsAdminSite(AdminSite):
                         "transit_processor_portal_url": transit_processor_portal_url,
                         "title": f"{agency.long_name} | {self.index_title} | {self.site_title}",
                         "start_url": (
-                            routes.IN_PERSON_ADDITIONAL_AGENCIES if agency.group_agencies() else routes.IN_PERSON_ELIGIBILITY
+                            routes.IN_PERSON_ADDITIONAL_AGENCIES
+                            if agency.group_agencies(only_active=False)
+                            else routes.IN_PERSON_ELIGIBILITY
                         ),
                     }
                 )
--- a/benefits/core/views.py
+++ b/benefits/core/views.py
@@ -239,11 +239,12 @@ class AdditionalAgenciesView(AgencySessionRequiredMixin, TemplateView):
     """View handler for nearby/additional providers."""
 
     template_name = "core/additional-agencies.html"
+    only_active_agencies = True
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         agency = self.agency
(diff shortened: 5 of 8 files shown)


benefits/core/models/transit.py
```diff
@@ -197,28 +197,34 @@ def customer_service_group_name(self):
...
-    def group_agency_short_names(self):
+
+        agencies_in_group = TransitAgency.objects.filter(
+            transitagencygroup__in=list(self.transitagencygroup_set.all())
+        ).distinct()
+
+        if only_active:
+            agencies_in_group = agencies_in_group.exclude(active=False)
+
+        return list(agencies_in_group.exclude(pk=self.pk).order_by("short_name"))
```
Scotchester: Suggestion: Move the self exclusion to the original `agencies_in_group` assignment so that the return statement is only adjusting the form (list and sorting) of what's being returned.

```suggestion
        ).distinct().exclude(pk=self.pk)

        if only_active:
            agencies_in_group = agencies_in_group.exclude(active=False)

        return list(agencies_in_group.order_by("short_name"))
```

Scotchester (dismissed):
This all looks and works great! In addition to the four testing scenarios you outlined in the description, I also confirmed that in-person enrollment for ungrouped CST works as expected.

I just left one stylistic suggestion in line, but totally cool if you prefer it as-is.

benefits/core/models/transit.py
lalver1: Thanks @Scotchester! I like how this makes it more readable, made the change in 59eadfa2c7bff402d14b88aed1a65c87fdc7c083 👍 

Also, thanks for testing with CST and confirming it works!

Scotchester (approved):

benefits/core/models/transit.py
```diff
@@ -197,28 +197,36 @@ def customer_service_group_name(self):
...
             TransitAgency.objects.filter(transitagencygroup__in=list(self.transitagencygroup_set.all()))
             .distinct()
-            .exclude(active=False)
             .exclude(pk=self.pk)
-            .order_by("short_name")
         )
 
-    def group_agency_short_names(self):
+        if only_active:
+            agencies_in_group = agencies_in_group.exclude(active=False)
```
jgravois, line 216: forgive the drive-by question, but why don't we want this method to sort results using 'short_name' when inactive agencies are filtered out?

is it because they aren't presented in the UI in a list?

Scotchester: In both cases they are sorted in the `return` statement below :)

jgravois: wow. the wrong one of us is taking a programming class this week. 😆

lalver1: No worries! 😄 All good questions, thanks for reviewing it!

angela-tran:
I was planning to review this today but didn't get to it yet, so I'll take a look at this tomorrow :eyes:

benefits/core/models/transit.py
angela-tran: I like this suggestion too. I think we could've gone further and moved `.order_by` up too so that the return statement simply triggers the evaluation of the QuerySet when it calls `list()`. Based on @jgravois's question in https://github.com/cal-itp/benefits/pull/3804#discussion_r3423619857, the `.order_by` being snuck in at the last second in the return statement could easily be missed when reading this code.

I guess you could argue that having the `.order_by` at the end makes it follow SQL syntax more. Anyways, I think I'm veering into minor-quibbles territory, so I'll say this is good and readable as-is -- no change needed.

benefits/core/mixins.py
```diff
@@ -16,15 +16,16 @@ class AgencySessionRequiredMixin:
...
 
-    If the session is not configured with an agency, return a user error.
+    Allows inactive agencies ONLY if the user is an authenticated staff member, otherwise returns a user error.
     """
 
     def dispatch(self, request, *args, **kwargs):
-        if session.active_agency(request):
-            self.agency = session.agency(request)
+        agency = session.agency(request)
+        if agency and (agency.active or getattr(request.user, "is_staff", False)):
```
angela-tran, line 24: @lalver1 and I noted yesterday while pairing on looking at this that this code assumes self-service will always only be working with active agencies and therefore the condition of `agency.active` will always be `True` for self-service and short-circuit the need to check the second condition of `getattr(request.user, "is_staff", False)`.

The question I had originally was if it's possible for `request.user` to be None, which I was thinking might raise an `AttributeError`:

```python
# just testing out what would happen if `request.user` were somehow `None`

> getattr(None, "is_staff")
Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'is_staff'
```

_edit: ok, so if you supply a default value, which our code does, the `AttributeError` does not get raised_
> `> getattr(None, "is_staff", False)`
> `False`

_end edit_

Testing it out myself in the app though, using a breakpoint at line 24, I see that:
- if you are _logged out_ of Benefits Admin, then go to the self-service side (`/`), and choose a transit agency from the drop-down (taking you to `/eligibility`), then `request.user` will be something like this:
```python
> request.user
<SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x7ee9c43ec7d0>>
```
and `getattr(request.user, "is_staff", False)` will return `False` and not throw an error.
- if you are _logged in_ to Benefits Admin, then go to `/`, and choose a transit agency from the drop-down (taking you to `/eligibility`), then `request.user` will be something like:
```python
> request.user
<SimpleLazyObject: <User: cst-user>>
```
and `getattr(request.user, "is_staff", False)` will return `True` and not throw an error.

Either way, `request.user` will be available and not throw an error, so all good.

angela-tran: Another separate note on this code: @lalver1 noted that he decided to use the `is_staff` attribute on `request.user` as the way to infer if the user is in Benefits Admin or not, as opposed to using the `is_authenticated` [property](https://github.com/django/django/blob/aa0efc9e221e023c21ca0ff17f1b1300a3d4ebe2/django/contrib/auth/models.py#L634-L636).

That decision was based off how confusing and self-contradictory the [Django documentation](https://docs.djangoproject.com/en/6.0/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated) on `is_authenticated` reads to us. You would think this method would be a reliable way to tell if the user is currently authenticated or not, but this documentation does not inspire confidence, and we didn't / haven't taken the time to simply test the behavior ourselves. (correct me if I'm wrong on that @lalver1)

Anyways, I think checking `is_staff` works just fine for what we need, but I wanted to note that technically, `is_staff` doesn't semantically tell you anything about the user's current authentication status in Benefits Admin; it simply tells you that this user is someone who is allowed to log in to Benefits Admin. The fact that the `User` we're inspecting is from the request is what makes it safe to assume they are authenticated.

angela-tran (commented):
Leaving some notes on the code changes for future reference

tests/pytest/core/test_mixins.py
```diff
@@ -25,24 +25,35 @@ def view(self, app_request):
...
 
-        assert view.agency == {"agency": "123"}
+        assert view.agency == model_TransitAgency
+
+    @pytest.mark.django_db
+    def test_dispatch_with_inactive_agency_staff_user(self, view, app_request, mocker, model_TransitAgency):
+        mock_session = mocker.patch("benefits.core.mixins.session")
+        model_TransitAgency.active = False
+        mock_session.agency.return_value = model_TransitAgency
+        app_request.user = mocker.Mock(is_staff=True)
```
angela-tran, line 52: Nice 💯

angela-tran:
## My testing of these changes

Thanks @lalver1 for [writing out the cases](https://github.com/cal-itp/benefits/pull/3804#issue-4669328779) to try out (_"Verify the following cases: ..."_). 

I wanted to clearly understand which of those cases were the ones broken in `main`, so I tested them all out in `main` before testing the fix in this branch. I also noticed we were missing a 5th case, namely the case of an inactive, non-regional agency.

### Behavior before the fix

_(Scenarios 1-4 are in the same order that @lalver1 wrote his cases in, just rephrased slightly.)_

| Scenario | Description | Working as expected? | Notes |
| --- | --- | --- | --- |
| 1 | 2 regional agencies, A and B - A is active, B is active | Yes :white_check_mark: | |
| 2 | 2 regional agencies, A and B - A is active, B is inactive | Yes :white_check_mark: | |
| 3| 2 regional agencies, A and B - A is inactive, B is active | No :x: | In-person enrollment is broken. :frowning_face:</br>The `/in_person/providers` page shows the "You may have reached this page on accident." screen. |
| 4 | 2 regional agencies, A and B - A is inactive, B is inactive | No :x: | In-person enrollment is broken. :frowning_face:</br>The `/in_person/providers` page is skipped (so doesn't throw an error, but is not the correct behavior), and then once you get past the eligibility check which tries to take you to `/in_person/enrollment`, you see the "You may have reached this page on accident." screen |
| 5 | 1 non-regional agency, C - C is inactive | No :x: | In-person enrollment is broken. :frowning_face: _(similar to Scenario 4)_</br>Once you get past the eligibility check which tries to take you to `/in_person/enrollment`, you see the "You may have reached this page on accident." screen|

<details><summary>
<em>Screenshot of the "user-error" page shown when trying to do inactive agency in-person enrollments</em>
</summary>
image
</details>


### Behavior after the fix

| Scenario | Description | In-person working as expected? | In-person testing notes | Self-service working as expected? | Self-service testing notes |
| --- | --- | --- | --- | --- | --- |
| 1 | 2 regional agencies, A and B - A is active, B is active | Yes - no regressions :white_check_mark: | | Yes - no regressions :white_check_mark: | |
| 2 | 2 regional agencies, A and B - A is active, B is inactive | Yes - no regressions :white_check_mark: | | Yes - no regressions :white_check_mark: |
| 3| 2 regional agencies, A and B - A is inactive, B is active | Yes - fixed :construction_worker: :hammer_and_wrench: :white_check_mark:  | <ul><li>`/in_person/providers` shows correctly</li><li>I was able to get through enrollment</li><li>success page **mentions the 2 active regional agencies**</li></ul> | Yes - no regressions :white_check_mark: | <ul><li>correctly shows the **1 active regional agency** in the drop-down</li><li>does not show the `/providers` page</li><li>the success page is **only for the 1 active regional agency**</li></ul> |
| 4 | 2 regional agencies, A and B - A is inactive, B is inactive | Yes - fixed :construction_worker: :hammer_and_wrench: :white_check_mark: | <ul><li>`/in_person/providers` shows correctly</li><li>I was able to get through enrollment</li><li>the success page **mentions the 2 active regional agencies**</li></ul> | Yes - no regressions :white_check_mark: | **neither inactive agency** is available in the dropdown, as expected |
| 5 | 1 non-regional agency, C - C is inactive | Yes - fixed :construction_worker: :hammer_and_wrench: :white_check_mark: | <ul><li>`/in_person/providers` shows correctly</li><li>I was able to get through enrollment</li><li>the success page is **only for the 1 active non-regional agency**</li></ul> | _N/A_ | _N/A_ |

angela-tran:
Something I realized as a part of all this testing is that there's a weird inconsistency we're introducing by supporting inactive agency in-person enrollment -- that is, on the self-service side, if you enroll for a regional agency whose agency group contains an inactive agency, you won't be told that your benefit will also work if you were to go tap-to-pay when riding public transit provided by the inactive agency. This is as opposed to the in-person enrollment experience, where you will be told. I guess since this whole inactive agency thing is more to support our Switchio production validation process where the hardware vendor does not hook their validator up to the ACC environment, and we don't anticipate actually doing long-term inactive agency in-person enrollment in `prod`, this is fine (for now).

angela-tran (approved):
Amazing work @lalver1! 👏 ❤️ 

Thanks for finding a way to make such a surgical and precise fix to get us past this issue without it being too hacky 🙂

jgravois:
> I guess since this whole inactive agency thing is more to support our Switchio production validation process where the hardware vendor does not hook their validator up to the ACC environment, and we don't anticipate actually doing long-term inactive agency in-person enrollment in prod, this is fine (for now).

i agree with this. i'm not sure if y'all tested it but i assume that the backdoor also opens us up to errors in the event that an inactive agency doesn't have enough information configured to be able to be activated.

if so, that's an acceptable trade-off too as far as i'm concerned.

angela-tran:
> i agree with this. i'm not sure if y'all tested it but i assume that the backdoor also opens us up to errors in the event that an inactive agency doesn't have enough information configured to be able to be activated.

Right, it's when saving a `TransitAgency` with `Active` set to true that a bunch of validation logic is run to make sure that all required configuration, including transit processor config, is in place. Good to call this out, and I agree this is a trade-off we are choosing to make.

lalver1:
Thanks for the reviews y'all! And thanks for the great documentation in this PR's comments @angela-tran; it's super helpful!

PR #3867, opened 2026-07-10
Refactor: N+1 query

### LINKED ISSUE #3859: Optimize database queries
We've seen Sentry report `N+1 Query` performance issues, with [this being the most recent one](https://cal-itp.slack.com/archives/C022HHSEE3F/p1781467473367939). It would be nice to modify our queries where needed so that we avoid these issues.

## Notes

- The [django-debug-toolbar](https://github.com/django-commons/django-debug-toolbar) is a handy tool to troubleshoot these types of issues since it can be ran locally to immediately see the raw SQL that is executed, and how changing the query affects performance. We could install the toolbar in the dev container if it will make it easier to review any PR related to this issue.
  Image
- Using the toolbar, it seems that the issue is related to the [`active_agencies` context processor](https://github.com/cal-itp/benefits/blob/main/benefits/core/context_processors.py#L45), although there may be other places too where database queries can be optimized.

## Acceptance Criteria

- [ ] The `N+1 Query` performance issue is avoided and we see no duplicate queries in the debug toolbar

lalver1:
Closes #3859

This PR installs the Django Debug Toolbar (DDT) to help diagnose `N+1 Query` issues. To turn on the toolbar, add a `DJANGO_DEBUG_TOOLBAR` environment variable to your `.env` and set it to `true`. It also makes the following changes to get rid of some redundant queries:

- `session` refactor 1 - `NULL` agency/flow (ceac1a1b51ac403568c9eded52138ba98c0f95bd and 223ea371ebf818e179ae1f89e0cd53b7ecae823e): If an agency or a flow isn't selected (the user hasn't reached that part of the application), we shouldn't query the database.
- `session` refactor 2 - `cache` agency/flow (b651c26f86f7e089ee96d700783b7acbe8437ec6 and d34230422a5fe0918ce8d94fcd59834083c2de5d): This one feels a bit hacky but it is restricted to only two places in the `session` module and it does remove a lot of redundant queries, especially in the `eligiblity` part of the application. Creating a sort of caching layer middleware would definitely be cleaner and maybe it's something we can come back to.
- Lazy evaluation (b4ad5597215f9d8f06a5083f26522487e694baa2): Simple change to remove casting the results of a QuerySet as a list (which forces a query evaluation) and keeping them as a QuerySet until the whole result needs to be evaluated.
- Use [`select_related()`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#select-related) (520f61ef59dc3be6a5ceb7a04f5d62f522ae9e5c): Implements a SQL JOIN in a single database query. It limits the related fields to `transit_processor_config` which is a `ForeignKey` and accessed by the `_agency_context` context processor.

## Interpreting SQL results from the DDT

To look for redundant queries, look for SQL items that say `Duplicated N times.`. For example, the screenshot below highlights a `SELECT` query duplicated 5 times in the DDT. Note that each instance (duplicate) of the query is shown as a unique item (the screenshot below has been truncated to two items, if you were to keep scrolling down you'd see all five items listed).

image

So a helpful method to troubleshoot redundant queries is:
- Look for items that show `Duplicated N times`.
- For each of these items, expand the result and look at the stack trace. The trace will show which `.py` file (or template) created the query.
- Typically the stack trace will be the same for each of the duplicates. These are the "easier" fixes. Sometimes though, the same query is generated at different places in the code so fixes need to be made in different places.
- After you make the fix, you should see the total number of SQL queries go down by the amount that had been shown in `Duplicated N times`.

## Reviewing

Turn on the DDT. Confirm the following drop in the number of queries by checking out the commit associated with each refactor and running through the application:

| page                              | `main` (or 950f3676b5f88611e264ec5759a0efe0537c82b7) | `NULL` agency/flow | `cache` agency/flow | lazy group agencies | `select_related()` |
|---------------------------------|--------|-----|-------|--------|--------|
| `/`                             |   25   | 16  |  16   |   14   |  5     |
| `/providers`                    |   35   | 31  |  24   |   20   |  11    |
| `/eligibility` (regional)       |   36   | 32  |  25   |   21   |  12    |
| `/eligibility`                  |   35   | 31  |  23   |   21   |  12    |
| `/eligibility/start` (regional) |   42   | 42  |  26   |   23   |  14    |
| `/eligibility/start`            |   43   | 43  |  25   |   23   |  14    |


benefits/settings.py
```diff
@@ -23,6 +23,9 @@ def _filter_empty(ls):
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DJANGO_DEBUG", "false").lower() == "true"
 
+# SECURITY WARNING: don't run with debug_toolbar turned on in production!
+DEBUG_TOOLBAR = os.environ.get("DJANGO_DEBUG_TOOLBAR", "false").lower() == "true"
```
thekaveman: I know you're still working on this. Just a question: any reason we can't just use the existing `DJANGO_DEBUG` env var / `DEBUG` setting, rather than introducing another env var / setting?

pyproject.toml
```diff
@@ -35,6 +35,7 @@ dev = [
     "flake8",
     "isort",
     "pre-commit",
+    "django-debug-toolbar"
```
thekaveman, line 38: I think we'd want to install this as part of the regular deps, right? In case we need/want to turn it on e.g. in the `dev` environment?

benefits/settings.py
lalver1: Yep, that's a good question. I originally had set it up that way, using the existing `DEBUG` setting to activate the toolbar. But the main problem is that the toolbar interferes with pytest since the toolbar adds itself to the templates and this made a few tests fail. So I introduced an environment variable for pytest, and depending on the combination of `DEBUG` and this new env variable the toolbar would automatically turn off when running the tests (both in VS Code and the coverage `run.sh` script). But this setup felt too complicated (I had to install the pytest-env plugin for pytest exclusively to use this new env variable).

So I thought it was probably better to turn on/off the toolbar independently of `DEBUG` since we run tests and examine the rendered page when verifying styles/layouts (and the toolbar gets in the way) more frequently than we use the toolbar. I envision most of the time a dev will have `DJANGO_DEBUG_TOOLBAR=false`, and will turn it on only for specific debugging scenarios.

pyproject.toml
lalver1: I also had the same question but I ended up thinking that we'll only want to use it locally. The toolbar has a setting called `INTERNAL_IPS` which may get tricky to configure for a Cloud environment, although I didn't look into it much. This setting also made me feel uneasy about turning it on in the Cloud, even if it's only for the `dev` environment.

But if we feel that it would be worthwhile and safe to turn on in `dev` I can definitely look more into the `INTERNAL_IPS` configuration.

benefits/settings.py
Scotchester: If we are open to having the toolbar on all the time locally, we could have it be automatically disabled when running tests: https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#disable-the-toolbar-when-running-tests-optional

But if we don't want it on all the time locally (as you were envisioning), we _could_ use [`SHOW_TOOLBAR_CALLBACK`](https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#show-toolbar-callback) to define some way to enable it that is possibly less onerous than changing an env var. The first idea that came to mind for me was using a query param like `?ddt` in the URL, but I realized it would be annoying to pass that from page to page. I suppose we could create a new config field in the DB for it, but that maybe feels too heavy-handed and not that much less onerous than updating an env var.

pyproject.toml
Scotchester: I could see it being convenient to turn it on in the `dev` environment, though I think it could expose secrets unless we figure out a way to only show it to authorized users. Probably not worth the effort.

benefits/settings.py
thekaveman: My thought is this is good enough for now, to get us started. Let's see how it is dealing with it for a few weeks before we try to engineer a more complex solution.

benefits/core/session.py
```diff
@@ -39,9 +39,17 @@
...
+    if not agency_id:
+        return None
+
+    if getattr(request, "_cached_agency", None) and request._cached_agency.id == agency_id:
+        return request._cached_agency
     try:
-        return models.TransitAgency.by_id(request.session[_AGENCY])
-    except (KeyError, models.TransitAgency.DoesNotExist):
+        agency = models.TransitAgency.by_id(agency_id)
+        request._cached_agency = agency
```
thekaveman, line 50: Is there a more "Django" way to do this caching? Attaching a random prop on to the `request` _feels_ a little hacky and like it would be easy to get out of control.

benefits/core/session.py
```diff
@@ -124,9 +132,17 @@ def enrollment_reenrollment(request):
...
+    if not flow_id:
+        return None
+
+    if getattr(request, "_cached_flow", None) and request._cached_flow.id == flow_id:
+        return request._cached_flow
     try:
-        return models.EnrollmentFlow.by_id(request.session[_FLOW])
-    except (KeyError, models.EnrollmentFlow.DoesNotExist):
+        flow = models.EnrollmentFlow.by_id(flow_id)
+        request._cached_flow = flow
```
thekaveman, line 143: Same question as above

benefits/core/session.py
lalver1: Completely agree, I had the same feeling. I'll do a bit more searching for a cleaner, more "Django" way to do this. I think it's worth the effort it since it does lower the duplicated queries by a good bit especially on the `/eligibility` part.

benefits/settings.py
```diff
@@ -107,6 +110,20 @@ def RUNTIME_ENVIRONMENT():
 if DEBUG:
     MIDDLEWARE.append("benefits.core.middleware.DebugSession")
 
+if DEBUG_TOOLBAR:
+    INSTALLED_APPS.append("debug_toolbar")
+    MIDDLEWARE.insert(0, "debug_toolbar.middleware.DebugToolbarMiddleware")
+    INTERNAL_IPS = ["127.0.0.1"]
```
Scotchester: I believe this has no effect when overriding `SHOW_TOOLBAR_CALLBACK` completely. If I comment it out or change the IP, DDT still appears.

```suggestion
```

If we weren't using Docker, I think we could _not_ override `SHOW_TOOLBAR_CALLBACK` and rely only on `INTERNAL_IPS`, but Docker makes that complicated. There's [a utility to try and help with that](https://github.com/django-commons/django-debug-toolbar/pull/2153), but it's not well documented and [apparently doesn't work with Docker Compose](https://github.com/django-commons/django-debug-toolbar/pull/2295).

benefits/settings.py
```diff
@@ -107,6 +110,20 @@ def RUNTIME_ENVIRONMENT():
...
+
+    def show_toolbar(request):
+        # Show the toolbar when in local development mode
+        return DEBUG
+
+    DEBUG_TOOLBAR_CONFIG = {
+        "SHOW_TOOLBAR_CALLBACK": show_toolbar,
+    }
```
Scotchester: This may not be worth the readability tradeoff, but you could simplify this with a lambda function:

```suggestion
    DEBUG_TOOLBAR_CONFIG = {
        "SHOW_TOOLBAR_CALLBACK": lambda request: DEBUG,
    }
```

benefits/settings.py
```diff
@@ -23,6 +23,9 @@ def _filter_empty(ls):
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = os.environ.get("DJANGO_DEBUG", "false").lower() == "true"
 
+# SECURITY WARNING: don't run with debug_toolbar turned on in production!
```
Scotchester: For extra caution?

```suggestion
# SECURITY WARNING: don't run with debug_toolbar turned on in any non-local environment!
```

I also wonder if it'd be nice to move this down with the activation code? Everything else up here are core Django settings.

pyproject.toml
Scotchester: It looks like it does a pretty good job of masking things that might be sensitive, but not sure if it's exhaustive. Still maybe not worth the risk.

Scotchester (commented):
A few comments on DDT loading before I dig into the meat of the PR.

Scotchester:
I agree with the feedback to look for a better caching mechanism, but besides that, the query optimizations are great!

benefits/core/session.py
lalver1: Unfortunately I couldn't find a more "Django way" of doing this. The closest I could get to it was to define a `SessionCache` middleware in 7ce78e3fb841ac8deb6e3c07cf0f6f759f3106ca to try to make it clear that the `request` object has 2 properties to hold cached (within a request/response cycle) session values. That way, this cache is not expressed only in the [session.py helper](https://github.com/cal-itp/benefits/blob/b651c26f86f7e089ee96d700783b7acbe8437ec6/benefits/core/session.py).

There are plenty of Django resources on caching backends and caching whole rendered templates and [model property caching](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.functional.cached_property), but none of these seemed to apply to our core issue that is using the session helper (with its database queries) in different parts of the application within the same request/response cycle.

I'm not sure if we want to try something else here 🤔

benefits/settings.py
lalver1: Good point, Scott, thanks. Re-read the docs more carefully and tested removing this line and the DDT still worked so I removed it in 990a8e6d48473871515201e59acd7d0764446838

benefits/settings.py
lalver1: I like it, thanks. Refactored it in 96044f62eba895ff3c2c039aa989652d07f45331 and also made a change to actually make the DDT independent from the value of `DEBUG` in 6c50254f858bf701a1d2da19eb538227755d7709. While I was at it, I noticed that I should've used `settings.DEBUG_TOOLBAR` instead of the module's variable since this is the more correct way to do it.

benefits/settings.py
lalver1: Good suggestion, I moved the code down in 209d6bb46851e84fd26c69033601e7d9bdfd1d31 to have it all close to the activation code.

benefits/core/session.py
thekaveman: Thank you for looking into this @lalver1. I don't know that the `SessionCache` middleware is adding a lot of value and it sort of obscures the actual work that is still happening in the session code. I think we can ideally handle this more cohesively as part of #2800. Your original approach is probably fine to get us started on cleaning up the DB queries.

Scotchester: Maybe an avenue to explore in #2800 is just caching more data on the session? Most of what we are looking to get when we re-query for a `TransitAgency` object rarely changes after production configuration, so I think it'd be pretty safe to just store it all at the start of the session.

lalver1: Thanks for the suggestions y'all and the reminder about #2800. I agree, the middleware did not add much on its own so I removed it in c77cac12aa799d922d394ead14dad78ba3b73e92

benefits/core/session.py
```diff
@@ -39,9 +39,17 @@
 
 def agency(request):
     """Get the agency from the request's session, or None"""
+    agency_id = request.session.get(_AGENCY)
```
jgravois, line 42: previously we called ~`request.session.get(_AGENCY)`~ `request.session[_AGENCY]` from within a try/catch to guard against a `KeyError`. based on my own local testing, i think its fine to be less defensive. 👍 

```python
request.session.get("invalid")
> None
```

benefits/core/session.py
```diff
@@ -39,9 +39,17 @@
...
+    if not agency_id:
+        return None
+
+    if getattr(request, "_cached_agency", None) and request._cached_agency.id == agency_id:
+        return request._cached_agency
     try:
-        return models.TransitAgency.by_id(request.session[_AGENCY])
-    except (KeyError, models.TransitAgency.DoesNotExist):
+        agency = models.TransitAgency.by_id(agency_id)
```
jgravois, line 49: entering this block indicates that the `session` is hydrated with an agency that doesn't match the cache.

this is an ignorant question, but perhaps we should attempt to correct that mismatch from within the `update()` function too?

benefits/core/session.py
jgravois, line 1: we should clear `request._cached_agency` explicitly from within the `reset()` function in this file too, yes?

jgravois (commented):
~70% reduction in SQL queries. awesome! 💃

benefits/core/session.py
lalver1: That's right @jgravois, thanks for checking this. And the other change is that we now call `.get()` on `request.session`, instead of directly using `request.session[_AGENCY]`. This way it returns `None` for a key which is not in the `request.session` dictionary-like object instead of raising a `KeyError`.

benefits/core/session.py
lalver1: Ah, the only thing is that in this block, the "cached session" (the `_cached_agency` property on the `request` object) is the one being hydrated. The `update()` function is actually updating the "real session", i.e., `request.session[_AGENCY]` which is different from the "cached session". Also, we're limiting ourselves to just caching reading the session, the `update()` function is about setting the session. Hopefully I didn't misunderstand your question, but if I did please let me know.

thekaveman: I think maybe another way to think about it (that Luis helped me understand) is the "cached session" is only for a single request: when different parts of the code (say, analytics, a middleware, a view, etc.) all need to get the agency out of the session during the same request -- this "caching" helps shortcut some duplicate DB calls.

The `update` / `reset` (i.e. the real session code) is for the lifetime of the user's session (across and between different requests), to track what agency/flow/etc. they have selected and where they are in the process.

benefits/core/session.py
lalver1: I think this is similar to the comment above? Where the `reset()` function is clearing `request.session[_AGENCY]` instead of `request._cached_agency`. Any changes to the "real session" will eventually affect `request._cached_agency` within the same request/response cycle.

benefits/core/session.py
jgravois: i don't want to belabor this because i trust both of your intuition here more than my own here, but just to be explicit, i'm suggesting the two _additional_ guards below:
```python
def update(
    # ...
):
    """Update the request's session with non-null values."""
    if agency is not None and isinstance(agency, models.TransitAgency):
        request.session[_AGENCY] = agency.id
        # in the unlikely event that a different agency was cached previously, nuke
        if getattr(request, "_cached_agency") and request._cached_agency.id != agency_id:
            del request._cached_agency
          
def reset(request):
    """Reset the session for the request."""
    logger.debug("Reset session")
    request.session[_AGENCY] = None
    del request._cached_agency
```

it sounds like you're both saying that there's no need for this because `_cached_agency` is something derived on the fly for each individual request.

if this is indeed the case, why do i see evidence that `_cached_agency` is hydrated within `update()` when selecting a flow and navigating from /eligibility to /eligibility/start?

```python
if flow is not None and isinstance(flow, models.EnrollmentFlow):
    request._cached_agency
    # i was suprised to see 👆 hydrated
```
when i inspected this i made sure to set another breakpoint inside `def agency()` and i'm reasonably certain its not generating a fresh cached value immediately beforehand.

happy to chat more if it'd be helpful, but if none of what i'm saying makes sense or its just too "boot and suspenders", we can just put a pin in it too. i 100% do _not_ have a pulse on a sequence of interactions that leaves us prone to corrupt sessions or stale/mismatched cache.

thekaveman: Thanks for pushing on this John, I think it is important we all have clarity here since the session is so central to how the app works.

> because `_cached_agency` is something derived on the fly for each individual request

I think this is the crux of the issue, and yes this is correct: `_cached_agency` lives and dies for a single request/response cycle. The idea is, within generation of a given response, we may be asking the DB multiple times for the same information (e.g. the agency that was stored in session at the beginning of the cycle).

`_cached_agency` shortcuts subsequent asks for that particular cycle. It says nothing about the _next_ request/response cycle.

> why do i see evidence that `_cached_agency` is hydrated within `update()` when selecting a flow and navigating from `/eligibility` to `/eligibility/start`?

If you look at the _very first time_ `session.agency` is called in the navigation from `/eligibility` to `/eligibility/start`, `_cached_agency` is indeed `None` and is set at that moment.

Here's an illustration (I tried to be accurate with the details, forgive me if I missed something):

## Consider a number of request/response cycles

- `cycle1`: Landing on `/`
- `cycle2`: Submitting agency choice and landing on `/eligibility/`
- `cycle3`: Submitting flow choice and landing on `/eligibility/start/`

### `cycle1`

1. `core.IndexView.get` calls `session.reset()`
2. `context_processors.agency` calls `session.agency`, returning None
3. `analytics.ViewedPageEvent` calls `session.agency`, returning None
4. DB call to get all agencies

### `cycle2`

1. `core.IndexView.form_valid` calls `session.reset()` (again)
2. `core.IndexView.form_valid` calls `session.update()` to store the agency ID
3. `core.AgencySessionRequired` middleware (from `eligibility.IndexView`) checks the agency stored in session (hydrating) to ensure one was selected; **`request._cached_agency` is set here**
4. `context_processors.agency` calls `session.agency`, returning `request._cached_agency`
6. `eligibility.IndexView.get` uses the already hydrated `self.agency` (from `AgencySessionRequired`) to get its flows from the DB
7. `analytics.ViewedPageEvent` calls `session.agency`, using `request._cached_agency`

### `cycle3`

_This is a new request, hence `request._cached_agency` is `None`_

1. `core.AgencySessionRequired` middleware (from `eligibility.IndexView`) checks the agency stored in session (hydrating) to ensure one was selected; **`request._cached_agency` is set here**
2. `eligibility.IndexView.form_valid` calls `session.update()` to store the flow ID
3. `analytics.SelectedFlowEvent` calls `session.agency`, returning the `_cached` value
4. `core.FlowSessionRequired` middleware (from `eligibility.StartView`) checks the flow stored in session (hydrating) to ensure one was selected; **`request._cached_flow` is set here**
5. `context_processors.agency` calls `session.agency`, returning `request._cached_agency`
6. `eligibility.StartView.get` calls `session.update()` to store `eligible=False`

It seems your debug point was on step 6. of `cycle3`, at which point `session.agency` has already been called multiple times during that cycle, hence the cached value being hydrated.

thekaveman: > happy to chat more if it'd be helpful

Me too BTW, if it would be more helpful to walk through this in a live debug session (no pun intended :wink:)

thekaveman: And I just have to state again for the record, it took @lalver1 explaining all of this to me 1:1 on a call to get it sunk in -- I was definitely not following this change before he did that. So I was right there with you @jgravois!

jgravois: _very_ elucidating. thank you!

i'll play around with the debugger some more tomorrow to make sure i can traverse the code paths you've outlined confidently. if not, I'd love to take a look at this together at LACI sometime soon. 💛

jgravois (approved):
thanks for humoring me y'all. good learning opportunity for me. this PR is a _big_ step in the right direction as far as i'm concerned.

benefits/core/session.py
lalver1: Not at all John, these are very good points, thanks for being thorough. This discussion has helped me too to double check this approach.

And thanks for the explanation Kegan, I was writing this up and didn't get a chance to post it yesterday.

I'll just add a few more notes in case these help:

* In cycle 2 (but similar behavior happens in cycle 3, step 6), `_cached_agency` is already hydrated inside `update()` because on `/eligibility`, [`IndexView` inherits the `AgencySessionRequiredMixin`](https://github.com/cal-itp/benefits/blob/main/benefits/eligibility/views.py#L22) early on when processing the request, and in that Mixin we call [`session.agency(request)`](https://github.com/cal-itp/benefits/blob/main/benefits/core/mixins.py#L23). This sets `request._cached_agency`, so when in `IndexView` we get to the actual code the view runs in `get`, we see that `_cached_agency` is already set when you step into `session.update()`.
* The two additional guards are a good observation, but are probably not needed. For example, in `update()`, this function updates `request.session[_AGENCY] = agency.id`. If during the same cycle there is a call to `session.agency(request)`, this function acts as its own guard because the newly updated `agency_id` from the session won't match the old `_cached_agency.id`, the function bypasses the cache, does a database lookup, and overwrites `request._cached_agency` with the new object. For `reset()` it is similar, it sets `request.session[_AGENCY] = None` and if any there are any calls to `session.agency(request)`, the function immediately aborts and returns `None` before even looking at the cache since `agency_id` will be `None`.
* It's important though that everywhere in the code we keep on only using `session.agency(request)` to get the agency because this ensures that the cached agency is up to date given the checks inside `session.agency(request)`. We should not use `request._cached_agency` directly since we'd be bypassing these checks and we could end up using a stale value.

We can definitely look at this in a call too and stepping through with the debugger, just let me know!

Scotchester: Super helpful discussion; thanks all!

Issue #3897, opened 2026-07-20
Code changes to move infrastructure CI/CD from DevOps Pipelines to GitHub Actions


lalver1:
Move the CI/CD from the Azure DevOps Pipelines defintion in 

- `terraform/pipeline/deploy.yml`
- `terraform/pipeline/tag.py`
- `terraform/pipeline/workspace.py`
- `terraform/azure-pipelines.yml`

to the GH Actions ["Deploy"](https://github.com/cal-itp/benefits/blob/main/.github/workflows/deploy.yml) workflow, and add a "Terraform plan" workflow to handle previewing infrastructure updates via `terraform plan`. This new "Terraform plan" workflow substitutes the automatic Azure DevOps Pipeline trigger that currently runs on all pull requests.

## Acceptance Criteria

- [x] Add a GH Actions "Terraform plan" workflow
- [x] Update the GH Actions ["Deploy"](https://github.com/cal-itp/benefits/blob/main/.github/workflows/deploy.yml) workflow
- [x] Import into Terraform the `azurerm_key_vault_access_policy`s for the new Service Principals/Security Group, similar to the current [Service Principal](https://github.com/cal-itp/benefits/blob/main/terraform/key_vault.tf#L82) or [Engineering Security Group](https://github.com/cal-itp/benefits/blob/main/terraform/key_vault.tf#L68).
- [x] Remove redundant Azure DevOps Pipelines-related code
- [ ] Remove outdated Azure DevOps Pipelines GitHub secrets

jgravois:
i'll branch off of #3913 now to deprecate the code that underpins our ADO pipeline runs.

PR #3902, opened 2026-07-21
chore: run 'terraform plan' via github action

jgravois:
adds a new pull_request driven github action to run `terraform plan` anytime the contents of the `terraform/` directory have changed and disables the pull request triggered ADO pipeline run that did the same thing formerly.

part of #3897

as part of this effort, DevSecOps created a new service principal and federated credentials on our behalf and we worked together to grant the runner needed privileges.

### DIFF
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -43,4 +43,4 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ github.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
-        run: terraform plan -lock=false
+        run: terraform plan
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -43,4 +43,4 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ github.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
-        run: terraform plan
+        run: terraform plan -lock-timeout=5m
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -1,0 +1,46 @@
+name: Terraform plan
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "terraform/**"
+      - ".github/workflows/terraform-plan.yml"
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  terraform-plan:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: terraform
+
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Log in to azure using federated identity credentials
+        uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
+        with:
+          client-id: ${{ secrets.AZURE_SP_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_SP_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+
+      - name: Setup terraform
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4
+        with:
+          terraform_version: 1.14.5
+
+      - name: Initialize terraform and select workspace
+        run: ./init.sh "dev"
+
+      - name: Terraform Plan
+        env:
+          TF_VAR_ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+          TF_VAR_CONTAINER_TAG: ${{ github.sha }}
+          TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
+          TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+        run: terraform plan -lock=false
--- a/terraform/azure-pipelines.yml
+++ b/terraform/azure-pipelines.yml
@@ -1,6 +1,6 @@
 trigger: none
-# it will be triggered "manually" from a GitHub Actions workflow
-# but will still automatically run on pull requests
+pr: none
+# all previous triggers are now disabled but "manual" runs from a GitHub Actions workflow are still supported
 # https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/trigger?view=azure-pipelines#trigger-none
 
 stages:
--- a/terraform/azure-pipelines.yml
+++ b/terraform/azure-pipelines.yml
@@ -1,7 +1,7 @@
 trigger: none
 pr: none
-# all previous triggers are now disabled but "manual" runs from a GitHub Actions workflow are still supported
-# https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/trigger?view=azure-pipelines#trigger-none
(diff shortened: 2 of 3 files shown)


.github/workflows/terraform-plan.yml
jgravois, line 1: in this new action i took the route of pinning each dependency to a specific SHA as that is now a security best-practice.

my thinking is that we can/should hold off on expanding this pattern in other workflows until after we see just how seamlessly dependabot is able to suggest updates as additional releases of the dependencies here are tagged.

.github/workflows/terraform-plan.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Initialize terraform and select workspace
+        run: ./init.sh "dev"
+
+      - name: Terraform Plan
+        env:
+          TF_VAR_ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+          TF_VAR_CONTAINER_TAG: ${{ github.sha }}
+          TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
+          TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+        run: terraform plan -lock=false
```
jgravois: setting `-lock=false` just ensures that the plan run on CI/CD doesn't prevent anyone anywhere else from running a plan simultaneously.

thekaveman: Is this correct? I think we have always had locking plans (the default, right?)

I.e. the ADO pipeline wasn't set up to _not_ lock during a plan? And when devs run plan locally, we don't typically/ever? include `-lock=false`, at least to my knowledge.

We have these docs that speak to addressing lock situations: https://docs.calitp.org/benefits/guides/troubleshooting/#terraform-lock

Have we discussed wanting to turn this off now? I'm fuzzy on all the implications of locking vs. not.

jgravois: > I'm fuzzy on all the implications of locking vs. not.

this isn't a hill i'd die on by any means, but the documentation you linked to is a great example of why locking is such a PITA.

anytime you _might_ act on a `plan` and use the information to make changes to a deployed environment. i'd call it a necessary evil. locking is what ensures you can trust that terraform will stick to the plan and that no one else can pull the rug out from underneath you.

in the case of the pipeline and this new github action. i consider the `terraform plan` to be an ephemeral/throwaway artifact taken at a snapshot in time.

ie: it wouldn't be problematic to rerun the job and see a different output if someone else deployed a change of their own nearly simultaneously and there's no reason why the creation of these plans should block anyone else who happens to be considering deploying a change of their own (imo).

that said, if we keep the default locking behavior, 99% of the time the lock will be assigned and cleared automatically. the other 1% of the time we can either wait for the lock to clear on its own (or nuke it if somebody 💩s the 🛏️).

thekaveman: Yeah I agree the linked docs make this seem like an annoying problem that has to be dealt with. I don't recall more than like, once or twice, ever having to actually deal with it. 

I'm not wanting to die on this hill either FWIW. 

> anytime you _might_ act on a plan and use the information to make changes to a deployed environment. i'd call it a necessary evil. locking is what ensures you can trust that terraform will stick to the plan and that no one else can pull the rug out from underneath you.

I will just say that, by definition a PR `plan` is something we _will_ act on by _merging the PR_: `dev` will be get the `apply` as soon as the merge completes. So these aren't quite as ephemeral in my mind, especially with the path filters... if a `plan` runs, it signals that an `apply` will run on merge. Right?

> it wouldn't be problematic to rerun the job and see a different output if someone else deployed a change of their own nearly simultaneously and there's no reason why the creation of these plans should block anyone else who happens to be considering deploying a change of their own

Yup agreed.

I guess maybe the issue I'm considering (which, on reflection/writing this out, isn't necessarily related to this PR/change) is: if one assumes their (stale) `plan` from before another `plan`/merge with changes was still correct and merged, only to discover their `plan`/`apply` now actually fails for some reason. I know you're working on the `apply` part of this now. Probably more relevant there. Maybe this is another argument for requiring branches to be up to date? I dunno. We can also just try this out for a while and see how it goes... it's not like the infra changes ALL that often.

jgravois: > a PR plan is something we will act on by merging the PR

i take your point, but most of the time _many_ minutes will pass between the plan associated with the last commit pushed to a PR that touches infrastructure and when that change is actually applied on merge.

regardless, i did a little more research and decided an abundance of caution is appropriate here. i had forgotten that even a plan writes to terraform state files, and not taking a lock opens us up to the (small) possibility of corruption. a32a5cdb304a8236260cb8ed1990b30375800a0f

https://developer.hashicorp.com/terraform/language/state/locking

lalver1: These are all good points, and I agree that out of an abundance of caution we can keep `plan`'s default to lock the state. Since we rarely had any issues with `plan` in ADO, I expect this behavior to carry on, especially since rarely there have been concurrent changes to the infra. I would suggest though, that we add [`-lock-timeout=5m`](https://developer.hashicorp.com/terraform/cli/commands/plan#lock-timeout-duration) like we [originally had in the ADO Pipeline](https://github.com/cal-itp/benefits/blob/main/terraform/pipeline/deploy.yml#L53). Setting this option is probably a good way to reduce the chance of getting ourselves in the [situation described in our docs](https://docs.calitp.org/benefits/guides/troubleshooting/#terraform-lock).

terraform/azure-pipelines.yml
```diff
@@ -1,6 +1,6 @@
 trigger: none
-# it will be triggered "manually" from a GitHub Actions workflow
-# but will still automatically run on pull requests
+pr: none
+# all previous triggers are now disabled but "manual" runs from a GitHub Actions workflow are still supported
```
lalver1: In the comment maybe we can say something like

```
# automatic pull request validation triggers are also now disabled but "manual" runs from a GitHub Actions workflow are still supported
# https://learn.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#pr-triggers
```

to remind ourselves that this change was focused on turning off the automatic pull request validation automatically ran by the Azure Pipelines GitHub App.

.github/workflows/terraform-plan.yml
```diff
@@ -0,0 +1,46 @@
...
+          tenant-id: ${{ secrets.AZURE_SP_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+
+      - name: Setup terraform
+        uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4
+        with:
+          terraform_version: 1.14.5
+
+      - name: Initialize terraform and select workspace
+        run: ./init.sh "dev"
```
lalver1, line 38: When this script (that's mostly meant for running terraform from a local machine) runs in this workflow, we might be running redundant commands. Such as getting the subscription via `az account list` and setting the subscription via `az account set`, since we already authenticated in the "Log in to azure using federated identity credentials" step. But, maybe that's ok because on the other hand, running the script keeps some of the logic DRY. I don't feel strongly either way so we can keep it like this 👍 Just wanted to note it in case we do come up with other reasons to bypass using `init.sh` which I might be overlooking now.

jgravois: you're not wrong. i've been operating under the assumption that its safe because the federated credential only has access to the subscription that we want it to use, but it _is_ a little overkill.

personally it feels like a good place to start to me, but i'm definitely open to breaking out the explicit commands needed if anyone else feels it would add value.

.github/workflows/terraform-plan.yml
jgravois: good call. f57ef83e83f64a0c576cb84d2850d24120062f52

terraform/azure-pipelines.yml
jgravois: i dig it 05fc05cd9ec28fff26a9a6f4539caf46322ebbbc

.github/workflows/terraform-plan.yml
lalver1: I agree, it's a good place to start, especially since we know that the `azure/login` step does not interfere with `init.sh`, so running those extra `az` commands in `init.sh` doesn't hurt.

lalver1:
Nice job, this looks good to me @jgravois! I think that if we merge this right now, the `terraform plan` will fail due to an active lock, probably related to the `terraform apply` PR testing. That's the only reason why I hesitate to approve it, we'll miss an opportunity to get a clean workflow run since we may need to manually intervene to remove the lock first.

I have to step out for the afternoon but I wanted to mention this so we can look at it in the morning.

jgravois:
> I think that if we merge this right now, the terraform plan will fail due to an active lock, probably related to the terraform apply PR testing.

that was certainly correct for much of this afternoon. 😅 I had a whole slew of hung attempts to `terraform apply` in https://github.com/cal-itp/benefits/pull/3913 that resulted in locks that needed to be cleared manually. all is well now though. i just confirmed by re-running the job associated with the most recent commit here manually.

https://github.com/cal-itp/benefits/actions/runs/30042334889

lalver1 (approved):
Great job @jgravois!

Issue #3892, opened 2026-07-17
Define new GitHub Actions environment-dependent secrets and variables


lalver1:
The following _environment_-dependent secrets will be used by GitHub Actions workflows to select the appropriate federated credentials, Terraform state file, and Terraform workspace (the environments need to match the configuration of the federated credentials of the Service Principals in #3890):

| Name                 | Environment | Notes                                                       | Created?  |
|----------------------|-------------|-------------------------------------------------------------|---|
| `AZURE_SP_CLIENT_ID` | `dev`       | Development Service Principal/Federated Credential | ✅ |
| `AZURE_SP_CLIENT_ID` | `test`      | Test Service Principal/Federated Credential | ✅ |
| `AZURE_SP_CLIENT_ID` | `prod`      | Production Service Principal/Federated Credential | ✅  |

The following _repository_-level secrets need to be set for the "Terraform Plan" workflow (since the "Terraform Plan" workflow does not set a GitHub Actions environment) and for the updated "Deploy" workflow since the Azure subscription and tenant ID are global for all Service Principals/Federated Credentials. The DevSecOps Service Principal and Cal-ITP Engineering AD group object IDs are also independent of environment:

| Name                       | Notes                                                              | Created? |
|----------------------------|--------------------------------------------------------------------|---|
| `AZURE_SP_CLIENT_ID`       | Pull Request / Terraform Plan Service Principal/Federated Credential    | ✅ |
| `AZURE_SP_SUBSCRIPTION_ID` | same value applies to all Service Principals/Federated Credentials | ✅ |
| `AZURE_SP_TENANT_ID`       | same value applies to all Service Principals/Federated Credentials | ✅ |
| `TF_VAR_DEVSECOPS_OBJECT_ID` | object ID of the DevSecOps Service Principal | ✅ |
| `TF_VAR_ENGINEERING_GROUP_OBJECT_ID` | object ID of the Cal-ITP engineering Active Directory Group | ✅ |

The following _environment_-dependent secret and variable are temporarily used to deploy the App Service to the `dev` environment, both will be removed when we fully cut over to Container Apps:

| Name                 | Environment | Notes                                                       | Created?  |
|----------------------|-------------|-------------------------------------------------------------|---|
| `AZURE_WEBAPP_PUBLISH_PROFILE` | `dev` | Used to deploy the App Service | ✅ |
| `AZURE_WEBAPP_NAME`            | `dev` | Used to deploy the App Service | ✅ |

[...]

PR #3913, opened 2026-07-23
chore: run 'terraform apply' via github action

jgravois:
part of #3897

https://github.com/cal-itp/benefits/actions/runs/30120264772 deployed 2b7a751 to the `dev` [container app](https://ca-cdt-pub-vip-calitp-d-web.wittyplant-56d5c43a.westus.azurecontainerapps.io/static/sha.txt).

i feel good enough about the changes here to merge and tag an RC after verifying that the deploy to `main` is successful, but we could easily create an RC tag on this branch too.

## artifacts

I didn't want to propose too many changes here, but i plan to investigate persisting the output of either terraform plan or apply as a [release artifact](https://docs.github.com/en/actions/concepts/workflows-and-actions/workflow-artifacts) in a subsequent PR.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,6 +15,10 @@ defaults:
   run:
     shell: bash
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions:
   id-token: write
   contents: write
@@ -86,9 +90,6 @@ jobs:
         check-dynamic-version,
       ]
     if: (!cancelled())
-    concurrency:
-      group: ${{ needs.setup.outputs.env_name }}
-      cancel-in-progress: true
     environment: ${{ needs.setup.outputs.env_name }}
     steps:
       - name: Checkout
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,40 +15,81 @@ defaults:
   run:
     shell: bash
 
-concurrency:
-  # this ternary operator like expression gives us the name of the deployment environment (see https://docs.github.com/en/actions/learn-github-actions/expressions#example)
-  group: ${{ github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}
-  cancel-in-progress: true
+permissions:
+  id-token: write
+  contents: write
+  pull-requests: write
 
 jobs:
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      env_name: ${{ steps.set-env.outputs.env_name }}
+      is_tag: ${{ steps.set-env.outputs.is_tag }}
+      is_release: ${{ steps.set-env.outputs.is_release }}
+    steps:
+      - id: set-env
+        run: |
+          # 1. Determine if its a tag
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            IS_TAG="true"
+          else
+            IS_TAG="false"
+          fi
+          echo "is_tag=$IS_TAG" >> $GITHUB_OUTPUT
+
+          # 1. Determine environment name
+          if [[ $IS_TAG == "false" ]]; then
+            ENV="dev"
+          elif [[ "${{ github.ref }}" == *"-rc"* ]]; then
+            ENV="test"
+          else
+            ENV="prod"
+          fi
+          echo "env_name=$ENV" >> $GITHUB_OUTPUT
+
+          # 2. Determine if it's a final release tag
+          if [[ "$IS_TAG" == "true" && "${{ github.ref }}" != *"-rc"* ]]; then
+            echo "is_release=true" >> $GITHUB_OUTPUT
+          else
+            echo "is_release=false" >> $GITHUB_OUTPUT
+          fi
+
   tests-ui:
+    needs: [setup]
     uses: ./.github/workflows/tests-ui.yml
-    if: github.ref_type == 'tag'
+    if: ${{ needs.setup.outputs.is_tag == 'true' }}
 
   tests-pytest:
+    needs: [setup]
     uses: ./.github/workflows/tests-pytest.yml
-    if: github.ref_type == 'tag'
+    if: ${{ needs.setup.outputs.is_tag == 'true' }}
 
   check-migrations-and-messages:
+    needs: [setup]
     uses: ./.github/workflows/check-migrations-and-messages.yml
-    if: github.ref_type == 'tag'
+    if: ${{ needs.setup.outputs.is_tag == 'true'}}
 
   check-dynamic-version:
+    needs: [setup]
     uses: ./.github/workflows/check-dynamic-version.yml
-    if: github.ref_type == 'tag'
+    if: ${{ needs.setup.outputs.is_tag == 'true' }}
 
   deploy:
     runs-on: ubuntu-latest
     needs:
       [
+        setup,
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -103,74 +144,47 @@ jobs:
...
-            elif [[ "$PIPELINE_STATUS" == "cancelling" || "$PIPELINE_STATUS" == "notStarted" || "$PIPELINE_STATUS" == "inProgress" ]]; then
-              sleep 10
-            else
-              echo "Azure DevOps Pipeline finished with unexpected status: $PIPELINE_STATUS"
-              exit 1
-            fi
-          done
+      - name: Terraform apply
+        working-directory: terraform
+        run: terraform apply -lock-timeout=5m tfplan
```
jgravois, line 183: initially i expected setting `-input=false` (as we do in [terraform/pipeline/deploy.yml](https://github.com/cal-itp/benefits/blob/178c5df2acb9dd87a19c483337755754841382b5/terraform/pipeline/deploy.yml#L70)) to disable the interactive input and make it possible to apply changes without typing 'yes' as a confirmation. this is not the case.

the terraform documentation suggests that [saved plan mode](https://developer.hashicorp.com/terraform/cli/commands/apply#saved-plan-mode) is considered preferable in CI/CD, so i took that tack, but i'd be fine with using `-auto-approve` instead too if anyone else thinks that's a better idea.

.github/workflows/deploy.yml
```diff
@@ -15,40 +15,81 @@ defaults:
...
+            echo "is_release=true" >> $GITHUB_OUTPUT
+          else
+            echo "is_release=false" >> $GITHUB_OUTPUT
+          fi
+
   tests-ui:
+    needs: [setup]
     uses: ./.github/workflows/tests-ui.yml
-    if: github.ref_type == 'tag'
+    if: ${{ needs.setup.outputs.is_tag == 'true' }}
```
jgravois, line 64: the technique i'm proposing here to centralize defining the environment name and determining whether or not the run is associated with the tag push is a bit awkward (writing conditional logic in bash, only being able to pass along outputs as strings).

if anyone else has an idea to improve, i'm all ears. 👂

.github/workflows/deploy.yml
```diff
@@ -16,39 +16,81 @@ defaults:
     shell: bash
 
 concurrency:
-  # this ternary operator like expression gives us the name of the deployment environment (see https://docs.github.com/en/actions/learn-github-actions/expressions#example)
-  group: ${{ github.ref_type != 'tag' && github.ref_name || contains(github.ref, '-rc') && 'test' || 'prod' }}
+  group: ${{ github.workflow }}-${{ github.ref }}
```
jgravois, line 19: based on what i'm reading about [concurrency](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency) i believe it will result in equivalent behavior to skip the custom logic related to tags and just name the concurrency group by concatenating the workflow and the SHA, but if i'm missing something we could easily revert (we just can't access the new shared logic in 'setup').

.github/workflows/deploy.yml
lalver1: This looks good to me. I'm seeing if there's another way to do this too, but if there isn't, I don't think that the bash code is too complicated so this should be fine.

.github/workflows/deploy.yml
```diff
@@ -16,39 +16,81 @@ defaults:
...
   cancel-in-progress: true
 
+permissions:
+  id-token: write
+  contents: write
+  pull-requests: write
```
lalver1: I was wondering if we need the `pull-requests` permission for this workflow once we merge it to `main`? I guess we might have been using it to test this PR.

.github/workflows/deploy.yml
lalver1: I think defining the concurrency group this way will work 👍 The main scenario we want to be careful about is pushing to `main`. If a PR gets merged, deploy is called and the concurrency group is `deploy-refs/heads/main`. Then, if another PR is merged, the current `deploy` will cancel and the newest one will start. For deploying tags, it'll still work but is less critical since it is very unlikely that we'll get 2 or more workflows running concurrently since the concurrency group looks like`deploy-refs/tags/2026.07.1` and tags don't get repeated. (I was talking out loud mostly to make sure I understood the `concurrency` key 😅).

lalver1 (commented):
Looks great @jgravois! I only had one question, the rest makes sense to me 👍 .

.github/workflows/deploy.yml
jgravois: this is a great catch. i encountered an error indicating that the permission was needed for one of the prerequisite steps and it didn't occur to me that it was only because i was temporarily running the workflow on push to the `chore/gh-action-apply` branch.

fixed in f520a0e89

.github/workflows/deploy.yml
```diff
@@ -16,39 +16,80 @@ defaults:
...
+          if [[ $IS_TAG == "false" ]]; then
+            ENV="dev"
+          elif [[ "${{ github.ref }}" == *"-rc"* ]]; then
+            ENV="test"
+          else
+            ENV="prod"
+          fi
+          echo "env_name=$ENV" >> $GITHUB_OUTPUT
+
+          # 2. Determine if it's a final release tag
```
thekaveman: Nit, but the numbering in these comments is off (1., 1., 2.)

.github/workflows/deploy.yml
```diff
@@ -16,39 +16,80 @@ defaults:
...
+  setup:
+    runs-on: ubuntu-latest
+    outputs:
+      env_name: ${{ steps.set-env.outputs.env_name }}
+      is_tag: ${{ steps.set-env.outputs.is_tag }}
+      is_release: ${{ steps.set-env.outputs.is_release }}
+    steps:
+      - id: set-env
+        run: |
+          # 1. Determine if its a tag
```
thekaveman:
```suggestion
          # 1. Determine if it's a tag
```

.github/workflows/deploy.yml
```diff
@@ -16,39 +16,80 @@ defaults:
...
+            ENV="dev"
+          elif [[ "${{ github.ref }}" == *"-rc"* ]]; then
+            ENV="test"
+          else
+            ENV="prod"
+          fi
+          echo "env_name=$ENV" >> $GITHUB_OUTPUT
+
+          # 2. Determine if it's a final release tag
+          if [[ "$IS_TAG" == "true" && "${{ github.ref }}" != *"-rc"* ]]; then
```
thekaveman: Would this be implied from the `ENV` above being `prod` or no?

thekaveman (dismissed):

.github/workflows/deploy.yml
jgravois: > Would this be implied from the `ENV` above being `prod`

indeed! 9b23cec19

lalver1 (approved):
This looks ready @jgravois! I looked over it one last time and found no issues. Our GitHub environment secrets are ready too as documented in #3892.

.github/workflows/deploy.yml
jgravois: turns out this _was_ needed. https://github.com/cal-itp/benefits/pull/3917  😇

if i had understood the landscape better yesterday i would have realized that the reason this wasn't flagged in f520a0e is because at that point the workflow wasn't running each and every time i pushed a commit to `chore/gh-action-apply` anymore.

PR #3916, opened 2026-07-27
chore: remove unused ADO pipeline code and update relevant documentation

jgravois:
part of #3897

* ~destroy the (now unneeded) key vault access policy for our defunct ADO pipeline user~
* upgrade to the latest stable `terraform` release
* update relevant documentation to refer to github actions/container apps consistently
* remove the outdated `restart-app` bash script


.github/workflows/deploy.yml
```diff
@@ -175,8 +175,7 @@ jobs:
         env:
           TF_VAR_ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
           TF_VAR_CONTAINER_TAG: ${{ github.sha }}
-          TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
-          TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+          TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ vars.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
```
jgravois: `terraform plan` is failing here because i haven't made the change ☝️ yet.

i propose that we make `TF_VAR_ENGINEERING_GROUP_OBJECT_ID` a github actions variable (rather than a secret) so that contributors can see the associated value (in the interest of hydrating their own `terraform.tfvars` file).

docs/explanation/infrastructure.md
```diff
@@ -160,7 +160,7 @@ Use the following shorthand for conveying the Resource Type as part of the Resou
    ./init.sh <env>
    ```
 
-1. Create a local `terraform.tfvars` file (ignored by git) from the sample; fill in the `*_OBJECT_ID` variables with values from the Azure Pipeline definition.
+1. Create a local `terraform.tfvars` file (ignored by git) from the sample; `ARM_SUBSCRIPTION_ID` refers to our <kbd>CDT/ODI Production</kbd> Azure subscription. `ENGINEERING_GROUP_OBJECT_ID` is stored as a GitHub actions environment variable.
```
jgravois: see https://github.com/cal-itp/benefits/pull/3916/changes#r3661428039 for more info about fully enacting the proposed change.

docs/tutorials/secrets.md
```diff
@@ -44,4 +44,4 @@ The steps are:
 The effects of following those steps should be:
 
 - A restart of the App Service is triggered.
-- The next time that our Azure infrastructure pipeline is run, the value of `change_me_to_refresh_secrets` is set back to the value defined in our Terraform file for the App Service resource.
+- The next time that GitHub action deploy workflow is run, the value of `change_me_to_refresh_secrets` is set back to the value defined in our Terraform file for the App Service resource.
```
thekaveman: Funny this note is still hanging around. 

We deleted this secret back in #1948 / #2138, so I think this could be removed entirely.

jgravois: how could i have forgotten! https://github.com/cal-itp/benefits/pull/3717

jgravois: > I think this could be removed entirely.

done in ~71a9ed7ca~ f438e0a

.github/workflows/deploy.yml
jgravois: @lalver1 and i discussed this and it turns out there _is_ a 🤸 we could do to [avoid printing](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow) `ENGINEERING_GROUP_OBJECT_ID` (as a variable) in plain text in the logs.

```yaml
- name: Mask variable
  env:
    # Bind to an environment variable first
    UNMASKED_VALUE: ${{ vars.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }} 
  # Reference the shell variable ($UNMASKED_VALUE) instead of the expression syntax
  run: echo "::add-mask::$UNMASKED_VALUE"
```

this feels like more trouble than its worth to me though. so i reverted my original suggestion in 48de1613b

terraform/key_vault.tf
```diff
@@ -78,21 +78,6 @@ resource "azurerm_key_vault_access_policy" "engineering" {
...
-  key_vault_id = azurerm_key_vault.main.id
-  tenant_id    = data.azurerm_client_config.current.tenant_id
-  object_id    = var.DEVSECOPS_OBJECT_ID
-
-  key_permissions    = local.all_key_permissions
-  secret_permissions = local.all_secret_permissions
-
-  # This ensures the Key Vault itself is created before trying to attach a policy.
-  depends_on = [azurerm_key_vault.main]
-}
```
jgravois, line 94: i'll open this PR up for review after DevSecOps confirms that this service principal isn't used for anything other than running terraform commands from ADO pipeline runs.

jgravois: DevSecOps didn't provide any clarity so @lalver1 and i agreed to wait and remove this access policy when we tear down the application services.

.github/workflows/deploy.yml
thekaveman: Interesting to know this solution exists though...

lalver1 (approved):
This looks great @jgravois! I guess we can always come back to https://github.com/cal-itp/benefits/pull/3916#discussion_r3661428039 if we find a better way to manage these values, but this looks ready to me.

PR #3927, opened 2026-07-30
Fix: error when uploading agency logo in the Container App

### LINKED ISSUE #3861: Error uploading transit agency logo (Container app only)
## To Reproduce

1. login to the [dev container app](https://ca-cdt-pub-vip-calitp-d-web.wittyplant-56d5c43a.westus.azurecontainerapps.io/admin)
2. attempt to update the logo for the CST transit agency (or create a new throwaway transit agency)
3. a fatal error is encountered

> PermissionError at /admin/core/transitagency/4/change/
Operation not permitted: '/home/calitp/app/data/uploads/agencies/cst_piStIby.png'
Request Method: | POST
5.2.15
PermissionError
/calitp/.local/lib/python3.12/site-packages/django/core/files/storage/filesystem.py, line 156, in _save
django.contrib.admin.options.change_view

note: the file in the error message _is_ uploaded to the mounted volume.

## Screenshots
Image

Image

lalver1:
Closes #3861.

This PR adds the [FILE_UPLOAD_PERMISSIONS](https://docs.djangoproject.com/en/6.0/ref/settings/#file-upload-permissions) Django setting and sets it to `None` so that Django uses operating-system dependent behavior.

The error occurs when Django saves the logo using its [`_save` method which has a call to `os.chmod()`](https://github.com/django/django/blob/6c8eee41f778da87bdd17e27381d3705ed2d068b/django/core/files/storage/filesystem.py#L156). According to an LLM, this only becomes an issue in the Azure File Share used in the Container App (and not in the App Service which uses the `WEBSITES_ENABLE_APP_SERVICE_STORAGE` feature) due to how these volume mounts are implemented in Azure. The App Service mount knows how to handle `os.chmod()` calls because it uses a proprietary CIFS/SMB mount, while the Container App uses a standard Azure File CSI (Container Storage Interface) which does not allow Django to change permissions. This explains the need for setting `FILE_UPLOAD_PERMISSIONS = None` to avoid calling `os.chmod()` when saving the logo.

## Reviewing

- [x] Ensure uploading an agency logo still works locally by reproducing #3861.
- [x] When this PR is merged to `main`, reproduce https://github.com/cal-itp/benefits/issues/3861 in the `dev` environment's App Container by:
  - [x] Updating a transit agency's logo
  - [x] Creating a throwaway transit agency with a logo
  - [x] For completeness, ensuring that you can delete the agency logo that was created for the throwaway agency after the agency has been deleted

### DIFF
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -300,6 +300,10 @@ MEDIA_ROOT = os.path.join(STORAGE_DIR, "uploads/")
 
 MEDIA_URL = "/media/"
 
+# Prevent Django from attempting to use os.chmod() when saving agency logos since it causes permission errors
+# on the Container App's Azure File Share
+FILE_UPLOAD_PERMISSIONS = None
+
 # Logging configuration
 LOG_LEVEL = os.environ.get("DJANGO_LOG_LEVEL", "DEBUG" if DEBUG else "WARNING")
 LOGGING = {


jgravois (approved):
i confirmed locally this isn't error inducing and made sure i can still repro the error in the `dev` container app.

i don't expect any problems, but i'll be curious to see what permissions are set for newly uploaded files in the container app after this is merged.

lalver1:
I was also curious about permissions in the Container App so I used the throwaway agency you created @jgravois and saw the same behavior, no problem with uploading a logo 👍. I also was able to update the logo in the Django admin and delete the previous one using Azure Portal.

Listing the contents of the directory from the Container App's console shows

```bash
ls -la /home/calitp/app/data/uploads/agencies/
-rwxrwxrwx 1 root root  4009 Jul 30 23:26 toaks_DkWwrGi.png
```

so `root` owns it. I guess that is the "operating system dependent behavior". But I don't think this is a problem because we can delete the file directly from the Storage Account in Azure Portal.

jgravois:
thanks for digging in @lalver1!

its a little wild that the default is read/write/execute for _everyone_, but with regard to uploaded logos, i definitely don't consider it a problem. 😆

PR #3938, opened 2026-08-03
chore: gh actions service principal keyvault access policies

### LINKED ISSUE #3897: Code changes to move infrastructure CI/CD from DevOps Pipelines to GitHub Actions

jgravois:
resolves #3897

## Explanation

this PR introduces two new terraform variables:

### `sp_plan_object_id`

the ObjectId of the Service Principal that runs `terraform plan` from GH Actions and needs an access policy that allows it to READ/LIST KeyVault secrets in each of our resource groups.

this one will be added to our existing `terraform.tfvars` file.
```tfvars
# terraform.tfvars 
sp_plan_object_id = "990dc63..." # see LastPass 'Azure env vars' note for full id
```

### `sp_apply_object_id`

the ObjectId of the Service Principal that has permission to run `terraform apply` from GH Actions in _one resource group only_. this one needs exhaustive secret permissions.

this environment-specific  variable can either be supplied interactively or persisted in a new _environment-specific_ .tfvars file

```tfvars
# env/dev.tfvars 
sp_apply_object_id = "bb48b01..."
```

```bash
# use the new helper/wrapper script
./plan.sh

# or load the new environment-specific tfvars file explicitly yourself
terraform plan -var-file="./env/$(terraform workspace show).tfvars"
```

## Misc

While i was digging in the closet I couldn't resist the urge to prove out whether lowercase variables are supported in GH Actions. they are indeed. the casing i'm proposing in this PR is more idiomatic, but if y'all prefer to shout, i'm not fussed.

I also updated a few existing GitHub secret names in the interest of clarity.

## How to test

right now running [`terraform plan`](https://github.com/cal-itp/benefits/actions/runs/30960906036/job/92164403074#step:7:107) locally against `dev` will propose that two brand new access policies be added to our KeyVault.

immediately prior to merging the PR, i will re-import the existing access policies to ensure they are updated in place instead.

~it would have been nice to leave them imported and allow folks to verify the plan themselves locally, but doing that would result in the existing azure access policies being destroyed as soon as any other PR was merged to `main`.~ 

@lalver1 and i verified the correct behavior during a pairing session corresponding with a code-freeze.

## The plan itself

Since the access policies already exist, you'd expect the (post-import) [`terraform plan`](https://github.com/cal-itp/benefits/actions/runs/30856853328/job/91829763242#step:7:135) for this PR to show 'no changes'. it doesn't because: 

* i took the liberty of downgrading the secret permission set for `sp_plan_object_id` back to GET,LIST. (we confirmed this is sufficient and then DevSecOps granted additional permissions when we were testing the apply runner)
* i'm proposing here that we remove _all_ KeyVault key permissions from the existing access policies. @lalver1 and i are under the impression that we aren't using keys in Benefits.

## `TODO:`
- [x] `terraform import` existing access policies in `dev`
- [x] ensure that the only discrepancy in the plan is downgraded permissions for the tf planner 
- [x] `terraform state rm azurerm_key_vault_access_policy.[name]`
- [x] re-run `terraform import` in `dev`
- [x] re-run `terraform import` in `test` and `prod`

after this PR is merged, i will 🗑️ the redundant/outdated GitHub actions secrets below.

* <kbd>AZURE_SP_SUBSCRIPTION_ID</kbd> ✅
* <kbd>AZURE_SP_TENANT_ID</kbd> ✅
* <kbd>AZURE_SP_CLIENT_ID</kbd> ✅

### DIFF
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -34,9 +34,9 @@ jobs:
       - name: Log in to azure using federated identity credentials
         uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
         with:
-          client-id: ${{ secrets.AZURE_SP_CLIENT_ID }}
-          tenant-id: ${{ secrets.AZURE_SP_TENANT_ID }}
-          subscription-id: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+          client-id: ${{ secrets.AZURE_SP_PLAN_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
 
       - name: Setup terraform
         uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4
@@ -51,6 +51,8 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ steps.get_main_sha.outputs.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+          TF_VAR_sp_apply_object_id: ${{ secrets.AZURE_SP_DEV_CLIENT_ID }}
+          TF_VAR_sp_plan_object_id: ${{ secrets.AZURE_SP_PLAN_OBJECT_ID }}
         run: terraform plan -out=tfplan -lock-timeout=5m
 
       # tf-summarize generates a course-grained summary w/ no sensitive infra details
--- a/.github/workflows/terraform-plan.yml
+++ b/.github/workflows/terraform-plan.yml
@@ -51,7 +51,7 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ steps.get_main_sha.outputs.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
-          TF_VAR_sp_apply_object_id: ${{ secrets.AZURE_SP_DEV_CLIENT_ID }}
+          TF_VAR_sp_apply_object_id: ${{ secrets.AZURE_SP_DEV_OBJECT_ID }}
           TF_VAR_sp_plan_object_id: ${{ secrets.AZURE_SP_PLAN_OBJECT_ID }}
         run: terraform plan -out=tfplan -lock-timeout=5m
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -159,8 +159,8 @@ jobs:
         uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3
         with:
           client-id: ${{ secrets.AZURE_SP_CLIENT_ID }}
-          tenant-id: ${{ secrets.AZURE_SP_TENANT_ID }}
-          subscription-id: ${{ secrets.AZURE_SP_SUBSCRIPTION_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
 
       - name: Setup terraform
         uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4
@@ -177,6 +177,8 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ github.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+          TF_VAR_sp_apply_object_id: ${{ secrets.AZURE_SP_OBJECT_ID }}
+          TF_VAR_sp_plan_object_id: ${{ secrets.AZURE_SP_PLAN_OBJECT_ID }}
         run: terraform plan -out=tfplan -lock-timeout=5m
 
       - name: Terraform apply
--- a/docs/explanation/infrastructure.md
+++ b/docs/explanation/infrastructure.md
@@ -161,6 +161,7 @@ Use the following shorthand for conveying the Resource Type as part of the Resou
    ```
 
 1. Create a local `terraform.tfvars` file (ignored by git) using the `terraform.tfvars.sample` file.
+1. Create local `env/[env].tfvars` files (ignored by git) using `env/dev.tfvars.sample` for any environments you plan on interacting with.
 
 ### Development process
 
@@ -170,7 +171,7 @@ When configuration changes to infrastructure resources are needed, they should b
 1. Preview the changes, as necessary.
 
    ```sh
-   terraform plan
+   ./plan.sh # runs terraform plan with necessary variables hydrated
    ```
 
 1. [Submit the changes via pull request.](development/commits-branches-merging.md)
--- a/terraform/apply.sh
+++ b/terraform/apply.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+if [[ $(terraform workspace show) != "dev" ]]; then
+  echo "Running terraform apply against environments other than dev is not supported"
+  exit 1
+fi
+
+echo "Running terraform apply using an interactively supplied SHA..."
+
+terraform apply \
+  -var-file="./env/$(terraform workspace show).tfvars"
--- a/terraform/env/dev.tfvars.sample
+++ b/terraform/env/dev.tfvars.sample
@@ -1,0 +1,2 @@
+# create test.tfvars and/or prod.tfvars to terraform plan in those envs
+sp_apply_object_id = "object-id"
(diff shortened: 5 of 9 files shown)


jgravois:
@lalver1 if you want to let this one simmer until our dev workshop on wed, i'm all for it.

thekaveman (commented):
## Summary of discussion

We discussed this change at the Dev Workshop today and the different options as @lalver1 laid them out:

1. Combine 4 new variables (1 plan, 3 env-specific apply) into the one (existing) `terraform.tfvars` file. Use logic in the modules etc. to pick the right variable to use given the environment
2. Split the existing `terraform.tfvars` file into 3 env-specific files and use `-var-file=...` locally, with GitHub env secrets in CI/CD (what this PR suggests).

I think we all landed on Option 1 being overall better and following our existing patterns more closely.

Option 2 adds a (very minimal, really, negligible in the grand scheme) amount of overhead for the local experience (needing to pass the `-var-file=...` param) which can be partially mitigated by the helper @jgravois suggests e.g. `-var-file="./env/$(terraform workspace show).tfvars"`

However Option 2 also requires ~duplicating~ triplicating the 1 new plan-only variable, plus the existing `ENGINEERING_GROUP_OBJECT_ID` variable, plus any later "global" variables we end up adding -- all while splitting the management across 3 local files. This seems like it could introduce more of a maintenance burden than keeping a single `terraform.tfvars` file updated locally and pushing the "management" into Terraform logic that we already employ.

E.g. we could define a `local.sp_apply_object_id` to encapsulate the logic of picking the right env-specific variable and not having to spread that around the codebase.

We noted _one possible_ argument for Option 2 being superior is that segregating the CI/CD secrets into GitHub environments theoretically reduces the attack surface of a possible leak of these secrets compared to Option 1, but that didn't feel compelling enough to warrant the additional overhead.

## Changes

Whatever the option, this PR should also update the `terraform.tfvars.sample` (or its new triplicate sample) with the current set of vars.

jgravois:
thanks for summarizing @thekaveman. i'm kicking myself for missing this discussion but i'm happy to defer to the group and pursue Option 1 in this branch.

a few points of clarification though:

> Option 2 adds ... overhead for the local experience (needing to pass the -var-file=... param)

this isn't technically a requirement. individual devs could also:
1. call plain old `terraform plan` and supply the value interactively
1. call `TF_VAR_sp_apply_object_id="..." terraform plan` to supply the value interactively
1. write `sp_apply_object_id` into the `terraform.tfvars` with the value that corresponds to the currently active workspace.

(the last option is a bit of a footgun, but since @angela-tran (wisely) proposed that none of us run `terraform apply` locally against any environment except `dev`, the risk is minimal.)

> Option 2 also requires triplicating the 1 new plan-only variable

this is not the case. we can/should continue to store any variables that are not environment-specific in the existing `terraform.tfvars` file. if/when, the command below is called, variables from both files are read and merged at runtime.

```
terraform plan -var-file="./env/$(terraform workspace show).tfvars"
```

> segregating the CI/CD secrets into GitHub environments theoretically reduces the attack surface of a possible leak of these secrets compared to Option 1, but that didn't feel compelling enough to warrant the additional overhead.

i agree with this. the ObjectIds are not secrets. i don't _love_ passing two unused variables on every GH Actions run, but i definitely don't consider it a security risk.

thekaveman:
> this is not the case. we can/should continue to store any variables that are not environment-specific in the existing `terraform.tfvars` file. if/when, the command below is called, variables from both files are read and merged at runtime.

Ah-ha. That maybe changes things. I don't think we understood that this was the case. 

@lalver1 @Scotchester @angela-tran any more thoughts to add from our discussion, or changing opinions given what John shared above?

jgravois:
> this PR should also update the terraform.tfvars.sample (or its new triplicate sample) with the current set of vars.

good catch. i tackled this in ~6c05cb65414f5e9f68eaa81a14c5bc2abbe4fb29~ ab02f3e2a882994ef6c5005af308ed576d56154c and added a convenience script to wrap calls to `terraform plan` for good measure. 

(even if we choose Option 1, it's a bit of a chore to supply the CONTAINER_TAG interactively.)

Scotchester:
Given John's clarifications, I think I am entirely indifferent at this point, as long as it's clearly documented.

jgravois:
i prefer this option so i'll go ahead and open this PR up for review.

i don't want to oversell it though, the work for Option 1 is done too. we can close this one and refine https://github.com/cal-itp/benefits/pull/3956 if @lalver1 feels better about it. i'm truly not fussed either way.

lalver1:
Thanks for the summary @thekaveman and the clarification @jgravois! Yesterday I overlooked the fact that you can use [`-var-file=FILENAME` multiple times](https://developer.hashicorp.com/terraform/cli/commands/plan#var-file-filename), that way we can pass the "global" `terraform.tfvars` and `./env/$(terraform workspace show).tfvars` to avoid triplicating the "global" input variables. 

Thinking about this some more, I think Option 2 would be better since

- The `tfvars`-related change is not the large overhead I initially thought given that we can have a "global" and an "environment" `tfvars` file.
- It is nice that in GH Actions, we can place the `TF_VAR_sp_apply_object_id` secret in the appropriate environment and not at the repo level.
- Upon more thinking, the scenario we are tackling here is slightly different from our established pattern of interpolating Terraform resource references/names with `${local.env_letter}` so on second thought I don't feel that we are really deviating from an established pattern. We touched on this a little during our discussion, but this PR deals with dynamically passing in just the required _input variables_ to Terraform. This is a new case we hadn't encountered before since `DEVSECOPS_OBJECT_ID` and `ENGINEERING_GROUP_OBJECT_ID` needed/need to be passed always. So it feels like this is really a new use case we are developing for.

Also, that's a good point @Scotchester, we need to make sure it's clearly documented how this is set up.

I only have the following suggestions:

- We may also need to add a helper for `apply` since we will likely be applying to `dev` and it'd be nice to have a script similar to the `plan` helper. This probably means that the `plan` helper outputs the plan and the `apply` helper would use that same plan.
- In the helper script, I think we need to add another `-var-file=` option to handle the "global" variables, correct? Or does Terraform automatically use `terraform.tfvars` and so you only need to add the option for "environment"-dependent variables?
- We may want to add `test.tfvars.sample` and `prod.tfvars.sample`. Or if not, since locally we only want to work against `dev`, we just need to make sure to document the usage of these files in case we need to deviate from our workflow in exceptional circumstances. I'm not sure what those would be, but at least documenting the usage may be helpful.

jgravois:
> We may also need to add a helper for `apply`

i like this idea. lets skip generating/consuming plan files though. if we're going to `apply` locally we can always inspect the preliminary plan to ensure it looks the way we expect.

> does Terraform automatically use `terraform.tfvars`?

indeed it does! (ie: the behavior is the same as calling `terraform plan` with _no_ input params)

> We may want to add `test.tfvars.sample` and `prod.tfvars.sample`

how about we just include a breadcrumb in `dev.tfvars.sample` as a compromise? 😇 

ab02f3e2a882994ef6c5005af308ed576d56154c

thekaveman:
Just for completeness, here's the docs page that discusses how variables are assigned values: https://developer.hashicorp.com/terraform/language/values/variables#assign-values-to-variables

(relevant parts):

> You can assign values to root module variables through multiple methods, each with different precedence levels... 
> ... if the root module receives multiple values for the same variable name from different sources, Terraform uses the following order of precedence:
>
> 1. Any `-var` and `-var-file` options on the [command line](https://developer.hashicorp.com/terraform/language/values/variables#command-line-variables) in the order provided and variables from HCP Terraform
> 2. Any `*.auto.tfvars` or `*.auto.tfvars.json` files in lexical order
> 3. The `terraform.tfvars.json` file
> 4. The `terraform.tfvars` file
> 
> ...
> 
> Terraform automatically loads variable definition files if it detects any of the following:
> - File names ending in `.auto.tfvars` or `.auto.tfvars.json`
> - A file named `terraform.tfvars.json`
> - A file named `terraform.tfvars`

.github/workflows/terraform-plan.yml
```diff
@@ -51,6 +51,8 @@ jobs:
           TF_VAR_CONTAINER_TAG: ${{ steps.get_main_sha.outputs.sha }}
           TF_VAR_DEVSECOPS_OBJECT_ID: ${{ secrets.TF_VAR_DEVSECOPS_OBJECT_ID }}
           TF_VAR_ENGINEERING_GROUP_OBJECT_ID: ${{ secrets.TF_VAR_ENGINEERING_GROUP_OBJECT_ID }}
+          TF_VAR_sp_apply_object_id: ${{ secrets.AZURE_SP_DEV_OBJECT_ID }}
```
thekaveman, line 54: Would it be possible to cut down on one more secret, and reuse the `dev` environment secret here, by configuring this workflow to run against that environment?

Or am I misunderstanding what is going on (entirely possible)?

jgravois: i had the same thought (and tried it). unfortunately defining the 'dev' environment for this workflow run automatically invokes the corresponding environment service principal during azure login.

jgravois:
i'll push up one more commit to update the documentation on the page below:
https://docs.calitp.org/benefits/explanation/infrastructure/

lalver1 (approved):
This looks great @jgravois! Thanks for walking through the details with me 👍

PR #3968, opened 2026-08-07
fix: ensure deploy workflow aborts subsequent jobs when a prerequisite explicitly fails

### LINKED ISSUE #3967: A failed pytest run should result in an abandoned deployment
Because of a GitHub Actions outage, we deployed [`2026.08.1`](https://github.com/cal-itp/benefits/releases/tag/2026.08.1) to production today via a manual workflow run after the tag had already been pushed.

https://github.com/cal-itp/benefits/actions/runs/31200039700

This resulted in some unexpected behavior. `tests-pytest.yml` [failed](https://github.com/cal-itp/benefits/actions/runs/31200039700/job/92937619791), but the deployment proceeded anyway.

this leaves us with some cleanup/hardening for us to do.

## Expected behavior

1. when any of the pre-requisite steps in the `deploy.yml` workflow fail, subsequent jobs should be abandoned
2. if/when `tests-pytest.yml` is triggered from a manual workflow run it should not attempt to post a coverage report PR summary.

jgravois:
resolves #3967

* skip posting a comment with code coverage results in `tests-pytest` when the action was triggered by an upstream `workflow_dispatch`
* distinguish failed prerequisite jobs from both those that were skipped and those that succeeded

| prerequisite | deploy | gh action run | `sha` |
| -- | -- | -- | -- |
| `tests-pytest` ✅ (via `workflow_dispatch`) | ✅ | [runs/31220198968](https://github.com/cal-itp/benefits/actions/runs/31220198968) | fda26243ee19c182d824eea6e3d002f568ad0a24 |
| `tests-pytest` ✅ / `tests-ui` ⏭️ | ✅ | [runs/31220062984](https://github.com/cal-itp/benefits/actions/runs/31220062984) | fda26243ee19c182d824eea6e3d002f568ad0a24 | 
| `tests-pytest` ❌ / `tests-ui` ⏭️ | `skipped` | [/runs/31220336744](https://github.com/cal-itp/benefits/actions/runs/31220336744) | 3423f867e723d792298799b2c62987001ea0dc0d | 

~there are three throwaway commits on this branch that allowed me to trigger mock deploys and make this nifty chart. i'll `-f` push and eradicate them prior to merge.~

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -69,6 +69,8 @@ jobs:
     needs: [setup]
     uses: ./.github/workflows/tests-pytest.yml
     if: ${{ needs.setup.outputs.is_tag == 'true' }}
+    with:
+      parent_event: ${{ github.event_name }}
 
   check-migrations-and-messages:
     needs: [setup]
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -92,7 +92,8 @@ jobs:
         check-migrations-and-messages,
         check-dynamic-version,
       ]
-    if: (!cancelled())
+    # this convoluted syntax allows skipped and successful prerequisite runs and disallows failures
+    if: (always() && !cancelled() && !failure())
     environment: ${{ needs.setup.outputs.env_name }}
     steps:
       - name: Checkout
--- a/.github/workflows/tests-pytest.yml
+++ b/.github/workflows/tests-pytest.yml
@@ -1,6 +1,13 @@
 name: Pytest
 
-on: [push, pull_request, workflow_call]
+on:
+  push:
+  pull_request:
+  workflow_call:
+    inputs:
+      parent_event:
+        type: string
+        required: true
 
 jobs:
   pytest:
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -90,7 +92,11 @@ jobs:
         check-migrations-and-messages,
         check-dynamic-version,
       ]
-    if: (!cancelled())
+    # this convoluted syntax treats both skipped and successful runs as valid prerequisites
+    if: >-
+      always() &&
+      !cancelled() &&
+      !contains(needs.*.result, 'failure')
```
jgravois: the crux of the issue is that `(!cancelled())` all by itself overrides the default check which requires that the job explicitly run (and succeed).

gemini helped me pinpoint the error and identify alternative syntax that would also be satisfied by skipped prerequisite jobs.

ref: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#example-not-requiring-successful-dependent-jobs

jgravois: i ended up simplifying a bit further based on other linked github documentation.


```yml
if: (always() && !cancelled() && !failure())
```
      
https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions

thekaveman (approved):
👍

jgravois:
the `-f` push i'm about to do shouldn't change the diff at all so i'm expecting @thekaveman's approval to stand.

famous last words.

Issue #3962, opened 2026-08-06
Teardown Application service terraform infrastructure and associated ADO cleanup


jgravois:
These tasks were previously listed in individual cutover tickets #3637, #3638, and #3639, but @lalver1 and I think it makes more sense to tackle them centrally now.

_After_ `prod` has been cutover and several days have passed with no adverse side-effects, we will be ready to tackle the chores below in all three environments simultaneously. 

## Acceptance Criteria

- [x] Remove the raw Container App URL from the 2 Google services
  - [x] `dev` reCAPTCHA (`benefits-recaptcha` project)
  - [x] `test` reCAPTCHA
  - [x] `prod` reCAPTCHA
  - [x] `dev` SSO (`benefits-admin` project)
  - [x] `test` SSO
  - [x] `prod` SSO
- [x] Clean up Terraform by removing the App Service and DevSecOps SP
- [x] Remove the App Service's URL from the following 
  - [x] `django-trusted-origins`
  - [x] `django-allowed-hosts`
- [x] Update/Replace all mentions of App Service in Mkdocs
- [x] Remove defunct [GitHub environments](https://github.com/cal-itp/benefits/settings/environments) ($ENV_NAME, refactor/cicd-container-apps, main)
- [x] Remove `AZURE_WEBAPP_PUBLISH_PROFILE` github secrets
- [x] Remove `AZURE_ADO_*` github secrets
- [x] Remove `TF_VAR_DEVSECOPS_OBJECT_ID` github actions secret and dependabot secret
- [x] Update log urls in [this google doc](http://docs.google.com/document/d/11EPDIROBvg7cRtU2V42c6VBxcW_o8HhcyORALNtL_XY/edit?tab=t.0)
- [x] Remove application service IP addresses from [`eligibility-server`](https://dev.azure.com/mstransit/courtesy-cards/_build?definitionId=1&_a=summary) firewalls
- [x] Resolve `DisallowedHost` sentry issue
- [x] Remove Azure Pipelines's 'GitHub App' repository access
- [x] Request that DevSecOps:
  - [x] remove all members of the engineering team from the ADO azure tenant
  - [x] remove all members of the engineering team from the "CDT/ODI Development" subscription?
  - [x] ~destroy all Service Principals associated with [ADO service connections](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_settings/adminservices)~ (Amitoz confirmed they aren't single purpose)

thekaveman:
[...]

PR #3975, opened 2026-08-10
chore: tear down application services and simplify the remaining container application

### LINKED ISSUE #3963: Deprecate USE_POSTGRES environment variable
Once our Application Services have been retired permanently, it will no longer be possible to back the deployed application with SQLite.

This will allow us to clean up a couple conditional blocks that are no longer needed.

## Acceptance Criteria

- [ ] PostgreSQL is the sole backend for the Django application locally and in all deployed environments.

### LINKED ISSUE #3964: Consolidate/cleanup mounted drive application directory structure
I'm a little fuzzy on the details for this one, so take this description with a grain of salt. 😇 

Right now, the working directory for the Container application is:
```
/calitp/app
```

Files uploaded to the Container app's mounted volume are accessed (from within the container) via:
```
/home/calitp/app/data
```

my understanding is that this was the most expeditious/convenient option in order to support both Application services and Container apps simultaneously.

Once we've retired our Application services, we can/should remove the `/home` prefix from the mounted drive.

## Acceptance Criteria

- [ ] the mounted drive is accessible via `/calitp/app/data`
- [ ] our [postgresql guide](https://docs.calitp.org/benefits/guides/postgresql/#copy-a-backup) reflects the latest guidance

jgravois:
handles the bulk of #3962
closes #3963
closes #3964

## `TODO:`
- [x] update docs
- [x] modularize existing django key vault secret resource
- [x] ensure pytest uses sqlite ~or postgresql, not both~
- [x] run appcontainer to test NGINX alias 
- [x] ensure neither of the two new resources flagged in the plan are problematic
(ie: will they result in a freshly generated random password? if so, will the container app continue to function correctly?)
- [x] confirm we should stop short of 💣 the [`/home/calitp`](https://github.com/cal-itp/docker-python-web/blob/65985bf70c2c370c9d5cbbb86aba45856191a7ae/appcontainer/Dockerfile#L31) directory [entirely](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/appcontainer/Dockerfile#L24)

## Post-merge
- [x] delete `TF_VAR_DEVSECOPS_OBJECT_ID` github secret (and dependabot copy)

this one won't be safe to merge until after `prod` is cutover to the Container app and has proven stable.


test
jgravois, line 1: not sure how this got appended. i've yanked the file in fb69c56 and will push as soon as GitHub authentication failures subside.

docs/guides/troubleshooting.md
```diff
@@ -75,13 +75,4 @@ A common problem that causes Littlepay API failures is that the certificate expi
...
-
-1. Grab the `outbound_ip_ranges` `output` values from the most recent Benefit [deployment](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build?definitionId=828) to the relevant environment.
-1. Update the IP ranges
-   1. Go to the [Eligibility Server Pipeline](https://dev.azure.com/mstransit/courtesy-cards/_build?definitionId=1&_a=summary)
-   1. Click `Edit`
-   1. Click `Variables`
-   1. Update the relevant variable with the new list of CIDRs
-
-Note there is nightly downtime as the Eligibility Server restarts and loads new data.
+Note there is nightly downtime as the [Eligibility Servers](https://docs.calitp.org/eligibility-server/) restart and load new data.
```
lalver1, line 79: We may still want to leave some of the original explanation here since it still applies to Container Apps. So we can change

> If the Benefits application gets a 403 error when trying to make API calls to the [Eligibility Server](https://docs.calitp.org/eligibility-server/), it may be because the outbound IP addresses changed, and the Eligibility Server firewall is still restricting access to the old IP ranges.
> 1. Grab the `outbound_ip_ranges` `output` values from the most recent Benefit [deployment](https://calenterprise.visualstudio.com/CDT.OET.CAL-ITP/_build?definitionId=828) to the relevant environment.
> 1. Update the IP ranges
>   a. Go to the [Eligibility Server Pipeline](https://dev.azure.com/mstransit/courtesy-cards/_build?definitionId=1&_a=summary)
>   b. Click `Edit`
>   c. Click `Variables`
>   d. Update the relevant variable with the new list of CIDRs

to 

If the Benefits application gets a 403 error when trying to make API calls to the [Eligibility Server](https://docs.calitp.org/eligibility-server/), it may be because the outbound IP address changed, and the Eligibility Server firewall is still restricting access to the old IP.

1. Grab the _IP address_ value from the _Public IP address_ resource in the Azure Portal for the relevant environment.
1. Update the IP address
   1. Go to the [Eligibility Server Pipeline](https://dev.azure.com/mstransit/courtesy-cards/_build?definitionId=1&_a=summary)
   1. Click `Edit`
   1. Click `Variables`
   1. Update the relevant variable with the new public IP address

jgravois: i like this suggestion, although i'd like to try and make it clear that it'd be more of an aberration than it was in the age of application services.

jgravois: done in 5856927

jgravois: today i combed through the doc more closely and updated the last of the stale references to 'App Service's in 304d0d6a5

thekaveman (changes requested):
> ensure neither of the two new resources flagged in the plan are problematic
(ie: will they result in a freshly generated random password? if so, will the container app continue to function correctly?)

Yes I think there are some issues here that need to be corrected before we merge.

So just to be clear, we're talking about the **adds** for:

- `module.application.azurerm_key_vault_secret.django_db_password`
- `module.application.random_password.django_db_password`

And corresponding **deletes** for:

- `azurerm_key_vault_secret.django_db_password`
- `random_password.django_db_password`

Now part of this is due to moving _where in the Terraform files_ these definitions exist, from the top-level down into the `application` module. Hence the expected resource naming change. I might also expect that we'd need to `terraform import` the existing Azure resources into the new Terraform location/ID, so that Terraform knows these newly moved/renamed definitions actually point to existing resources.

---

**BUT** I think part of the issue also stems from this PR (erroneously) reusing a secret name `local`:

## On `main`

[`azurerm_key_vault_secret.django_db_password = local.django_db_password_secret_name`](https://github.com/cal-itp/benefits/blob/ca801812c69143bf1f056fb92630d6e69dfc1b5b/terraform/app_service.tf#L124)

and

[`local.django_db_password_secret_name = "django-db-password"`](https://github.com/cal-itp/benefits/blob/ca801812c69143bf1f056fb92630d6e69dfc1b5b/terraform/app_service.tf#L2)

## On this PR

[`module.application.azurerm_key_vault_secret.django_db_password = var.django_db_password_secret_name`](https://github.com/cal-itp/benefits/blob/390699fb73ed689d1f282930364f03ee8978f8a1/terraform/modules/application/app_web.tf#L109)

and

[`var.django_db_password_secret_name = local.postgres_admin_password_secret_name`](https://github.com/cal-itp/benefits/blob/390699fb73ed689d1f282930364f03ee8978f8a1/terraform/modules.tf#L35)

and

[`local.postgres_admin_password_secret_name = "postgres-admin-password"`](https://github.com/cal-itp/benefits/blob/390699fb73ed689d1f282930364f03ee8978f8a1/terraform/database.tf#L2)

So this is going to wind up with **two secrets in the same KV that have the same name**, being a duplicate of:

[`azurerm_key_vault_secret.postgres_admin_password`](https://github.com/cal-itp/benefits/blob/390699fb73ed689d1f282930364f03ee8978f8a1/terraform/database.tf#L50)

which of course will throw an error.

_Just in case it isn't clear, these are in fact different passwords for different accounts. One is for the "Postgres superuser" needed to setup the database server itself. The other is for the Postgres user that owns the Django database within the server, and how the Django application connects/reads/writes to said database._

terraform/app_service.tf
```diff
@@ -1,138 +0,0 @@
...
-
-    # Sentry
-    "SENTRY_DSN"                = "${local.secret_prefix}sentry-dsn)",
-    "SENTRY_ENVIRONMENT"        = local.env_name,
-    "SENTRY_REPORT_URI"         = "${local.secret_prefix}sentry-report-uri)",
-    "SENTRY_TRACES_SAMPLE_RATE" = "${local.secret_prefix}sentry-traces-sample-rate)",
-  }
-
-  lifecycle {
-    prevent_destroy = true
```
lalver1, line 101: This may be problematic since I think it [won't allow us to destroy the resource by simply deleting this file](https://developer.hashicorp.com/terraform/tutorials/state/resource-lifecycle#prevent-resource-deletion). We may need to do it in 2 steps, changing it to `false` and then removing the file 🤔, but we'll see once we start running `plan`.

jgravois:
> part of the issue also stems from this PR (erroneously) reusing a secret name

great catch. this was an oversight on my part. i struggled a bit exposing the necessary references from within the module and ended up losing the plot entirely. i _think_ d2046f55e should do the trick.

> I might also expect that we'd need to terraform import the existing Azure resources into the new Terraform location/ID,

i don't know why, but it genuinely didn't occur to me that terraform would need a nudge to understand that these two resources refer to the exact same thing in Azure. after adding eaa98cf03 and running `import` locally, the plan is looking a lot more like what i'd expect. that said, i have a copy of the existing db secret value persisted locally in case the `apply` does something unexpected.

thx for the detailed notes @thekaveman!

terraform/app_service.tf
jgravois: good eyes @lalver1. i _think_ because the block is no longer present, the `apply` will succeed, but the only way i know of to test that theory is running `terraform apply` locally once we're all happy with the published `plan`.

> we'll see once we start running `plan`.

you meant `apply`, yes?

lalver1: So I actually thought that the `plan` might have reported something but it doesn't seem like it. Yep, we'll really see the actual behavior when we do the `apply` locally 👍

terraform/modules.tf
```diff
@@ -31,7 +34,8 @@ module "application" {
 
   # App Config
   azure_communication_connection_string_name = local.azure_communication_connection_string_name
-  django_db_password_secret_name             = azurerm_key_vault_secret.django_db_password.name
+  key_vault_id                               = azurerm_key_vault.main.id
+  django_db_password_secret_name             = local.django_db_password_secret_name
```
lalver1, line 38: Depending on what you are seeing in the current `plan`s @jgravois, it could help (especially thinking about `test` and `prod` later) to have something like

```terraform
moved {
  from = random_password.django_db_password
  to   = module.application.random_password.django_db_password
}

moved {
  from = azurerm_key_vault_secret.django_db_password
  to   = module.application.azurerm_key_vault_secret.django_db_password
}
```

at the end of this `modules.tf` file. I think this would make the `plan` simpler and safer since Terraform won't recreate these actual Azure resources, it'll know it's just an [internal move](https://developer.hashicorp.com/terraform/language/block/moved).

EDIT: I think you accomplished this using `import` so it's probably not needed but thought about mentioning it anyway.

jgravois: i appreciate the nudge here. i indeed accomplished this via `terraform import` but i think what you're suggesting is a worthwhile improvement because it canonizes something in code that we'd otherwise have to do interactively by running the same import in `test` and `prod`.

i'll try adding this and running a plan against `test` and push the change up if it does what we'd expect.

afterwards, we can/should delete any `moved` declarations that remain after the changes have been applied in _all_ environments.

sound good?

lalver1: Sounds good! And yep, removing `moved` later I think would be good since there isn't much value in leaving those statements there for the long term. The important thing is that they more explicitly showed the change for this PR.

jgravois: after running the command below to _un_-import the resource in `dev`, @lalver1's suggestion worked a treat. aafc6f3

```bash
terraform state rm module.application.azurerm_key_vault_secret.django_db_password
```

.github/workflows/tests-pytest.yml
```diff
@@ -12,12 +12,13 @@ on:
...
   pytest:
     runs-on: ubuntu-latest
+
     permissions:
       # Gives the action the necessary permissions for publishing new
       # comments in pull requests.
       pull-requests: write
       # Gives the action the necessary permissions for pushing data to the
-      # python-coverage-comment-action branch, and for editing existing
+      # python-coverage-comment-action-data branch, and for editing existing
```
lalver1, line 21: Good catch on this typo 👍

appcontainer/nginx.conf
```diff
@@ -74,7 +74,7 @@ http {
 
     # path for uploaded files
     location /media/ {
-      alias /home/calitp/app/data/uploads/;
+      alias /calitp/app/data/uploads/;
```
lalver1, line 77: Tested this locally by rebuilding the image for the `client` Compose service and I was able to see uploaded logos being saved to this new location 👍

docs/reference/environment-variables.md
lalver1: We can probably remove this part too?

terraform/modules/application/locals.tf
lalver1: We can probably remove `USE_POSTGRES` here too?

.github/workflows/tests-pytest.yml
```diff
@@ -40,7 +41,7 @@ jobs:
 
       - name: Run setup
         run: |
-          python manage.py migrate
+          python manage.py migrate --settings=tests.pytest.settings
```
lalver1, line 44: I think this change snuck in in 71b564ce09b9f3e1eac4ef39921dc977114bd738, but we don't need it?

docs/explanation/deployment.md
```diff
@@ -1,17 +1,17 @@
 # Deployment
 
-The Benefits app is currently deployed into a Microsoft Azure account provided by [California Department of Technology (CDT)'s Office of Enterprise Technology (OET)][oet], a.k.a. the "DevSecOps" team. More specifically, it uses [custom containers][app-service-containers] on [Azure App Service][app-service]. [More about the infrastructure.](infrastructure.md)
+The Benefits app is currently deployed into a Microsoft Azure account provided by [California Department of Technology (CDT)'s Office of Enterprise Technology (OET)][oet], a.k.a. the "DevSecOps" team. More specifically, [Azure Container Apps][container-apps]. [More about the infrastructure.](infrastructure.md)
```
lalver1: Small typo, missing "it uses" so it should be:

> More specifically, it uses [Azure Container Apps][container-apps].

terraform/modules.tf
thekaveman: This is so much better than opaque `import` calls against random environments. Thanks for the reminder of this functionality @lalver1

docs/explanation/infrastructure.md
```diff
@@ -59,11 +59,9 @@ flowchart LR
     subgraph Azure
         frontdoor --> NGINX
 
-        subgraph App Service
-            subgraph Custom container
-                direction TB
-                NGINX --> django
-            end
+        subgraph Container App
```
lalver1, line 62: Minor note, we could still keep the 2 levels of details as before, just replacing _App Service_ with _Container App_ and keeping _Custom container_. But simplifying to only _Container App_ as shown here also works.

thekaveman (dismissed):
Nice job on this @jgravois, thanks for sticking it with it! I know this is all pretty complex so I'm glad we had some time/more eyes on it. 

Will feel great to finally put all this to bed in all 3 envs 🚀

docs/explanation/infrastructure.md
```diff
@@ -130,16 +128,15 @@ The DevSecOps team sets the following naming convention for Resources:
...
-| Virtual Network  | `VNET`     |
-| Resource Group   | `RG`       |
-| Virtual Machine  | `VM`       |
-| Database         | `DB`       |
-| Subnet           | `SNET`     |
-| Front Door       | `FD`       |
+| Resource        | Convention |
+| --------------- | ---------- |
+| Container App   | `CA`       |
+| Virtual Network | `VNET`     |
```
lalver1: Another minor thing, we could replace _App Service Plan (ASP)_ with _Container Apps Environment (CAE)_ since both Azure services play a similar role.

docs/guides/troubleshooting.md
```diff
@@ -8,33 +8,29 @@ We have [ping tests](https://docs.microsoft.com/en-us/azure/azure-monitor/app/mo
 
 ### Logs
 
-#### Azure App Service Logs
+#### Azure Container App Logs
 
 [Open the `Logs` for the environment you are interested in.](https://docs.google.com/document/d/11EPDIROBvg7cRtU2V42c6VBxcW_o8HhcyORALNtL_XY/edit#heading=h.6pxjhslhxwvj) The following tables are likely of interest:
```
lalver1: We can probably replace this:
> [Open the Logs for the environment you are interested in.](https://docs.google.com/document/d/11EPDIROBvg7cRtU2V42c6VBxcW_o8HhcyORALNtL_XY/edit#heading=h.6pxjhslhxwvj) The following tables are likely of interest:

    AppServiceConsoleLogs: stdout and stderr coming from the container
    AppServiceHTTPLogs: requests coming through Container App
    AppServicePlatformLogs: deployment information

With:

> In the Azure Portal, in the Container App, you can find real time logs under _Monitoring > Log stream_ and historical logs under _Monitoring > Logs_

We can fill in more details in a followup PR or in this one, I think either way would work.

lalver1 (commented):
This looks great @jgravois! Thanks for this cleanup, it's a lot of stuff!

I only had some minor comments, the things related to the documentation are not critical so feel free to ignore them. Besides those, I think there are just a couple of more environment variable cleanups below.

Oh and just one more thing, does ff7a58a2d8735b4c8a8b6e531738e78064a2e463 need to be removed? It looks like a partial fixup happened in that commit.

.github/workflows/tests-pytest.yml
jgravois: actually, passing through the settings file explicitly is how we're convincing pytest to use SQLite instead of Postgres.

docs/explanation/deployment.md
jgravois: it took me longer than i'd like to admit to figure out that this is just my original diff and not you supplying an explicit 'suggestion' 🤦‍♂️

thekaveman:
@lalver1 

> Oh and just one more thing, does https://github.com/cal-itp/benefits/commit/ff7a58a2d8735b4c8a8b6e531738e78064a2e463 need to be removed? It looks like a partial fixup happened in that commit.

We were having a conversation about this the other day, this is part of the expected workflow: https://github.com/cal-itp/benefits/pull/3994#pullrequestreview-4963367472

docs/explanation/infrastructure.md
jgravois: as long as you were referring to the "Sample Names" directly above, i implemented a fix in 29c3ce48eaa33e70c108b7dc4756be0c9c9deea3

jgravois:
> this is part of the expected workflow

exactly. now that i'm done futzing around with the branch i ran `git rebase -i --autosquash` locally to clean up the history and get in sync with `main` in one fell swoop.

thank you both for the eagle eyes, i'll plan on running `terraform apply` locally tomorrow mornin 🌅 

edit:
> now that i'm done futzing

famous last words! 9972be071

(in the interest of being explicit, i plan on squashing out ☝️ _after_ approval)

.github/workflows/tests-pytest.yml
lalver1: Ah! Yes! Can't believe I got that mixed up 🤦‍♂️

docs/explanation/infrastructure.md
lalver1: Thanks, I actually hadn't noticed the _Sample Names_ section, good catch making the change there.

So sorry, super nit, I was more thinking about adding a row to the table

| Resource | Convention |
|----------|------------|
|Container Apps Environment | `CAE` | 

to replace the removed 

| Resource         | Convention |
| ---------------- | ---------- |
| App Service Plan | `ASP`      |

But it's no big deal, I know we want to test this out and this is an extermely minor part of the PR.

lalver1 (dismissed):
Looks great @jgravois!

docs/explanation/infrastructure.md
jgravois: ah, of course. not sure how i missed that. i need to rebase one last time anyway so i'll just tack this onto the last commit and ask you to re-approve one last time.

jgravois: done in 096c3ec823169d33e8c825edbd1f416982ad2eff

lalver1 (approved):
👍 🚀

PR #3977, opened 2026-08-12
Fix: enrollment expiration handling

### LINKED ISSUE #3822: Displayed expiration date for an expiring enrollment is off
In the process of QA testing the 2026.06.2 release on `test`, I was surprised to see this expiration date after enrolling in an expiring enrollment flow:

Enrollment success message saying 'Your benefit will expire on August 22, 2025'

This is surprising for two reasons:

- The date is in the past.
- Given that the flow is set to expire in 365 days, it's odd that it would say **August**, rather than June.

## To Reproduce

1. Go to https://test-benefits.calitp.org
2. Select CST from the agency dropdown
3. Select the CalFresh Cardholder flow
4. Verify eligibility through Login.gov sandbox with one of our two CalFresh recipient test users
5. Proceed through enrollment with a test card number
6. See the expiration date presented

## Expected behavior

The expiration date presented is 365 days in the future.

## Additional context

It's very possible that this test card was previously enrolled for this CST group in Littlepay QA, but if so, then the expected behavior would be to receive a message saying it's not time to re-enroll yet, or that it is time to re-enroll, and the new expiration is 365 days in the future.

@angela-tran [posted this theory in Slack](https://cal-itp.slack.com/archives/C037Y3UE71P/p1782158224457309?thread_ts=1782148794.124029&cid=C037Y3UE71P):

> I think you might've caught an existing bug.
>
> It seems like you're running into the ["already enrolled" scenario](https://github.com/cal-itp/benefits/blob/2b187e845677fa66cbd65e8c569c13bbdcdcf5f4/benefits/enrollment_littlepay/enrollment.py#L86-L87), which stores the existing expiration date on the session to use when calculating things like the reenrollment date. The code does check if the existing expiration date has passed and [updates the enrollment](https://github.com/cal-itp/benefits/blob/2b187e845677fa66cbd65e8c569c13bbdcdcf5f4/benefits/enrollment_littlepay/enrollment.py#L112-L119) using the transit processor's API, but it looks like it doesn't update the session.

This makes sense, however, I then would have expected the date to be updated and correct upon a second successful enrollment of the same card in the same flow, and it remained August 22, 2025.

thekaveman:
Closes #3822 

---

# Updated description

The main thrust of this PR was of course to fix the referenced bug (_see the `Original description` below for more details about the nuance of the bug, which was deeper than just what date was displayed to the user_).

But since the bug was present in the enrollment handling logic duplicated across Littlepay and Switchio specific `enroll` handlers (and thus had to be fixed in both places), this PR also refactors that duplicated logic into a central place for re-use.

We introduce an `EnrollmentDecision` class, populated by a new helper in the shared `benefits.enrollment.enrollment` module: 

```python
def resolve_enrollment_decision(
  flow: models.EnrollmentFlow,
  already_enrolled: bool,
  existing_expiry: datetime | None
) -> EnrollmentDecision
```

This helper encapsulates the (provider-agnostic) enrollment decision based on:

- If the flow supports expiration or not
- If the user is already enrolled or not
- If the enrollment has an existing expiry or not
- If the existing expiry is valid, within the reenrollment window, or expired

The provider-specific `enroll()` functions then use that decision to guide their interaction with the specific APIs, to e.g. enroll the user, and set or remove an expiration date with the enrollment.

# Original description

_The following was for an earlier iteration of this PR that is no longer relevant. Leaving for historical posture._

<details>
<summary>Original description</summary>

It turns out this bug was deeper than just the expiration date displayed to the user at the end of a successful enrollment, and it was present for both Littlepay and Switchio. This PR fixes expiring enrollments for both processors and updates the tests accordingly to cover the (now squashed) bug.

_No changes were made to enrollments that don't support expiration, as this was already working as expected._

### Problem sequence

Prior to this PR, the sequence when a flow supported expiration was:

1. Get the "enrollment" e.g. funding source (Littlepay) / group (Switchio)
2. Mark if the previous step returned a value (`already_enrolled = True`) or not (`already_enrolled = False`)
3. Store an expiration date in session:
    1. If `already_enrolled` and an expiration date is present on the enrollment, store it in session;
    1. Otherwise store a newly calculated expiration date in session
4. Determine how to enroll:
   1. If not `already_enrolled`, enroll with the stored expiration date
   1. If `already_enrolled` _without_ an expiration date, update the enrollment with the stored expiration date
   1. If `already_enrolled` _with_ an expiration date, if the expiration date has passed or is within the re-enrollment window, update the enrollment with the stored expiration date
   1. Otherwise throw a re-enrollment error

The problem originally starts with step `3.i.` above. Regardless of _when_ the expiration date was, it was stored in session once and considered good thereafter. This was true even for long past expiration dates (such as the original example from #3822, `August 22, 2025`)

So by the time we get to step `4.`, no matter which branch the logic went down, the stored expiration date could have been long past useful, yet it was being sent in all cases where an enrollment call to the processor was being made.

The end result was an expiration loop that was impossible to get out of: the existing enrollment had expired, but then we updated it with the old expiration (expired) date. Rinse and repeat. 

_**Whether this would have resulted in a problem for a user trying to tap, or re-enroll, is debatable I think**_. We assume that once the expiration date passes, the enrollment is purged from the processor's system. Which would result in the `3.ii.` case above.

However for testing/demos, this was certainly a visible issue. And the logic should be sound regardless.

### Fixed sequence

This PR updates the sequence in a few key ways:

1. Get the "enrollment" e.g. funding source (Littlepay) / group (Switchio)
2. Mark if the previous step returned a value (`already_enrolled = True`) or not (`already_enrolled = False`)
3. Capture the existing expiration if present on the enrollment; initialize a flag to mark if we `should_update_expiry = True`
4. If we have an existing expiration:
    1. Store it in session
    1. If stored value is NOT expired AND outside of the re-enrollment window, throw a re-enrollment error and change the flag `should_update_expiry = False`
5. If `should_update_expiry`:
   1. Calculate a new expiry
   1. Store it in session
   1. Enroll or update an existing enrollment with the newly calculated expiry

We start by assuming the expiration needs to be updated, and move the re-enrollment error check up to handle that case earlier, which is the only way to invalidate the initial assumption. 

If the assumption holds, a new expiration is calculated and stored, and an enrollment is performed (or updated) with the newly calculated expiration.

</details>

# Cases to review

_(Both Littlepay and Switchio for all cases)_

See this helpful table / discussion for the source of all the cases we need to cover/consider: https://github.com/cal-itp/benefits/pull/1989#issue-2211450180

- Non-expiring flow (e.g. Senior) with a new card :arrow_right: enrollment success
- Non-expiring flow (e.g. Senior) with an already enrolled card (e.g. redo the one from the prior case) :arrow_right: enrollment success
- CalFresh flow and a new card :arrow_right: enrollment success with expiration date in the future
- CalFresh flow and a card that is already enrolled, outside the re-enrollment window, and not yet expired (e.g. redo the same one from the prior case) :arrow_right: re-enrollment error
- CalFresh flow and a card that is already enrolled and inside the re-enrollment window :arrow_right: enrollment success with updated expiration
  1. Make the following code change to generate an expiry that will be inside the re-enrollment window 
  ```python
  def _calculate_expiry(expiration_days: int):
    # existing code ...
    # temp generate an expiry that would be within a 14 day re-enrollment window
    now = timezone.now()
    expiry_datetime = expiry_date.replace(day=now.day + 1, month=now.month, year=now.year)
  
    return expiry_datetime
  ```
  2. Enroll with a brand new card to set this as the initial expiration
  3. Remove the temp code change
  4. Enroll with the same card, expect an updated expiration
- CalFresh flow and a card that is already enrolled with an expiration date in the past :arrow_right: enrollment success with updated expiration
  1. Follow the same sequence above but instead of overriding to calc an expiry inside the re-enrollment window, calc an expiry in the past, e.g. 
    ```python
    def _calculate_expiry(expiration_days: int):
      # existing code ...
      # temp generate an expiry in the past
      now = timezone.now()
      expiry_datetime = expiry_date.replace(year=now.year - 1)
    
      return expiry_datetime
    ```
- CalFresh flow (with temp code change below) and a card that is already enrolled (any expiration date)
  - :arrow_right: Switchio, enrollment success, expiry is removed (_you'll still see `Your benefit will expire on` but it should be blank_)
  - :arrow_right: Littlepay, `NotImplementedError` raised

  ```python
  def resolve_enrollment_decision(
    flow: models.EnrollmentFlow, already_enrolled: bool, existing_expiry: datetime | None
  ) -> EnrollmentDecision:
     # existing docstrings ...
     # flow does not support expiration
     if True:
       # existing code
  ```

### DIFF
--- a/benefits/enrollment_littlepay/enrollment.py
+++ b/benefits/enrollment_littlepay/enrollment.py
@@ -2,6 +2,7 @@ import re
 from dataclasses import dataclass
 
 from littlepay.api.client import Client
+from littlepay.api.funding_sources import FundingSourceResponse
 from requests.exceptions import HTTPError
 
 from benefits.core import session
@@ -55,7 +56,7 @@ def request_card_tokenization_access(request) -> CardTokenizationAccessResponse:
     )
 
 
-def enroll(request, card_token) -> tuple[Status, Exception]:
+def enroll(request, card_token) -> tuple[Status, Exception, FundingSourceResponse]:
     """
     Attempts to enroll this card into the transit processor group for the flow in the request's session.
 
@@ -78,60 +79,45 @@ def enroll(request, card_token) -> tuple[Status, Exception]:
     exception = None
     try:
         group_funding_source = _get_group_funding_source(client=client, group_id=group_id, funding_source_id=funding_source.id)
-
         already_enrolled = group_funding_source is not None
+        has_expiration = already_enrolled and group_funding_source.expiry_date is not None
 
         if flow.supports_expiration:
-            # set expiry on session
-            if already_enrolled and group_funding_source.expiry_date is not None:
-                session.update(request, enrollment_expiry=group_funding_source.expiry_date)
-            else:
-                session.update(request, enrollment_expiry=_calculate_expiry(flow.expiration_days))
+            should_update_expiry = True
+            expiry_date = group_funding_source.expiry_date if has_expiration else None
 
-            if not already_enrolled:
-                # enroll user with an expiration date, return success
-                client.link_concession_group_funding_source(
-                    group_id=group_id, funding_source_id=funding_source.id, expiry=session.enrollment_expiry(request)
-                )
-                status = Status.SUCCESS
-            else:  # already_enrolled
-                if group_funding_source.expiry_date is None:
-                    # update expiration of existing enrollment, return success
-                    client.update_concession_group_funding_source_expiry(
-                        group_id=group_id,
-                        funding_source_id=funding_source.id,
-                        expiry=session.enrollment_expiry(request),
+            if expiry_date:
+                session.update(request, enrollment_expiry=expiry_date)
+                if not (
+                    _is_expired(expiry_date)
+                    or _is_within_reenrollment_window(expiry_date, session.enrollment_reenrollment(request))
+                ):
+                    status = Status.REENROLLMENT_ERROR
+                    should_update_expiry = False
+
+            if should_update_expiry:
+                new_expiry = _calculate_expiry(flow.expiration_days)
+                session.update(request, enrollment_expiry=new_expiry)
+                if not already_enrolled:
+                    client.link_concession_group_funding_source(
+                        group_id=group_id, funding_source_id=funding_source.id, expiry=new_expiry
                     )
-                    status = Status.SUCCESS
                 else:
-                    is_expired = _is_expired(group_funding_source.expiry_date)
-                    is_within_reenrollment_window = _is_within_reenrollment_window(
-                        group_funding_source.expiry_date, session.enrollment_reenrollment(request)
+                    client.update_concession_group_funding_source_expiry(
+                        group_id=group_id, funding_source_id=funding_source.id, expiry=new_expiry
                     )
+                status = Status.SUCCESS
 
-                    if is_expired or is_within_reenrollment_window:
-                        # update expiration of existing enrollment, return success
-                        client.update_concession_group_funding_source_expiry(
-                            group_id=group_id,
-                            funding_source_id=funding_source.id,
-                            expiry=session.enrollment_expiry(request),
-                        )
-                        status = Status.SUCCESS
-                    else:
-                        # re-enrollment error, return enrollment error with expiration and reenrollment_date
-                        status = Status.REENROLLMENT_ERROR
-        else:  # eligibility does not support expiration
+        else:  # expiration not supported
             if not already_enrolled:
-                # enroll user with no expiration date, return success
                 client.link_concession_group_funding_source(group_id=group_id, funding_source_id=funding_source.id)
                 status = Status.SUCCESS
-            else:  # already_enrolled
-                if group_funding_source.expiry_date is None:
-                    # no action, return success
-                    status = Status.SUCCESS
-                else:
-                    # remove expiration date, return success
-                    raise NotImplementedError("Removing expiration date is currently not supported")
+            elif not has_expiration:
+                # already enrolled, without an expiration date -> no action, return success
+                status = Status.SUCCESS
+            else:
+                # already enrolled with an expiration date -> remove expiration date, return success
+                raise NotImplementedError("Removing expiration date is currently not supported")
 
     except HTTPError as e:
         if e.response.status_code >= 500:
--- a/benefits/enrollment_switchio/enrollment.py
+++ b/benefits/enrollment_switchio/enrollment.py
@@ -167,36 +167,26 @@ def enroll(
     try:
         group = _get_group_for_token(client, pto_id, group_id, token)
         already_enrolled = group is not None
+        has_expiration = already_enrolled and group.expiresAt is not None
 
         if flow.supports_expiration:
-            # set expiry on session
-            if already_enrolled and group.expiresAt is not None:
-                session.update(request, enrollment_expiry=group.expiresAt)
-            else:
-                session.update(request, enrollment_expiry=_calculate_expiry(flow.expiration_days))
+            should_update_expiry = True
+            expiry_date = group.expiresAt if has_expiration else None
 
-            if not already_enrolled:
-                # enroll user with an expiration date, return success
-                client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
+            if expiry_date:
+                session.update(request, enrollment_expiry=expiry_date)
+                if not (
+                    _is_expired(expiry_date)
+                    or _is_within_reenrollment_window(expiry_date, session.enrollment_reenrollment(request))
+                ):
+                    status = Status.REENROLLMENT_ERROR
+                    should_update_expiry = False
+
+            if should_update_expiry:
+                new_expiry = _calculate_expiry(flow.expiration_days)
+                session.update(request, enrollment_expiry=new_expiry)
+                client.add_group_to_token(pto_id, group_id, token, expiry=new_expiry)
                 status = Status.SUCCESS
-            else:  # already_enrolled
-                if group.expiresAt is None:
-                    # update expiration of existing enrollment, return success
-                    client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
-                    status = Status.SUCCESS
-                else:
-                    is_expired = _is_expired(group.expiresAt)
-                    is_within_reenrollment_window = _is_within_reenrollment_window(
-                        group.expiresAt, session.enrollment_reenrollment(request)
-                    )
-
-                    if is_expired or is_within_reenrollment_window:
-                        # update expiration of existing enrollment, return success
-                        client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
-                        status = Status.SUCCESS
-                    else:
-                        # re-enrollment error, return enrollment error with expiration and reenrollment_date
-                        status = Status.REENROLLMENT_ERROR
         else:  # flow does not support expiration
             if not already_enrolled:
                 # enroll user with no expiration date, return success
@@ -207,20 +197,19 @@ def enroll(
                     timeout=settings.REQUESTS_TIMEOUT,
                 )
                 status = Status.SUCCESS
-            else:  # already enrolled
-                if group.expiresAt is None:
-                    # no action, return success
-                    status = Status.SUCCESS
-                else:
-                    # remove expiration date, return success
-                    # (when you don't include an expiration date, Switchio will set the expiration date to null.)
-                    client.add_group_to_token(
-                        pto_id=pto_id,
-                        group_id=group_id,
-                        token=token,
-                        timeout=settings.REQUESTS_TIMEOUT,
-                    )
-                    status = Status.SUCCESS
+            elif not has_expiration:
+                # already enrolled, without an expiration date -> no action, return success
+                status = Status.SUCCESS
+            else:
+                # remove expiration date, return success
+                # (when you don't include an expiration date, Switchio will set the expiration date to null.)
+                client.add_group_to_token(
+                    pto_id=pto_id,
+                    group_id=group_id,
+                    token=token,
+                    timeout=settings.REQUESTS_TIMEOUT,
+                )
+                status = Status.SUCCESS
     except HTTPError as e:
         if e.response.status_code >= 500:
             status = Status.SYSTEM_ERROR
--- a/tests/pytest/enrollment_littlepay/test_enrollment.py
+++ b/tests/pytest/enrollment_littlepay/test_enrollment.py
@@ -31,6 +31,16 @@ def mocked_funding_source():
 
 
 @pytest.fixture
+def mocked_group_funding_source_future_expiry(mocked_funding_source):
+    return GroupFundingSourceResponse(
+        id=mocked_funding_source.id,
+        created_date="2023-01-01T00:00:00Z",
+        updated_date="2023-01-01T00:00:00Z",
+        expiry_date="2099-01-01T00:00:00Z",
+    )
+
+
+@pytest.fixture
 def mocked_group_funding_source_no_expiry(mocked_funding_source):
     return GroupFundingSourceResponse(
         id=mocked_funding_source.id,
@@ -41,7 +51,7 @@ def mocked_group_funding_source_no_expiry(mocked_funding_source):
 
 
 @pytest.fixture
-def mocked_group_funding_source_with_expiry(mocked_funding_source):
+def mocked_group_funding_source_past_expiry(mocked_funding_source):
     return GroupFundingSourceResponse(
         id=mocked_funding_source.id,
         created_date="2023-01-01T00:00:00Z",
@@ -287,7 +297,6 @@ def test_enroll_success_flow_supports_expiration(
     app_request,
     card_token,
     mocked_funding_source,
-    mocked_session_enrollment_expiry,
     mocked_session_flow,
     mocked_session_group,
     mocked_session_update,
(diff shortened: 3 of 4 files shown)


benefits/enrollment/enrollment.py
```diff
@@ -26,18 +27,27 @@ class Status(Enum):
     REENROLLMENT_ERROR = 4
 
 
+@dataclass
+class EnrollmentDecision:
+    status: Status
+    expiry_to_store: datetime | None = None
+    expiry_to_send: datetime | None = None
+    should_enroll: bool = False
+    should_remove_expiry: bool = False
```
thekaveman, line 36: `dataclass` models the decisions needed to drive provider-specific (Littlepay, Switchio) API calls, and session interaction

benefits/enrollment/enrollment.py
```diff
@@ -46,6 +56,58 @@ def _calculate_expiry(expiration_days: int):
     return expiry_datetime
 
 
+def _calculate_reenrollment_start(expiry: datetime, reenrollment_days: int) -> datetime:
+    """Returns the first date at which reenrollment can occur (renenrollment_days before the expiry)."""
+    return expiry - timedelta(days=reenrollment_days)
```
thekaveman, line 61: New helper essentially duplicates the logic from https://github.com/cal-itp/benefits/blob/1fffd359e2b8e579b8db2d46df05eee7a4e4078a/benefits/core/session.py#L122-L128

without relying on `session` data.

I was trying to see if we could remove that `session` helper entirely, but it is used on the re-enrollment error page (and at that point, the session should have valid expiration info).

LMK what you think, should we try and consolidate the logic? Try and remove the original session helper entirely?

benefits/enrollment/enrollment.py
```diff
@@ -46,6 +56,58 @@ def _calculate_expiry(expiration_days: int):
...
+
+
+def resolve_enrollment_decision(
+    flow: models.EnrollmentFlow, already_enrolled: bool, existing_expiry: datetime | None
+) -> EnrollmentDecision:
```
thekaveman, line 66: New helper centralizes the logic for the various branches / cases (see table in: https://github.com/cal-itp/benefits/pull/1989#issue-2211450180).

This is used in the provider-specific `enroll()` functions (Littlepay, Switchio) to drive the decisions around which API endpoints to call.

benefits/enrollment/enrollment.py
```diff
@@ -46,6 +56,58 @@ def _calculate_expiry(expiration_days: int):
...
+    if not already_enrolled or existing_expiry is None:
+        return EnrollmentDecision(
+            status=Status.SUCCESS,
+            expiry_to_store=new_expiry,
+            expiry_to_send=new_expiry,
+            should_enroll=True,
+        )
+
+    # expired enrollment or within reenrollment window
+    reenrollment_start = _calculate_reenrollment_start(existing_expiry, flow.expiration_reenrollment_days)
```
thekaveman, line 101: Using that new helper with logic extracted/duplicated from the `session` (see [comment](https://github.com/cal-itp/benefits/pull/3977#discussion_r4039733137) above).

benefits/enrollment_littlepay/enrollment.py
```diff
@@ -78,60 +79,30 @@ def enroll(request, card_token) -> tuple[Status, Exception]:
...
-                        funding_source_id=funding_source.id,
-                        expiry=session.enrollment_expiry(request),
-                    )
-                    status = Status.SUCCESS
-                else:
-                    is_expired = _is_expired(group_funding_source.expiry_date)
-                    is_within_reenrollment_window = _is_within_reenrollment_window(
-                        group_funding_source.expiry_date, session.enrollment_reenrollment(request)
-                    )
+        decision = resolve_enrollment_decision(flow, already_enrolled, existing_expiry)
```
thekaveman, line 85: Run the centralized decision logic

benefits/enrollment_littlepay/enrollment.py
```diff
@@ -78,60 +79,30 @@ def enroll(request, card_token) -> tuple[Status, Exception]:
...
-                    if is_expired or is_within_reenrollment_window:
-                        # update expiration of existing enrollment, return success
-                        client.update_concession_group_funding_source_expiry(
-                            group_id=group_id,
-                            funding_source_id=funding_source.id,
-                            expiry=session.enrollment_expiry(request),
-                        )
-                        status = Status.SUCCESS
+        if decision.expiry_to_store is not None:
+            session.update(request, enrollment_expiry=decision.expiry_to_store)
```
thekaveman, line 89: Store the expiration once in session, if needed.

The difference here from the bugged case is:

- We update the session before any usage of the data in session is invoked
- We refactored most of the logic from the various `enrollment` modules to _not_ rely on session data (e.g. `views` still uses session data).

benefits/enrollment_littlepay/enrollment.py
```diff
@@ -78,60 +79,30 @@ def enroll(request, card_token) -> tuple[Status, Exception]:
...
+
+        if status is Status.SUCCESS:
```
thekaveman, line 91: Use decision to guide interaction with Littlepay API

benefits/enrollment_switchio/enrollment.py
```diff
@@ -165,62 +170,32 @@ def enroll(
...
-            else:  # already_enrolled
-                if group.expiresAt is None:
-                    # update expiration of existing enrollment, return success
-                    client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
-                    status = Status.SUCCESS
+        token_group = _get_group_for_token(client, pto_id, group_id, token)
+        already_enrolled = token_group is not None
+        existing_expiry = token_group.expiresAt if already_enrolled else None
+
+        decision = resolve_enrollment_decision(flow, already_enrolled, existing_expiry)
```
thekaveman, line 177: Run the centralized decision logic

benefits/enrollment_switchio/enrollment.py
```diff
@@ -165,62 +170,32 @@ def enroll(
...
+        status = decision.status
+
+        if decision.expiry_to_store is not None:
+            session.update(request, enrollment_expiry=decision.expiry_to_store)
```
thekaveman, line 181: Store the expiration once in session, if needed.

The difference here from the bugged case is:

- We update the session before any usage of the data in session is invoked
- We refactored most of the logic from the various `enrollment` modules to _not_ rely on session data (e.g. `views` still uses session data).

benefits/enrollment_switchio/enrollment.py
```diff
@@ -165,62 +170,32 @@ def enroll(
...
+
+        if status is Status.SUCCESS:
```
thekaveman, line 183: Use decision to guide interaction with Switchio API

thekaveman (commented):
Some more detailed line-by-line explanation

benefits/enrollment_switchio/enrollment.py
```diff
@@ -165,62 +170,32 @@ def enroll(
...
-            if not already_enrolled:
-                # enroll user with an expiration date, return success
-                client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
-                status = Status.SUCCESS
-            else:  # already_enrolled
-                if group.expiresAt is None:
-                    # update expiration of existing enrollment, return success
-                    client.add_group_to_token(pto_id, group_id, token, expiry=session.enrollment_expiry(request))
-                    status = Status.SUCCESS
+        token_group = _get_group_for_token(client, pto_id, group_id, token)
```
thekaveman, line 173: Renamed this from `group` to `token_group`, since we already take a `group` parameter in this function, and so the variable was overriding/hiding the incoming parameter.

This didn't seem like a direct problem, since we store the `group_id` from the incoming `group` param, and that is all we need... but still something that could bite us later if we weren't paying attention.

benefits/enrollment/templates/enrollment/reenrollment-error.html
```diff
@@ -24,5 +24,5 @@
 {% endblock paragraphs %}
 
 {% block call-to-action-button %}
-  {% include "core/includes/button--origin.html" %}
+  {% include "core/includes/button--index.html" %}
```
jgravois, line 27: 👍

benefits/enrollment/enrollment.py
```diff
@@ -46,6 +56,61 @@ def _calculate_expiry(expiration_days: int):
...
+    """
+    # flow does not support expiration
+    if not flow.supports_expiration:
+        if not already_enrolled:
+            # not yet enrolled
+            return EnrollmentDecision(status=Status.SUCCESS, should_enroll=True)
+
+        if existing_expiry is not None:
+            # already enrolled with existing expiry: remove
+            return EnrollmentDecision(status=Status.SUCCESS, should_remove_expiry=True)
```
jgravois, line 83: this is solely to catch out the (unlikely) case that a flow previously supported expiration, but no longer does, correct?

benefits/enrollment/enrollment.py
```diff
@@ -26,18 +27,27 @@ class Status(Enum):
     REENROLLMENT_ERROR = 4
 
 
+@dataclass
+class EnrollmentDecision:
+    status: Status
+    expiry_to_store: datetime | None = None
+    expiry_to_send: datetime | None = None
```
jgravois, line 34: just to make sure i'm understanding:

`expiry_to_store` - this is the value we write to the session. 

for `REENROLLMENT_ERROR`, using the existing expiration allows us to display UI to the user indicating that they need to try again later.

Image

when re-enrollment _is_ possible, we use the proposed expiration in order to display that value to the user on success and send analytics associated with the successful re-enrollment.

(just typing this out helped me wrap my mind around this 😂)

jgravois (commented):
the only reason this consolidating PR is more green than red is because of all the new great test coverage. 🥳

i tested all the (many) permutations outlined in the PR description and the behavior was exactly as i expected. thanks for the thorough writeup.

i agree that it makes sense to consolidate this logic in anticipation of a third looming transit processor.

jgravois (approved):

benefits/enrollment/enrollment.py
thekaveman: Something like that, yeah :sweat_smile: 

It is hard to believe we would ever run down this code path, but... it is a possibility, so we have it handled Just In Case :tm:

benefits/enrollment/enrollment.py
thekaveman: > `expiry_to_store` - this is the value we write to the session.

Exactly :+1: 

> for `REENROLLMENT_ERROR` ...

That's right. And in that case, the "existing expiration" _becomes_ `expiry_to_store`.

> when re-enrollment _is_ possible ...

Yep, in that case we have `expiry_to_store == expiry_to_send`

Basically we should always have one of:

* `expiry_to_store` populated, `expiry_to_send` empty (re-enrollment error)
* `expiry_to_store == expiry_to_send` (standard enrollment, or successful re-enrollment)
* both empty (expiration not supported)

benefits/enrollment/enrollment.py
angela-tran: Hmm yeah... I could see removing this `enrollment_reenrollment` helper and having the 2 consumers (the self-service `ReenrollmentErrorView` and the in-person `ReenrollmentErrorView`) calculate the reenrollment date themselves. Actually, the self-service view could just have a method to calculate it for use by itself and the in-person view.

It seems like having this "global" helper for getting the reenrollment date from any possible context (e.g. flow doesn't support expiration, flow hasn't been selected, etc.) isn't that useful.

I notice some technical debt / cruft with the `ReenrollmentErrorView`s and their templates, so perhaps it'd be better to do this refactor (removing `enrollment_reenrollment` helper) as a follow-up PR?

angela-tran: Hmm yeah... I could see removing this `enrollment_reenrollment` helper and having the 2 consumers (the self-service `ReenrollmentErrorView` and the in-person `ReenrollmentErrorView`) calculate the reenrollment date themselves. Actually, the self-service view could just have a method to calculate it for use by itself and the in-person view.

It seems like having this "global" helper for getting the reenrollment date from any possible context (e.g. flow doesn't support expiration, flow hasn't been selected, etc.) isn't that useful.

I notice some technical debt / cruft with the `ReenrollmentErrorView`s and their templates, so perhaps it'd be better to do this refactor (removing `enrollment_reenrollment` helper) as a follow-up PR?

benefits/enrollment_littlepay/enrollment.py
```diff
@@ -78,60 +79,30 @@ def enroll(request, card_token) -> tuple[Status, Exception]:
...
-                    status = Status.SUCCESS
+                        client.link_concession_group_funding_source(
+                            group_id=group_id, funding_source_id=funding_source.id, expiry=decision.expiry_to_send
+                        )
                 else:
-                    # remove expiration date, return success
-                    raise NotImplementedError("Removing expiration date is currently not supported")
+                    client.update_concession_group_funding_source_expiry(
+                        group_id=group_id, funding_source_id=funding_source.id, expiry=decision.expiry_to_send
+                    )
```
angela-tran, line 105: This is so much more readable! Awesome work with extracting the key pieces of data that determine the decision.

I thought about how this code, which consumes the populated `EnrollmentDecision`, is duplicated across Littlepay and Switchio. I think it makes sense to leave it. It might end up being the same for INIT as well when we get to implementing it, and if so, we could look into making something reusable then if it makes sense. Probably best to not prematurely optimize it

angela-tran (approved):
This is really great work @thekaveman! Thank you for cleaning this up. I see how a major contributor to the bug was the conflation of the need to store an expiration date for various messaging with the need to send an expiration date. That was great how you first fixed the bug without removing the conflation and then split out the two concepts entirely.

tests/pytest/enrollment/test_enrollment.py
```diff
@@ -317,3 +331,117 @@ def test_handle_enrollment_results_reenrollment_error(app_request, mocked_analyt
...
+@pytest.mark.parametrize(
+    "already_enrolled,existing_expiry,should_enroll,should_remove_expiry",
+    [
+        (False, None, True, False),
+        (True, None, False, False),
+        # the specific existing_expiry shouldn't matter here, since this flow doesn't support expiration anyway
+        (True, timezone.datetime(2026, 9, 15), False, True),
+    ],
+)
+def test_resolve_expiry_decision__expiration_not_supported(
```
angela-tran, line 357: Noticed a small typo here

```suggestion
def test_resolve_enrollment_decision__expiration_not_supported(
```

benefits/enrollment/enrollment.py
lalver1: Since we know that currently it's only used in `ReenrollmentErrorView` and the context for that page, I agree that it looks like we could remove `session.enrollment_reenrollment`. But I'm still wondering if in the near future we could have a need for it, will we ever need to pass the reenrollment date to another page 🤔. In any case, a follow-up PR makes sense 👍

lalver1 (dismissed):
Great job @thekaveman! 

I like how this fix abstracts the enrollment decision but you can still easily follow the basic logic: 1) supports expiration? 2) funding source already enrolled in group? 3) update expiration?

I tested all the cases in the PR's description both with Littlepay and Switchio and saw no issues 👍 Thanks for the tips on how to review the cases!

thekaveman:
Fixed a typo that @angela-tran caught out, dismissing the reviews.

Going to set this to auto-merge and whoever can get to it first is fine :smile: Thanks y'all!

jgravois (approved):
🌈🌈

PR #3979, opened 2026-08-12
Fix: admin portal link text

### LINKED ISSUE #3774: In person: agency staff landing page incorrectly mentions Littlepay when agency is configured for Switchio
## To Reproduce

1. login to https://dev-benefits.calitp.org/admin as cst-user

## Expected behavior

I'd expect to see a more generic label appropriate to all transit processors.

## Screenshots

Image

thekaveman:
Closes #3774 

---

Fairly straightforward fix to figure out which `TransitProcessorConfig` subclass is in use based on the logged in user's agency. From there, just hard-code the correct portal link text in context and use it on the template.

More interesting in this PR are the test changes: I feel like we've run across this a few times where we wanted to `pytest.mark.parameterize` with a list of _fixtures_, i.e. inject different fixtures for different instances of the test (vs. copying a test over and over for each fixture). Well it turns out there is a mostly non-hacky way to do it: 

`pytest` has a built-in fixture called [`request`](https://docs.pytest.org/en/latest/reference/reference.html#request) with a helper function [`getfixturevalue(argname)`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.FixtureRequest.getfixturevalue):

> Dynamically run a named fixture function.
>
> Declaring fixtures via function argument is recommended where possible. But if you can only decide whether to use another fixture at test setup time, you may use this function to retrieve it inside a fixture or test function body.

So rather than have multiple instances of the same test that pass a different fixture for `model_LittlepayConfig` and `model_SwitchioConfig`, this PR parameterizes the existing test via the above to keep things DRY :+1:

---

I went with naming the processor vs. something more generic (as indicated in the original ticket) because it isn't clear to me if something like `"Launch Transit Processor Control Portal"` would be recognizable to an agency staffer. E.g. is the phrase `"Transit Processor"` widely understood? This avoids that question, and related questions like _Well what text should we use instead?_ and just keeps things simple.

## Review

- Sign in to the admin as an agency user with a Littlepay agency, see the expected portal link text
- Sign in to the admin as an agency user with a Switchio agency, see the expected portal link text

### DIFF
--- a/benefits/admin.py
+++ b/benefits/admin.py
@@ -59,13 +59,20 @@ class BenefitsAdminSite(AdminSite):
 
                 if agency.transit_processor_config is not None:
                     transit_processor_portal_url = agency.transit_processor_config.portal_url
+                    transit_processor_portal_text = "Control Portal"
+                    if agency.littlepay_config:
+                        transit_processor_portal_text = "Littlepay " + transit_processor_portal_text
+                    elif agency.switchio_config:
+                        transit_processor_portal_text = "Switchio " + transit_processor_portal_text
                 else:
                     transit_processor_portal_url = ""
+                    transit_processor_portal_text = ""
 
                 context.update(
                     {
                         "has_permission_for_in_person": has_permission_for_in_person,
                         "transit_processor_portal_url": transit_processor_portal_url,
+                        "transit_processor_portal_text": transit_processor_portal_text,
                         "title": f"{agency.long_name} | {self.index_title} | {self.site_title}",
                         "start_url": (
                             routes.IN_PERSON_ADDITIONAL_AGENCIES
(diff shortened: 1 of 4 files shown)


jgravois (approved):
i tested locally and this branch does what it says on the tin.

> E.g. is the phrase `"Transit Processor"` widely understood? This avoids that question, and related questions like _Well what text should we use instead?_ and just keeps things simple.

i definitely agree with this sentiment. Personally i would have taken the tack of just calling it 'Launch Control Panel' because I think the clarification text above does a pretty good job of setting context for the link all on its own. far from a hill i'd die on though. i'm allergic to conditionally generated copy but sometimes its a fact of life.

regardless, i dig the pytest fixture helper you summoned for this.

thekaveman:
> i'm allergic to conditionally generated copy but sometimes its a fact of life.

yeah I hear you for sure. I guess if I would have looked closely, this box has the heading of `"Transit Processor"` and with the context you point out, it should be obvious to just state `"Launch Control Portal"`.

New PR incoming :smiling_imp:

PR #3982, opened 2026-08-12
CI: notify Sentry of new releases

### LINKED ISSUE #3685: Notify Sentry about Benefits releases
Sentry can be notified about specific released versions of the app, which brings with it a lot of additional features and insights. Read more: https://docs.sentry.io/product/releases/

We can make use of the [`getsentry/action-release`](https://github.com/getsentry/action-release) step in workflow once the new instance is configured.

## Acceptance Criteria

- [ ] All production releases are tracked in Sentry using their version tag
- [ ] _(? should `dev`/`test` deploys also be sent to Sentry ?)_

thekaveman:
Closes #3685 

---

~Per Sentry docs guidance, notify of the release _before_ it is deployed: https://docs.sentry.io/product/releases/~

~> We recommend notifying Sentry about a new release before deploying it. But if you don't, Sentry will automatically create a release entity in the system the first time it sees an event with that release identifier.~

Notify of the release as a separate job, in between the existing `deploy` (e.g. Docker and Terraform) and `release` (e.g. GitHub release) jobs.

Only notify for `test` and `prod`, using the tag as the release version, with a `benefits@` project prefix since releases are _global across a Sentry org._

The referenced environment secret and variables have been configured.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -216,7 +216,7 @@ jobs:
           SENTRY_PROJECT: ${{ vars.SENTRY_PROJECT }}
         with:
           environment: ${{ needs.setup.outputs.env_name }}
-          release: ${{ github.ref_name }}
+          release: benefits@${{ github.ref_name }}
 
   release:
     needs: deploy
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -216,7 +216,7 @@ jobs:
           SENTRY_PROJECT: ${{ vars.SENTRY_PROJECT }}
         with:
           environment: ${{ needs.setup.outputs.env_name }}
-          release: ${{ github.ref }}
+          release: ${{ github.ref_name }}
 
   release:
     needs: deploy
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -197,6 +197,27 @@ jobs:
           skip-comment: true
           working-directory: terraform
 
+  sentry:
+    needs: [setup, deploy]
+    if: ${{ needs.setup.outputs.is_tag == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      # notify Sentry of new release
+      - name: Create Sentry release
+        uses: getsentry/action-release@ff07929a6537bac57790c3451cf4d364aca38528 # v3.7.0
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: ${{ vars.SENTRY_ORG }}
+          SENTRY_PROJECT: ${{ vars.SENTRY_PROJECT }}
+        with:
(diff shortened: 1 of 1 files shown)


jgravois (commented):
nit of all nits, but personally i'd prefer to see this step come _after_ the `terraform apply`.

1. if it fails mysteriously, i'd hate for it to short-circuit the actual deployment
1. if the deployment fails, we'd be better off _not_ notifying sentry.

thekaveman:
> nit of all nits, but personally i'd prefer to see this step come _after_ the `terraform apply`.

I mean, I definitely hear you. I'm also following explicit guidance from Sentry (linked and copied above) to notify _before_ deployment occurs. I'm not sure I totally understand that guidance so let me do a little more digging to see if I can find any better reasoning / explanation.

jgravois:
if they've got some sugar wired up that expects/requires the notification to come first, i'm happy to adhere to their guidance.

we might consider making sure the new step is (non-blocking) prerequisite, but i'm not _that_ fussed about it.

if the actual deployment is what's failing, the sentry notification is the least of our worries.

thekaveman:
Seems we aren't the only ones confused by some of the docs guidance: https://github.com/getsentry/sentry-docs/issues/5458 There is actually conflicting guidance depending on where you look, per links in that issue.

And the answer? https://github.com/getsentry/sentry-docs/issues/5458#issuecomment-1243827325

> finalization doesn't affect deploys so both orders will work the same way

:+1: :face_exhaling:

While we aren't using the `sentry-cli deploy` mentioned specifically in that issue / docs, I read all this to mean that, it actually doesn't matter :facepalm: 

Since the [`sentry-release`](https://github.com/marketplace/actions/sentry-release#parameters) GHA step defaults to `finalize = true` (and uses `sentry-cli release` under the hood) I'm inclined to follow our gut feeling and move this step _after_ the `terraform apply` as you suggest.

But I'm glad I spent a little more time on this, because I came across this in the [CLI docs](https://docs.sentry.io/cli/releases/) which I hadn't noticed before:

> Note that releases are global per organization. If you want the releases in different projects to be treated as separate entities, make the version name unique across the organization. For example, if you have projectA and projectB that share version numbers, you can name the releases `projectA-1.0` and `projectB-1.0` respectively.

Which tells me we should be pre-pending `benefits` here to the version number. It is entirely possible we have a project (either `eligibility-server` now, or another in the future) that uses the same versioning scheme and is tracked in Sentry, and it sounds like we'd be SOL if we ever released multiple during the same month. So I'll make that amendment as well.

lalver1:
I found some more relevant details about when to run this Action at the very bottom of this [documentation page](https://docs.sentry.io/product/releases/setup/release-automation/github-actions/). Seems like the note about running before a deploy is only applicable to JavaScript applications?

> It's important the Sentry Release GitHub Actions runs after any build steps in your pipeline but before deployment and to deploy the files that have been built after the action runs so that Sentry can properly un-minify stack traces for you.

thekaveman:
This has me wondering if we need to align the `release` we configure [with the SDK setup](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/sentry.py#L93) with this notified release, and I assume the answer would be yes...

PR #3983, opened 2026-08-13
Fix: correctly parse hyphen in agency paths

### LINKED ISSUE #3863: Allow hyphen in transit agency slugs
> Re:
> 
> Slug, phone number, card schemes added above (cc [@jgravois](https://github.com/jgravois) [@angela-tran](https://github.com/angela-tran) [@indexing](https://github.com/indexing)) Still waiting on logo.
> 
> Benefits kick-off call is now scheduled for July 20th.
> 
> _(from https://github.com/cal-itp/benefits/issues/3753#issuecomment-4744433175)_
> 
> and 
> 
> - **Slug:** Simi-valley-transit (or similar, such as simivalleytransit)
> 
> _(from the issue description)_,
> 
> it appears our code will [raise an error](https://cal-itp.slack.com/archives/C022HHSEE3F/p1783553595360919) if the `slug` contains hyphens, so I will update the Simi Valley slug in `test` and `prod` to `simivalleytransit`. 

 _Originally posted by @angela-tran in [#3753](https://github.com/cal-itp/benefits/issues/3753#issuecomment-4920172154)_

## More detail

This is specifically a problem with the `TransitAgencyPathConverter` that parses a path value and tries to lookup a Transit Agency. The regex is overly restrictive:

https://github.com/cal-itp/benefits/blob/3a930abe827601a8b541a0d464648f6d8979eb3a/benefits/core/urls.py#L29

It should allow hyphens, which are perfectly valid in URL slugs.

thekaveman:
Closes #3863 

---

[`TransitAgency.slug` is a `SlugField`](https://github.com/cal-itp/benefits/blob/9789d4b9f614320088daaf8bc37c3be4cd28fa44/benefits/core/models/transit.py#L69) and so allows configuring an agency with hyphens in its slug, e.g. 
`simi-valley-transit`.

However the URL path parsing in [`TransitAgencyPathConverter`](https://github.com/cal-itp/benefits/blob/9789d4b9f614320088daaf8bc37c3be4cd28fa44/benefits/core/urls.py#L29) failed to account for hyphens, leading to an error for agencies so configured, e.g.

```log
Reverse for 'agency_index' with arguments '('simi-valley-transit',)' not found. 
1 pattern(s) tried: ['(?P<agency>[a-zA-Z]{3,})\\Z']
```

This PR fixes the regex in `TransitAgencyPathConverter` to correctly allow hyphens, and adds test coverage for this class as well.

## Review

- Edit any active `TransitAgency` instance to have a `slug` with hyphens, e.g. `c-s-t`
- Navigate to the agency index URL e.g. `/c-s-t`
- See the expected agency-specific entry page

### DIFF
--- a/benefits/core/urls.py
+++ b/benefits/core/urls.py
@@ -26,7 +26,7 @@ class TransitAgencyPathConverter:
     """Path converter to parse valid TransitAgency objects from URL paths."""
 
     # used to test the url fragment, determines if this PathConverter is used
-    regex = "[a-zA-Z]{3,}"
+    regex = "[a-zA-Z-]{3,}"
 
     def to_python(self, value):
         """Determine if the matched fragment corresponds to an active Agency."""
--- a/tests/pytest/core/test_urls.py
+++ b/tests/pytest/core/test_urls.py
@@ -1,0 +1,47 @@
+"""
+h/t to this blog post for the inspiration for these tests
+https://adamj.eu/tech/2025/08/01/django-custom-url-converter-string/
+"""
+
+import pytest
+from django.urls import path, reverse
+
+# Import forces registration
+from benefits.core.urls import TransitAgencyPathConverter  # noqa: F401
+
+urlpatterns = [
+    path(
+        "<agency:agency>",
+        lambda *args, **kwargs: None,  # dummy view
+        name="sample",
+    )
+]
+sample_path = urlpatterns[0]
+
+
+@pytest.mark.django_db
+class TestTransitAgencyPathConverter:
+    def test_active_agency(self, model_TransitAgency):
+        result = sample_path.resolve(model_TransitAgency.slug)
+        assert result.kwargs["agency"] == model_TransitAgency
+
+        result = reverse("sample", urlconf=__name__, kwargs={"agency": model_TransitAgency})
+        assert result == f"/{model_TransitAgency.slug}"
+
+    def test_active_agency__hyphen_slug(self, model_TransitAgency):
+        model_TransitAgency.slug = "c-s-t"
+        model_TransitAgency.save()
+
+        result = sample_path.resolve(model_TransitAgency.slug)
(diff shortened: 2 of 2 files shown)


jgravois (approved):
looks good!

not germane to this PR specifically, but i noticed testing this PR that a trailing slash also induces an error.

http://localhost:11369/foo-bar/

Image

i wonder if there's a way to get Django to be more resilient to that too...

thekaveman:
> i wonder if there's a way to get Django to be more resilient to that too...

It turns out this is pretty straightforward to allow for, but it does change the behavior slightly.

There is an [`APPEND_SLASH`](https://docs.djangoproject.com/en/5.2/ref/settings/#append-slash) setting, which is `True` by default, and:

> When set to `True`, if the request URL does not match any of the patterns in the URLconf and it doesn’t end in a slash, an HTTP redirect is issued to the same URL with a slash appended.

So if we amend our [existing `path`](https://github.com/cal-itp/benefits/blob/9789d4b9f614320088daaf8bc37c3be4cd28fa44/benefits/core/urls.py#L63) like so:

```diff
- path("<agency:agency>", AgencyIndexView.as_view(), name=routes.name(routes.AGENCY_INDEX)),
+ path("<agency:agency>/", AgencyIndexView.as_view(), name=routes.name(routes.AGENCY_INDEX)),
```

Then it would all "just work": requests to `/some-agency` get redirected to `/some-agency/` which matches the actual path pattern (so of course direct requests work as well).

So the difference is, these URLs would always end up with a trailing slash. A slight change, but a change nonetheless. For better or worse, it would _only_ be this path that enforces the trailing slash.

I don't know how often (if ever?) we see this as a problem, so I'm less inclined to tackle it.

jgravois:
> I don't know how often (if ever?) we see this as a problem, so I'm less inclined to tackle it.

good to know its a straightforward change. i'm happy to let sleeping dogs lie until/unless we see the error crop up in sentry logs.

PR #3994, opened 2026-08-17
Change typeface to Noto Sans throughout Benefits

### LINKED ISSUE #3961: Update Benefits typography to Noto Sans
This breaks off the font change from #3660 

## Acceptance Criteria

- [ ] The font used throughout the entire Benefits app (self-service and in-person) is `Noto Sans`

Scotchester:
Closes #3961 

---

There are four (significant) commits here:

1. Straightforward replacement of Roboto with Noto Sans
2. Remove Public Sans, which was previously used only for the "Get started with Login.gov" button, which was updated to also use Noto Sans in the current designs
3. Eliminating all extended letter-spacing rules, which are not desired with Noto
4. Adjusting the weight of some emphasized non-heading text elements

@cmajel, let me know if you'd like to screenshare and get a live demo.

### Screenshots

Before | After
------ | -----
Benefits front end with Roboto as the primary typeface | Benefits front end with Noto Sans as the primary typeface


benefits/static/css/styles.css
```diff
@@ -1,14 +1,5 @@
 @import "variables.css";
 
-@font-face {
-  font-family: "Public Sans";
-  font-weight: var(--bold-font-weight);
-  font-style: normal;
-  src:
-    local("PublicSans"),
-    url("../fonts/PublicSans-Bold.woff") format("woff");
```
jgravois, line 9: any reason _not_ to yank the compressed font itself too?

https://github.com/cal-itp/benefits/tree/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/static/fonts

jgravois (commented):
i'm all for you scheduling a live demo and/or @cmajel taking this branch for a spin locally, but a few before/after screenshots for posterity in the PR description would be sublime too. 😇

benefits/static/css/styles.css
Scotchester: None other than it was a silly oversight 😁

jgravois: unused true type font, you... _shall... not.. pass_.

gandalf

Scotchester:
> a few before/after screenshots for posterity in the PR description would be sublime too. 😇

Good call :) Added!

benefits/static/css/variables.css
```diff
@@ -11,17 +11,12 @@
   --medium-font-weight: 500;
   --bold-font-weight: 700;
   --bs-body-font-size: calc(18rem / 16);
-  --bs-font-sans-serif:
-    Roboto, system-ui, -apple-system, "Segoe UI", sans-serif,
-    "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  --bs-font-sans-serif: "Noto Sans", system-ui;
```
jgravois, line 15: just curious, why get rid of all the other fallbacks?

(i'm having a bit of deja-vu feeling like once in the past i proposed we get rid of fallbacks that i thought were redundant and you were the one to talk me off the ledge)

jgravois (commented):
this looks great to me and i love how much custom CSS you're yanking, but i'll hold off on blessing this until after we've received a benediction from @cmajel.

benefits/static/css/variables.css
Scotchester: Ah yes, forgot to comment on this.

I think [this is the conversation you're remembering](https://github.com/cal-itp/calitp.org/pull/534#discussion_r2582164550), but it was about the monospace font stack in the DSDL.

When setting up sans-serif font stacks for the DSDL, Christine agreed with using a system font stack (search Compiler Slack for `system-ui`). I determined that `system-ui` support was widespread enough and we have no current emoji needs that `system-ui` was the only fallback we need now. I applied that same line of thinking to Benefits here.

jgravois: ah, thanks for clarifying 👑

cmajel:
👋🏻 I'm having trouble getting my local branch up so a demo walkthrough would be awesome. I'll schedule a time

One adjustment from what I can see visually. It looks like the body links are set to `font-weight:bold/700`and for Noto I adjusted them to `font-weight:600`. My basic logic is headings are bold/700, other emphasized elements are semibold.

thekaveman (commented):
Will also leave the final approval to @cmajel, but this looks great to me. I absolutely love a PR that deletes whole swaths of CSS 😁 👏 

---

Tiniest of notes: I see commit 7ca01c1a1172ba3563029dda7f18699b01966e97 is intended to be rebased away into the commit that removed the styles for Public Sans (a74aacb7439ed0679bdcb28a2b84e4e1d669d62e), and I have no issue with how this was approached.

I just want to make sure you are aware of the built-in git feature to handle this sort of thing. 

E.g. in making the second commit `7ca01c1` you can say:

```shell
git commit --fixup a74aacb7439ed0679bdcb28a2b84e4e1d669d62e
```

_Docs for [`git commit --fixup`](https://git-scm.com/docs/git-commit.html#Documentation/git-commit.txt---fixupamendrewordcommit)._

This creates a commit message for you like

```log
fixup! style: Remove Public Sans
```

Then later, when you're ready to clean things up, call:

```
git rebase --interactive --autosquash
```

_Docs for [`git rebase --autosquash`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash)._

This will automatically reorder your `fixup!` commits where they should be, mark them to be squashed/fixup'd into their respective previous commit, and allow you to review everything before continuing.

FWIW, I use some aliases to make this all a little simpler: 

- [`git fix <SHA>`](https://github.com/thekaveman/dotfiles/blob/da87f069ae55b09cd00b1cd78707ca048aae47ec/git/gitconfig#L9)
- [`git autosquash`](https://github.com/thekaveman/dotfiles/blob/da87f069ae55b09cd00b1cd78707ca048aae47ec/git/gitconfig#L5)

jgravois:
thanks for sharing this chestnut @thekaveman. i've done more interactive rebases that could have used `--autosquash` than i could count and i frequently waffle between pushing followup commits that feel extraneous and rewriting history when it feels premature.

going forward i'll definitely give this a try and save the rebase for immediately pre-merge. 😎

Scotchester:
Whoa, that's an awesome tip @thekaveman! This feels like something I heard about a long time ago but never put into practice and forgot about it. I'll definitely be giving it a try!

cmajel (dismissed):

Scotchester:
Sorry, @cmajel, I pushed a fix for the button line-height issue we saw on the call and it dismissed your review. Please approve again?

jgravois (approved):
🌈🌈

cmajel (approved):

Issue #3645, opened 2026-04-01
UX spike for additional in-person enrollment options


cmajel:
We now have a clearer sense of the eligibility policies we'll be pursing for groups like disabled veterans, youth, and people with disabilities.

In person enrollment is an immediate place where we can offer a way to enroll certain rider groups while we investigate digital integrations for self service.

This would be a spike to investigate:
* How we would add disabled veterans to in person enrollment
* Future forward design for adding more categories 

In person enrollments we expect to prioritize:
* Up first:
   * PwD: Disabled veterans 
   * Calfresh cardholders
* Future facing:
   * Additional PwD categories (medical certification, DMV plates/placards, SSI/SSDI)
   * Youth / under 18

This will also be an opportunity to evaluate admin UI design primitives (#2648) and evaluate moving shared components a central library (https://github.com/cal-itp/benefits/issues/2649)  

## Acceptance Criteria

- [x] Designs are shared with the product and engineering lead
- [ ] Design review scheduled with developers to review annotations and open questions

cmajel:
I've added a series of explorations on this topic to this Figma branch: [added pathways + primitives ](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/branch/f8u8mOaGvxFfeXGszyq2Lo/Cal-ITP-Benefits-Admin?node-id=3691-26354&t=AbShJMeeDQncIkdI-0)

cc @indexing and @thekaveman for thoughts.

There are a number of explorations here in addition to the immediate need, so I'll note below which need decisions. After you all review perhaps the three of us could review together re immediate vs future plan. I've indicated my preferred options with a 🟢 

* Adding disabled veterans and Calfresh (needs decision)
  * I have an option that [leverages existing patterns](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/branch/f8u8mOaGvxFfeXGszyq2Lo/Cal-ITP-Benefits-Admin?node-id=3655-9173&t=AbShJMeeDQncIkdI-0), and one that explores [adding a distinct `verification` page](https://www.figma.com/design/Y7fffLWV4cvZMgLM0fLvtP/branch/f8u8mOaGvxFfeXGszyq2Lo/Cal-ITP-Benefits-Admin?node-id=3655-9174&t=AbShJMeeDQncIkdI-0) (which I recommend). 
    * As we support pathways with criteria, we start to outgrow the current checkbox on the `eligibility` page, and this gives us more real estate.
[...]

PR #3995, opened 2026-08-18
Refactor: move in-person policy language to model

### LINKED ISSUE #3996: In-person: move policy language to a model field
When working through #3986, with the UX changes envisioned in #3645, and the general understanding that additional In-person enrollment options will be coming online -- it is clear we need more flexibility and agility in how these are configured within the app.

One area to tackle now: the policy language, next to the checkbox where agency staff affirm the rider's eligibility. Currently this is hard-coded in the [`InPersonEligibilityForm.eligibility_policies`](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/in_person/forms.py#L35) attribute, necessitating a code change and release to add/update that language.

Since the In-person enrollment experience, and Benefits Administrator more generally, are not translated, we can easily move these text strings to be backed by a model field and managed in the database. The effect(s) would be:

- A more streamlined configuration process for new In-person flows, like #3986 
- A simpler experience for maintenance updates to the language, such as those outlined in #3645
- Opening the possibility for non-engineers (e.g. Product) to be able to update policy language as needs arise

This, coupled with #3499, would get us to a place where an In-person flow could be entirely configured using the Benefits Administrator, requiring zero code changes or deployments, which would be a major efficiency win for the product.

## Acceptance Criteria

- [ ] A new field on `EnrollmentFlow` captures the policy language currently found in the `InPersonEligibilityForm`
- [ ] All existing `EnrollmentFlow` policy language is migrated, requiring zero maintenance or downtime of existing In-person flows
- [ ] The In-person UX does not otherwise change for agency staff nor riders: policy language is still displayed for the selected flow, next to the confirmation checkbox as it is today

thekaveman:
Closes #3996 

---

The main thrust of this PR is to add a new field to the `EnrollmentFlow` model called `in_person_policy`, a simple text field that holds the language we display in the Admin In-person flow, next to the checkbox where agency staff affirm the rider's eligibility.

## Before

Before this PR, to configure a (new) flow for In-person Enrollment, the process was roughly:

1. Engineers add a new [`SystemName` entry](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/core/models/enrollment.py#L26) for the flow
2. Engineers update the [`SUPPORTED_IN_PERSON_FLOWS` list](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/core/models/enrollment.py#L111) with the new `SystemName` entry
3. Engineers add new policy language to the [`InPersonEligibilityForm.eligibility_policies`](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/in_person/forms.py#L35) attribute
4. Engineers deploy/release the code changes
5. (Someone) configures a new `EnrollmentFlow` instance and links to the relevant `TransitAgency` instances

## After

After this PR, to configure a (new) flow for In-person Enrollment, the process will roughly be:

1. Engineers add a new [`SystemName` entry](https://github.com/cal-itp/benefits/blob/0b93440194c02a25f1022b2f40dcce223b6189ba/benefits/core/models/enrollment.py#L26) for the flow
2. Engineers deploy/release the code changes
3. (Someone) configures a new `EnrollmentFlow` instance, including a value for the `in_person_policy` field, and links to the relevant `TransitAgency` instances

## Review

- On the `main` branch, setup the database with local fixtures. Then swap to this branch and run `python manage.py migrate`. Verify existing flows get their policy language migrated in.
- Verify the field error/validation when saving a flow marked for In-person without policy language.
- Verify the expected policy language is displayed when running an In-person enrollment, and swapping between flow options.

## Extra

Since I was touching this area of the code, I figured I would consolidate the `EnrollmentFlow.clean()` logic into `EnrollmentFlowForm.clean()` given we had a note to do that at some point, and it was annoying trying to figure out where the best place to do this validation was. The tests were updated for consistency here as well.

After merging, it is advised that engineers update their local fixtures with policy language, since the data migration only works when _not_ fully resetting the database from scratch (i.e. for deployed environments). I will update the LastPass fixtures accordingly after merge as a point of reference.


benefits/core/migrations/0006_enrollmentflow_in_person_policy.py
```diff
@@ -0,0 +1,50 @@
...
+                    "possess a valid Medicare card."
+                case SystemName.OLDER_ADULT:
+                    flow.in_person_policy = "I confirmed this rider’s identity using a government-issued ID and verified they "
+                    "are age 65 or older."
+                case SystemName.REDUCED_FARE_MOBILITY_ID:
+                    flow.in_person_policy = "I confirmed this rider’s identity using a government-issued ID and verified they "
+                    "possess an SBMTD Reduced Fare Mobility ID."
+                case SystemName.GCTD_CARD:
+                    flow.in_person_policy = "I confirmed this rider’s identity using a government-issued ID and verified they "
+                    "possess a GCTD Reduced Fare ID."
```
Scotchester: If these multiline strings aren't enclosed in parentheses, the resulting field value is just the first line.

```suggestion
                case SystemName.COURTESY_CARD:
                    flow.in_person_policy = (
                        "I confirmed this rider’s identity using a government-issued ID and verified they "
                        "possess an MST Courtesy Card."
                    )
                case SystemName.MEDICARE:
                    flow.in_person_policy = (
                        "I confirmed this rider’s identity using a government-issued ID and verified they "
                        "possess a valid Medicare card."
                    )
                case SystemName.OLDER_ADULT:
                    flow.in_person_policy = (
                        "I confirmed this rider’s identity using a government-issued ID and verified they "
                        "are age 65 or older."
                    )
                case SystemName.REDUCED_FARE_MOBILITY_ID:
                    flow.in_person_policy = (
                        "I confirmed this rider’s identity using a government-issued ID and verified they "
                        "possess an SBMTD Reduced Fare Mobility ID."
                    )
                case SystemName.GCTD_CARD:
                    flow.in_person_policy = (
                        "I confirmed this rider’s identity using a government-issued ID and verified they "
                        "possess a GCTD Reduced Fare ID."
                    )
```

Scotchester (changes requested):
This is a great refactor! Everything worked perfectly once I addressed a little bug described below.

benefits/core/migrations/0006_enrollmentflow_in_person_policy.py
thekaveman: Huh, I saw this auto-correction (to wrap the text in `(...)`) on _one_ of the lines but not the others, and instead I just manually corrected it to be like how it was in the original form.

Thanks for double-checking :+1:

Scotchester: Interesting. I guess Python handles it differently when it's setting a parameter (as in the original form) vs. assigning a variable. But in your original commit, I was surprised it would just silently ignore a line with nothing but a string on it.

Scotchester (approved):
Ran updated migration and all is well 👍

benefits/core/models/enrollment.py
```diff
@@ -234,28 +229,6 @@ def eligibility_verifier(self):
...
-            errors.append(ValidationError(f"Template not found: {t}"))
-
-        if supports_in_person and self.system_name not in SUPPORTED_IN_PERSON_FLOWS:
-            errors.append(
-                ValidationError(f"{self.system_name} not configured for in-person enrollment. Please uncheck to continue.")
-            )
-
-        if errors:
-            raise ValidationError(errors)
-
```
jgravois, line 258: i dig this cleanup

benefits/in_person/forms.py
```diff
@@ -32,28 +32,6 @@ class InPersonEligibilityForm(forms.Form):
...
-        ),
-        models.SystemName.REDUCED_FARE_MOBILITY_ID.value: EligibilityPolicy(
-            details="I confirmed this rider’s identity using a government-issued ID and verified they possess an SBMTD "
-            "Reduced Fare Mobility ID."
-        ),
-        models.SystemName.GCTD_CARD.value: EligibilityPolicy(
-            details="I confirmed this rider’s identity using a government-issued ID and verified they possess a GCTD Reduced "
-            " Fare ID."
-        ),
-    }
```
jgravois, line 55: i dig this extraction too.

benefits/core/migrations/0006_enrollmentflow_in_person_policy.py
```diff
@@ -0,0 +1,60 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="enrollmentflow",
+            name="in_person_policy",
+            field=models.TextField(
+                blank=True,
+                default="",
+                help_text="The policy language used by transit agency staff to verify a user's eligibility in-person.",
```
jgravois, line 56: this is pure pedantry, but i think of this language as being an attestation of the user's eligibility.

```suggestion
                help_text="The policy language used by transit agency staff to attest a user's eligibility in-person.",
```

jgravois (commented):
i didn't test locally, but i love what i see here.

benefits/core/migrations/0006_enrollmentflow_in_person_policy.py
thekaveman: This change needs to be made in the model as well, or else this (generated) migration won't match and we'll get an error / notice that a migration needs to be generated.

I anticipate we'll come back to this as part of the UX changes envisioned in #3645, so I propose we address it then?

jgravois: sounds good 🤙

lalver1 (approved):
This looks great! I tested locally a few different scenarios and had no issues 👍

thekaveman:
Thanks y'all for the feedback / review. Are we OK to merge so it is included in the cutover tomorrow? Or would we rather wait? I'm good either way.

cc @lalver1 @jgravois

jgravois:
> Are we OK to merge so it is included in the cutover tomorrow? Or would we rather wait?

we're reseeding the db regardless so i say :shipit:

## clarification

when we cutover tomorrow, the code will still only live on `dev`, but i remain bullish about the release we're going to need to do to tear down the application services a few days later.

thekaveman:
Oh, right 😅 

I'll wait til after tomorrow just to keep the noise to a minimum

thekaveman:
Combined fixtures in LP were updated with policy language for the 2 configured `in_person` flows (Senior, Medicare).

lalver1:
We may have overlooked a small issue when reviewing this PR (related to consolidating the `clean()` logic), so I just wanted to get some confirmation if other devs are also seeing this behavior before possibly creating a ticket:

While working on #4003, I wanted to see how the new Metro Mobility Wallet flow would look like in the Admin but when I tried saving it, I got a `Template not found: eligibility/includes/selection-label--.html` error (even though I created a dummy template called `selection-label--metro_mobility_wallet.html`). So I guess that we may have introduced a bug in the Admin's form validation. As far as I can tell, [`self.instance.selection_label_template`](https://github.com/cal-itp/benefits/blob/main/benefits/core/admin/enrollment.py#L70) is not fully formed due to [`self.system_name`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/enrollment.py#L195) being empty. Something like the model instance hasn't been saved yet when it is called from `clean()`. You can also reproduce this if you try to create a new Older Adult or Medicare or any other existing flow, you'll still see an issue with the path to the template not being correct.

Scotchester:
Good catch, Luis! Yeah, since `selection_label_template` is a property, `self.instance.selection_label_template` can't be retrieved if the instance doesn't exist yet.

I've got a fix sussed out locally. I'll open a PR.

lalver1:
That's great @Scotchester! I started thinking about a simple fix but didn't get very far so thanks for the PR! 🙏

Issue #3686, opened 2026-04-28
Spike: architecture for LA Metro Mobility Wallet registration


thekaveman:
The work to support LA Metro's Mobility Wallet Phase 3 pilot is almost upon us. Let's take a moment to review the high-level objectives and constraints and think about how we might implement this set of features.

## Objectives and constraints

- [ ] Users do not interact with the primary Benefits rider discounts experience: Metro is not shown in the agency selector drop-down or homepage; a direct entry URL allows users to go straight into the Metro experience.
- [ ] ~An email address is collected from the user before they attempt eligibility verification, stored in Metro's [Submittable](https://www.submittable.com/) instance.~
- [ ] Eligibility verification is performed by the Identity Gateway via Login.gov and CalFresh, similar to the existing rider discounts self-service experience.
- [ ] The results of the eligibility check will contain the yes/no flag, and potentially some user info from Login.gov (first, last, email address, etc.)
- [ ] If eligible, users will have a chance to review/update their info.
- [ ] If eligible, the user is registered into Metro's [Submittable](https://www.submittable.com/) instance.

## Acceptance Criteria

- [ ] We have a plan for implementation that considers the above objectives and constraints.
- [ ] The plan considers how existing functionality from Benefits will be integrated (e.g. models for configuration, analytics instrumentation).
- [ ] The plan details a task breakdown that can serve as the basis for future Sprint planning.
- [ ] Any key open questions or blockers are highlighted.

## Additional context

- [Submittable API docs](https://submittable-api.submittable.com/docs/v4/index.html)
- [Deck with more info about the approach](https://docs.google.com/presentation/d/1-jKrrdYdfjIbvCie9xLmQWwPBMjwiDLVuecFX7zIoiM/edit?slide=id.g34c8e1ebe23_0_447#slide=id.g34c8e1ebe23_0_447)

thekaveman:
# Notes from 2026-08-04

## Assumptions

- (Soft) launch by December 1, application window open through March 2027 (roughly 3 months)
- We don’t need/want to integrate Mobility Wallets functionality with standard discount enrollments functionality
- Design for this flow will be completely different from normal Benefits flow
- Standalone (Django) app
- We are not implementing in-person for Metro Mobility Wallet using Benefits

## Open questions / discussion

### Data modeling and config

[...]

PR #3997, opened 2026-08-18
Feat: Metro Mobility Wallet Django app

### LINKED ISSUE #3945: Metro Mobility Wallet: Create and register a Django app
This ticket is to create the folder / app structure to hold logic, templates, etc. for the LA Metro Mobility Wallet implementation.

We aren't trying to be clever with defining _mobility wallet_ as a feature more generally here, as we don't know enough about if/when this type of feature will be used more broadly.

## Acceptance Criteria

- [ ] A top-level app folder under `benefits/` exists for the LA Metro Mobility Wallet work
- [ ] The folder contains an `__init__.py` to register it as a Python module
- [ ] The folder contains an `apps.py` to define the Django `AppConfig`
- [ ] The folder contains a `views.py` to define a single `IndexView` CBV rendering a template with placeholder content
- [ ] The folder contains an `urls.py` to define a `urlpatterns` list with a single `path` entry for the `IndexView`
- [ ] The new Django app is registered in [`settings.py`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/settings.py#L53)
- [ ] The new URLs are registered in the top-level [`urls.py`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/urls.py#L32)
- [ ] At least one test exists for the `IndexView` CBV in an appropriately named folder under `test/pytest/`

lalver1:
Closes #3945

This PR creates the folder / app structure to hold logic, templates, etc. for the LA Metro Mobility Wallet implementation.

### DIFF
--- a/benefits/metro_mobility_wallet/urls.py
+++ b/benefits/metro_mobility_wallet/urls.py
@@ -1,0 +1,13 @@
+"""
+The metro_mobility_wallet application: URLConf for the metro_mobility_wallet app.
+"""
+
+from django.urls import path
+
+from . import views
+
+app_name = "metro_mobility_wallet"
+urlpatterns = [
+    # /metro_mobility_wallet
+    path("", views.IndexView.as_view(), name="metro_mobility_wallet:index"),
+]
--- a/benefits/metro_mobility_wallet/urls.py
+++ b/benefits/metro_mobility_wallet/urls.py
@@ -9,5 +9,5 @@ from . import views
 app_name = "metro_mobility_wallet"
 urlpatterns = [
     # /metro_mobility_wallet
-    path("", views.IndexView.as_view(), name="metro_mobility_wallet:index"),
+    path("", views.IndexView.as_view(), name="index"),
 ]
--- a/benefits/metro_mobility_wallet/urls.py
+++ b/benefits/metro_mobility_wallet/urls.py
@@ -8,6 +8,6 @@ from . import views
 
 app_name = "metro_mobility_wallet"
 urlpatterns = [
-    # /metro-mobility-wallet
+    # /metro-mobility-wallet/
     path("", views.IndexView.as_view(), name="index"),
 ]
--- a/benefits/metro_mobility_wallet/urls.py
+++ b/benefits/metro_mobility_wallet/urls.py
@@ -8,6 +8,6 @@ from . import views
 
 app_name = "metro_mobility_wallet"
 urlpatterns = [
-    # /metro_mobility_wallet
+    # /metro-mobility-wallet
     path("", views.IndexView.as_view(), name="index"),
 ]
--- a/benefits/metro_mobility_wallet/apps.py
+++ b/benefits/metro_mobility_wallet/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class MetroMobilityWalletConfig(AppConfig):
+    name = "benefits.metro_mobility_wallet"
+    label = "metro_mobility_wallet"
--- a/benefits/metro_mobility_wallet/templates/metro_mobility_wallet/index.html
+++ b/benefits/metro_mobility_wallet/templates/metro_mobility_wallet/index.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="description"
+          content="The Mobility Wallet pilot program encourages transportation equity by promoting access to opportunities,
+          reducing gas emissions and improving economic and health outcomes for people in Los Angeles County.">
+    <title>Metro Mobility Wallet Pilot Program</title>
+  </head>
+  <body>
+    <main>
+      Metro Mobility Wallet Pilot Program
+    </main>
+  </body>
+</html>
--- a/benefits/metro_mobility_wallet/views.py
+++ b/benefits/metro_mobility_wallet/views.py
@@ -1,0 +1,7 @@
+from django.views.generic import TemplateView
+
+
+class IndexView(TemplateView):
+    """View for the Metro Mobility Wallet landing page."""
+
+    template_name = "metro_mobility_wallet/index.html"
--- a/benefits/settings.py
+++ b/benefits/settings.py
@@ -68,6 +68,7 @@ INSTALLED_APPS = [
     "benefits.eligibility",
     "benefits.oauth",
     "benefits.in_person",
+    "benefits.metro_mobility_wallet",
 ]
 
 GOOGLE_SSO_CLIENT_ID = os.environ.get("GOOGLE_SSO_CLIENT_ID", "secret")
--- a/benefits/urls.py
+++ b/benefits/urls.py
@@ -38,6 +38,7 @@ urlpatterns = [
     path("in_person/", include("benefits.in_person.urls")),
     path("littlepay/", include("benefits.enrollment_littlepay.urls")),
     path("switchio/", include("benefits.enrollment_switchio.urls")),
+    path("metro-mobility-wallet/", include("benefits.metro_mobility_wallet.urls")),
 ]
 
 if settings.RUNTIME_ENVIRONMENT() == settings.RUNTIME_ENVS.LOCAL:
--- a/tests/pytest/metro_mobility_wallet/test_views.py
+++ b/tests/pytest/metro_mobility_wallet/test_views.py
@@ -1,0 +1,14 @@
+import pytest
+
+import benefits.metro_mobility_wallet.views as views
+
+
+class TestIndexView:
+    @pytest.fixture
+    def view(self, app_request):
+        v = views.IndexView()
+        v.setup(app_request)
+        return v
+
+    def test_template_name(self, view):
+        assert view.template_name == "metro_mobility_wallet/index.html"


benefits/metro_mobility_wallet/urls.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+
+from django.urls import path
+
+from . import views
+
+app_name = "metro_mobility_wallet"
+urlpatterns = [
+    # /metro_mobility_wallet
+    path("", views.IndexView.as_view(), name="metro_mobility_wallet:index"),
```
lalver1: Since the app name is kind of long, I could add it to [`routes.py`](https://github.com/cal-itp/benefits/blob/main/benefits/routes.py), something like

```python
@property
    def METRO_INDEX(self):
        """The landing page for the Metro Mobility Wallet program."""
        return "metro_mobility_wallet:index"
```

We'll probably end up with 4 or 5 similar properties but they'll all be next to each other so it would be easy to clean up this file if we need to remove these `METRO_` properties.

thekaveman: Generally I think we want to avoid modifying top-level code where we don't have to. This would seem like [another example](https://github.com/cal-itp/benefits/issues/3686#issuecomment-5321514111) of there not being a lot of value in mingling Mobility Wallet into rider discounts (i.e. how much routing/reference from rider discounts to Mobility Wallet do we _actually_ anticipate needing to do? Why would we need Metro routes here?) 

We should make every attempt to keep it all self-contained in this app. E.g. a `routes.py` in this app could be helpful if we want to avoid typing it out in various places (I'm not sure if we'll need it? Probably).

jgravois: i'm seeing this locally:
```
?: (urls.W003) Your URL pattern '' [name='metro_mobility_wallet:index'] has a name including a ':'. Remove the colon, to avoid ambiguous namespace references.
```
i see the same pattern elsewhere in benefits but i'm not sure if there's a substantive difference or we are just suppressing the warning somehow.

since `app_name` is defined explicitly immediately above, i assume it'd be safe to 👇, but i'd call that _more_ ambiguous, not less.


```suggestion
    path("", views.IndexView.as_view(), name="index"),
```

benefits/metro_mobility_wallet/urls.py
```diff
@@ -0,0 +1,13 @@
...
+The metro_mobility_wallet application: URLConf for the metro_mobility_wallet app.
+"""
+
+from django.urls import path
+
+from . import views
+
+app_name = "metro_mobility_wallet"
+urlpatterns = [
+    # /metro_mobility_wallet
```
jgravois: underscores aren't genuinely supposed to resolve, are they?

```suggestion
    # /metro-mobility-wallet
```

and personally i'd advocate for appending trailing slashes in this new app in the name of maximum resilience.

related: https://github.com/cal-itp/benefits/pull/3983#issuecomment-5284990301

benefits/metro_mobility_wallet/urls.py
lalver1: Ah, thanks for catching this @jgravois! I had forgotten/missed that in the `urls.py` files we use the [`name` static method](https://github.com/cal-itp/benefits/blob/main/benefits/routes.py#L197), which is why the warning is not present for other URLs 🤦‍♂️

jgravois: ah, of course!

benefits/metro_mobility_wallet/urls.py
lalver1: Ah, that's definitely a typo on the comment 😅, thanks for catching that!
Looking over https://github.com/cal-itp/benefits/pull/3983#issuecomment-5284990301, I think we can append trailing slashes to the URLs for this app. But we'll prob start implementing that once we start building out the rest of the URLs. For the root URL, `""` is Django's recommendation (I couldn't find an explicit mention in the docs, but if I add a slash, I get this warning `?: (urls.W002) Your URL pattern '/' [name='index'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'.`).
Fixed in 1840baec6f7654bf540d226ad1935916abad6669

benefits/metro_mobility_wallet/urls.py
lalver1: > Generally I think we want to avoid modifying top-level code where we don't have to. This would seem like [another example](https://github.com/cal-itp/benefits/issues/3686#issuecomment-5321514111) of there not being a lot of value in mingling Mobility Wallet into rider discounts (i.e. how much routing/reference from rider discounts to Mobility Wallet do we _actually_ anticipate needing to do? Why would we need Metro routes here?)
> 
> We should make every attempt to keep it all self-contained in this app. E.g. a `routes.py` in this app could be helpful if we want to avoid typing it out in various places (I'm not sure if we'll need it? Probably).

I agree, I'll leave it as is since we don't have a need for it yet. If we do end up needing it, then we can add it but for now it feels premature to do so (and especially because the name should not have `metro_mobility_wallet:`, so there is no issue anymore with the length of the name).

thekaveman:
@lalver1 just a heads up, I manually cancelled this run https://github.com/cal-itp/benefits/actions/runs/32160913979/job/95789316927 that was started by commit [`6c34352`](https://github.com/cal-itp/benefits/pull/3997/commits/6c34352e6c0eeefd9a7f3b11b50d69a105977406) from this PR. 

It was hung for 5+ hours trying to setup the runner. Still some ghosts in the machine from recent GitHub outages, it seems :shrug: :ghost:

lalver1:
Thanks for that @thekaveman, I was in and out this afternoon and missed checking on the status of the run so I appreciate you taking a look at it!

benefits/metro_mobility_wallet/urls.py
jgravois: i was expecting we'd do this to enforce the trailing slash.

```python
path("", views.IndexView.as_view(), name="index/"),
```

because even though its the root for this Django app, its still served up from within a subdirectory. does that make sense?

jgravois: nm, rereading your reply i think we're already on the same page. makes perfect sense to me to wait until we build out additional urls.

thekaveman: Also, John, this `name` parameter here has nothing to do with how the URL shows up. That is the first parameter of the [`path()`](https://docs.djangoproject.com/en/5.2/ref/urls/#django.urls.path) function being called.

`name` is for reference elsewhere, like in calls to [`reverse()`](https://docs.djangoproject.com/en/5.2/ref/urlresolvers/#reverse) or in templates. 

More on URL naming: https://docs.djangoproject.com/en/5.2/topics/http/urls/#naming-url-patterns

jgravois: ah yes, derp!

benefits/metro_mobility_wallet/urls.py
```diff
@@ -0,0 +1,13 @@
...
+The metro_mobility_wallet application: URLConf for the metro_mobility_wallet app.
+"""
+
+from django.urls import path
+
+from . import views
+
+app_name = "metro_mobility_wallet"
+urlpatterns = [
+    # /metro-mobility-wallet
```
jgravois: i confirmed this is forcefully appended even when `path=""` 😂 

```suggestion
    # /metro-mobility-wallet/
```

jgravois (dismissed):
i approve whether you humor my tedious inline nit or not!

jgravois:
snooping the local Django logs i noticed the DEBUG statement below:
```
[18/Aug/2026 17:29:28] DEBUG benefits.core.models.transit:303 Get all active TransitAgency
[18/Aug/2026 17:29:28] INFO django.server:213 "GET /metro-mobility-wallet/ HTTP/1.1" 200 15870
```
not sure how we'd more cleanly separate concerns and avoid irrelevant db queries. perhaps we will just need to define a custom set of [`context_processors`](https://github.com/cal-itp/benefits/blob/57bf050d4f336b48c42212addd29bf5e5b050b8b/benefits/settings.py#L166) for the routes within this app specifically? i'll do a little bit more research into this tomorrow.

thekaveman:
> perhaps we will just need to define a custom set of [context_processors](https://github.com/cal-itp/benefits/blob/57bf050d4f336b48c42212addd29bf5e5b050b8b/benefits/settings.py#L166) for the routes within this app specifically?

Context Processors run for every request (regardless of which "app" is servicing the request), so the behavior you are seeing is expected from that point of view anyway.

Separately we could look at our existing processors and decide if we want to remove/clean up some of them to cut down on DB queries. But this was partially addressed with the N+1 query refinements / request caching, so I don't think it is necessary at this time.

Maybe creating a ticket to come back to it is worthwhile.

benefits/metro_mobility_wallet/urls.py
lalver1: No problem! Updated in b0e5b51b5cb0d582830a48b37e2be2a6490b0335

Sorry, this dismissed your review @jgravois. Appreciate a re-review 🙏

jgravois (approved):
🌈🌈

PR #4003, opened 2026-08-20
Feat: add Metro Mobility Wallet to `SystemName` enum

### LINKED ISSUE #3948: Metro Mobility Wallet: Add to `SystemName` enum
We've decided that the best way to track and consume configuration data for Metro Mobility Wallet is via an [`EnrollmentFlow`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/core/models/enrollment.py#L120) instance that links to the configured [`ClaimsVerificationRequest`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/4e263f017cc28ea1b5c64f2968b4b73f249ae22f/cdt_identity/models.py#L6) and [`IdentityGatewayConfig`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/4e263f017cc28ea1b5c64f2968b4b73f249ae22f/cdt_identity/models.py#L53) instances.

The first step is to update the [`SystemName`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/core/models/enrollment.py#L26) enumeration to add a new entry for Metro Mobility Wallet. This will give us a point of reference for the other logic that needs to use the corresponding `EnrollmentFlow` instance.

## Acceptance Criteria

- [ ] A new entry exists in the `SystemName` enum
- [ ] A migration is generated for this entry

## Additional context

In #3946 we'll make the first use of this new `SystemName` by having a helper to get the `EnrollmentFlow` instance.

lalver1:
Closes #3948 

This PR adds _Metro Mobility Wallet_ to the `SystemName` enum that will be used by [`EnrollmentFlow`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/core/models/enrollment.py#L120).

## Reviewing
Make sure that the migration successfully runs and that in the Admin you can see the new _Metro Mobility Wallet_ in the _Add enrollment flow_ > _System name:_ drop-down.


benefits/core/models/enrollment.py
```diff
@@ -27,6 +27,7 @@ class SystemName(models.TextChoices):
     COURTESY_CARD = "courtesy_card"
     GCTD_CARD = "gctd_card"
     MEDICARE = "medicare"
+    METRO_WALLET = "metro_mobility_wallet"
```
lalver1: Please feel free to suggest changes to the naming of `METRO_WALLET = "metro_mobility_wallet"` if this doesn't feel right.

angela-tran: Since it shows up in the UI, my vote is that we go ahead and use the full name, but I won't push further if others disagree

## Before
image

## After

```suggestion
    METRO_MOBILITY_WALLET = "metro_mobility_wallet"
```

image

angela-tran:
As an interesting note for our collective knowledge, apparently the migration generated when a `Choice` enum is updated is solely for documentation purposes. This is according to an [independently-run](https://django.wiki/about/) Django resource: https://django.wiki/tutorials/choices-and-enums/#migrations-and-choices

This answers why Django generates a migration even though `Choice` values aren't stored anywhere in the database. (a question I've had for a while now)

benefits/core/models/enrollment.py
angela-tran: (I guess the migration will need to be updated too, so please note that my code suggestion above is not the complete set of changes)

lalver1: Thanks @angela-tran! I went ahead and made the change (76d58815223169e24ae3125b4a105796cd565f45) 👍

jgravois (approved):
🍔

edit: given the amount of Metro work coming down the pike, might be cool to update the "[Combined] Benefits fixtures" in LastPass to add the new stubbed enrollment flow.

lalver1:
Good idea @jgravois 👍 I added the new _Metro Mobility Wallet_ flow to the "[COMBINED] Benefits fixtures" LP note so it will now show up in the Benefits Admin, but I left most of its fields `null`. This is ok due to how the model is set up, as long as we don't try to save the model from the Admin, these `null` values won't be problematic. Later we can start filling it out with actual values.

Issue #3949, opened 2026-08-04
Metro Mobility Wallet: define analytics events classes


thekaveman:
This ticket is for the definition of the events only. Don't worry about where/how the events are sent as part of this work.

## Acceptance Criteria

- [ ] Inside the folder from #3945, an `analytics.py` file exists
- [ ] New event classes exist for each new event identified in the plan: #3959 
- [ ] Helper functions exist to send each of the events for a `request`
- [ ] Test(s) exist for the above

## Additional context

#3768 is the same work for the more general OAuth `"failure to proof"` event that we are adding as part of this work.

Issue #3946, opened 2026-08-04
Metro Mobility Wallet: Create a Session class


thekaveman:
We want to define a session class for use within the LA Metro Mobility Wallet Django app, and as an exploration of what this kind of pattern could look like for Benefits writ-large.

This ticket is focused on getting the basic definition in place, knowing that we don't have all the information yet about what exactly we need to track in session for this effort.

## Acceptance Criteria

- [ ] Inside the folder from #3945, a `session.py` file exists with an appropriately named class (like `Session`)
- [ ] The session class has at least one helper/attribute to return the (hard-coded) `EnrollmentFlow` instance for Metro Mobility Wallet
- [ ] At least one test exists for the above

## Additional context

See the pattern as established in CDRC:

- https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/blob/main/web/core/session.py
- https://github.com/Office-of-Digital-Services/cdt-ods-disaster-recovery/blob/main/web/vital_records/session.py

PR #4004, opened 2026-08-20
Add MMW app-specific OAuth hooks and URL patterns

### LINKED ISSUE #3974: Metro Mobility Wallet: Setup app-specific OAuth config/hooks/URLs
From https://github.com/cal-itp/benefits/issues/3686#issuecomment-5185581751:

> - Currently, hooks are defined in [`oauth.hooks.OAuthHooks`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/hooks.py#L14) and registered in [`oauth.urls`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/urls.py#L13); these hooks are used by all flows that go through the IdG in Benefits
> - We may want/need different hooks for Mobility Wallets: the current [`claims_verified_eligible`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/hooks.py#L49) and [`claims_verified_not_eligible`](https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/hooks.py#L69) hooks rely on having an agency configured in the session and redirect into the standard discount eligibility/enrollment apps
> - A separate registration should be possible if we organize the app in #3945 in such a way as to allow for distinct URL and hook registrations. But this has implications for how @johnatstate sets up the client configuration for Benefits Metro Mobility Wallet (i.e. the callback URLs)

To build the Mobility Wallet feature as cleanly as possible, and since we need a separate OAuth client registration with IdG anyway, it makes sense to define a Metro Mobility Wallet-specific `OAuthHooks` class and corresponding URL registration(s).

## Acceptance Criteria

- [ ] Inside the folder from #3945, an `oauth.py` file exists with an implementation of [`cdt_identity.hooks.DefaultHooks`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/4e263f017cc28ea1b5c64f2968b4b73f249ae22f/cdt_identity/hooks.py#L44)
- [ ] The implementation should reuse and call existing [`benefits.oauth.analytics`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/analytics.py) events as appropriate, including the new `"failure to proof"` event from #3768, and anything relevant from #3949 
- [ ] The implementation should have placeholders to redirect into the appropriate Metro Mobility Wallet views
- [ ] The URL configs from #3945 should include a subset of `/<app>/oauth` registrations that utilize the above hooks implementation, and this set of URLs should be used to configure a new client registration with IdG

## Additional context

The hooks implementation can be largely based on the standard Benefits hooks implementation with key changes for view middleware and redirects:

https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/hooks.py#L14

The URL configs can be largely based on the standard Benefits OAuth URLs, with the appropriate hooks applied:

https://github.com/cal-itp/benefits/blob/e9bf8fcd96d6583a830a2c44b708c4f55acecb47/benefits/oauth/urls.py#L18

Scotchester:
Closes #3974 (eventually)

---

~~This draft is a little premature (sorta blocked by https://github.com/cal-itp/benefits/issues/3949 which will implement new MMW-specific analytics events), but I wanted to keep going in the OAuth area and see how well I'm understanding it.~~

~~I'll reach out to the team for a pairing session to see if I'm on the right track.~~

---

- Adds `metro_mobility_wallet.oauth`, providing `OAuthHooks` with an initial set of the ones we are fairly sure we'll want
  - Includes our first use of the `failure_to_proof` hook
  - Has redirects to theoretical MMW views where expected
  - Analytics calls are currently commented out pending those prerequisites
- Adds OAuth URLs to `metro_mobility_wallet.urls`
- Upgrades `django-cdt-identity` to support the new above

### DIFF
--- a/benefits/metro_mobility_wallet/oauth.py
+++ b/benefits/metro_mobility_wallet/oauth.py
@@ -10,7 +10,6 @@ from benefits.oauth import analytics
 # from django.utils.decorators import decorator_from_middleware, method_decorator
 
 
-
 class OAuthHooks(DefaultHooks):
     @classmethod
     def pre_login(cls, request):
--- a/benefits/metro_mobility_wallet/oauth.py
+++ b/benefits/metro_mobility_wallet/oauth.py
@@ -1,21 +1,21 @@
 import sentry_sdk
 from cdt_identity.hooks import DefaultHooks
 from django.shortcuts import redirect
-from django.utils.decorators import decorator_from_middleware, method_decorator
 
-from benefits.core import session
-from benefits.core.middleware import FlowSessionRequired
+# from benefits.core import session
+# from benefits.core.middleware import FlowSessionRequired
 from benefits.eligibility.views import analytics as eligibility_analytics
 from benefits.oauth import analytics
-from benefits.oauth.hooks import OAuthHooks as core_hooks
+
+# from django.utils.decorators import decorator_from_middleware, method_decorator
+
 
 
 class OAuthHooks(DefaultHooks):
-    # @classmethod
-    # def pre_login(cls, request):
-    #     super().pre_login(request)
-    #     analytics.started_sign_in(request)
-    pre_login = core_hooks.pre_login
+    @classmethod
+    def pre_login(cls, request):
+        super().pre_login(request)
+        analytics.started_sign_in(request)
 
     @classmethod
     def cancel_login(cls, request):
@@ -23,30 +23,28 @@ class OAuthHooks(DefaultHooks):
         analytics.canceled_sign_in(request)
         return redirect("metro_mobility_wallet:eligbility_unverified")
 
-    # @classmethod
-    # def pre_logout(cls, request):
-    #     super().pre_logout(request)
-    #     analytics.started_sign_out(request)
+    @classmethod
+    def pre_logout(cls, request):
+        super().pre_logout(request)
+        analytics.started_sign_out(request)
 
-    #     # the user is signed out of the app
-    #     session.logout(request)
-    pre_logout = core_hooks.pre_logout
-
-    # @classmethod
-    # def post_logout(cls, request):
-    #     super().post_logout(request)
-    #     analytics.finished_sign_out(request)
-
-    #     origin = session.origin(request)
-    #     return redirect(origin)
-    post_logout = core_hooks.post_logout
+        # the user is signed out of the app
+        session.logout(request)
 
     @classmethod
-    @method_decorator(
-        [
-            decorator_from_middleware(FlowSessionRequired),
-        ]
-    )
+    def post_logout(cls, request):
+        super().post_logout(request)
+        analytics.finished_sign_out(request)
+
+        origin = session.origin(request)
+        return redirect(origin)
+
+    @classmethod
+    # @method_decorator(
+    #     [
+    #         decorator_from_middleware(FlowSessionRequired),
+    #     ]
+    # )
     def failure_to_proof(cls, request):
         super().failure_to_proof(request)
         session.update(request, logged_in=True)  # QUESTION: Are they still considered "logged in" if they failed to prove?
@@ -60,12 +58,12 @@ class OAuthHooks(DefaultHooks):
         return redirect("metro_mobility_wallet:failure_to_proof")
 
     @classmethod
-    @method_decorator(
-        [
-            # decorator_from_middleware(AgencySessionRequired),
-            decorator_from_middleware(FlowSessionRequired),
-        ]
-    )
+    # @method_decorator(
+    #     [
+    #         # decorator_from_middleware(AgencySessionRequired),
+    #         decorator_from_middleware(FlowSessionRequired),
+    #     ]
+    # )
     def claims_verified_eligible(cls, request, claims_request, claims_result):
         super().claims_verified_eligible(request, claims_request, claims_result)
         session.update(request, logged_in=True)
@@ -80,12 +78,12 @@ class OAuthHooks(DefaultHooks):
         return redirect("metro_mobility_wallet:enrollment_index")
 
     @classmethod
-    @method_decorator(
-        [
-            # decorator_from_middleware(AgencySessionRequired),
-            decorator_from_middleware(FlowSessionRequired),
-        ]
-    )
+    # @method_decorator(
+    #     [
+    #         # decorator_from_middleware(AgencySessionRequired),
+    #         decorator_from_middleware(FlowSessionRequired),
+    #     ]
+    # )
     def claims_verified_not_eligible(cls, request, claims_request, claims_result):
         super().claims_verified_not_eligible(request, claims_request, claims_result)
         session.update(request, logged_in=True)
--- a/benefits/metro_mobility_wallet/oauth.py
+++ b/benefits/metro_mobility_wallet/oauth.py
@@ -1,0 +1,104 @@
+import sentry_sdk
+from cdt_identity.hooks import DefaultHooks
+from django.shortcuts import redirect
+from django.utils.decorators import decorator_from_middleware, method_decorator
+
+from benefits.core import session
+from benefits.core.middleware import FlowSessionRequired
+from benefits.eligibility.views import analytics as eligibility_analytics
+from benefits.oauth import analytics
+from benefits.oauth.hooks import OAuthHooks as core_hooks
+
+
+class OAuthHooks(DefaultHooks):
+    # @classmethod
+    # def pre_login(cls, request):
+    #     super().pre_login(request)
+    #     analytics.started_sign_in(request)
+    pre_login = core_hooks.pre_login
+
+    @classmethod
+    def cancel_login(cls, request):
+        super().cancel_login(request)
+        analytics.canceled_sign_in(request)
+        return redirect("metro_mobility_wallet:eligbility_unverified")
+
+    # @classmethod
+    # def pre_logout(cls, request):
+    #     super().pre_logout(request)
+    #     analytics.started_sign_out(request)
+
+    #     # the user is signed out of the app
+    #     session.logout(request)
+    pre_logout = core_hooks.pre_logout
+
+    # @classmethod
+    # def post_logout(cls, request):
+    #     super().post_logout(request)
+    #     analytics.finished_sign_out(request)
+
+    #     origin = session.origin(request)
+    #     return redirect(origin)
+    post_logout = core_hooks.post_logout
+
+    @classmethod
+    @method_decorator(
+        [
+            decorator_from_middleware(FlowSessionRequired),
+        ]
+    )
+    def failure_to_proof(cls, request):
+        super().failure_to_proof(request)
+        session.update(request, logged_in=True)  # QUESTION: Are they still considered "logged in" if they failed to prove?
+        analytics.failure_to_proof(request)
+
+        # TODO: Confirm whether this event is desired.
+        # Did they in fact "start eligibility", or does that only refer to the post-authentication piece?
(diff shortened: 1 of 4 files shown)


Scotchester:
I had a good session with Angela wherein she helped me realize a couple issues with my initial commit:

1. The code I borrowed from `oauth.hooks` and modified in `metro_mobility_wallet.oauth` was still relying on `core.session`, when we intend to create a new session class for MMW (#3946). I pushed a commit that makes that no longer the case, but the new session doesn't exist yet. #3946 is currently a stretch item for this sprint, but I'm going to mark as blocking this issue and change status to To Do. It can be started as soon as Luis's `SystemName` PR is done.
   - My update includes taking out the `FlowSessionRequired` middleware mixin decorations, since those also rely on `core.session` and we expect that the flow will be permanently stored on all MMW sessions. 
2. I needed to update the OAuth error URL to point to a MMW-specific `SystemErrorView`, which didn't exist yet, so I put a placeholder in `metro_mobility_wallet.views`.

Scotchester:
Paired with Kegan today to understand more about the class-based session implementation in DDRC. We decided to punt on actually using the `metro_mobility_wallet.session.Session` class at this time, wanting to think a little more about our preferred approach for building that out.

We also decided to comment out the analytics calls until we have more clarity on what is desired there.

What we're left with is a starter file for MMW's OAuth hooks that we'll iterate on as we finalize the application flow, and the OAuth URL additions.

I wanted to have tests for the new hooks, so I copied and modified the tests from `core.oauth.hooks`. That led me to setting up the additional URLs and view stubs that they would need. If this feels like a bridge too far, I can pull that out.

All that said, I think this is now ready for review!

benefits/metro_mobility_wallet/oauth.py
```diff
@@ -0,0 +1,84 @@
...
+        # analytics.finished_sign_out(request)
+
+        origin = session.origin(request)
+        return redirect(origin)
+
+    @classmethod
+    def failure_to_proof(cls, request):
+        super().failure_to_proof(request)
+        session.update(request, logged_in=True)
+        # QUESTION: Are they still considered "logged in" if they failed to prove?
```
jgravois: as per @johnatstate's explanation yesterday, i would say 'yes'.

benefits/metro_mobility_wallet/oauth.py
```diff
@@ -0,0 +1,84 @@
...
+        # Kegan thinks the only purpose of this may be to show the "log out of Login.gov" link
+
+        # analytics.failure_to_proof(request)
+
+        # Did they in fact "start eligibility", or does that only refer to the post-authentication piece?
```
jgravois: my guess is that it makes sense for the behavior of `claims_verified_not_eligible` and `failure_to_proof` to be the same in this regard, but that's just drive-by commentary.

benefits/metro_mobility_wallet/urls.py
```diff
@@ -2,12 +2,49 @@
...
+    path(f"oauth/{OAuthRoutes.cancel}", decorator(cdt_identity_views.cancel), kwargs, name=OAuthRoutes.cancel),
+    path(
+        f"oauth/{OAuthRoutes.failure_to_proof}",
+        decorator(cdt_identity_views.failure_to_proof),
+        kwargs,
+        name=OAuthRoutes.failure_to_proof,
+    ),
+    path(f"oauth/{OAuthRoutes.logout}", decorator(cdt_identity_views.logout), kwargs, name=OAuthRoutes.logout),
+    path(f"oauth/{OAuthRoutes.post_logout}", decorator(cdt_identity_views.post_logout), kwargs, name=OAuthRoutes.post_logout),
+    path("oauth/error", views.SystemErrorView.as_view(), name=routes.name(routes.OAUTH_SYSTEM_ERROR)),
```
jgravois: it would be nice to define the `oauth` prefix once like we do 👇 but i'd definitely file that under 'nice to have':

https://github.com/cal-itp/benefits/blob/1fffd359e2b8e579b8db2d46df05eee7a4e4078a/benefits/oauth/urls.py#L15-L17

jgravois (dismissed):
i still don't consider myself a Django SME, but i'd consider this mergeable as-is.

locally i pulled down this branch, rebuilt the dev container and used `pip freeze` to be absoutely sure i had the latest and greatest `django-cdt-identity` dep laid down.

afterward i went through a normal benefits e2e flow and short-circuited a second attempt by clicking <kbd>Cancel</kbd> in the middle of the eligibility check and the behavior in each case was 'as expected'.

benefits/metro_mobility_wallet/urls.py
```diff
@@ -2,12 +2,49 @@
...
+    # /metro-mobility-wallet/enrollment/...
+    path("enrollment", views.IndexView.as_view(), name=routes.name("enrollment_index")),
+    # path(
+    #     "enrollment/error/reenrollment",
+    #     views.ReenrollmentErrorView.as_view(),
+    #     name=routes.name("enrollment_reenrollment_error"),
+    # ),
+    # path("enrollment/retry", views.RetryView.as_view(), name=routes.name("enrollment_retry")),
+    # path("enrollment/success", views.SuccessView.as_view(), name=routes.name("enrollment_success")),
+    # path("enrollment/error", views.SystemErrorView.as_view(), name=routes.name(routes.ENROLLMENT_SYSTEM_ERROR)),
```
thekaveman: IMHO all the URL registrations above this line (and related view creation etc.) are premature and should not be in scope for this PR. We are far from knowing exactly the sequence of views/routes the user will flow through, the names we want to give them, etc.

It may be the case that _something_ is required in order to do things like (in the `Hooks` class):

```python
return redirect("metro_mobility_wallet:eligbility_unverified")
```

Since we have to come back and deal with analytics and many other things anyway, in the interim I'd rather see us do something more like:

```python
return redirect("metro_mobility_wallet:index")
```

benefits/metro_mobility_wallet/urls.py
thekaveman: Yeah this is an interesting point... we definitely have to be under the `/metro-mobility-wallet` namespace for all of this, but is there a way to structure this such that we can better reuse the `cdt_identity` registrations?

Bigger question, hinted at from this comment:

> so that the URL namespace matches what cdt_identity expects. 
> (e.g. when cdt_identity reverses routes)

What happens when these namespaces don't match up? Is this even going to work at all?

benefits/metro_mobility_wallet/oauth.py
thekaveman: We should get clarification with @indexing the sequence of events that we want to log here (and in standard Benefits as well).

My initial guess would be:

1. `started sign in`
2. `finished sign in`
3. `failure to proof`

E.g. no start/end eligibility for this case.

thekaveman (changes requested):

benefits/metro_mobility_wallet/urls.py
lalver1: I was thinking the same thing but hadn't come up with a solution for what to do in the interim, Kegan's suggestion looks good to me. I feel it'll be easier to come back and make updates once we know more about the user flow if we have fewer "placeholder" URLs and views.

There are also a few `eligbility` instead of `eligibility` typos, but there's probably no need to fix these if we do change the approach  here.

lalver1: I also thought about mentioning that since we don't know exactly the user flow, I think it would be ok to leave out tests that require too many assumptions (views, URLS) to make them pass in this PR. It feels like we are at a stage where there's not enough code for making tests pass, but we don't want to commit failing ones, so we can instead leave them out and it'll be clear what is missing later by looking at the test coverage.

benefits/metro_mobility_wallet/urls.py
Scotchester: Ahhh, that's a great question! I didn't realize `cdt_identity` was doing some reversing in that way.

I took a closer look and no, I don't think it will work. It seems like we will need to make those `reverse` calls dynamic in some way. Probably make the argument _to_ them dynamic, rather – able to handle a nested namespace situation.

I've started hacking in ways to try and test things, which confirmed that with cdt_identity's code as-is, it will reverse back to the existing non-MMW `/oauth/authorize` path when you successfully log in.

I'll keep investigating how to properly do nested URL namespaces here.

Scotchester: Hokay, I pushed a change that addresses this, when paired with Office-of-Digital-Services/django-cdt-identity#72.

I didn't include all of the hackery I did to test it out, but much of it was putting the `flow` session variable in place to satisfy the `FlowUsesClaimsVerificationSessionRequired` middleware that wraps all of the OAuth URLs. I assume we'll want to leave this in place, since we're planning to use a standard `EnrollmentFlow` model for MMW.

Scotchester:
Now updated to use the new release of django-cdt-identity that supports nested namespaces. If anyone wants to test it out for themselves, here's what I did. (assumes you have loaded the current combo fixtures and have Django Debug Toolbar enabled)

1. Rebuild container and confirm that django-cdt-identity `2026.09.2` is installed.
2. Add the following to `metro_mobility_wallet.urls`:
   ```py
   eligibility_patterns = [
       path("start", views.EligibilityStartView.as_view(), name=routes.name(routes.ELIGIBILITY_START)),
   ]

   ...

   url_patterns += [
       path("eligibility/", include((eligibility_patterns, "eligibility"), namespace="eligibility")),
   ]
   ```
3. Add the following to `metro_mobility_wallet.views`:
   ```py
   from django.urls import reverse

   from benefits.core.models.enrollment import EnrollmentFlow


   class EligibilityStartView(TemplateView):
       """CBV for the eligibility verification getting started screen."""

       template_name = "metro_mobility_wallet/eligibility/start.html"

       def get(self, request, *args, **kwargs):
           session.update(
               request,
               flow=EnrollmentFlow.objects.get(id=6),
               eligible=False,
               origin=reverse(f"metro_mobility_wallet:{routes.ELIGIBILITY_START}"),
           )
           return super().get(request, *args, **kwargs)
   ```
4. Copy the `eligibility` app's `start.html` template into `metro_mobility_wallet` and:
   1. Hardcode the CTA button route to `/metro-mobility-wallet/oauth/login`
   2. Hardcode something in the button text like "Go to Login.gov"
5. Start up your debug session
6. Go to the admin and edit the MMW enrollment flow to:
   1. Turn off in-person enrollment (just to avoid creating a policy template for now)
   2. Use the Login.gov oauth_config
   3. Use any claims_request
7. Go to `/` to ensure the session is cleared
8. Go to `/metro-mobility-wallet/eligibility/start`
9. Open Django Debug Toolbar, inspect the request, and confirm that the session has a `flow` value of `6`.
10. Check the DDT box to intercept redirects
11. Click the button to go to Login.gov
12. On the intercept page:
    1. You should be on URL `/metro-mobility-wallet/oauth/login`
    2. You should see that the redirect param in the URL you're being redirected to is showing `metro-mobility-wallet%2Foauth%2Fauthorize`
    3. Inspecting the request should show:
       1. It's using view function `cdt_identity.views.login`
       2. with kwargs `{'hooks': "<class 'benefits.metro_mobility_wallet.oauth.OAuthHooks'>"}`
       3. and the URL name is `metro_mobility_wallet:cdt:login`

PR #4007, opened 2026-08-20
Fix validation of new `EnrollmentFlow` objects

### LINKED ISSUE #4006: Validation error on any new `EnrollmentFlow` when checking its selection label template path
While working on #4003, I wanted to see how the new Metro Mobility Wallet flow would look like in the Admin but when I tried saving it, I got a `Template not found: eligibility/includes/selection-label--.html` error (even though I created a dummy template called `selection-label--metro_mobility_wallet.html`). So I guess that we may have introduced a bug in the Admin's form validation. As far as I can tell, [`self.instance.selection_label_template`](https://github.com/cal-itp/benefits/blob/main/benefits/core/admin/enrollment.py#L70) is not fully formed due to [`self.system_name`](https://github.com/cal-itp/benefits/blob/main/benefits/core/models/enrollment.py#L195) being empty. Something like the model instance hasn't been saved yet when it is called from `clean()`. You can also reproduce this if you try to create a new Older Adult or Medicare or any other existing flow, you'll still see an issue with the path to the template not being correct.

_Originally posted by @lalver1 in https://github.com/cal-itp/benefits/issues/3995#issuecomment-5357834073_

---

I reproduced this issue locally, as well.

Scotchester:
Fixes #4006

---

In moving the validation to the form, if we are creating a new `EnrollmentFlow` we now don't have the instance created before the template path is checked, so `not template_path(self.instance.selection_label_template)` was always true, failing the validation.

This adds a static method to the class that we can use to get the expected template path and check it against the selected system name.

### DIFF
--- a/benefits/core/admin/enrollment.py
+++ b/benefits/core/admin/enrollment.py
@@ -5,6 +5,7 @@ from django.core.exceptions import ValidationError
 
 from benefits.core import models
 from benefits.core.models.common import template_path
+from benefits.core.models.enrollment import EnrollmentFlow
 
 from .mixins import ProdReadOnlyPermissionMixin, StaffPermissionMixin, SuperuserPermissionMixin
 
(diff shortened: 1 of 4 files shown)


thekaveman (dismissed):
Thanks for flagging this @lalver1 and for the quick fix @Scotchester.

Can we get a few tests though?

- Something that fails with the existing behavior for `clean()`, but passes with this fix?
- A basic test for the static helper

jgravois (commented):
verified the fix locally 👍

Scotchester:
Tests pushed! The one for the admin does indeed fail on main and pass on this branch.

jgravois (approved):
🎸

i'll dismiss @thekaveman's request for changes since his feedback has been incorporated so we can merge and i can tag another RC. 😇

Scotchester:
Argh, I thought I was doing a rebase update but must have messed it up on my phone. I'm now out and about, could someone else please rebase the backwards merge commit away and merge the PR?

jgravois:
> could someone else please rebase the backwards merge commit away and merge the PR?

i'd be happy to. 🧞

PR #4013, opened 2026-08-24
Fix: more graceful Switchio response parsing

### LINKED ISSUE #4012: TypeError: RegistrationStatus.__init__() got an unexpected keyword argument
Sentry Issue: [BENEFITS-2G](https://compiler.sentry.io/issues/7689673605/?referrer=github_integration)

```
TypeError: RegistrationStatus.__init__() got an unexpected keyword argument 'merchantRestInfo'
  File "benefits/enrollment_switchio/enrollment.py", line 115, in get_registration_status
    registration_status = client.get_registration_status(
  File "benefits/enrollment_switchio/api.py", line 167, in get_registration_status
    return RegistrationStatus(**response.json())
```

thekaveman:
Fixes #4012 

---

This PR adapts the changes that were made on the Littlepay side in https://github.com/cal-itp/littlepay/pull/69 over to the Switchio side.

We implement a `from_kwargs(**kwargs)` class method helper to parse all responses coming back from Switchio API calls in order to more gracefully handle the case where an unexpected field is included in a response.

This PR adds tests for unexpected fields in responses, while also refactoring all Switchio API tests into a class-based format.

### DIFF
--- a/benefits/enrollment_switchio/api.py
+++ b/benefits/enrollment_switchio/api.py
@@ -1,12 +1,46 @@
 import hashlib
 import hmac
 import json
+import logging
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from enum import Enum
+from inspect import signature
 from tempfile import NamedTemporaryFile
 
 import requests
+
+logger = logging.getLogger(__name__)
+
+
+class BaseDataClass:
+
+    @classmethod
+    def from_kwargs(cls, **kwargs):
+        """
+        @classmethod for instantiating a dataclass that allows unexpected fields
+
+        See https://stackoverflow.com/a/55101438
+        """
+        # fetch the constructor's signature
+        class_fields = {field for field in signature(cls).parameters}
+
+        # split the kwargs into native ones and new ones
+        native_args, new_args = {}, {}
+        for name, val in kwargs.items():
+            if name in class_fields:
+                native_args[name] = val
+            else:
+                new_args[name] = val
+
+        # use the native ones to create the class ...
+        instance = cls(**native_args)
+
+        # ... and log any unexpected args
+        for new_name, new_val in new_args.items():
+            logger.info(f"Unexpected arg parsing response for {cls}: {new_name} = {new_val}")
+
+        return instance
 
 
 @dataclass
--- a/benefits/enrollment_switchio/api.py
+++ b/benefits/enrollment_switchio/api.py
@@ -44,7 +44,7 @@ class BaseDataClass:
 
 
 @dataclass
-class Registration:
+class Registration(BaseDataClass):
     regId: str
     gtwUrl: str
 
@@ -62,7 +62,7 @@ class EshopResponseMode(Enum):
 
 
 @dataclass
-class RegistrationStatus:
+class RegistrationStatus(BaseDataClass):
     regState: str
     created: datetime
     mode: str
@@ -181,7 +181,7 @@ class TokenizationClient(Client):
 
         response.raise_for_status()
 
-        return Registration(**response.json())
+        return Registration.from_kwargs(**response.json())
 
     def get_registration_status(self, registration_id, timeout=5) -> RegistrationStatus:
         request_path = f"/api/v1/registration/{registration_id}"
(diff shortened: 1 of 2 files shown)


benefits/enrollment_switchio/api.py
```diff
@@ -1,16 +1,50 @@
...
 
 import requests
 
+logger = logging.getLogger(__name__)
+
+
+class BaseDataClass:
+
+    @classmethod
+    def from_kwargs(cls, **kwargs):
```
jgravois, line 19: i chatted with Gemini a little about this and the LLM suggested a _slightly_ different pattern, namely overriding the built-in `__init__` function.

```python
@dataclass
class User:
    name: str
    email: str

    def __init__(self, **kwargs):
        # Dynamically extract only the fields defined in the dataclass
        names = {f.name for f in fields(self)}
        for k, v in kwargs.items():
            if k in names:
                setattr(self, k, v)

User(**data)
```

i feel like there's merit in not having to remember to instantiate each instance of a class with `.from_kwargs()`, but i'm also willing to admit there may be converse drawbacks that i'm not considering.

tests/pytest/enrollment_switchio/test_api.py
```diff
@@ -18,6 +21,24 @@
...
+
+    @dataclass
+    class SampleDataClass(BaseDataClass):
+        fieldOne: int
+        fieldTwo: str
+
+    def test_from_kwargs(self, caplog):
+        response_json = {"fieldOne": 1, "fieldTwo": "two", "fieldThree": "three"}
+
+        with caplog.at_level(logging.DEBUG):
```
angela-tran, line 34: TIL about `caplog`. Nice!

benefits/enrollment_switchio/api.py
```diff
@@ -28,7 +62,7 @@ class EshopResponseMode(Enum):
 
 
 @dataclass
-class RegistrationStatus:
+class RegistrationStatus(BaseDataClass):
```
jgravois, line 65: seems like it'd result in more noise than signal to start logging `merchantRestInfo` for each and every card registration.

is it premature to add it as an explicit prop?

jgravois (commented):
CIL for a few questions, nits.

i tested this locally and confirmed it gets the job done.

benefits/enrollment_switchio/api.py
```diff
@@ -44,7 +44,7 @@ def from_kwargs(cls, **kwargs):
 
 
 @dataclass
-class Registration:
+class Registration(BaseDataClass):
```
angela-tran, line 47: Inheriting the class method seems much nicer than having it duplicated everywhere 😄  Nice improvement!

angela-tran (commented):
These code changes look good to me 🙌  I will work on testing it out locally

benefits/enrollment_switchio/api.py
thekaveman: Overriding the `__init__` is certainly possible, but it effectively nullifies any reason to use a `@dataclass`. The whole point of those is to implement a standard `__init__` for you based on the configured attributes.

Check out the linked SO discussion (from the comment) for more details/thoughts on different approaches: https://stackoverflow.com/a/55101438

benefits/enrollment_switchio/api.py
thekaveman: IMHO the noise in our logs is nothing to be too concerned about at this point. We don't go into those very often.

On the other hand, we don't actually yet know what this prop is, what data it contains, etc. and so I would say that it is premature to set it.

benefits/enrollment_switchio/api.py
jgravois: found a concise argument _against_ overriding `__init__` in another StackOverflow article.

https://stackoverflow.com/questions/54678337/how-does-one-ignore-extra-arguments-passed-to-a-dataclass

i found an answer that was a bit buried in the original SO thread linked in the code comment that mentions a helper introduced in Python 3.10 that might be useful, but personally i don't mind how explicitly what you're suggesting reads as-is.

https://stackoverflow.com/a/74459763/3019940

angela-tran (approved):
I confirmed the behavior for both before and after the fix:

**Before** ☹️ 

Image

Image

**After** ☺️ 

Image

jgravois (approved):
🚀

lalver1 (approved):
Looks good to me! I also saw the new `merchantRestInfo` in the new INFO log when testing locally 👍

thekaveman:
Thanks all for the reviews and comments!

benefits/enrollment_switchio/api.py
jgravois: premature indeed 😂

PR #4018, opened 2026-08-24
Chore: flag MMW URL registration in prod

thekaveman:
Don't activate these URLs/app in `prod` yet.


jgravois (approved):
verified locally that this still surfaces the route for normal development.

💂

Issue #643, opened 2022-06-06
Update failure flows


thekaveman:
Epic #635 

- [x] Create screen and design flow for Eligibility Not Verified
- [x] Create screen and design flow for Payment not Verified
- [x] Verify error flow for Login.gov
- [x] Verify cancel flow for Login.gov

srhhnry:
- [ ] Create screen and design flow for Eligibility Not Verified
- [ ] Create screen and design flow for Payment not Verified
- [ ] Verify error flow for Login.gov

srhhnry:
Cancel flow is interesting--I'm assuming that means the user has decided to opt out of the login.gov flow by canceling from within the Login.gov site. In that case I believe we should direct them to the eligibility error screen, as it has a direct way to contact the agency (and my guess is cancellation will come because the process is too arduous or the user doesn't have the documents) and a link to the intro/home page with more information

Updated the Figma boards
![Image](https://user-images.githubusercontent.com/5394204/172922174-629644fb-f234-4397-8022-2727bb494ea7.png)

srhhnry:
As for the error flow, the previous idea had been to direct the user back to the Get Started page. On further thought, I think we should also direct an errors in Login.gov to the same Eligibility error page. Given the length of the identity proofing flow, I feel like it's a good idea to give the user more information about what happened, even if it's vague (i.e. eligibility couldn't be verified). 

Updated the Figma board to reflect both of these pathways.

thekaveman:
🙏 Thank you for these pathway arrows @srhhnry very helpful!

srhhnry:
No problem! And really, there should be more! Now that I'm thinking about it, I'll add visualizations of the anchor links to the Help Page, as we're almost done there.

PR #4029, opened 2026-08-28
Refactor: OAuth `cancel_login` logging

### LINKED ISSUE #3666: Improve logging in OAuth cancel handler
## Background

We have the [`cdt_identity.hooks.cancel_login`](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/main/cdt_identity/hooks.py#L101) hook to allow consumers to perform actions when the user cancels their sign in request with the identity provider (IdP) via the Identity Gateway (IdG).

The hook is called in [`cdt_identity.views.cancel` view function](https://github.com/Office-of-Digital-Services/django-cdt-identity/blob/4e263f017cc28ea1b5c64f2968b4b73f249ae22f/cdt_identity/views.py#L110-L114).

In Benefits, we implement this hook to send the `canceled sign in` event to Amplitude: https://github.com/cal-itp/benefits/blob/979f743ac680fde14040d7ddd32a654a6ddb955f/benefits/oauth/hooks.py#L21-L24

And register the view function as our `/oauth/cancel` endpoint, called by IdG in these situations: 

https://github.com/cal-itp/benefits/blob/979f743ac680fde14040d7ddd32a654a6ddb955f/benefits/oauth/urls.py#L22

## Looking for improvements

We learned from CDT that the call to `/oauth/cancel` should contain information that allows us to determine if this is a true user cancellation from the IdP sign in vs. one of the anomalous cancellations we have [seen recently with users in the Medicare.gov flow specifically](https://cal-itp.slack.com/archives/C037Y3UE71P/p1773942303504169).

* Can we separate the true `canceled sign in` events from the anomalous sign in events in Amplitude?
* If so, what details can we send with the anomalous sign in events that might help with later debugging etc.?
* Are there corresponding / supporting improvements to the logging setup in e.g. the cancel hook implementation that can be made?

## Acceptance Criteria

- [ ] We send the `canceled sign in` Amplitude event only for true user-initiated cancellations at the IdP
- [ ] We send a separate Amplitude event for the anomalous cancellations with as much detail as possible

## Additional context

CDT says that a true user-initiated cancellation will come with a message related to or mentioning `"Access denied"`. This is indicative of the IdG's failure to obtain access (from the user) to their details stored in the IdP, because the user canceled their sign in.

When the user does not actively cancel their sign in, we should _not_ see an `"Access denied"` type message.

lalver1:
Closes #3666

This PR 

- Changes when the CanceledSignInEvent is sent so that it is only sent for true user-initiated cancellations at the IdP, i.e., when the query string contains `error=access_denied`
- Creates a new analytics event to handle anomalous oauth cancellations (those that do not contain `error=access_denied` in the query string but do contain other properties in it) by logging details as event properties
- Refactors the [`app_request` Pytest fixtures](https://github.com/cal-itp/benefits/blob/main/tests/pytest/conftest.py#L30) to make them more DRY

## Notes
- The standard [OAuth error codes](https://www.oauth.com/oauth2-servers/authorization/the-authorization-response/) are: `invalid_request`, `access_denied` (the only one we explicitly handle), `unauthorized_client`, `unsupported_response_type`, `invalid_scope`, `server_error`, and `temporarily_unavailable`.
- The [`cancel_login` hook](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/hooks.py#L21) is called by the IdG (after the user is redirected by the IdP) when either the user
  - clicks on _Cancel_ in _Connect your Medicare claims data to State of California?_ during a Medicare flow
  - clicks on _Back to Dev Cal-ITP Benefits_ in _Dev Cal-ITP Benefits is using Login.gov to allow you to sign in to your account safely..._ during a login.gov  flow
- Given that the `remoteFailure` URL in an authentication scheme only allows a single URL, the IdG will always redirect you to `https://dev-benefits.calitp.org/oauth/cancel` even if you started a session in your local development environment. We can come back to this behavior later, but it's noted here since it affects how to review this PR.

## Reviewing

- In your local environment, start a new login.gov or medicare flow. You only need to go up to `/eligibility/start` to prepare the session so that you can [pass the `FlowUsesClaimsVerificationSessionRequired` middleware and be able to successfully call `/oauth/cancel`](https://github.com/cal-itp/benefits/blob/main/benefits/oauth/urls.py#L22) locally.
- Manually enter in the address bar (due to the note above)
  - `https://localhost:PORT/oauth/cancel?error=access_denied&message=error_description%20is%20null&cid=123` for login.gov flows
  - `https://localhost:PORT/?error=access_denied` for the Medicare flow
- And confirm that you see a `canceled sign in` event followed by a `returned eligibility` event in your local server's logs for each of the cases above
- Manually enter in the address bar (due to the note above)
  - `https://localhost:PORT/oauth/cancel?error=other_error&message=error_description%20is%20null&cid=123` for login.gov flows
  - `https://localhost:PORT/?error=other_error&other_property=property` for the Medicare flow
- And confirm that you see a `canceled sign in anomaly` (with appropriate `cancel_sign_in_anomaly` `event_properties`) event followed by a `returned eligibility` event in your local server's logs for each of the cases above

### DIFF
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -20,7 +20,9 @@ class OAuthHooks(DefaultHooks):
     @classmethod
     def cancel_login(cls, request):
         super().cancel_login(request)
-        analytics.canceled_sign_in(request)
+        # getlist() returns [] if "error" is missing
+        if "access_denied" in request.GET.getlist("error"):
+            analytics.canceled_sign_in(request)
         return redirect(routes.ELIGIBILITY_UNVERIFIED)
 
     @classmethod
--- a/benefits/oauth/hooks.py
+++ b/benefits/oauth/hooks.py
@@ -23,6 +23,8 @@ class OAuthHooks(DefaultHooks):
         # getlist() returns [] if "error" is missing
         if "access_denied" in request.GET.getlist("error"):
             analytics.canceled_sign_in(request)
+        else:
+            analytics.canceled_sign_in_anomaly(request)
         return redirect(routes.ELIGIBILITY_UNVERIFIED)
 
     @classmethod
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -65,6 +65,25 @@ def app_request_post(rf):
 
 
 @pytest.fixture
+def app_request_query(rf):
+    """
+    Fixture creates and initializes a new Django request object similar to a real application request with a URI query.
+    """
+    # create a request for the path, initialize
+    app_request = rf.get("/some/arbitrary/path", data={"tag": ["tag_1", "tag_2"], "property_2": "key_2"})
+
+    # https://stackoverflow.com/a/55530933/358804
+    middleware = [SessionMiddleware(lambda x: x), LocaleMiddleware(lambda x: x)]
+    for m in middleware:
+        m.process_request(app_request)
+
+    app_request.session.save()
+    session.reset(app_request)
+
+    return app_request
+
+
+@pytest.fixture
 def model_User():
     return User.objects.create(is_active=True, is_staff=True, first_name="Test", last_name="User")
 
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -26,14 +26,10 @@ def pytest_runtest_setup():
     disable_socket()
 
 
-@pytest.fixture
-def app_request(rf):
+def _setup_request_middleware_session(app_request):
     """
-    Fixture creates and initializes a new Django request object similar to a real application request.
+    Helper function to apply middleware and initialize session for mock requests.
     """
-    # create a request for the path, initialize
-    app_request = rf.get("/some/arbitrary/path")
-
     # https://stackoverflow.com/a/55530933/358804
     middleware = [SessionMiddleware(lambda x: x), LocaleMiddleware(lambda x: x)]
     for m in middleware:
@@ -43,6 +39,15 @@ def app_request(rf):
     session.reset(app_request)
 
     return app_request
+
+
+@pytest.fixture
+def app_request(rf):
+    """
+    Fixture creates and initializes a new Django request object similar to a real application request.
+    """
+    request = rf.get("/some/arbitrary/path")
+    return _setup_request_middleware_session(request)
 
 
 @pytest.fixture
@@ -50,18 +55,8 @@ def app_request_post(rf):
     """
     Fixture creates and initializes a new Django POST request object similar to a real application request.
     """
-    # create a request for the path, initialize
-    app_request = rf.post("/some/arbitrary/path")
-
-    # https://stackoverflow.com/a/55530933/358804
-    middleware = [SessionMiddleware(lambda x: x), LocaleMiddleware(lambda x: x)]
-    for m in middleware:
-        m.process_request(app_request)
-
-    app_request.session.save()
-    session.reset(app_request)
-
-    return app_request
+    request = rf.post("/some/arbitrary/path")
+    return _setup_request_middleware_session(request)
 
 
 @pytest.fixture
@@ -69,18 +64,8 @@ def app_request_query(rf):
     """
     Fixture creates and initializes a new Django request object similar to a real application request with a URI query.
     """
-    # create a request for the path, initialize
-    app_request = rf.get("/some/arbitrary/path", data={"tag": ["tag_1", "tag_2"], "property_2": "key_2"})
-
-    # https://stackoverflow.com/a/55530933/358804
-    middleware = [SessionMiddleware(lambda x: x), LocaleMiddleware(lambda x: x)]
-    for m in middleware:
-        m.process_request(app_request)
-
-    app_request.session.save()
-    session.reset(app_request)
-
-    return app_request
+    request = rf.get("/some/arbitrary/path", data={"tag": ["tag_1", "tag_2"], "property_2": "key_2"})
+    return _setup_request_middleware_session(request)
 
 
 @pytest.fixture
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -64,7 +64,7 @@ def app_request_query(rf):
     """
     Fixture creates and initializes a new Django request object similar to a real application request with a URI query.
     """
-    request = rf.get("/some/arbitrary/path", data={"error": "error_code", "message": "error_description", "cid": "123"})
+    request = rf.get("/some/arbitrary/path", data={"error": "error_code", "message": "error_description"})
     return _setup_request_middleware_session(request)
 
 
--- a/tests/pytest/conftest.py
+++ b/tests/pytest/conftest.py
@@ -64,7 +64,7 @@ def app_request_query(rf):
     """
     Fixture creates and initializes a new Django request object similar to a real application request with a URI query.
     """
-    request = rf.get("/some/arbitrary/path", data={"tag": ["tag_1", "tag_2"], "property_2": "key_2"})
+    request = rf.get("/some/arbitrary/path", data={"error": "error_code", "message": "error_description", "cid": "123"})
     return _setup_request_middleware_session(request)
 
 
--- a/benefits/oauth/analytics.py
+++ b/benefits/oauth/analytics.py
@@ -37,6 +37,15 @@ class CanceledSignInEvent(OAuthEvent):
         super().__init__(request, "canceled sign in")
 
 
+class CanceledSignInAnomalyEvent(OAuthEvent):
+    """Analytics event representing the anomalous canceling of application sign in."""
+
+    def __init__(self, request):
+        super().__init__(request, "canceled sign in anomaly")
+        details = dict(request.GET.lists())
+        self.update_event_properties(cancel_sign_in_anomaly=details)
+
+
 class FinishedSignInEvent(OAuthEvent):
     """Analytics event representing the end of the OAuth sign in flow."""
 
@@ -81,6 +90,11 @@ def started_sign_in(request):
 def canceled_sign_in(request):
     """Send the "canceled sign in" analytics event."""
     core.send_event(CanceledSignInEvent(request))
+
+
+def canceled_sign_in_anomaly(request):
+    """Send the "canceled sign in anomaly" analytics event."""
+    core.send_event(CanceledSignInAnomalyEvent(request))
 
 
 def finished_sign_in(request, error=None):
--- a/benefits/oauth/analytics.py
+++ b/benefits/oauth/analytics.py
@@ -42,8 +42,10 @@ class CanceledSignInAnomalyEvent(OAuthEvent):
 
     def __init__(self, request):
         super().__init__(request, "canceled sign in anomaly")
-        details = dict(request.GET.lists())
-        self.update_event_properties(cancel_sign_in_anomaly=details)
+        error = request.GET.get("error")
+        message = request.GET.get("message")
+        cid = request.GET.get("cid")
+        self.update_event_properties(error=error, message=message, cid=cid)
 
 
 class FinishedSignInEvent(OAuthEvent):
--- a/benefits/oauth/analytics.py
+++ b/benefits/oauth/analytics.py
@@ -44,8 +44,7 @@ class CanceledSignInAnomalyEvent(OAuthEvent):
         super().__init__(request, "canceled sign in anomaly")
         error = request.GET.get("error")
         message = request.GET.get("message")
-        cid = request.GET.get("cid")
-        self.update_event_properties(error=error, message=message, cid=cid)
+        self.update_event_properties(error=error, message=message)
 
 
 class FinishedSignInEvent(OAuthEvent):
--- a/tests/pytest/oauth/test_analytics.py
+++ b/tests/pytest/oauth/test_analytics.py
@@ -2,6 +2,7 @@ import pytest
 
 import benefits.core.analytics
 from benefits.oauth.analytics import (
+    CanceledSignInAnomalyEvent,
     CanceledSignInEvent,
     FailureToProofEvent,
     FinishedSignInEvent,
@@ -11,6 +12,7 @@ from benefits.oauth.analytics import (
     StartedSignInEvent,
     StartedSignOutEvent,
     canceled_sign_in,
+    canceled_sign_in_anomaly,
     error,
     failure_to_proof,
     finished_sign_in,
@@ -61,6 +63,12 @@ def test_FinishedSignInEvent_with_error(app_request):
 
 
 @pytest.mark.django_db
+def test_CanceledSignInAnomalyEvent(app_request_query):
+    event = CanceledSignInAnomalyEvent(app_request_query)
+    assert event.event_properties["cancel_sign_in_anomaly"] == {"tag": ["tag_1", "tag_2"], "property_2": ["key_2"]}
+
+
+@pytest.mark.django_db
 def test_FinishedSignInEvent_without_error(app_request):
     event = FinishedSignInEvent(app_request)
     assert "error_code" not in event.event_properties
@@ -102,6 +110,18 @@ def test_canceled_sign_in(app_request, spy_send_event):
     call_arg = spy_send_event.call_args[0][0]
     assert isinstance(call_arg, CanceledSignInEvent)
     assert call_arg.event_type == "canceled sign in"
+
+
+@pytest.mark.django_db
+def test_canceled_sign_in_anomaly(app_request, spy_send_event):
+    canceled_sign_in_anomaly(app_request)
+
+    # event should have been sent
+    spy_send_event.assert_called_once()
+    # the first arg of the first (and only) call
+    call_arg = spy_send_event.call_args[0][0]
+    assert isinstance(call_arg, CanceledSignInAnomalyEvent)
+    assert call_arg.event_type == "canceled sign in anomaly"
 
 
 @pytest.mark.django_db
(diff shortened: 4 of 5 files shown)


thekaveman:
@indexing this PR improves how we handle OAuth sign-in cancellation to distinguish between true cancellations and those anomalous cancellations we were seeing with some Medicare users.

@lalver1 please correct anything below that I misstate. I want to ensure @indexing has a chance to weigh in on these changes before we merge.

### Event naming

This PR adjusts when we send the existing `canceled sign in` event and adds the new `canceled sign in anomaly` event.

Does that naming work and match your expectations?

### Event payload

This PR sends additional error details with `canceled sign in anomaly` events in a `cancel_sign_in_anomaly` property. The value of this property comes from the URL querystring parameters. So for example we might get a request with a querystring that looks like:

```
/oauth/cancel?error=other_error&message=error_description%20is%20null&cid=123
```

And the value of the event parameter would look like:

```python
cancel_sign_in_anomaly=[('error', ['other_error']), ('message', ['error_description is null'), ('cid', ['123'])]
```

Does this high-level roll-up work and match your expectations?

_IMHO this is really verbose and not useful to work with from an analytics point of view. I'd rather see us break this up into individual event props or at least a JSON structure under a `details` prop in the event._

indexing:
Separating user-initiated cancellations from other OAuth errors is the key first step, so this PR is a nice first iteration.

I agree with @thekaveman , however, that ideally we would have handling for each of the individual Oauth errors in our analytics. This separation of events/properties will help us pinpoint exactly how an eligibility check fails with Medicare.gov. So perhaps we use an event separate from `canceled sign in` called `oauth error` with the individual properties representing each exception (`invalid_request`, `access_denied`, `unauthorized_client`, `unsupported_response_type`, `invalid_scope, server_error`)?

I also want to ask: how do we log the user action when someone clicks the Cancel button on the Medicare.gov consent screen? For me, this action is a distinct event from the set of Oauth errors. The person is simply canceling the flow by clicking the Cancel button; they aren’t actually denying access. (This same event is mirrored in the Login.gov flow.) Moreover, the description for `access_denied` is broader than a user action, “the user _or authorization server_ denied the request.” If we can separate cancellation from denying access, we gain more insight. In my mind, the user might, for example, deny the request by unchecking the consent to share their information. Again, this is a distinct choice from choosing to cancel the flow. Alternately, the authorization server might deny the request and we know this wasn't the user canceling the flow. 

I hope I'm in the ballpark here with the details and y'all get the gist of how I see the analytics functioning to give us better insight into how the eligibility flow ends.

indexing (dismissed):
See my "clarifying questions" comment: https://github.com/cal-itp/benefits/pull/4029#issuecomment-5515529016

lalver1:
Thanks for your comments @thekaveman and @indexing, they helped clear up some things for myself. I made a few changes outlined below:

> IMHO this is really verbose and not useful to work with from an analytics point of view. I'd rather see us break this up into individual event props or at least a JSON structure under a `details` prop in the event.

I completely agree, now I see that initially I mixed up logging and analytics in my mind. To try to satisfy both requirements I just stuffed everything that we could possibly see in the URL's querystring into an event property, but this clearly made for awkward analytics. In 60ecf92d1a47d3f12cda212a9440d1f0e600216c I instead focused on the parameters that we've seen that the IdG adds to the URL's querystring (`error`, `message`, `cid`) and made each of those an individual event property. If we become aware of other parameters coming up as edge cases, we can add them to the analytics, but at this point it feels premature to do so and it just makes for analytics that are not useful to work with.

One followup question @thekaveman, should I even include `cid` as an event property? It may be some sort of correlation id property that the IdG adds to internally trace a request (it's not part of the OAuth standard), but this is going into the logging realm and probably does not belong in analytics? 

> I agree with @thekaveman , however, that ideally we would have handling for each of the individual Oauth errors in our analytics. This separation of events/properties will help us pinpoint exactly how an eligibility check fails with Medicare.gov. So perhaps we use an event separate from `canceled sign in` called `oauth error` with the individual properties representing each exception (`invalid_request`, `access_denied`, `unauthorized_client`, `unsupported_response_type`, `invalid_scope, server_error`)?

With the change in 60ecf92d1a47d3f12cda212a9440d1f0e600216c, your description @indexing matches how the new event will behave. The only difference is that the event would be named `canceled sign in anomaly` instead of `oauth error` to be consistent with the `oauth/cancel` endpoint of Benefits which is where this event is triggered from. And in the event properties you would see individual properties for `error`, `message`, and `cid`. So a `canceled sign in anomaly` event will look like

```
{
  "app_version": ...,
  ...,
  "event_properties": {
    "path": "/oauth/cancel",
    "transit_agency": "...",
    "eligibility_verifier": "...",
    "enrollment_method": "...",
    "enrollment_flows": [
      "..."
    ],
    "claims_provider": "...",
    "error": "server_error",
    "message": "error_description is null",
    "cid": "123"
  },
  "event_type": "canceled sign in anomaly",
  ...
}
```

Would this work for you @indexing?

> I also want to ask: how do we log the user action when someone clicks the Cancel button on the Medicare.gov consent screen? For me, this action is a distinct event from the set of Oauth errors. The person is simply canceling the flow by clicking the Cancel button; they aren’t actually denying access. (This same event is mirrored in the Login.gov flow.) Moreover, the description for `access_denied` is broader than a user action, “the user _or authorization server_ denied the request.” If we can separate cancellation from denying access, we gain more insight. In my mind, the user might, for example, deny the request by unchecking the consent to share their information. Again, this is a distinct choice from choosing to cancel the flow. Alternately, the authorization server might deny the request and we know this wasn't the user canceling the flow.

When a user

- clicks on _Cancel_ in _Connect your Medicare claims data to State of California?_ during a Medicare flow or
- clicks on _Back to Dev Cal-ITP Benefits_ in _Dev Cal-ITP Benefits is using Login.gov to allow you to sign in to your account safely..._ during a login.gov flow

after a series of redirects from the IdP and the IdG, the user will send a request to Benefits similar to `https://benefits.calitp.org/oauth/cancel?error=access_denied&...`. It is until this point where we actually log the `canceled sign in` event when the `/oauth/cancel` endpoint is called with `error=access_denied`. Unfortunately, we completely rely on the IdP and IdG for these redirects (and the URL that the redirect sends the user to) so we can't get more granularity/insight. As noted in the first bullet item above, when the user clicks the Cancel button on the Medicare.gov consent screen, this will get logged in analytics as a `canceled sign in` because of how the IdP/IdG generate the redirects. Maybe the root issue is that the OAuth spec could have been more granular? Instead of defining `access_denied: the user or authorization server denied the request` they could have broken this up into 2 separate errors, one for the auth server and another for the user, which I think it is what you are after.

indexing:
Thanks @lalver1 . To ensure I understand the current status:

- the `canceled sign in` analytics event allows us to log the user action of clicking the Cancel button, or using the browser back button to return to Cal-ITP Benefits, by associating those actions with the specific Oauth error `access_denied`. Ideally, we _would_ distinguish these "navigation" actions from Oauth errors, as they are not actually errors, but thanks for confirming we are dependent on the IdG response. (Perhaps we could request an enhancement on the IdG to provide us this differentiation.)
- the `canceled sign in anomaly` event allows us to log exceptions when the IdG cannot complete any eligibility check due to these Oauth errors: `invalid_request`, `unauthorized_client`, `unsupported_response_type`, `invalid_scope`, and `server_error`. 

I'm still a bit confused about the event naming. I'm also wondering about the value of separating these errors into different analytic events. 

Assuming the above is accurate (we cannot differentiate cancellations from Oauth errors), perhaps we simply consolidate both of the above events into a single `Oauth error` analytic event? When we see, the `access_denied` property for this event, we can assume it's likely a user-driven interruption of the flow. When/if we do have the ability to log cancellations as a distinct user action, we can bring back the `canceled sign in` event. (I note there is a stale 'Oauth event` in Amplitude currently.)

This is simply a thought without considering level of effort or other ramifications of this change. I can certainly work with the current event naming.

thekaveman:
@lalver1 

> should I even include `cid` as an event property

I would say probably not? I agree it is more relevant in a logging context. We are logging these errors as well, right?

@indexing 

> I'm still a bit confused about the event naming. I'm also wondering about the value of separating these errors into different analytic events.
> ...
> perhaps we simply consolidate both of the above events into a single `Oauth error` analytic event? When we see, the `access_denied` property for this event, we can assume it's likely a user-driven interruption of the flow.

I'm still confused/nonplussed about the naming as well. I would agree with the approach laid out above: use props to differentiate the different types of errors. 

To Luis' point though, this is all driven through our `/oauth/cancel` endpoint which must be registered with the IdG in order to have a place to send the user for all these cases. So I'd advocate for keeping the (catch-all) event name as `canceled sign in`

indexing:
@thekaveman Is the `/oauth/cancel` endpoint named that way using Oauth standards, or is that URL naming determined by an implementation choice on the IdG? I'm wondering how faithful we need to be to the endpoint naming versus conceptually what's happening in the application.

indexing:
Adding another log to the fire. This example in the Oauth documentation clearly demonstrates how I would expect the `access_denied` exception to work. Cancelling or navigating back in the flow is just different that not authorizing. 

Screenshot 2026-09-04 at 11 57 36 AM

thekaveman:
> Is the `/oauth/cancel` endpoint named that way using Oauth standards, or is that URL naming determined by an implementation choice on the IdG? I'm wondering how faithful we need to be to the endpoint naming versus conceptually what's happening in the application.

It's a good question @indexing, and the standard(s) don't say anything about "OAuth cancellation" per se. 

@johnatstate has mentioned needing our app to register with the IdG:

> ... a single remoteFailure URL in an authn scheme

And that is currently pointing at our `/oauth/cancel` endpoint (if my Slack sleuthing can be believed, this was from back in the day before _failure to proof_ was fully supported/specified by Login.gov AND because we found we needed to specifically handle the "user declines to provide consent / cancels sign in" use case, hence our naming).

While we know IdG implements OpenID Connect (OIDC, a superset of OAuth 2.0) _to talk with Login.gov and other IDPs_, there are some implementation details for Benefits as a _client_ of IdG that do not map cleanly to the OIDC or OAuth 2.0 spec, like this `RemoteFailureRedirectUrl` registration that IdG requires.

It is important to keep that in mind when looking at the spec, we can't necessarily map our interaction with IdG onto it, though our interaction is _mostly_ according to OIDC spec. What your screenshot shows is how the IdG would process a response from the IDP (e.g. Login.gov), not necessarily exactly how it would flow back to Benefits.

---

All this to say, I think @indexing you are probably correct in the direction you are pointing: there is **not a good reason** for us to call _our_ endpoint `/oauth/cancel` beyond the history I shared above. Given we are attempting to better handle these situations with more granularity, perhaps we should be renaming the endpoint to `/oauth/error` and the event to `oauth error`, like you are suggesting.

The finer grain we would get from the work @lalver1 is doing in this PR would allow us to "calculate" an event like `canceled sign in` (from the props) while also having more information / nuance about the other error scenarios.

And finally, this is all of course separate from `failure to proof`, which is yet another endpoint registration with IdG that we have yet to complete. @Scotchester built the handler for it in https://github.com/Office-of-Digital-Services/django-cdt-identity/pull/69 but we haven't implemented that handler in Benefits yet.

lalver1:
Thanks for the good discussion y'all. I agree with most of what has been said, just one small clarification @indexing, clicking the browser's back button to go from the login.gov Sign in page back to Benefits' `eligibility/start` will not trigger the `/oauth/cancel` endpoint so "Back button clicks" wouldn't get logged in analytics. But everything else in how you understand the current status in https://github.com/cal-itp/benefits/pull/4029#issuecomment-5544210964 is correct.

@thekaveman I removed `cid` from the event's properties in 68e79c222b7f43dc9c033d1a84a83def4320d7a8 since the whole `GET /oauth/cancel` request (with querystring) is available in the logs. At first I was tempted to add a `logger` statement for clarity but it's not even required. If we do see a `canceled sign in anomaly`, we can just search the logs for `/oauth/cancel` if we ever need these `cid` values.

thekaveman:
Here are some of the earliest discussions I could find related to this topic:

- https://github.com/cal-itp/benefits/issues/643 <-- design issue where we first described the logic
- https://github.com/cal-itp/benefits/issues/713 <-- dev issue where we first implemented the endpoint/view
- https://cal-itp.slack.com/archives/C037Y3UE71P/p1654793877924689 <-- first mention in Slack of cancellation (also, Failure to Proof :facepalm:)
- https://cal-itp.slack.com/archives/C037Y3UE71P/p1655937253397919 <-- confirmation of the endpoint we wanted to be registered at the IdG
- https://cal-itp.slack.com/archives/C037Y3UE71P/p1656006501308629 <-- follow-up conversation + screen recording of Angela testing cancellation

lalver1:
Per our discussion on error handling, I removed `canceled sign in anomaly` (`CanceledSignInAnomalyEvent`) and implemented sending the `oauth error` event instead when `/oauth/cancel` is called to handle any requests where the `error` parameter in the query string is not `access_denied`. These are situations that are not triggered by the user clicking on Cancel or Back in the medicare and login.gov consent page. The naming `canceled sign in anomaly` assumed too much (there could be other reasons why `/oauth/cancel` was called) and was confusing, especially since we already have an `oauth error` event that is better suited for this.

The idea is to monitor the analytics for these events, look into the logs if needed to get more context if an `oauth error` is identified (for example if there is a key in the query string that is not `error` or `error_description`), and refine the handler as we collect more information.

benefits/oauth/hooks.py
```diff
@@ -20,7 +20,13 @@ def pre_login(cls, request):
     @classmethod
     def cancel_login(cls, request):
         super().cancel_login(request)
```
jgravois: i'd expect a breakpoint within this funcction to be hit when:
1. selecting Medicare
1. logging in successfully
1. clicking <kbd>Cancel</kbd> on the new BB v3 consent screen instead of <kbd>Share</kbd>

instead locally i'm redirected to https://localhost:30300/error.html?error=

is this consistent with what you're experiencing @lalver1?

benefits/oauth/hooks.py
```diff
@@ -20,7 +20,13 @@ def pre_login(cls, request):
...
-        analytics.canceled_sign_in(request)
+        # getlist() returns [] if "error" is missing
+        if "access_denied" in request.GET.getlist("error"):
```
jgravois: this is a nit (at best) but wouldn't 👇 be safe even if `error` was missing from the query string entirely?

```python
error_code = request.GET.get("error")
if error_code == "access_denied":
  analytics.canceled_sign_in(request)
else: 
  error_description = request.GET.get("error_description")
  analytics.error(request, message=error_description, operation=error_code)
return redirect(routes.ELIGIBILITY_UNVERIFIED)
```

tests/pytest/conftest.py
jgravois, line 1: this deduplication you snuck in looks like an improvement to me!

jgravois (commented):
CIL for one nit but i'd co-sign this as-is.

tests/pytest/conftest.py
lalver1: Ah thanks for noticing that @jgravois! It was more valuable when we had the `CanceledSignInAnomalyEvent` (we were saving having the same code in 3 places, now we're only saving having it in 2 places) but it felt ok to leave in since the work was already done and it's still useful to have.

benefits/oauth/hooks.py
lalver1: No, that's a good point! You're right that ☝️ would still be safe (and simpler). I used

```python
if "access_denied" in request.GET.getlist("error"):
```

because I was still in "we don't know a whole lot about the query parameters we'll see in requests to `/oauth/cancel`" mode. This was making me think, what if we see multiple `error` keys in the querystring? Technically this is possible and `request.GET.get("error")` only [returns the last one](https://docs.djangoproject.com/en/6.1/ref/request-response/#django.http.QueryDict.get) and we could miss `access_denied`. So checking for `in list` felt safer. But after our latest conversations, this seems like an unlikely situation, and I would agree with your proposed change @jgravois. I can make this change if y'all agree that the code doesn't need to be so defensive as I was initially thinking.

jgravois: > `request.GET.get("error")` only [returns the last one](https://docs.djangoproject.com/en/6.1/ref/request-response/#django.http.QueryDict.get)

in my mind, it would be appropriate to call `analytics.error()` if we ever encountered a query string like `?error=access_denied&error=something_else...`.

not a strong opinion though.

thekaveman: I'd agree with @jgravois. If we saw more than one, we can't know which one is the "right" one and an error is the safest thing to log.

lalver1: I went back and forth on this, but like y'all mentioned, I believe John's suggestion that I implemented in 6e6e11022b7d09201ab04ac486e17a91eb98148a is the better approach 👍

The main reason is that during our conversation about the IdG's behavior, more than 1 `error` key in the query string was actually never mentioned, so it's just been an assumption of mine that we could see more than 1 `error`. Since seeing more than 1 `error` is very unlikely, the alternative below is unnecessarily complicated (and not quite right, the `else` branch is missing logic to select all the potential error codes and send them all by either using a new `analytics.other_error` function and new `OAuthErrorEvent` or making several refactors to the existing one):

```python
@classmethod
    def cancel_login(cls, request):
        super().cancel_login(request)
        is_single_error = len(request.GET.getlist("error")) == 1
        error_code = request.GET.get("error")
        if error_code == "access_denied" and is_single_error:
            analytics.canceled_sign_in(request)
        else:
            error_description = request.GET.get("error_description")
            analytics.error(request, message=error_description, operation=error_code)
        return redirect(routes.ELIGIBILITY_UNVERIFIED)
```

Hence my thinking that the approach in 6e6e11022b7d09201ab04ac486e17a91eb98148a is the better option given what we know for certain.

jgravois (approved):
🍔

thekaveman (approved):
Very nice work 🎸

lalver1:
sorry @angela-tran, I got caught up with closing this ticket and forgot that you said that you were catching up on this PR today 😅 But if you have any comments I'm more than happy to come back to it, just let me know!

Issue #3834, opened 2026-06-30
VCTC: enable U.S. Veteran and GCTD Reduced Fare ID enrollment flows


jgravois:
Immediately before the Ventura region goes live in production, we need to enable the U.S. Veteran enrollment flows for Ventura County Transportation Commission to match all their sibling agencies.

The GCTD Reduced Fare ID enrollment flow was added to VCTC's flows as a part of production validation testing.

## Acceptance Criteria

- [x] <kbd>U.S. Veteran should</kbd> appear as an enrollment option in the self-service list
- [x] <kbd>GCTD Reduced Fare ID</kbd> should appear as an option in the in-person list

## Additional context

originally posted in https://github.com/cal-itp/benefits/issues/3691#issuecomment-4566686164

related: #3703

indexing:
@angela-tran Our main point of contact at VCTC, and coordinator for the Ventura regional effort, plans to do some production validation beginning as early as tomorrow. He plans to use in-person enrollment to register cards for testing. While he could use the older adult or Medicare cardholder in-person enrollment options that are currently active, we'd like to test an enrollment using the GCTD Reduced Fare ID in-person option (primarily to confirm proper logging in our analytics). 

I note in this issue enabling GCTD Reduced Fare ID in-person option in slated to happen just before launch. Are you able to make this config change now? Enabling now will also align with the production validation config for the other Ventura county agencies. (Note: U.S. Veterans self-service will remain off until the full launch.)

cc// @laneymangan

angela-tran:
> I note in this issue enabling GCTD Reduced Fare ID in-person option in slated to happen just before launch. Are you able to make this config change now? Enabling now will also align with the production validation config for the other Ventura county agencies. (Note: U.S. Veterans self-service will remain off until the full launch.)

@indexing - Yep, I can do that. I'll log into prod Benefits in just a bit and add the GCTD Reduced Fare ID to VCTC's flows.

angela-tran:
Hey @indexing - I thought about this a little more, and I believe that technically, an alternative approach to confirm proper logging in our analytics would be to use one of the other Ventura county agencies that already have GCTD Reduced Fare ID enabled as a flow. 

My guess though is that by adding it to VCTC, our contact at VCTC can log in with their existing `prod` Benefits Administrator account and do production validation testing via in-person enrollment (as opposed to creating a new dummy account for them associated to one of the other agencies or trying to mess around with swapping which agency their existing account is associated with).

[...]

PR #4038, opened 2026-08-31
docs: marks agencies in ventura county LIVE

indexing:
Per the following issues:
- #3695 
- #3737 
- #3747
- #3758
- #3782  
- https://github.com/cal-itp/benefits/issues/3834


jgravois (changes requested):
lookin good. just need to tick off 'U.S. Veteran' for VCTC too. 😇

angela-tran (commented):
@indexing - should we also update VCTC to have a :white_check_mark:  for the U.S. Veterans benefits option in this PR (per #3834)? Or were you planning on doing that in a separate PR?

indexing:
@angela-tran @jgravois VCTC should be marked with `veterans`. Good catch! Making that change now.

angela-tran (approved):

jgravois (approved):
😎

PR #4030, opened 2026-08-28
docs(newsletters): Add GHA workflow to add newsletter to archive

Scotchester:
Closes #3960 

---

This PR adds a GitHub Actions workflow to add a new newsletter to our archive in the docs.

The workflow is triggered manually, taking the HubSpot ID of a marketing email. The script fetches the web version's HTML and images, adds them to the appropriate places in the docs folder, creates the shell page, and inserts a new entry into the index page. The shell page and the index page have some TK spots for @indexing to decide how he wants to summarize the content of that particular newsletter.

It commits the work to a new branch and then opens a PR.

See it being tested at: https://github.com/compilerla/actions-testing/pull/170 (Not an exact representation – some minor tweaks have been made in the code submitted here).

PR #4041, opened 2026-09-01
fix(newsletters): Add insertion point to archive index

Scotchester:
Follow-up to #4030

The automation for adding a new card to the index won't work correctly without this.


jgravois (approved):
good catch. i missed this entirely.

Scotchester:
No worries; it's easy to miss when you can't run it for real until it's merged!

jgravois:
> it's easy to miss

one of the rare drawbacks of using a playground repo for testing. it makes it really hard to diff the full set of changes you end up proposing.

PR #4050, opened 2026-09-04
chore: generate placeholder enrollment_init app

### LINKED ISSUE #4045: INIT integration: set up `enrollment_init` app
This task adds the needed foundational structure, i.e. a `enrollment_init` Django app, to Benefits. This Django app is where most of the code will be for integrating with INIT for transit benefit enrollment.

The scope of this task is to simply create the Django app. Having this foundation in place will allow us to work on other tasks, such as defining models and their Django admin configuration, etc., in parallel.

The Switchio analog to this issue is #2902, and the PR for that issue ([#2911](https://github.com/cal-itp/benefits/pull/2911/commits)) can be referenced as a way of getting started. As the first commit on that PR says, you can get an initial Django app in place by running
```
python manage.py startapp enrollment_init

mv enrollment_init/ benefits/
```

## Acceptance Criteria

- [ ] An `enrollment_init` Django app is defined under the `benefits` directory
- [ ] `benefits.enrollment_init` is included in `INSTALLED_APPS`

## Additional context

I considered having this task include the definition of models that we will need, but decided against it so that it is a separate task and will not block having the `enrollment_init` Django app in place.

jgravois:
closes #4045

### DIFF
--- a/benefits/enrollment_init/admin.py
+++ b/benefits/enrollment_init/admin.py
@@ -1,0 +1,3 @@
+# from django.contrib import admin
+
+# Register your models here.
--- a/benefits/enrollment_init/apps.py
+++ b/benefits/enrollment_init/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class EnrollmentInitConfig(AppConfig):
+    name = "benefits.enrollment_init"
+    label = "enrollment_init"
--- a/benefits/enrollment_init/models.py
+++ b/benefits/enrollment_init/models.py
@@ -1,0 +1,3 @@
+# from django.db import models
+
+# Create your models here.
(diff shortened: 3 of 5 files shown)


benefits/enrollment_init/tests.py
```diff
@@ -0,0 +1,3 @@
+# from django.test import TestCase
+
+# Create your tests here.
```
thekaveman: We don't typically add tests at the `benefits/` level, instead we use the `tests/pytest/` directory (which largely mirrors `benefits/` as it goes down).

thekaveman (changes requested):
Very minor nit

benefits/enrollment_init/tests.py
jgravois: 👍 i'll just yank this autogenerated file altogether.

thekaveman (approved):

Issue #4040, opened 2026-09-01
tests-ui.yml github action workflow fails unexpectedly intermittently


jgravois:
lately i've noticed more intermittent [flakiness](https://github.com/cal-itp/benefits/actions/workflows/tests-ui.yml?query=is%3Afailure) in `tests.ui.yml` that i've been able to workaround via manually rerunning the workflow.

i took a quick look at the lighthouse docs and noticed that our attempt to trigger [3 runs](https://github.com/cal-itp/benefits/blob/c16c0258f3ddf7b2b1057f4251953c376435e04c/.github/lighthouserc.json#L2-L3) isn't being honored, likely because the configuration syntax changed after @machikoyasuda last touched this back in #202.

https://github.com/cal-itp/benefits/blob/c16c0258f3ddf7b2b1057f4251953c376435e04c/.github/lighthouserc.json#L2-L3

## Our options

1. start triggering 3 runs again (as proposed in https://github.com/cal-itp/benefits/pull/4039).

to be perfectly honest its possible that one failure out of three will still fail the entire job, but it'd be a cheap test to just merge it and 'wait and see'. the only drawback being that the total runtime increases ~1 minute.

2. continue to run lighthouse only once, remove the unused configuration setting, and tolerate/ignore the intermittent errors

3. put some effort into diagnosing why calls to `/` and `/help` in the container are failing more often than they used to. is it a timing thing? an external dependency that doesn't load consistently?

## Acceptance Criteria

- [ ] `tests-ui.yml` passes in gh action runs associated with unrelated PRs 100% of the time.
- [ ] our lighthouse config file contains no ignored/invalid declarations

PR #4051, opened 2026-09-04
docs: marks Yolo as live

indexing:
Per https://github.com/cal-itp/benefits/issues/3881#issuecomment-5543315652, I've updated the status of Yolo.


jgravois (approved):
unrelated to this PR, but looking at the updated table has me wondering, do we want to advertise that the Ventura region is configured for (in-person only) agency cards?

jgravois:
Screenshot 2026-09-04 at 9 38 46 AM

#4040 strikes again!

indexing:
@jgravois Interestingly, I was just thinking the same re: agency cards. This will also apply for SC Metro, so I will look to make this change next week.

Issue #3832, opened 2026-06-29
Agency onboarding: consider linking SwitchioGroup to SwitchioConfig instead of TransitAgency


jgravois:
we have made a _lot_ of progress simplifying how agencies and their associated transit processors are modeled. with regard to the looming Ventura region, only one duplicative configuration step remains.

in both `test` and `prod` all Ventura agencies will share a single `SwitchioConfig` and support the same enrollment flows and card schemes. the way the relationships are modeled currently, it is necessary to create a duplicate `SwitchioGroup` for each individual supported enrollment flow for each individual agency.

unless the relationship back to the `TransitAgency` itself serves a purpose that i'm unaware of (telemetry, logging, etc.) it would be superior if the region required only 6 `SwitchioGroups` (3 flows linked to one `SwitchioConfig` in two envs) instead of 30 (3x5 = 15 in `test` and 15 more in `prod`).

PR #4053, opened 2026-09-04
chore: define an initial 'INIT config' model

### LINKED ISSUE #4046: INIT integration: define `InitConfig` model
As previously alluded to in https://github.com/cal-itp/benefits/issues/4010#issuecomment-5444492825, we need a model to store the details for connecting to Collect.js for tokenization and to MOBILEvario for registration.

For Collect.js. we know we need:
- a URL to the script, which can be used for any agency and environment
- an API key, which is agency-specific and environment-specific

For MOBILEvario, we know we need:
- a base API URL, which is agency/region-specific and environment-specific
- the header value for Basic HTTP Authentication, which is also agency/region-specific (as a function of it being specific to an API instance) and environment-specific

The definition of `InitConfig` should take all that into consideration and use patterns from `LittlepayConfig` and `SwitchioConfig`. It should also define its own `ModelAdmin` subclass and honor the same add/change/delete/view permissions as the `LittlepayConfig`/`SwitchioConfig` models.

## Acceptance Criteria

- [ ] `InitConfig` model is defined with respect to the details listed above
- [ ] `InitConfig` models can be created / deleted / viewed / changed in Benefits Administrator with the expected permissions respected for our user types of superuser, Cal-ITP staff user, and transit agency staff user
- [ ] Sample fixtures updated
- [ ] Shared LastPass fixtures updated

## Additional context

This is all based on our conversations with the INIT team in the context of implementing this for PSTA and their riders. We understand that other agencies that use INIT may not use Collect.js for tokenization and that the MOBILEvario API instances for those agencies may use different mechanisms for authentication. We recognize that the `InitConfig` we define in this task may not work as-is for the next INIT integration, and that is ok.

jgravois:
closes #4046

This PR introduces the `InitConfig` model but does not hook it up to be used on `TransitAgency` yet.

image

<details><summary>Screenshot of older iteration</summary>

Screenshot 2026-09-08 at 9 14 11 AM

</details>


benefits/enrollment_init/admin.py
```diff
@@ -1,3 +1,14 @@
...
+from benefits.core.admin.mixins import StaffPermissionMixin
+from benefits.enrollment_init import models as benefitsModels
+
+
+@admin.register(benefitsModels.InitConfig)
+class InitConfigAdmin(StaffPermissionMixin, admin.ModelAdmin):
+    # use a TextInput for inherited fields instead of a Textarea
+    formfield_overrides = {
+        models.TextField: {"widget": TextInput(attrs={"size": "80"})},
+    }
```
jgravois: this is just a 'quality of life' tweak to display the Label/Portal Url in a plain old `<input type="text">` instead of `type="textarea"`

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,54 @@
...
+
+
+class InitConfig(TransitProcessorConfig):
+    """Configuration for connecting to INIT, an entity that applies transit agency fare rules to rider transactions."""
+
+    tokenization_api_key = models.CharField(help_text="The Collect.js API key used for tokenization.", default="", blank=True)
+
+    registration_base_url = models.URLField(
+        help_text="The absolute base url of the MOBILEvario API instance.", default="", blank=True
+    )
```
jgravois: it seemed preferable to me to use a [URLField](https://docs.djangoproject.com/en/6.1/ref/models/fields/#urlfield) instead of relying on the helper text to try and ensure the value is prefaced with 'https'

one cool side-effect is that the existing value is displayed as a link alongside the proposed update.

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,54 @@
...
+
+    registration_username = models.CharField(
+        help_text="The username used to authenticate with MOBILEvario.",
+        default="",
+        blank=True,
+        max_length=50,
+    )
```
jgravois: [CharField](https://docs.djangoproject.com/en/6.1/ref/models/fields/#charfield) felt like a better fit here, but if there's an explicit reason why we've leaned harder on TextField historically, i'm happy to amend.

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,54 @@
...
+
+    def clean(self):
+        field_errors = {}
+
+        if self.pk and self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
+            message = "This field is required when this configuration is referenced by an active transit agency."
+            needed = dict(
+                tokenization_api_key=self.tokenization_api_key,
+            )
+            field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
```
jgravois, line 41: my thought is that we can/should wait to incorporate MOBILEvario props in model validation until we're closer to implementing something with them.

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,54 @@
...
+
+        if field_errors:
+            raise ValidationError(field_errors)
+
+    class Meta:
+        verbose_name = "INIT Config"
```
jgravois: this is just another quality of life thing to improve the display name in /admin

benefits/enrollment_init/admin.py
thekaveman: You can get this behavior (arguably the "right" way) with by using either a [`CharField`](https://docs.djangoproject.com/en/5.2/ref/models/fields/#charfield) or [`URLField`](https://docs.djangoproject.com/en/5.2/ref/models/fields/#urlfield) on the model, instead of a `TextField`.

I would advocate for the latter for the portal URL, to be the most correct.

thekaveman: LOL just seeing your notes below. Yeah I'd lean more on `Char` or `URL` over `Text` :+1:

jgravois: > Yeah I'd lean more on `Char` or `URL` over Text 👍

just to be explicit, the fields i'm overriding here are defined in `TransitProcessorConfig`, not my new model.

if y'all are game, i'd be happy to open a (different) PR to migrate them.

thekaveman: Oh, whoops. Yeah IMHO the overuse of `TextField` is not something we should continue doing. 

I'd take a PR to fix that for the parent model.

benefits/enrollment_init/models.py
lalver1: I was wondering about this and I agree, it's nice to keep _INIT_ capitalized in the admin. Extremely minor comment, maybe we can change _Config_ to _config_ to match the casing of the rest of the objects in the admin. But this is fine too, just something I noticed.

jgravois: i agree! 1edd96c27577b057443d47abfa7dad5c94889eac

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,50 @@
...
+    registration_username = models.CharField(
+        help_text="The username used to authenticate with MOBILEvario.",
+        default="",
+        blank=True,
+        max_length=50,
+    )
+
+    registration_password_secret_name = SecretNameField(
+        help_text="The name of the secret containing the password used to authenticate with MOBILEvario, "
+        "typically: [agency]-init-registration-password",
```
Scotchester: Trivial formatting nit: Can we use `# noqa: E501` here to allow a longer line and prevent this awkward string splitting?

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,50 @@
...
+        default="",
+        blank=True,
+        max_length=50,
```
Scotchester: Our custom `SecretNameField` has a `max_length` of 127. Did you intentionally reduce that here?

tests/pytest/enrollment_init/test_models.py
```diff
@@ -0,0 +1,57 @@
...
+from benefits.core.models import SecretNameField
+from benefits.enrollment_init.models import InitConfig
+
+
+@pytest.mark.django_db
+def test_InitConfig_defaults():
+    init_config = InitConfig.objects.create(environment="dev")
+
+    assert init_config.environment == "dev"
+
```
lalver1, line 17: Maybe we can add an `assert` for each field that has a `default` option. Looks like most of the asserts would look like `assert  init_config.tokenization_api_key== ""`

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,50 @@
...
+
+    def clean(self):
+        field_errors = {}
+
+        # TODO: verify the presence of an active transit agency when we're ready to associate the two
+        message = "This field is required when this configuration is referenced by an active transit agency."
+        needed = dict(
+            tokenization_api_key=self.tokenization_api_key,
+        )
+        field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
```
Scotchester: I agree with leaving out the MOBILEvario props for now, but I don't see any harm in going ahead and including the bit we have in `LittlepayConfig` and `SwitchioConfig` that checks whether this is actually associated with a transit agency. 

```suggestion
        if self.pk and self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
            message = "This field is required when this configuration is referenced by an active transit agency."
            needed = dict(
                tokenization_api_key=self.tokenization_api_key,
                # TODO: Add in MOBILEvario props
            )
            field_errors.update({k: ValidationError(message) for k, v in needed.items() if not v})
```

Scotchester (dismissed):
Looks good overall! I have a few non-blocking suggestions for your consideration.

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,50 @@
...
+
+    @property
+    def registration_password(self):
+        secret_field = self._meta.get_field("registration_password_secret_name")
+        return secret_field.secret_value(self)
+
+    def clean(self):
+        field_errors = {}
+
+        # TODO: verify the presence of an active transit agency when we're ready to associate the two
```
lalver1: I'm probably missing something, but I was wondering why we aren't adding something like

```python
if self.pk and self.transitagency_set and any([agency.active for agency in self.transitagency_set.all()]):
```

in this PR like we do for the other transit processors 🤔

lalver1 (commented):
It's looking good @jgravois! I just had a few comments/questions below.

benefits/enrollment_init/models.py
jgravois: i was thinking that it was premature only because it would obligate me to make corresponding changes in the `TransitAgency` model simultaneously. 

https://github.com/cal-itp/benefits/blob/46e70a13cd4a27248f995fc77d88d4ed2716725f/benefits/core/models/transit.py#L250

if not, i'm all for it.

Scotchester: I did a quick test and having that in place here doesn't obligate a change in `TransitAgency` yet. It will just remain the case that a `TransitAgency` can only be saved if it has selected a `LittlepayConfig` or `SwitchioConfig`, and an `InitConfig` won't be selectable until that change is made later.

jgravois: hehe. it took me longer than it ought, but i think i implemented this suggestion correctly in 1cb03de8a

benefits/enrollment_init/models.py
jgravois: i reverted this change. not sure what i was thinking.

Scotchester (dismissed):
LGTM now!

lalver1 (dismissed):
Looks good! 👍

jgravois:
thank you both. 🙏 

i'm assigning @angela-tran because i'd like her to sign off prior to merge as well.

i'll be out of the office by the time she's back so feel free to take the branch over entirely in my absence if any of you think of anything else.

benefits/enrollment_init/admin.py
jgravois: done in https://github.com/cal-itp/benefits/pull/4056. @Scotchester and i identified a whole slew of other columns to prune but I can whittle the scope back to these two specifically if that'd be preferred.

angela-tran (commented):
This overall looks good to me. Thanks to @Scotchester and @lalver1 for the great suggestions in their reviews

I have a nitpick about the field names that start with `registration_` -- since the `LittlepayConfig` and `SwitchioConfig` models include `_api_` in their field / property names, I don't see a reason to omit it here. I'll add a commit to make that change and request re-review from Scott and Luis

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,47 @@
...
-# Create your models here.
+from benefits.core.models import SecretNameField, TransitProcessorConfig
+
+
+class InitConfig(TransitProcessorConfig):
+    """Configuration for connecting to INIT, an entity that applies transit agency fare rules to rider transactions."""
+
+    tokenization_api_key = models.CharField(help_text="The Collect.js API key used for tokenization.", default="", blank=True)
+
+    registration_base_url = models.URLField(
```
angela-tran: Suggestion is for this to be `registration_api_base_url`

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,47 @@
...
+        help_text="The absolute base url of the MOBILEvario API instance.", default="", blank=True
+    )
+
+    registration_username = models.CharField(
```
angela-tran: Suggestion is for this to be `registration_api_username`

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,47 @@
...
+        help_text="The username used to authenticate with MOBILEvario.",
+        default="",
+        blank=True,
+        max_length=50,
+    )
+
+    registration_password_secret_name = SecretNameField(
```
angela-tran: Suggestion is for this to be `registration_api_password_secret_name`

benefits/enrollment_init/models.py
```diff
@@ -1,3 +1,47 @@
...
+        help_text="The name of the secret containing the password used to authenticate with MOBILEvario, typically: [agency]-init-registration-password",  # noqa: E501
+        default="",
+        blank=True,
+    )
+
+    @property
+    def registration_password(self):
```
angela-tran: Suggestion is for this to be `registration_api_password`

benefits/enrollment_init/models.py
angela-tran: Hmm, I feel the same argument could be made for the tokenization API key, so if we're already validating that, I feel like we should just go ahead and validate the MOBILEvario fields. 🤔

angela-tran (dismissed):
I read over this a few times and ended up with these requested changes (which I will also implement 😅):
- add `_api_` to the `registration_` fields / properties
- add validation of `registration_api_` fields (i.e. validate that they're not empty)
- ~~update sample fixtures file~~ - see https://github.com/cal-itp/benefits/pull/4053#issuecomment-5638483052

_edit_: on closer inspection of the sample fixtures file, I see now that it actually only has 1 [`core.transitagency` fixture](https://github.com/cal-itp/benefits/blob/f5e49e88d92480653233fb5f1747299569b5cb21/benefits/core/migrations/local_fixtures.json#L401-L418) which is configured against the `enrollment_littlepay.littlepayconfig` fixture. 

https://github.com/cal-itp/benefits/blob/f5e49e88d92480653233fb5f1747299569b5cb21/benefits/core/migrations/local_fixtures.json#L415

https://github.com/cal-itp/benefits/blob/f5e49e88d92480653233fb5f1747299569b5cb21/benefits/core/migrations/local_fixtures.json#L419-L426

There is [a `enrollment_switchio.switchioconfig` fixture](https://github.com/cal-itp/benefits/blob/f5e49e88d92480653233fb5f1747299569b5cb21/benefits/core/migrations/local_fixtures.json#L463-L471) defined, but it's not referenced by anything.



For me, this raises the question of "what is the expectation for this `local_fixtures.json` file?" Is it meant to simply be a starting point for someone who wants to quickly try out the app? Or is it supposed to comprehensively provide sample data for all supported integrations / features? I'm leaning towards the former option but only because it would allow me to ignore the cognitive dissonance that this file gives me. Really what I'm asking myself is: should we add an `enrollment_init.initconfig` fixture here, and what would the purpose of that be if it's not referenced by anything?

Scotchester:
> There is [a `enrollment_switchio.switchioconfig` fixture](https://github.com/cal-itp/benefits/blob/f5e49e88d92480653233fb5f1747299569b5cb21/benefits/core/migrations/local_fixtures.json#L463-L471) defined, but it's not referenced by anything.
> 
> For me, this raises the question of "what is the expectation for this `local_fixtures.json` file?" Is it meant to simply be a starting point for someone who wants to quickly try out the app? Or is it supposed to comprehensively provide sample data for all supported integrations / features? I'm leaning towards the former option but only because it would allow me to ignore the cognitive dissonance that this file gives me. Really what I'm asking myself is: should we add an `enrollment_init.initconfig` fixture here, and what would the purpose of that be if it's not referenced by anything?

I see some possible value in having unreferenced `SwitchioConfig` and `InitConfig` fixtures in that it makes it _somewhat_ easier to switch to testing one of those processors, but the groups will still have to be created separately, and that can be a bit annoying because this line will raise a `MultipleObjectsReturned` exception:

https://github.com/cal-itp/benefits/blob/6ef320ccc5f280d0b3b1d1ce5335e683775aefac/benefits/enrollment/views.py#L78

And furthermore, updated secrets are needed to actually test a full enrollment, so I don't really have a good answer to your question, "What is the expectation for this `local_fixtures.json` file?" It can load the homepage and present the eligibility start screen, but can't even get through eligibility with the IdG if it's not running on the special port. I suppose it provides a little bit of a version-controlled reference point for how we might want to update our private fixtures, but I'm not sure the value warrants the maintenance burden.

Maybe with the advent of our shared fixtures with secrets, we no longer really have a need for `local_fixtures.json`?

lalver1:
I think about `local_fixtures.json` as "a starting point for someone who wants to quickly try out the app" like you mentioned, Angela. So I don't think that we need to add an `enrollment_init.initconfig` fixture here. There are several other configurations that are required to be able to get to the INIT enrollment stage that are not in `local_fixtures.json`, so adding an `enrollment_init.initconfig` I feel wouldn't help much.

angela-tran:
> that can be a bit annoying because this line will raise a `MultipleObjectsReturned` exception

This is a great point to highlight @Scotchester. I think the refactor suggested in #3832 would help with this if not resolve it entirely.

> I suppose it provides a little bit of a version-controlled reference point for how we might want to update our private fixtures, but I'm not sure the value warrants the maintenance burden.

That's been the historical reason for `local_fixtures.json`, and I still think the idea behind it is good, but right, as the complexity and feature set of our app has grown, so has the maintenance burden for these sample fixtures. From what I remember from when the app was simpler, there was somewhat of an expectation that we would update these sample fixtures so that it included whatever new thing we were adding, which usually was a new method of eligibility verification, i.e. a new flow. It feels to me that we implicitly abandoned that practice long ago due to friction like what you raised above. I'm not sure yet if we should completely drop having the version-controlled `local_fixtures.json`, but I agree its purpose / value has become increasingly minimal as we continue to favor the shared fixtures in LastPass.


> I think about `local_fixtures.json` as "a starting point for someone who wants to quickly try out the app" like you mentioned, Angela. So I don't think that we need to add an `enrollment_init.initconfig` fixture here. There are several other configurations that are required to be able to get to the INIT enrollment stage that are not in `local_fixtures.json`, so adding an `enrollment_init.initconfig` I feel wouldn't help much.

Thanks @lalver1 - I will go ahead and remove adding an `enrollment_init.initconfig` from the list of todos for this PR.

benefits/enrollment_init/models.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/4053/commits/97cd4e7405f3c0bd1bfb92392909e610ab73cbda

thekaveman:
Thanks for the thoughts everyone. Just to add a little more context from my perspective:

`local_fixtures.json` has allowed us to maintain a "start the app from a `git clone`" MO. This isn't necessarily a comment on if we still _want_ to maintain this MO, but it was part of the history anyway. 

Further, at least one of our test suites (Lighthouse tests for a11y) rely on this file to get a basic app setup:

https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/.github/workflows/tests-ui.yml#L20-L23

I'd be perfectly fine to trim down the file to the absolute bare minimum to support use-cases like :point_up: and move forward with the understanding that we don't need to maintain this with new agencies, new flows, new features going forward in favor of the shared LastPass fixtures.

angela-tran:
Sorry, putting this into draft real quick to change a small thing in the tests

tests/pytest/conftest.py
```diff
@@ -205,6 +206,20 @@ def model_EnrollmentFlow_supports_sign_out(model_EnrollmentFlow):
...
+@pytest.fixture
+def model_InitConfig(model_TransitAgency):
+    init_config = InitConfig.objects.create(
+        environment=Environment.DEV,
+        tokenization_api_key="api_key",
+        registration_api_password_secret_name="agency-init-registration-password",
+    )
+
+    model_TransitAgency.transit_processor_config = init_config
+    model_TransitAgency.save()
```
angela-tran, line 218: `save` does not call `clean`, so it's fine to go ahead and assign an `InitConfig` to `TransitAgency`s in tests even though you wouldn't be able to using the actual Benefits Administrator interface.

https://docs.djangoproject.com/en/6.1/ref/models/instances/#what-happens-when-you-save

angela-tran:
This is ready for re-review

jgravois:
> I'd be perfectly fine to trim down the file to the absolute bare minimum

i agree with this (and everything else y'all said about local_fixtures.json)

> This is ready for re-review

i can't approve my own PR, but i definitely approve of what you did to this branch when your hand was in the cookie jar. 🍪

tests/pytest/conftest.py
lalver1: Related to this comment, good catch in 0938b248e1547484c8be9c174868a4cc17e4c099! I actually missed this the first time around.

From the commit message

> updating the test revealed that it is slightly different from its Littlepay counterpart, namely test_LittlepayConfig_clean, in that it calls InitConfig.clean rather than TransitAgency.clean.

I'm now wondering if the reason why the tests were different from their Littlepay counterparts was due to how the code looked like before the fix in c3140f997866b1260a4c8c2a602fbda22c26bd05. So for consistency maybe we can instead now implement the INIT tests pretty much identical to `test_LittlepayConfig_clean_inactive_agency` and `test_LittlepayConfig_clean`. This way we don't need to call `init_config.clean()`.

jgravois: you're suggesting this, correct? 👇

if so, i agree that it feels like an improvement.

```python
@pytest.mark.django_db
def test_InitConfig_clean(model_TransitAgency):
    init_config = InitConfig.objects.create(environment="dev")

    model_TransitAgency.transit_processor_config = init_config
    # clean() won't succeed until this is considered a valid TransitProcessorConfig
    model_TransitAgency.save()

    with pytest.raises(ValidationError) as e:
        init_config.clean()

    errors = e.value.error_dict

    assert len(errors) == 4
    # the error_dict contains 4 items...
    # ...
```
edit: rereading your post again i can see that you weren't. 😂 i'll keep poking.

jgravois: unless i'm missing something fundamental, the code comment i wrote in the snippet above explains what precludes us from taking the tack we took here: https://github.com/cal-itp/benefits/blob/9d8cad06ac07b25eb64ba6bbdbdd3850e6e5d56e/tests/pytest/enrollment_littlepay/test_models.py#L44-L57

angela-tran: I think @lalver1 is suggesting that we change

```python
 with pytest.raises(ValidationError) as e:
        init_config.clean()
```

to 

```python
 with pytest.raises(ValidationError) as e:
        model_TransitAgency.clean()
```

so that the test follows the same pattern as the tests for [Littlepay](https://github.com/cal-itp/benefits/blob/9d8cad06ac07b25eb64ba6bbdbdd3850e6e5d56e/tests/pytest/enrollment_littlepay/test_models.py#L56-L57) and [Switchio](https://github.com/cal-itp/benefits/blob/main/tests/pytest/enrollment_switchio/test_models.py#L56-L57)

jgravois: yup. and unless i'm missing something:
1. if the agency is `inactive` the InitConfig won't consider its own fields mandatory
1. if the agency is `active`, its own validation will disallow an `InitConfig` being assigned.

angela-tran: @lalver1 -- I think John is right that it's still the case that `TransitAgency.clean` won't succeed here. ~~His changes in https://github.com/cal-itp/benefits/commit/c3140f997866b1260a4c8c2a602fbda22c26bd05 to more fully implement `InitConfig.clean` doesn't allow us to call `model_TransitAgency.clean` in the test.~~ Still thinking over all this to make sure this is correct :thinking:

lalver1: Right, but by changing using `init_config.clean()` to `model_TransitAgency.clean()`, it may be necessary to add a `test_InitConfig_clean_inactive_agency` test and modify `test_InitConfig_clean` test. This will probably give us test coverage for the behavior you mentioned above @jgravois and will follow the same pattern as the Littlepay and Switchio tests. 

We're getting coverage right now, so my original comment may not be all that important, but it just felt like the tests for Littlepay, Switchio, and INIT could be almost identical, and this would be nice for consistency but not required (especially if I'm missing something).

angela-tran: Ok, I think I follow what you're suggesting @lalver1. If we were to do

```python
 with pytest.raises(ValidationError) as e:
        model_TransitAgency.clean()
```
the test will read the same as the ones for Littlepay and Switchio but of course will behave differently. This is because `TransitAgency.clean` has (intentionally) not been updated to expect the possibility of having an `InitConfig` as its transit processor config and to inspect such a model. So the `ValidationError` that the test would see would be the one from 

https://github.com/cal-itp/benefits/blob/48d83a395670c0e392070d4adceb6e0a62f97d5b/benefits/core/models/transit.py#L250-L251

instead of what we ultimately will want to assert, which is something like `"INIT configuration is missing fields..."` (e.g. the `ValidationError` for Littlepay - https://github.com/cal-itp/benefits/blob/48d83a395670c0e392070d4adceb6e0a62f97d5b/benefits/core/models/transit.py#L256-L259)

The assertions about the `error_dict` will reflect this difference and can be updated when we do add `InitConfig` validation to `TransitAgency.clean`. Did I more or less get it right?

jgravois: that's my interpretation of @lalver1's suggestion too. i'm not opposed to setting up the tests that way, but it would result in a (temporary) coverage gap.

Screenshot 2026-09-14 at 12 14 27 PM

angela-tran: The thought did cross my mind earlier of "wait why are the Littlepay and Switchio tests calling `model_TransitAgency.clean`? It seems more unit-test-y to call the `LittlepayConfig`/`SwitchioConfig`/`InitConfig`'s `clean`", which is in line with your point about losing the coverage @jgravois

angela-tran: I think it'd be good to make them consistent one way or another. Which way are y'all leaning?

jgravois: i think it makes sense to leave as-is here and followup afterwards to make the other two consistent.

lalver1: Ah, I see that I was missing the part about `TransitAgency.clean` has (intentionally) not been updated, and how this affects the tests. Thanks for pointing this out y'all, and sorry for overlooking it!

I agree that we can leave as-is and then make the tests for the other 2 processors consistent 👍

tests/pytest/enrollment_init/test_admin.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@pytest.mark.django_db
+class TestInitConfigAdmin:
+    @pytest.fixture(autouse=True)
+    def init(self):
+        self.model_admin = InitConfigAdmin(models.InitConfig, admin.site)
+
+    def test_permissions_mixin(self):
+        assert isinstance(self.model_admin, StaffPermissionMixin)
```
angela-tran, line 16: I guess because we were talking about the tests so much 😂, I realized we didn't have one for the `InitConfigAdmin` added in this PR and so added this test

angela-tran:
So I think we've agreed to leave `test_InitConfig_clean` as-is (https://github.com/cal-itp/benefits/pull/4053#discussion_r4009051693) to be addressed later (I can make a GH issue for it), and then I added a small test / file that I realized was missing.

Ready for re-review from devs who are not me or @jgravois :smile:

lalver1 (approved):
Looks great @jgravois and @angela-tran! And thanks again for the explanation regarding the feedback I had 🙏

tests/pytest/conftest.py
jgravois: > make the tests for the other 2 processors consistent

when i went to implement this in #4093 i realized that we're calling `TransitAgency.clean()` instead of `TransitConfig.clean()` because it gives us code coverage for both pieces of validation in one fell swoop. ⚡ 

it makes sense to me to reverse course on what i said yesterday and test `TransitAgency.clean()` consistently, but if either of y'all are interested in chatting about any of the alternative options below, i'm game for that too.

* test both `TransitConfig.clean()` and `TransitAgency.clean()` in `tests/pytest/enrollment_[provider]/test_models.py`
* test `TransitConfig.clean()` in `tests/pytest/enrollment_[provider]/test_models.py` and test `TransitAgency.clean()` in `tests/pytest/core/models/test_transit.py`

Issue #3193, opened 2025-09-19
Santa Cruz: Transit provider live in Cal-ITP Benefits `production` environment


indexing:
## Acceptance criteria

_The following configuration always happens in the Benefits `production` environment._

@angela-tran's tasks:
- [x] Create `TransitAgency` instance
- [x] Create `LittlepayConfig` instance
  - [x] Mark the environment as `production`
  - [x] Use API credentials for the agency's Littlepay `production` instance
- [x] Set the `TransitAgency`'s transit processor to the `LittlepayConfig` you just created
- [x] Create `LittlepayGroup` instance(s) using group ID(s) from the agency's Littlepay `production` instance

Account manager task:
- [x] Return to `TransitAgency` instance and check the **Active** box

angela-tran:
`prod` Benefits is now set up with an inactive Santa Cruz METRO `TransitAgency` that has all its needed Littlepay config in place.

It is ready to be set as `active` by the Account Manager whenever appropriate to do so. I'll assign @o-ram as the AM on this ticket since she's covering for Nina right now.

angela-tran:
Paused until August - see https://github.com/cal-itp/benefits/issues/3187#issuecomment-4871145803

indexing:
Per this comment (https://github.com/cal-itp/benefits/issues/3187#issuecomment-5360250738), I'm moving this issue forward to the September sprint.

jgravois:
from the looks of it @ninadinh-la can tick the last box off the list here and close this sucker out! 🚀

indexing:
Please do log the activation here @ninadinh-la . I'll reference it when I update the adoption table.

ninadinh-la:
Thanks all for taking care of tracking^ My notifications are limited on GitHub, will look at that soon to review so I dont miss these pings ;-) Closing now, over and out!

PR #4062, opened 2026-09-08
docs: marks scmetro LIVE

indexing:
Per https://github.com/cal-itp/benefits/issues/3193#issuecomment-5626282584


jgravois (approved):
🍔

indexing:
@jgravois As discussed last week, I added agency card designations for agencies that recently launched with in-person agency card options. I'd like to differentiate enrollment pathways based on method eventually, but the table is already not doing that well. For example, all agencies that offer enrollment for Veterans currently only have that option for self-service. So I believe the current version of the table is now accurate if not precise.

PR #4068, opened 2026-09-09
Tell robots not to index the app if it's not the production hostname

Scotchester:
Part of cal-itp/calitp.org#577

---

Adds an NGINX directive that sets the `X-Robots-Tag` header on all requests. If the hostname is `benefits.calitp.org`, the header value is `allow`, otherwise it defaults to `none` (equivalent to `noindex, nofollow`). The NGINX `map` directive is used to set a variable dynamically based on the hostname, which is then injected into the `add_header` directive.

References:

- [`X-Robots-Tag` at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Robots-Tag)
- [NGINX `map` module docs](https://nginx.org/en/docs/http/ngx_http_map_module.html)
- [NGINX `add_header` module docs](https://nginx.org/en/docs/http/ngx_http_headers_module.html)

### Testing instructions

1. Rebuild your app container with `bin/build.sh`.
2. Rebuild the dev container without cache.
3. Instead of launching the Django `runserver` process via the VS Code Run and Debug feature, open a shell on the dev container and run `bin/start.sh` to start up nginx.
4. Access the site with the usual `localhost:port` and check the response headers for the page request.

### DIFF
--- a/appcontainer/nginx.conf
+++ b/appcontainer/nginx.conf
@@ -96,4 +96,13 @@ http {
       include /calitp/run/proxy.conf;
     }
   }
+
+  map $http_host $allow_indexing {
+    hostnames;
+    default             none;
+
+    benefits.calitp.org all;
+  }
+
+  add_header X-Robots-Tag $allow_indexing;
 }


appcontainer/nginx.conf
```diff
@@ -96,4 +96,13 @@ http {
       include /calitp/run/proxy.conf;
     }
   }
+
+  map $http_host $allow_indexing {
+    hostnames;
+    default             none;
+
+    benefits.calitp.org allow;
```
lalver1: Should this instead be `benefits.calitp.org all;` per the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Robots-Tag#all)?

lalver1 (commented):
Locally it looks good to me! I tested using the `client` Compose service and I verified that the `X-Robots-Tag: none` header was added to all the responses 👍. Not that it matters, but it's interesting that the first letter of a header name is uppercased locally but is lowercased when the app runs on Azure.

I just had one question below about the directive we are using.

angela-tran (changes requested):
Thanks for including those helpful reference links @Scotchester 

I noticed that `allow` isn't listed as a [directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Robots-Tag#directives) on the MDN docs for `X-Robots-Tag`. I asked Gemini as well and was told that it's not a valid value. 

image


I'm not sure what behavior it would cause, but regardless, I think we should either use a valid directive or modify the NGINX config so that we don't add the header if it's prod.

Scotchester:
Dangit, yes, you're both correct. I should have used `all` there. Thanks for catching that!

thekaveman (approved):
Very nice solution, @Scotchester 👍

lalver1 (approved):
Awesome! It was really nice to see how to do this using nginx 👍

angela-tran (approved):
🎉

PR #4069, opened 2026-09-10
Add environment banner to the Admin in non-production envs

### LINKED ISSUE #3920: Create environment banner for `dev`/ `test` in the Admin
> ### Notes taken while working in Benefits Admin (`prod` environment)
>...
> 
> _(A nice-to-have: as I moved over to `prod` Benefits to do basically the exact same steps as I had just done in `test`, I was thinking it might be nice to have a big banner at the top of Benefits Admin if it's for `dev` or `test`. That way, you know that if you don't see a big banner at the top, you are in `prod` and should be careful. It's kind of inverted I guess, but Login.gov follows that same pattern for its sandbox environment.)_ 

 _Originally posted by @angela-tran in [#3881](https://github.com/cal-itp/benefits/issues/3881#issuecomment-5122468474)_

Scotchester:
Closes #3920 

---

- Adds new context processor `runtime_env` to include the value of `settings.RUNTIME_ENVIRONMENT()` in the context
- Adds new context processor `is_prod` to check whether we're in the production environment or not
- Adds a new `env_banner.html` template that extends `alert-bar.html` to display "This is the `<ENV>` environment."
- Includes `env_banner.html` at the top of the Admin `base_site.html` template when `is_prod` is `False`

#### Login screen

The login screen with a red banner saying 'This is the LOCAL environment.'

#### Main dashboard

The main dashboard with a red banner saying 'This is the LOCAL environment.'


benefits/templates/admin/base_site.html
```diff
@@ -20,6 +20,13 @@
   <link href="https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;500;700&display=swap" rel="stylesheet">
 {% endblock extrahead %}
 
+{% block header %}
+  {% if debug %}
```
angela-tran, line 24: I think this admin banner would be more helpful if we always showed it for all environments except for production. So rather than checking for `debug` here, could we check if the `runtime_env` is not `RUNTIME_ENVS.PROD`?

Scotchester: You know what, I had it in my head that `debug` was enabled for the `test` environment, but taking another look, I see that it's not!

This makes sense, since @indexing uses it for demos, and the big debug banner we see on `dev`/local would be pretty intrusive. I suppose I'd like to ask you, Andy, if you ever demo the Admin side for in-person enrollments, and whether such a banner would feel disruptive in that case.

indexing: I do indeed demo in-person enrollment from time to time, but I'm fine with this banner. I typically verbally alert people that I am in the `test` environment. This banner will make that crystal clear.

Scotchester: Thank you, Andy! @angela-tran, I will proceed with your suggested change.

Scotchester: @angela-tran Do you think it's fine to hardcode `runtime_env != "prod"`, or would you prefer to get that value into a context var?

angela-tran: Hmm, yeah decisions... I think it'd be best to avoid hard-coding `"prod"` in the template, so my instinct is that we should change something in the context processor. I don't have strong feelings on whether that is to add a `"prod": settings.RUNTIME_ENVS.PROD` context var or to replace `runtime_env` with `is_prod` or something else.

Scotchester: I like the idea of making an `is_prod` var :+1:

thekaveman: That lines up with our Terraform as well, so I second (third??) that idea 👍 

https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/terraform/environment.tf#L2

Scotchester:
`is_prod` check is now implemented! Reopening for review.

benefits/templates/admin/base_site.html
jgravois: if only NGINX could access the context too 😂

jgravois (approved):
neat! 🍔

(just need to update your PR description to explain that you aren't checking 'debug' anymore)

PR #4071, opened 2026-09-11
Research: MMW multi-language selector

### LINKED ISSUE #3957: Metro Mobility Wallet: scaffold multi-language selector
The Metro Mobility Wallet feature will bring additional language support beyond the existing two (English, Spanish). Thus we need to implement a more comprehensive language selector, since our current binary button will no longer suffice.

@cmajel has indicated they will largely base this component design on the excellent guidance and prior art from the USWDS.

Even ahead of a confirmed design, we can start spiking on how this component might function within the Metro Mobility Wallet app.

**Note:** we are _not_ attempting to bring this component into the general Benefits app/user flow. This should be restricted to the Metro Mobility Wallet work only at this time.

## Acceptance Criteria

- [ ] We have a general sense of how to implement a multi-language selector component in the front-end
- [ ] We have a general sense of how to configure and swap between more than two languages on the back-end
- [ ] The solution for the multi-language selector has no impact on the existing, English/Spanish only selector, for the "main" Benefits app

## Additional context

[USWDS Language selector component for three or more languages](https://designsystem.digital.gov/components/language-selector/#language-selector-with-three-or-more-languages).

lalver1:
Closes #3957 (but will not be merged)

## Background

Django's `i18n` framework is configured in [Benefits' settings](https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/benefits/settings.py#L245) and applies to all Django apps in the project. The relevant settings for this PR, with their current Benefits configuration, are listed below:

- `LANGUAGE_CODE = "en"`: [Provides the default/fallback language](https://docs.djangoproject.com/en/6.1/ref/settings/#language-code)
- `LANGUAGES = [("en", "English"), ("es", "Español")]`: [Used to restrict language selection for Benefits to a subset of the Django-provided languages](https://docs.djangoproject.com/en/6.1/ref/settings/#languages)
- `LOCALE_PATHS = [os.path.join(BASE_DIR, "benefits", "locale")]`: [A list of directories where Django look for a translation](https://docs.djangoproject.com/en/6.1/ref/settings/#std-setting-LOCALE_PATHS)

Maybe the simplest solution to implementing the multi-language selector (and keeping Benefits' and MMW's translation behavior separate) would be to simply modify the [`LANGUAGES` setting](https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/benefits/settings.py#L256) conditionally based on if we are on a Benefits page or a MMW page. But, [since settings should not be modified at runtime](https://docs.djangoproject.com/en/6.1/topics/settings/#altering-settings-at-runtime), this stops us from using this straightforward implementation.

On the front-end, most of the translation functionality is covered by the template tags

- `{% load i18n %}`
- `{% get_current_language as LANGUAGE_CODE %}`
- `{% get_available_languages as LANGUAGES %}`
- `{% get_language_info %}`

as seen in the [lang-selector.html template](https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/benefits/core/templates/core/includes/lang-selector.html).

## MMW Prerequisites

- A list of the MMW-specific languages stored in `metro_mobility_wallet`
- A `locale` folder in `metro_mobility_wallet` containing a folder [following the locale name notation](https://docs.djangoproject.com/en/6.1/topics/i18n/translation/#how-django-discovers-translations) for each MMW-specific language. This means there's no need to modify `LOCALE_PATHS` since [Django looks for a `locale` directory in each of the installed Django apps](https://docs.djangoproject.com/en/6.1/topics/i18n/translation/#how-django-discovers-translations) so we keep a clean separation between Benefits and MMW.
- A `makemessages.sh` helper script similar to [`benefits/bin/makemessages.sh`](https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/bin/makemessages.sh)

## ~Solution 1: Add a custom template tag for MMW~

The idea was to define a new template tag in `metro_mobility_wallet` called `get_metro_languages` used to get the MMW-specific languages. Then we would replace `{% get_available_languages as LANGUAGES %}` with `{% get_metro_languages as LANGUAGES %}`.

**Problem:** On the front end, a form rendered correctly for each language in the `LANGUAGES` template variable. However, only the English and Spanish translated text worked as expected. Clicking on any other language would set the current language to English (the default). The reason was that MMW-specific languages are not in `settings.LANGUAGES`; and Django's i18n `set_language` view (which is POSTed to when the user needs to change languages) uses `settings.LANGUAGES`.

**Possible solution:** We could define a view similar to Django's built-in `set_language` that uses the MMW-specific language list for `metro_mobility_wallet` but we'd have to make sure we implement everything Django's `i18n` framework expects, and this didn't feel right.

### Back-end

Since Django's i18n `set_language` view uses `settings.LANGUAGES`, we
add all languages globally here (by extending the current list with MMW-specific languages). Then we create a template tag for Benefits' language selector so that it is only aware of English and Spanish. Across all the codebase, this is only used in [`lang-selector.html`](https://github.com/cal-itp/benefits/blob/324f59ebd90818f9727ccae957c69c49cf50a96a/benefits/core/templates/core/includes/lang-selector.html). The `MMWCurrentLanguage` middleware is added to protect the English/Spanish button if a user is on a MMW page and then goes to a Benefits page with a non-Benefits language selected. Without this middleware, the English/Spanish button gets messed up.

These scenarios all work as expected

1. Start in Benefits, English/Español translation and button works
2. Start in Benefits, go to MMW, language selection is preserved from Benefits since MMW includes English and Spanish
3. Start in MMW, select English or Spanish, go to Benefits, language selection is preserved and translation and button works
4. Start in MMW, select a language other than English or Spanish, go to Benefits, language is set to English (default) and translation and button works

### Front-end

Currently, a form for each language is rendered with no styling (it looks terrible 😅). However, we have all the data we need in the template to render the component and the sample string is being successfully translated and rendered.

### Loose ends

- It would be nice for the middleware to not run when getting resources like `/media/agencies/cst.png`. From the logs, it looks like at least the new `ResetUnsupportedLanguage` middleware is running in these situations.
- See if the workflow `check-migrations-and-messages` needs to be updated
- Get around permissions issues when running `makemessages.sh`


lalver1:
This is ready for an initial review. Although some parts can probably be more DRY and naming can be improved, the overall strategy is ready.

Most of the files changed are new `.po` files, the actual refactor is spread across a few files only. I'm still having some issues with file permissions and being able to completely run `makemessages.sh` but this shouldn't block us from an initial review. The front-end styling is also not included but all the data that is needed is available in the template.

benefits/settings.py
```diff
@@ -255,7 +257,7 @@ def RUNTIME_ENVIRONMENT():
 LANGUAGE_COOKIE_SAMESITE = "Lax"
 LANGUAGE_COOKIE_SECURE = True
 
-LANGUAGES = [("en", "English"), ("es", "Español")]
+LANGUAGES = [("en", "English"), ("es", "Español")] + metro_conf.LANGUAGES
```
jgravois, line 260: forgive me if this is a dumb question, but what would be the drawback to:

1. declaring the exhaustive list for the entire application
1. generating new .po files for all existing content (without translations for now)
1. using runtime logic in benefits to avoid displaying the dropdown and only allow toggling between english and spanish

in my mind, it'd result in a lot of unused boilerplate but it would simplify the business logic significantly and set us up for the future to add new languages to benefits one at a time anytime we're ready...

lalver1: Not at all, thanks for the good question @jgravois.

In general, this PR has taken the approach of keeping translation-related code as separate as possible for Benefits and MMW, with the idea that it should be easy to remove MMW code once the MMW sign-up period is over. That's why we have, for example, a separate `metro_conf.LANGUAGES` variable, a separate `locale` folder, etc. I agree that treating these components instead in a more "app-wide" manner would in some parts simplify the business logic.

I feel that the new middleware and template tag are in a sense doing the work of item 3 though. First the middleware checks where the request is going to and a template tag filters the languages for the Benefits application. This may feel a bit hacky because we are essentially stretching the use of Django's i18n framework to handle two different language sets within the same Django project. Seeing how the framework is configured has led me to believe that it is meant to behave the same for all apps in a Django project (this is especially clear where Django's `set_language` view uses `settings.LANGUAGES`), but in our case we need to work with apps that use different language sets in the same project.

I definitely welcome more feedback though in case I'm missing anything!

jgravois: i appreciate the additional explanation. 💟 

> Seeing how the framework is configured has led me to believe that it is meant to behave the same for all apps in a Django project (this is especially clear where Django's `set_language` view uses `settings.LANGUAGES`)

i think that's the crux of it. our philosophical goal is to sequester as much of the MMW work as possible. in my mind both the approach you outlined in this PR and my hand-wavey alternative represent bleed into the app at large. 

perhaps your proposal adheres to our philosophical goal a little more rigorously and mine is more 'with the grain' for Django itself? is that a fair assessment?

at the end of the day i'm comfortable with either option. i don't think we're setting ourselves up for problems reverting either one.

thekaveman: > perhaps your proposal adheres to our philosophical goal a little more rigorously and mine is more 'with the grain' for Django itself? is that a fair assessment?

I think this is probably accurate John. I also don't _necessarily_ see a problem with what you are proposing.

I wouldn't want us to hew so far into the philosophy of "making it easy to rip out MMW when the signup period is over", that we have to reinvent a bunch of Django wheels along the way. That doesn't feel like the right place to be spending our effort.

On the other hand, and to John's suggestion/point (and this PR's current implementation), either way ("in the grain" vs. "in the philosophy"), we still need to be able to ensure our standard Benefits app language selector only displays (and allows switching between) English and Spanish, right? 

- Is the easiest way to do that a custom template tag as shown here?
- Does it make the main app less messy to keep a separate `locale` directory inside MMW? 
  - _I'm thinking of, e.g., does doing this prevent us from having a huge number of placeholders for all the standard app strings that aren't going to be translated for MMW?_

benefits/core/middleware.py
```diff
@@ -148,3 +149,20 @@ def process_request(self, request):
...
+class ResetUnsupportedLanguage(MiddlewareMixin):
+    """
+    Middleware reverts active language to English when a Benefits view is accessed with a MMW language active
+    and which is not a Benefits' supported language.
+    """
+
+    def process_request(self, request):
+        referer = request.META.get("HTTP_REFERER", "")
+        is_a_mmw_page = "metro-mobility-wallet" in request.path or "metro-mobility-wallet" in referer
+        if not is_a_mmw_page:
```
Scotchester, line 163: Question: What is the purpose of checking the referrer and considering it an MMW page if it contains `metro-mobility-wallet`?

I ask because if I add a link to the MMW homepage to go back to Benefits and click it, `is_a_mmw_page` is True, and the block below is not executing, resulting in this semi-broken state:

Benefits homepage with debug bar showing that the lang value is still zh-hans and both English and Spanish buttons are showing

Going directly to `/` via the browser address bar after changing the language on MMW _does_ trigger this as expected.

Scotchester (commented):
This is a great start, Luis! The overall approach seems sound to me.

I wasn't seeing any translated paragraphs on the MMW homepage, so I thought I'd try `makemessages` and see if that got it working.

Not yet realizing that you created a new `makemessages.sh` script specifically for MMW, I tried running the regular one in `bin/` and I wanted to note that it also touches the MMW English and Spanish PO files, so we'll want to figure out how to exclude those when that original helper is run.

Then I ran `benefits/metro_mobility_wallet/makemessages.sh` (no permissions problem) and it had some interesting behavior:

1. The console output only showed:
   ```
   processing locale es
   processing locale en
   ```
   despite it actually updating all of the languages.
2. The `POT-Creation_date` was updated _only_ on the English and Spanish PO files.
3. The `msgstr` in the Spanish PO was converted to a multiline format, but not the ones in any other language, despite them also having long lines.
4. In all PO files _except_ English and Spanish, the `Report-Msgid-Bugs-To` value had another `https://github.com/cal-itp/benefits/issues` appended to it. This appears to happen every time the MMW-specific `makemessages.sh` is run.
   - I can't see why it doesn't happen to the MMW English and Spanish PO files. These are actually having that value stripped entirely.
   - I don't know why the behavior is different for the original `bin/makemessages.sh`.

There is some other interesting stuff comparing the main Benefits Spanish PO to the MMW Spanish PO:

Diff of the top of the two Spanish PO files

- Español vs. Spanish
- 2021 vs 2023
- Fields not present in original:
  - `Project-Id-Version`
  - `PO-Revision-Date`
  - `Last-Translator`
  - `Language-Team`
- Different `Plural-Forms` config

I don't have deep knowledge of Django's translation system, but all these differences are pretty striking to me, given that `manage.py makemessages` seems to be invoked in the same way in both places.

angela-tran:
Reviewing this now :eyes:

lalver1:
Thanks @angela-tran, I'm about to push a few more commits to simplify updating message files (and removing some other unnecessary things).

lalver1:
Thanks y'all for the reviews and suggestions. I'm going to try to cover the updates I've made since opening this PR up for review:

## Translation framework

- Removed the unnecessary `HTTP_REFERER` check per Scott's [comment](https://github.com/cal-itp/benefits/pull/4071#discussion_r4030112168).
- Per John's and Kegan's comments, moved the list of MMW languages into `benefits/settings.py` to avoid importing app-level settings into project-level settings.
  - Kegan had a nice idea on how to do something more dynamic, like only adding MMW languages if the MMW app is activated. The idea was to use `AppConfig.ready()` in MMW's `apps.py` to override `settings.LANGUAGE` by doing something like 
  
    ```python
    LANGUAGES = [("en", "English"), ("es", "Español")] + LANGUAGES_METRO
    settings.configure(LANGUAGES=LANGUAGES)
    ```
  
    inside `ready()`, but I got a `RuntimeError: Settings already configured.` It looks like when `ready()` runs, the settings have already been loaded so we are not allowed to change them. So I had no choice but declaring the extra MMW languages directly in `benefits/settings.py`.

## Making message files

- Removed unnecessary metadata from the message files to match only the fields we use in Benefits' message files.
- Consolidated making message files into the current `bin/makemessages.sh` script so that we don't have to run a different one depending on what we are working on.
  - Processes Benefits message files by using the `--ignore` flag
  - Processes MMW message files (note that it changes directory and uses `django-admin` instead of `python manage.py` so that only message files in `benefits/metro_mobility_wallet/locale/` are updated)
  - Does the `sed` modification for Benefits and MMW
- Updated the GH `Check for up-to-date Django migrations and messages` workflow since the check changed due to the extra languages

## Trying out the changes

### Copy updates for Benefits

In a Benefits template, make a change to any string enclosed in a `{% translate %}` tag. Like the [Benefits' docs mention](https://docs.calitp.org/benefits/explanation/development/i18n/#updating-message-files), run `bin/makemessages.sh`. You will see that the `.po` files in `benefits/locale` have a new timestamp and the translation string was updated. You will also see that the `.po` files in `benefits/metro_mobility_wallet/locale` have a new timestamp, but none of the `.po` files had any translations leak into them from the change made in Benefits. Run `python manage.py compilemessages`, start your debugger, and confirm the translation renders properly.

### Copy updates for MMW

In a MMW template, make a change to any string enclosed in a `{% translate %}` tag. Like the [Benefits' docs mention](https://docs.calitp.org/benefits/explanation/development/i18n/#updating-message-files), run `bin/makemessages.sh`. You will see that the `.po` files in `benefits/metro_mobility_wallet/locale` have a new timestamp and the translation string was updated. You will also see that the `.po` files in `benefits/locale` have a new timestamp, but none of the `.po` files had any translations leak into them from the change made in MMW. Run `python manage.py compilemessages`, start your debugger, and confirm the translation renders properly.

## For further discussion

- For the case where a user is in a MMW page translated to a language other than English or Spanish, and then navigates to a Benefits page, this PR switches the user's language to the default (English) so that Benefits' pages still work. Should it even be possible for a user to go from MMW to Benefits? Maybe we just redirect the user instead of switching to the default language? @indexing and @cmajel (and anyone else 🙂), please let us know if you have any thoughts about this situation.

benefits/metro_mobility_wallet/locale/es/formats.py
```diff
@@ -0,0 +1,2 @@
+# https://docs.djangoproject.com/en/stable/ref/templates/builtins/#date-and-time-formatting-specifiers
+DATE_FORMAT = "F j, Y"
```
jgravois: since all these custom formats are identical and we aren't actually registering the module where they reside, perhaps we'd be better of just omitting them altogether until/unless we find that we need more fine-grained control than Django's built-in formats provide?

https://github.com/cal-itp/benefits/blob/19694a15f98e1724265f64947f7e4d1a13b5c019/benefits/settings.py#L271-L274

angela-tran:
I'm currently seeing that the `/metro-mobility-wallet` page is not getting translated. :thinking: 

It looks like the sequence of requests and logic when I switch languages via one of the buttons is:
- request to `/i18n/setlang` -> `is_a_mmw_page` is False -> middleware activates language to `English`, and then the view sets the language cookie to whatever I selected
-  request to `/metro-mobility-wallet` -> `is_a_mmw_page` is True
- request to `/favicon.ico` -> `is_a_mmw_page` is False -> middleware activates language to `English` and finally the page is rendered in English.

I'm still testing to get a better understanding of the details and to see if I'm doing something wrong.

Scotchester:
Yeah, as with my previous look at this, I'm still not seeing the translated paragraph on the MMW homepage, either. That was an interesting hypothesis, Angela, but I agree with your edit that that can't be right.

I **do** see the `lang` attribute getting updated on the `<html>` element, and the way that Firefox renders the default font changes, but the text itself is not changing.

lalver1:
I remember Scott mentioned that he wasn't seeing the translated paragraph in the MMW page too and I thought that maybe it was related to missing the `manage.py compilemessages` step since I didn't mention running it before. I'm taking a look at this again because I see the text get translated on my browser but there may be something else I'm missing.

Scotchester:
> I thought that maybe it was related to missing the manage.py compilemessages step since I didn't mention running it before

Ohhhh snap, that was it. I overlooked that in your new testing instructions 🤦 I see the translations now.

Now I need to go refresh my memory on how that command works.

Your other updates look good. Unfortunate that we can't exclude the unneeded PO timestamp updates in the app that hasn't changed, but I think it's a reasonable thing to deal with in exchange for not having a separate `makemessages.sh`.

angela-tran:
I had the same thought right before your comment @lalver1 that I need to run `compilemessages`. That used to be run as a part of `./bin/init.sh` when that was a thing, i.e. in the SQLite days. Trying it out now

_update_: I see the translation working now :+1:

lalver1:
Nice! I'm glad the translated text is working 👍 Yeah, sorry I didn't mention `manage.py compilemessages` originally. Now that command only runs automatically in `start.sh` (so only when the appcontainer is built), we need to run it manually in the dev container.

benefits/metro_mobility_wallet/locale/es/formats.py
lalver1: I completely agree John, thanks for catching this. I _think_ these files got created automatically when I first ran `python ../../manage.py makemessages -l en -l es -l "zh_Hans" -l "zh_Hant" -l ko -l ja -l th -l vi -l ru -l hy  --no-obsolete --no-location` and I didn't notice that all the formats were identical and that they're not even registered in `FORMAT_MODULE_PATH`.

Yep, when we implement the MMW templates and copy, we can remove these if there's no need for them.

benefits/core/templatetags/benefits_languages.py
```diff
@@ -0,0 +1,22 @@
...
+    Usage::
+
+        {% get_benefits_languages as LANGUAGES %}
+        {% for language in LANGUAGES %}
+        ...
+        {% endfor %}
+
+    This puts English and Spanish only into the named variable.
+    """
+    return [("en", "English"), ("es", "Español")]
```
thekaveman: I still think this can be a distinct variable in `settings`, so it doesn't have to be hard-coded anywhere else, right?

benefits/core/middleware.py
```diff
@@ -148,3 +149,19 @@ def process_request(self, request):
...
+    Middleware reverts active language to English when a Benefits view is accessed with a MMW language active
+    and which is not a Benefits' supported language.
+    """
+
+    def process_request(self, request):
+        is_a_mmw_page = "metro-mobility-wallet" in request.path
+        if not is_a_mmw_page:
+            logger.debug(f"MMW check triggered for path: {request.path}")
+            current_lang = translation.get_language()
+            if current_lang not in ["en", "es"]:
```
thekaveman: Another place where the distinct `settings` variable would be useful

benefits/settings.py
```diff
@@ -255,7 +256,18 @@ def RUNTIME_ENVIRONMENT():
...
+    ("zh-hant", "Chinese traditional"),
+    ("ko", "Korean"),
+    ("ja", "Japanese"),
+    ("vi", "Vietnamese"),
+    ("th", "Thai"),
+    ("ru", "Russian"),
+    ("hy", "Armenian"),
+]
+
+LANGUAGES = [("en", "English"), ("es", "Español")] + LANGUAGES_METRO
```
thekaveman: In reference to my other comments, it would be preferable here to say:

```python
LANGUAGES_CORE = [("en", "English"), ("es", "Español")]
LANGUAGES_METRO = [...]

LANGUAGES = LANGUAGES_CORE + LANGUAGES_METRO
```

Then elsewhere you can do:

```python
from django.conf import settings

# use settings.LANGUAGES_CORE instead of hard-coding
```

thekaveman (changes requested):

angela-tran (commented):
This approach looks good to me @lalver1, nice work!

I agree with @thekaveman's comments that there are a few values hard-coded around that should be coming from a more centralized / canonical source.

I take it that this PR was more for research and laying out the high-level approach, so maybe it's assumed / implied we wouldn't actually have the `[en, es]` list hard-coded in so many places. Even so, it's good to note it explicitly and work out what the canonical source will look like ahead of the actual implementation that is to be merged.

Scotchester:
Personally, if we're happy with this approach, I don't see any reason not to merge this (once Kegan's comments are resolved).

cmajel:
> For the case where a user is in a MMW page translated to a language other than English or Spanish, and then navigates to a Benefits page, this PR switches the user's language to the default (English) so that Benefits' pages still work.

> Should it even be possible for a user to go from MMW to Benefits?

We are not planning for there to be any links between MMW and Benefits, and are treating them as distinct products from the user's perspective (esp since Metro isn't on Benefits). The only reason I could see us reconsidering is if we heard a need for more Cal-ITP brand recognition, and I don't think that is likely. 

If a person is using a language only supported by MMW, resetting to English if they visited Benefits feels like a reasonable approach.

benefits/settings.py
lalver1: For sure, thanks for reminding me about this part Kegan, I made a note to make this change and somehow overlooked it 🤦‍♂️

lalver1:
Alright, I think this is ready for a final review, thanks for all the comments!

I made a few more changes, especially since we decided to merge this PR:
- Use a centralized language list
- Cleanups (remove the `formats.py` files for MMW and the initial `.md` file that had the list of languages and some other notes)
- Add a test for the `ResetUnsupportedLanguage` middleware

I also tested switching to this branch locally and rebuilding and running the `client` service to verify that the startup scripts indeed compile the message files. Both Benefits and MMW looked fine in terms of translation behavior.

thekaveman:
> > For the case where a user is in a MMW page translated to a language other than English or Spanish, and then navigates to a Benefits page, this PR switches the user's language to the default (English) so that Benefits' pages still work.
> 
> > Should it even be possible for a user to go from MMW to Benefits?
> 
> We are not planning for there to be any links between MMW and Benefits, and are treating them as distinct products from the user's perspective (esp since Metro isn't on Benefits). The only reason I could see us reconsidering is if we heard a need for more Cal-ITP brand recognition, and I don't think that is likely.
> 
> If a person is using a language only supported by MMW, resetting to English if they visited Benefits feels like a reasonable approach.

@cmajel one idea the engineers discussed (which we can address in a follow-up if warranted) is to redirect users back to the MMW homepage if they leave MMW and try to get to standard Benefits. Totally up to you and @indexing if this is the UX we want or not.

Issue #4083, opened 2026-09-15
INIT integration: implement Enrollment Index for `enrollment_init` app


angela-tran:
This task is to implement the INIT version of the Enrollment Index.

## Acceptance Criteria

- [ ] A class-based view called `IndexView` is defined in `enrollment_init/views.py` and provides the `context` needed for the template to render its contents (which includes a button that links to the card details form page)
- [ ] `enrollment_init/urls.py` is updated so that this view is reachable
- [ ] `index.html` template is implemented and matches what Figma shows - _todo: add more specifics here once designs are finalized_
  - [ ] As @jgravois and I discussed, our approach here will actually be to have a **base template** with the common UI elements / copy and a **child template** which extends that and includes the INIT-specific UI elements / copy. This is so at some point we can refactor the Littlepay and Switchio templates to also get the new design improvements without necessitating that we do it for this ticket.

## Additional context

Littlepay and Switchio's Enrollment Index pages serve 2 purposes:
1. Tells the user what things they need to have on hand to go through the next step of enrollment
2. Performs tokenization and registration

INIT's Enrollment Index page will only do the first of those. This is because there will be a separate page, which so far we've been calling the ["Card details page"](https://github.com/cal-itp/benefits/issues/3950), which performs tokenization and registration.

**Here's a screenshot of what's in Figma right now -- note that it's subject to change until designs are finalized / translations are done**

<img width="1977" height="1364" alt="Image" src="https://github.com/user-attachments/assets/55b312ac-0f42-4f97-8100-067a386c08e5" />

Issue #4087, opened 2026-09-15
INIT integration: apply design improvements to other pages


angela-tran:
The design improvements that came about from #3954 (e.g. the contextual "eyebrow" UI element) are desired not only for the Enrollment Index page, but also for a few other pages of the Benefits app. The justification is that:
- it would be a little odd for only the new INIT pages to have the design improvements
- it's not too much of an engineering lift to update these 3 additional pages
- the reasoning for the design improvements apply to these pages as well

This GH issue lists all 3 pages, but feel free to implement as separate PRs if that would make more sense.

## Acceptance Criteria

- [ ] Design improvements are implemented for
  - [ ] `/eligibility/start`
  - [ ] `/eligibility/confirm`
  - [ ] `/enrollment/success`

## Additional context

**Here are screenshots of what's in Figma right now -- note that it's subject to change until designs are finalized / translations are done**

<img width="2472" height="1153" alt="Image" src="https://github.com/user-attachments/assets/d9fc0210-90fd-4e91-b4c5-6326ae7f5ccf" />

<img width="1482" height="1595" alt="Image" src="https://github.com/user-attachments/assets/088da885-6538-4380-a16f-37187c240ea8" />

angela-tran:
This issue and #4083 should coordinate on the shared styles, i.e. we wouldn't want 2 separate implementations of the eyebrow, etc.

Scotchester:
Update from @cmajel today: Designs are finalized (but not translations yet); some of the changes for this issue are indexed here: https://github.com/cal-itp/benefits/issues/3954#issuecomment-5702948463

PR #4099, opened 2026-09-17
feat: enrollment_init landing page

jgravois:
closes #4083

more TK


benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+{% block headline %}
+  {% translate "You are eligible for a reduced fare!" as headline_text %}
+  {% include "core/includes/badged-headline.html" with headline=headline_text %}
+{% endblock headline %}
+
+{% block inner-content %}
+  {% block paragraphs %}
+
+    {% translate "The last step is to register a bank card so you get a reduced fare when you tap." as last_step %}
+    <p>{{ last_step }}</p>
```
Scotchester: IMO, we don't need the variable here.


```suggestion
    <p>{% translate "The last step is to register a bank card so you get a reduced fare when you tap." as last_step %}</p>
```

benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+    <p class="mt-3">
+      {% translate "You won't be charged, and we don't store your information. Your card details will be sent to our payment partner " %}
```
Scotchester: Fancy quotes :)

```suggestion
      {% translate "You won’t be charged, and we don’t store your information. Your card details will be sent to our payment partner " %}
```

benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+      <a href="modal--transit-processor"
+         class="{{ classes }}"
+         data-bs-toggle="modal"
+         data-bs-target="#modal--transit-processor">{{ transit_processor.name }}</a>
+      .
```
Scotchester: Putting the period on its own line gives a bit too much of a visual gap between it and the question mark icon, I think.

```suggestion
         data-bs-target="#modal--transit-processor">{{ transit_processor.name }}</a>.
```

Scotchester: Since we no longer need the `partner_post_link`, I believe this can be replaced with the standard `modal-trigger` include directive:

```suggestion
      {% include "core/includes/modal-trigger.html" with modal="modal--transit-processor" text=transit_processor.name period=True %}
```

benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+    <div class="mt-3">
+      {% if alert_include %}
+        {% include alert_include %}
+      {% else %}
+        <p>{% translate "Please use a contactless debit or credit card by:" %}</p>
+        {% include "core/includes/list--card-schemes.html" with classes="mt-3" %}
+        <a href="modal--transit-processor"
+           class="{{ classes }}"
+           data-bs-toggle="modal"
+           data-bs-target="#modal--transit-processor">{% translate "Learn more about the bank cards we accept" %}</a>
```
Scotchester: This one can also use the include:

```suggestion
        {% translate "Learn more about the bank cards we accept" as text %}
        {% include "core/includes/modal-trigger.html" with modal="modal--contactless" text=text classes="p" %}
```

`classes="p"` added for sizing

benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+      {% endif %}
+    </div>
+    {% include "enrollment/includes/modal--transit-processor.html" with id="modal--transit-processor" size="modal-md" header="p-md-2 p-3" body="pb-md-3 mb-md-3 mx-md-3 py-0 pt-0 absolute-top" transit_processor=transit_processor %}
```
Scotchester: Contactless modal markup needs to be added, as well.

```suggestion
    {% include "enrollment/includes/modal--transit-processor.html" with id="modal--transit-processor" size="modal-md" header="p-md-2 p-3" body="pb-md-3 mb-md-3 mx-md-3 py-0 pt-0 absolute-top" transit_processor=transit_processor %}
    {% include "eligibility/includes/modal--contactless.html" with id="modal--contactless" size="modal-lg" header="p-md-2 p-3" body="pb-md-3 mb-md-3 mx-md-3 py-0 pt-0 absolute-top" %}
```

I'm not sure about the difference in the size parameter; I copied this line from `eligibility/start.html`.

benefits/enrollment/templates/enrollment/index-init.html
```diff
@@ -0,0 +1,55 @@
...
+  {% endblock paragraphs %}
+
+  {% block enrollment-form %}
+  {% endblock enrollment-form %}
+{% endblock inner-content %}
+
+{% block call-to-action-button %}
+  <a href="{% url cta_button %}"  class="btn btn-lg btn-primary">{% translate "Register your card" %}</a>
+{% endblock call-to-action-button %}
```
Scotchester: I think this is fine for this PR, but when `index-base.html` subsumes these changes, we'll have to retain its current setup for Littlepay and Switchio, right?

Or perhaps we leave it as it is in `index-base` here and override `{% block call-to-action-button %}` in `enrollment_init/index.html`?

benefits/enrollment_init/views.py
```diff
@@ -1,3 +1,34 @@
...
+    template_name = "enrollment_init/index.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        context.update(
+            {
+                # routes.ENROLLMENT_INIT_REGISTER?
+                "cta_button": routes.ENROLLMENT_INIT_INDEX,
+                "flow": self.flow,
```
Scotchester: Just acknowledging that this will need to be translated, but I'm looking into that, so this is fine for now.

benefits/core/templates/core/includes/badged-headline.html
```diff
@@ -0,0 +1,4 @@
+<div class="d-flex flex-column gap-3 align-items-start mb-4 eyebrowed-headline-container">
+  <div class="p fw-semibold eyebrow-badge">{{ flow }}</div>
```
Scotchester: The `p` class makes this text too large.


```suggestion
  <div class="fw-semibold eyebrow-badge">{{ flow }}</div>
```

Optional suggestion: If we're already doing the majority of the styling on `.eyebrow-badge` in `styles.css`, I'd rather have the weight applied there, too. (Presumably no badge will ever not be semibold.)

benefits/core/templates/core/includes/badged-headline.html
```diff
@@ -0,0 +1,4 @@
+<div class="d-flex flex-column gap-3 align-items-start mb-4 eyebrowed-headline-container">
```
Scotchester: I think I'd just call this `badged-headline` to match the template name?

```suggestion
<div class="d-flex flex-column gap-3 align-items-start mb-4 badged-headline">
```

Or, rename this `eyebrowed-headline.html`. Either way, I think the `-container` is unnecessary.

Scotchester (changes requested):
Thanks, John! I dig the badged headline component.

Found a few tweaks needed for design alignment, a couple things that can be simplified, and an issue with the new modal not having its correct content.

benefits/enrollment_init/views.py
jgravois: yup. definitely intentional. 🤗

benefits/enrollment/templates/enrollment/index-init.html
jgravois: heh. i got a little overzealous because the one above needed to be passed through as a variable. 6db82f0a68ce54e1f614236bd576ab19bb776bd9

benefits/enrollment/templates/enrollment/index-init.html
jgravois: 💅 6db82f0a68ce54e1f614236bd576ab19bb776bd9 (and d8c5bf1ce886d258c21e9df948d66bfd636710c4)

benefits/enrollment/templates/enrollment/index-init.html
jgravois: i like it! 6db82f0a68ce54e1f614236bd576ab19bb776bd9

benefits/enrollment/templates/enrollment/index-init.html
jgravois: less is more! 6db82f0a68ce54e1f614236bd576ab19bb776bd9

benefits/enrollment/templates/enrollment/index-init.html
jgravois: as i see it there will be two different deviations in the Littlepay/Switchio variants.

1. the language will be slightly different to account for the redirects
1. the button action itself will be different

i figured it makes sense to punt on working through the particulars of how to most elegantly templatize something universally useful until the next pass, but i'm happy to dive in and think through it now if anyone is keen to.

benefits/core/templates/core/includes/badged-headline.html
jgravois: good call. 🤙

benefits/enrollment_init/views.py
```diff
@@ -1,3 +1,34 @@
...
+from benefits.core import models
+from benefits.core.mixins import AgencySessionRequiredMixin, EligibleSessionRequiredMixin
+from benefits.enrollment.views import IndexContextMixin
+
+from .routes import routes
+
+logger = logging.getLogger(__name__)
+
+
+class IndexView(AgencySessionRequiredMixin, EligibleSessionRequiredMixin, IndexContextMixin, FormView):
```
angela-tran: I don't think we're using anything from `IndexContextMixin` for the new template. Can we remove it from this view?

benefits/enrollment/templates/enrollment/index-init.html
Scotchester: Sorry, I did an incomplete suggestion here. The entire `<a>` can be replaced with the `include`, as above.

benefits/enrollment/templates/enrollment/index-init.html
Scotchester: Yeah, I'm cool to punt it down the road.

benefits/enrollment/templates/enrollment/index-init.html
jgravois: derp! if i knew i was doing i would have caught out that what i did resulted in _two_ ❓ marks.

13a032e53

benefits/enrollment_init/views.py
jgravois: great suggestion @angela-tran!

this caused me to take a deeper look and realize

1. without `IndexContextMixin` we need to substitute `FlowSessionRequiredMixin`
1. `TemplateView` probably makes more sense here than `FormView`
1. we aren't doing anything with `enrollment_method` yet either

13a032e53

that said, i'm learning as i go here, so if anything else seems sub-optimal, i've got thick skin.

angela-tran: All good points, nice!

Somewhat separate note: I'm finding that the intermediary `fixup!` commits are not as helpful or as easy to review as compared to if the commits were just autosquashed into where they will eventually land. Would it work for your workflow to go ahead with doing the autosquash before pushing up?

jgravois: > Would it work for your workflow to go ahead with doing the autosquash before pushing up?

absolutely. i'm always wary of `-f` pushing when i've asked someone for code review but i can appreciate that the !fixup commits are only transiently useful (at most).

i've cleaned up the commit history and am happy to take that tack with you (and anyone else that prefers it) going forward.

benefits/enrollment_init/views.py
angela-tran: Something else to note here is that `self.flow` here is the `EnrollmentFlow` object, and Django decides to render that using the model's `label`. That `label` so far has only been intended for display in in-person enrollment since it's just a field with untranslated text, so here we're adding a new use case for that field.

One other thing I notice is that the `EnrollmentFlow` `label`s in our environments are in title case while Figma shows that the eyebrow should be sentence case for some of the flows. I'm wondering how we should handle this. Figma also shows that for the agency card flows, the eyebrow should include the `agency name` + the `name of the agency card`.

It seems there's quite a bit to figure out here, so maybe it'd be best to work on this more in a follow-up PR?

jgravois: > so maybe it'd be best to work on this more in a follow-up PR?

that makes sense to me. I know @Scotchester wanted to investigate how best to store translated labels so I figure it makes sense to refine w/ today's use cases in mind in #4087

PR #4100, opened 2026-09-17
Spike: transit processor abstraction

### LINKED ISSUE #4095: INIT integration: add supporting code ("plumbing") needed in `enrollment` to reach `enrollment_init`
## Acceptance Criteria

- [ ] With an INIT-configured `TransitAgency` selected, the user can go through the Benefits app and reach `enrollment_init` without having to do any manual URL navigation

## Additional context

#4086 and #4089 can be worked on and manually tested ahead of this issue being done by going through the Benefits app up to the point of enrollment and then manually putting their URLs in the URL bar.

Regardless of whether this issue is done before or after #4086 and #4089, this task is to implement the supporting code needed in the `enrollment` app so that we can actually reach the `enrollment_init` app.

Notice the patterns used to do this for reaching `enrollment_littlepay` and `enrollment_init`:

https://github.com/cal-itp/benefits/blob/20518a217e936067ee5950b9e160945b0ff42f06/benefits/enrollment/views.py#L67-L74

https://github.com/cal-itp/benefits/blob/20518a217e936067ee5950b9e160945b0ff42f06/benefits/core/models/transit.py#L183-L193

The easiest thing to do would be to add another branch in the `if` statement on `TransitAgency.enrollment_index_route`. Given that this is the third transit processor integration being added to Benefits, we could take a time-boxed look at if would help to add an abstraction somewhere so that we don't have so many `if littlepay; else if switchio; else if init` type of statements everywhere.

angela-tran:
Part of #4095 

As part of needing to add an INIT branch to `TransitAgency.enrollment_index_route`, this spike is to explore ideas for abstracting away all specific transit processor code details from the more generic parts of the Benefits app.

The thought here is that any transit processor implementation should be able to "plug in" without the rest of the app needing to statically import anything from the implementation.


benefits/core/session.py
angela-tran, line 159: Need to move all this to each transit processor's `AppConfig`. I had been thinking of dynamically importing the `group_model`, which is why the dict in the registry has the `module_name`, but now I realize each `AppConfig` can just have an actual reference to its `EnrollmentGroup` child class. So I'll probably remove the module name

benefits/core/context_processors.py
angela-tran, line 124: This no longer has the Littlepay- and Switchio-specific routes, so we should refactor their templates (`enrollment_littlepay/index.html` and `enrollment_switchio/index.html`) to be given the routes they need rather than assuming it's in the global `routes` context variable.

angela-tran:
I realize that this PR impacts a lot of our code _and_ will probably have merge conflicts with the changes in #3977... so I'm not sure how much of this we'll actually merge, but it was good for me to explore as much of this refactor as possible to help check for any cases where this approach wouldn't work.

angela-tran:
@thekaveman - could you review this and let me know what you think of this approach? 

There are still things to do to finish out this refactor, and there's also manual testing we'd need to do to make sure the refactors didn't cause any bugs, but I thought it'd be good to get your feedback before doing any more work here.

benefits/core/session.py
```diff
@@ -150,7 +150,7 @@ def group(request) -> models.EnrollmentGroup | None:
     """Get the EnrollmentGroup from the request's session, or None"""
 
     if agency(request):
-        match agency(request).transit_processor:
+        match agency(request).transit_processor_system_name:
```
thekaveman, line 153: Not the worst case if this is the only place that's leftover, but this is another example (albeit, a different structure) of the `if ... elif ...` type pattern this PR attempts to get rid of.

benefits/enrollment/registry.py
```diff
@@ -0,0 +1,44 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class TransitProcessorRegistry:
+    """Registry which stores information about all registered TransitProcessor modules."""
+
+    entries = {}
+
+    def add(self, system_name, module_name):
```
thekaveman, line 11: Since `entries` is already defined at the class-level, it is _not_ an instance variable.

Hence I might suggest marking `add()` as a `@classmethod`. This might have some other nice downstream effects:

- The helpers below (I think) could be incorporated directly into this class as other `@classmethod`s
- This class becomes a true `singleton` like your initial comment had indicated, no need to instantiate it anymore
- Which maybe simplifies some of the usage elsewhere e.g. `EnrollmentAppConfig` doesn't need a `transit_processors` instance? You can just use the `TransitProcessorRegistry` class itself, everywhere?

benefits/enrollment_init/routes.py
```diff
@@ -0,0 +1,13 @@
+from benefits.routes import Routes as BenefitsRoutes
+
+
+class Routes(BenefitsRoutes):
```
thekaveman, line 4: I don't understand why these need to inherit from `BenefitsRoutes`? It doesn't appear that this inheritance is being used at all?

To be clear, I like the idea of separating the routes into their individual apps 👍

benefits/enrollment_littlepay/routes.py
```diff
@@ -0,0 +1,28 @@
+from benefits.routes import Routes as BenefitsRoutes
+
+
+class Routes(BenefitsRoutes):
```
thekaveman, line 4: [Same comment](https://github.com/cal-itp/benefits/pull/4100#discussion_r4050945925) about inheritance

benefits/enrollment_switchio/routes.py
```diff
@@ -0,0 +1,28 @@
+from benefits.routes import Routes as BenefitsRoutes
+
+
+class Routes(BenefitsRoutes):
```
thekaveman, line 4: [Same comment](https://github.com/cal-itp/benefits/pull/4100#discussion_r4050945925) about inheritance

thekaveman (commented):
Wow, this is a really nice vision for consolidating logic across the transit processors and making maintenance much simpler going forward.

I have a few minor (in the grand scheme) comments below. I'm particularly interested in the nuance of making `TransitProcessorRegistry` function as a singleton through use of `@classmethod` and class-level variables.

RE: #3977 and potential conflicts -- it seems to me that even if there end up being line-by-line conflicts, the resolution should be pretty straightforward. This PR seems to just change how the API clients are initialized, and those lines weren't touched at all in my PR.

This is really nice!! 👏

benefits/enrollment_init/routes.py
angela-tran: Good question - I was prompted to make them inherit from `BenefitsRoute` because the individual apps also need to access the Benefits core routes (e.g. to route to the success page or error pages), and it's simpler to access them from the same imported `route` instance rather than having to import two and alias one. It also makes it so we have access to the methods on `BenefitsRoute`, which are `to_dict` and `name`. I don't think we need the former, but `name` is relevant for when we define the `urlpatterns`.

thekaveman: That makes sense -- maybe I missed this usage though? I thought I was seeing the "import two and alias one" style approach. Maybe this was a follow up that hasn't been done yet?

benefits/core/session.py
angela-tran: Agreed. Finished out the refactor for this in https://github.com/cal-itp/benefits/pull/4100/commits/f0b2c5d2e74db0cfeb3668c0afe53d3402e8544b

benefits/core/session.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/f0b2c5d2e74db0cfeb3668c0afe53d3402e8544b. See similar comment in https://github.com/cal-itp/benefits/pull/4100#discussion_r4050835527

benefits/enrollment/registry.py
angela-tran: Thanks for pointing that out! Those downstream effects sounds good to me, so I'll work on applying your suggestion

benefits/enrollment_init/routes.py
angela-tran: One usage is in `benefits.enrollment_littlepay.views`:

https://github.com/cal-itp/benefits/blob/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9/benefits/enrollment_littlepay/views.py#L16-L18

https://github.com/cal-itp/benefits/blob/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9/benefits/enrollment_littlepay/views.py#L55-L65

The `IndexView` needs routes from `core` as well as from `enrollment_littlepay` (e.g. `route_system_error` and `route_tokenize_success`, respectively).

If this Littlepay `Routes` class didn't extend `BenefitsRoutes`, I would have to do something like:

```python
from benefits.core.routes import routes
from benefits.enrollment_littlepay.routes import routes as littlepay_routes

...

class IndexView( ...

...

     route_system_error = routes.ENROLLMENT_SYSTEM_ERROR 
     route_tokenize_success = littlepay_routes.ENROLLMENT_LITTLEPAY_INDEX
```

We unfortunately do have a place where we have to do the "import two and alias one" style, for now -- it's the `benefits.in_person.views`. Because the `in_person` app currently imports Littlepay and Switchio directly and does not have any sort of separation of them from the rest of its code, we'll have to live with this for now:

https://github.com/cal-itp/benefits/blob/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9/benefits/in_person/views.py#L19-L26

https://github.com/cal-itp/benefits/blob/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9/benefits/in_person/views.py#L101-L102

https://github.com/cal-itp/benefits/blob/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9/benefits/in_person/views.py#L188-L189

I think we can eventually refactor this out, following a similar approach as this PR (move the views into their respective apps, have the `in_person` app allow the transit processors to register themselves as an in-person enrollment implementation). Maybe that refactor could be done when we get to doing the INIT in-person implementation, In any case, it feels out of scope for this PR.

angela-tran:
> RE: #3977 and potential conflicts -- it seems to me that even if there end up being line-by-line conflicts, the resolution should be pretty straightforward. This PR seems to just change how the API clients are initialized, and those lines weren't touched at all in my PR.

Thanks, good point! This PR started out as a spike, but based on your feedback there and on reading back over the changes, I think this spike may be clean enough to actually merge rather than just serving as a throw-away proof-of-concept to be re-implemented. I'm not sure if I'm biased and favoring merging this work because of sunk costs. 

I'll keep working towards getting this merged, but please let me know if we should abandon this PR and re-implement the approach at a later time.

benefits/core/context_processors.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/37cacc3ffc9dd44dbe9fd3b43095e6c5dc0439b9

benefits/enrollment/registry.py
angela-tran: I found this interesting discussion about using the class-level approach: https://discuss.python.org/t/any-extra-reason-to-not-use-classes-for-singletons/19243

Nothing in there makes me think we shouldn't do this, so really I'm linking it here just to have the notes about what you cannot do by taking the class approach and to document that we are ok with it.

> 
> A lesser known alternative is to just use the class itself as the one and only value, but as you point out, it has disadvantages:
> 
> - you have to make all the methods `classmethod`;
> - `property` will not work;
> - this design is so rare that it doesn’t have a catchy name, or even a boring name;
> - will likely confuse linters and automated code checkers;
> - many programmers are like cats, and hate and fear anything different, and this is certainly different!

thekaveman: Hmm, that is interesting discussion. I take their points about using classes like this, maybe it is more of a quirk/feature of Python (there are some equivalent concepts in C#, but not fully...)

I think the main point I was getting at was, is it possible to avoid code like this (which feels a little clunky):

```python
from django.apps import apps
from benefits.enrollment.apps import EnrollmentAppConfig

enrollment_app_config = apps.get_app_config(EnrollmentAppConfig.label)

# do something with enrollment_app_config.transit_processors
```

Maybe there isn't a better way to handle this, and it does make sense to store the instance ref on the `EnrollmentAppConfig` and look it up each time?

thekaveman:
> I'll keep working towards getting this merged

Yep I think this is a great direction and refactor, keep going :+1:

angela-tran:
@thekaveman - you can ignore what I said in our check-in earlier today about maybe not needing the registry. We do need it. Django has some inconsistent behavior that was confusing me.

If anyone is curious, here are the details, but also feel free to simply accept that we do need the registry if you'd rather save your brainpower for something else:

- When you have a `TransitAgency` instance and say `agency.transit_processor_config`, that gives you the actual "typed" object, e.g.
  ```
  > agency.transit_processor_config
   <LittlepayConfig: (cst)>
  ```
  - This made me think "wait, then why did we need those `@property`s like `littlepay_config` that say `return self.transit_processor_config.littlepayconfig` in the first place??". This left me thinking we never needed them and that we don't need the registry.
- However, within the methods of `TransitAgency`, when you say `self.transit_processor_config`, that gives you the object with the base class (`TransitProcessorConfig`) as the type, e.g.
  ```
  > self.transit_processor_config
   <TransitProcessorConfig: (cst)>  # exact same object, but with the base class as the type
  ```

That's why we needed those `@property`s. And in this PR, the registry makes it so we have a single `@property` that abstractly gives you the actual "typed" object.

benefits/enrollment/registry.py
angela-tran: > I think the main point I was getting at was, is it possible to avoid code like this (which feels a little clunky):

Agreed, I was able to remove that type of code in https://github.com/cal-itp/benefits/pull/4100/commits/d0cf0bbaa44c9bc13fd769d64dfb0469fc736e80 and that looks better to me.

benefits/enrollment/registry.py
```diff
@@ -0,0 +1,33 @@
...
+    @classmethod
+    def add(cls, system_name):
+        cls.entries.append(system_name)
+        logger.debug(f'Registered "{system_name}" as a transit processor')
+        logger.debug(f"Currently registered transit processors: {cls.entries}")
+
+    @classmethod
+    def get_transit_processor_config(cls, transit_agency):
+        if transit_agency.transit_processor_config:
+            for system_name in TransitProcessorRegistry.entries:
```
thekaveman: Nit: from _within_ the class, I think we should be consistent on using `cls` instead of the class name.

This is so (in the extremely unlikely event) inheritance still works as expected.

tests/pytest/enrollment/test_registry.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from benefits.enrollment.registry import TransitProcessorRegistry
+from benefits.enrollment_littlepay.models import LittlepayConfig
+
+
+def test_TransitProcessorRegistry_add():
```
thekaveman: Nit: when we're setting up a brand new test suite, especially one that is largely testing a class, it is our newer/better practice to wrap in a class e.g.

```python
class TestTransitProcessorRegistry:

    def test_add(self):
       # ...
```

For one, you can wrap the entire class in `@pytest.mark.django_db` and not have to repeat that for each test that needs the DB.

For another, the fixtures/setup can be more streamlined: use class props to setup necessary data, `autouse=True` fixtures scoped within the test class, etc.

E.g.

```python
@pytest.mark.django_db  # do this once for the entire suite
class TestTransitProcessorRegistry:

    @pytest.fixture(autouse=True)  # automatically used for every test
    def setup(self):
        self.system_name = "blablabla"  # do any setup you need on self

    def test_something(self):
        # self.system_name is available
```

thekaveman (changes requested):
Just some more (possibly too early) nits

benefits/enrollment/registry.py
angela-tran: Good catch!

angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/d64bc844700c8842382b3d3670482a87a0b401f1

tests/pytest/enrollment/test_registry.py
angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/86657ceb7daf0fd26b8e723ae5b7fe131951eddf

benefits/enrollment/registry.py
```diff
@@ -0,0 +1,33 @@
...
+        if transit_agency.transit_processor_config:
+            for system_name in cls.entries:
+                transit_processor_config = cls.get_transit_processor_config_for(transit_agency, system_name)
+                if transit_processor_config:
+                    return transit_processor_config
+
+            return None
+
+    @classmethod
+    def get_transit_processor_config_for(cls, transit_agency, system_name):
```
thekaveman: Consider this the final nit to an overall great PR:

Is there value in having this logic split up across these 2 methods? I didn't see any explicit calls to `get_transit_processor_config_for` aside from the call in `get_transit_processor_config`... so it doesn't seem like callers need both ways of access?

Maybe especially in the absence of docstrings, it could later be confusing which one a caller is supposed to use? I'd advocate for just inlining this one into the main `get_transit_processor_config` since it's only 5 more lines. But I wouldn't block the PR because of it.

angela-tran: > Is there value in having this logic split up across these 2 methods? ...
> 
> Maybe especially in the absence of docstrings, it could later be confusing which one a caller is supposed to use?

That is a great point. Very helpful to look at it from that perspective :100:

angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/ef19692d6f02e97ef2c08e58151d06023cdc31c0

benefits/enrollment/registry.py
```diff
@@ -0,0 +1,30 @@
...
+                config_model_name = system_name + "config"
+                if hasattr(transit_agency.transit_processor_config, config_model_name):
+                    transit_processor_config = getattr(transit_agency.transit_processor_config, config_model_name)
+                else:
+                    transit_processor_config = None
+
+                if transit_processor_config:
+                    return transit_processor_config
+
+            return None
```
angela-tran, line 30: Hmm is this indented one too many levels? Or maybe I need another `return None`?

I'll test this out

jgravois (commented):
i'll leave it to @thekaveman to bless this one. i just wanted to say that i played around with this branch yesterday with an eye toward implementing the transit processor info in the help pages dynamically and the results were very promising.

nice work!

benefits/enrollment/registry.py
angela-tran: Just to be clear, my comment above is talking about how there's no return statement for essentially the `else` branch for the `if agency.transit_processor_config` statement on line 19. 

I tested it out and read Python docs and found that Python will implicitly return `None`.

_From https://docs.python.org/3/tutorial/controlflow.html#defining-functions_:
> even functions without a [`return`](https://docs.python.org/3/reference/simple_stmts.html#return) statement do return a value, albeit a rather boring one. This value is called `None` (it’s a built-in name).

_And from https://docs.python.org/3/reference/simple_stmts.html#return_:
> **return_stmt**: "`return`" [[`expression_list`](https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-expression_list)]
> ...
> If an expression list is present, it is evaluated, else `None` is substituted.

I'll go ahead and add a `return None` at the same level as the `if` on line 19 so that the intention is clear we should return `None` if the agency doesn't have a `transit_processor_config`.

angela-tran: Done in https://github.com/cal-itp/benefits/pull/4100/commits/e0de24be37b97a0a6025143d96338db1947942e8

thekaveman:
Sorry! Will get to this in just a few minutes

thekaveman (approved):
Very nice 🎸

angela-tran:
Thanks @thekaveman and good call on the title renaming!

